PR #10, opened 2025-09-04
Enterprise Felt-is-Firefox with addons

lissyx:


Cargo.toml
```diff
@@ -96,13 +96,6 @@ opt-level = 1
 [profile.release.build-override]
 opt-level = 1
 
-# optimizing glsl more makes a big difference in swgl build times
```
gcp: This feels totally random? Merge error?

browser/extensions/felt/api.js
```diff
@@ -0,0 +1,218 @@
...
+        break;
+
+      case "FeltParent:FirefoxNormalExit":
+        Services.ppmm.removeMessageListener("FeltParent:FirefoxNormalExit", this);
+	Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit | Ci.nsIAppStartup.eConsiderQuit);
+        break;
+
+      case "FeltParent:FirefoxAbnormalExit":
+        Services.ppmm.removeMessageListener("FeltParent:FirefoxAbnormalExit", this);
+        // TODO: What should we do, restart Firefox?
```
gcp, line 211: The launcher needs to do this I think to set up the IPC channel correctly.

browser/extensions/felt/api.js
```diff
@@ -0,0 +1,218 @@
...
+    this.feltXPCOM.setPrefsForFelt();
+  }
+
+  onStartup() {
+    this.feltXPCOM = Cc["@mozilla.org/toolkit/library/felt;1"].getService(
+      Ci.nsIFelt
+    );
+
+    if (this.feltXPCOM.isFeltUI()) {
+      this.feltXPCOM.setConsoleAddr("http://localhost:8000");
```
gcp: Have we thought about how to supply this in production?

browser/extensions/felt/background.js
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* eslint-env webextensions */
+
+"use strict";
+
+browser.runtime.onUpdateAvailable.addListener(_details => {
+  // The New Tab built-in add-on is not designed to be updated at runtime.
```
gcp, line 10: New Tab? :smile:

browser/extensions/felt/content/FeltParent.sys.mjs
```diff
@@ -0,0 +1,188 @@
...
+            this.felt.sendStringPreference(name, value);
+            break;
+        }
+      });
+    }).then(() => {
+      console.debug(`firefox: waiting on proc PID ${this.proc.pid}`, this.proc);
+
+      console.debug(`Starting 30s timeout to show about:restartforeced`);
+      lazy.setTimeout(() => {
+        console.debug(`Triggered 30s timeout to show about:restartforeced`);
```
gcp, line 86: Typo

build/mozconfig.common.enterprise
```diff
@@ -1,4 +1,6 @@
 ac_add_options --with-branding=browser/branding/official
 ac_add_options --enable-bootstrap
 ac_add_options --enable-application=browser
+ac_add_options --enable-felt
 ac_add_options --enable-strip
+export MOZILLA_OFFICIAL=1
```
gcp: Is there something specific that requires this?

gcp (approved):

Cargo.toml
lissyx: it was rebased so i need to check that

browser/extensions/felt/api.js
lissyx: Not yet, that is something I'd like we discuss next monday

browser/extensions/felt/background.js
lissyx: yeah :p

build/mozconfig.common.enterprise
lissyx: Not yet but your telemetry changes will need it no?

gcp: The opposite, I decoupled Telemetry from this setting.

browser/extensions/felt/api.js
gcp: Can add it to the doc so we can go over it (probably Tuesday with fiji).

lissyx: One idea I had is that this is just setting a pref so it could be set at install time somehow ? https://github.com/mozilla/enterprise-firefox/pull/10/commits/ac21fc72229b3f0594a1ce436d8876cb7a56c217#diff-6d0fd0afe07ce9bcbb6ddb882c13066bcbb500ab4bfcc96792c8005fd4b6fd99R375

lissyx:
@1rneh I'd like your review on the UI part and maybe on the tests as well

lissyx:
@mkaply I'd like your review on the Live Policies changes

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -191,19 +201,22 @@ EnterprisePoliciesManager.prototype = {
...
-          this._schedulePolicyCallback(
-            timing,
-            policyCallback.bind(
-              policyImpl,
-              this /* the EnterprisePoliciesManager */,
-              parsedParameters
-            )
-          );
-        }
+  _maybeCallbackPolicy(policyImpl, parsedParameters = undefined) {
```
mkaply, line 221: Can you comment as to what this function does?Can you add a comment as to what this function does?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -676,6 +846,17 @@ class macOSPoliciesProvider {
     this._policies = lazy.macOSPoliciesParser.readPolicies(prefReader);
   }
 
+  onPoliciesChanges(handler) {
```
mkaply, line 878: Nit, but grammatically this doesn't make sense. What exactly is changing here?

mkaply (commented):
A few comments, but from my perspective this looks sound as long as tests pass.

I'm going to build it locally this week.

browser/extensions/felt/api.js
```diff
@@ -0,0 +1,218 @@
...
+      }
+
+      register(actorName, actor);
+    }
+  },
+
+  addJSWindowActors(actors) {
+    this._addActors(actors, "JSWindowActor");
+  },
+};
```
1rneh, line 105: We can use `ChromeUtils.registerWindowActor` in `onStartUp` like we do with `ChromeUtils.unregisterWindowActor` in `onShutDown`. It's not clear to me why we need to register and unregister the felt actors via preference?

browser/extensions/felt/api.js
1rneh: A repack?

browser/extensions/felt/content/FeltChild.sys.mjs
```diff
@@ -0,0 +1,44 @@
...
+  constructor() {
+    super();
+    // Notify ExtensionAPI side that we are loaded, we cannot do the
+    // Services.cpmm calls within constructor but the setTimeout delays
+    // enough that it works
+    setTimeout(() => {
+      Services.cpmm.addMessageListener("FeltMain:RedirectURL", this);
+      Services.cpmm.sendAsyncMessage("FeltChild:Loaded", {});
+    }, 0);
+  }
```
1rneh, line 11: We can use `actorCreated` here, so we don't need `setTimeout`.

browser/extensions/felt/content/FeltChild.sys.mjs
```diff
@@ -0,0 +1,44 @@
...
+
+  // Only handle DOMContentLoaded on https://sso.mozilla.com/dashboard
+  handleEvent(event) {
+    if (event.type !== "DOMContentLoaded") {
+      console.error(`Unexpected event.type=${event.type}`);
+      return;
+    }
+
+    this.sendAsyncMessage("FeltChild:StartFirefox", {});
```
1rneh, line 20: We should avoid that the user sees a glimpse of the dashboard in the sso frame before we start another Firefox instance.

browser/extensions/felt/content/FeltParent.sys.mjs
```diff
@@ -0,0 +1,197 @@
...
+            this.startFirefox();
+          })
+          .catch((err) => {
+            console.debug(`FELT: Killed failed: ${err}`);
+          });
+        break;
+
+      default:
+        break;
+    }
```
1rneh, line 225: Is the restart related code about the forced restart? If so, could you move any restart related logic to a separate patch to streamline this patch?

browser/extensions/felt/api.js
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  receiveMessage(message) {
+    console.debug(`FeltExtension: ${message.name} handling ...`);
+    switch (message.name) {
+      case "FeltChild:Loaded":
+        Services.ppmm.removeMessageListener("FeltChild:Loaded", this);
+        const redirect_after_sso = Services.prefs.getStringPref("browser.felt.redirect_after_sso");
+        Services.ppmm.broadcastAsyncMessage("FeltMain:RedirectURL", redirect_after_sso);
+        break;
```
1rneh, line 194: Is there a reason why `FeltChild` doesn't just get the preference once it receives the `FeltParent:Done` message? This way we could save ourselves the `FeltChildLoaded` and `FeltMain:RedirectURL` messages.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: This is code that was here and I moved out of the function because at some point it proved to be useful.

browser/extensions/felt/api.js
1rneh: I just realized that the child never receives the message `FeltParent:Done` because the parent doesn't send it anywhere, so we never redirect.

1rneh (commented):
Just some comments and questions. No blockers though, since our goal is to get this landed. But I'd like to re-visit this with a bit more time on our hands.

browser/extensions/felt/api.js
lissyx: That was complicated because we were depending on the `matches` pref and the `JSWindowActor` C++ code expects things to be set once. So we could not really register until we had the proper information.

With the recent evolutions, this might not be required anymore?

browser/extensions/felt/content/FeltChild.sys.mjs
lissyx: Thanks!

browser/extensions/felt/content/FeltChild.sys.mjs
lissyx: I think it's macOS specific as I dont get what you mean :)

browser/extensions/felt/content/FeltParent.sys.mjs
lissyx: Nope, restart forced is just an example/demo of this.

browser/extensions/felt/api.js
lissyx: Yep, I broke things at some point and the `FeltParent:Done` message went missing. Writing tests made me catch this.

For the prefs it's likely a consequences of the old way of fetching prefs from console. I'll see if we can simplify this right now.

browser/extensions/felt/api.js
1rneh: Yes, now we have all prefs on startup and we can register them right away.

browser/extensions/felt/api.js
```diff
@@ -0,0 +1,264 @@
...
+      }
+
+      register(actorName, actor);
+    }
+  },
+
+  addJSWindowActors(actors) {
+    this._addActors(actors, "JSWindowActor");
+  },
+};
```
1rneh, line 105: I don't see any use of `enablePreference` on our actors, so the whole actor registration logic can be streamlined to

```
registerActors() {
    ChromeUtils.registerWindowActor("Felt", {
      parent: {
        esModuleURI: "chrome://felt/content/FeltParent.sys.mjs",
      },

      child: {
        esModuleURI: "chrome://felt/content/FeltChild.sys.mjs",
        events: {
          DOMContentLoaded: {},
          load: {},
        },
      },
      allFrames: true,
      matches: Services.prefs.getStringPref("browser.felt.matches").split(","),
    });
  }
```

and then calling `this.registerActors()` instead of `ActorManagerParent.addJSWindowActors(JSWINDOWACTORS);` in `onStartUp`.

browser/extensions/felt/content/FeltChild.sys.mjs
1rneh: I'll raise this again once I merged my UI changes.

browser/extensions/felt/api.js
lissyx: yes, now we should be able to do it. I'll look at that in a follow up, or if you want to fold it in your patches it's fine as well

PR #13, opened 2025-09-04
Prepare Felt with crate

lissyx:


Cargo.toml
```diff
@@ -96,13 +96,6 @@ opt-level = 1
 [profile.release.build-override]
 opt-level = 1
 
-# optimizing glsl more makes a big difference in swgl build times
```
gcp: Wrong merge?

build/mozconfig.common.enterprise
```diff
@@ -1,4 +1,6 @@
 ac_add_options --with-branding=browser/branding/official
 ac_add_options --enable-bootstrap
 ac_add_options --enable-application=browser
+ac_add_options --enable-felt
 ac_add_options --enable-strip
+export MOZILLA_OFFICIAL=1
```
gcp: Let's do this separately

netwerk/cookie/CookieService.cpp
```diff
@@ -776,6 +776,23 @@ CookieService::AddForAddOn(const nsACString& aHost, const nsACString& aPath,
   return NS_OK;
 }
 
+NS_IMETHODIMP
```
gcp, line 780: `#if defined MOZ_WIDGET_FELT`

build/mozconfig.common.enterprise
gcp: I added `MOZ_ENTERPRISE` in my branch but we can make that imply `MOZ_WIDGET_FELT` or whatever.

gcp (approved):

build/mozconfig.common.enterprise
lissyx: Sure

PR #14, opened 2025-09-04
Fix path handling of firefox binary path

lissyx:
When mounting the dmg, it may mount to "/Volumes/Firefox x" with a space. Lack of quotes breaks.


gcp:
Squash this with the original pull?

lissyx:
> Squash this with the original pull?

https://github.com/mozilla/enterprise-firefox/pull/13 ? I can, if you r+ it :)

lissyx:
Adedd in #13

PR #16, opened 2025-09-08
FELT: Add an about:restartforced page

lissyx:
Differential Revision: https://phabricator.services.mozilla.com/D262889

### DIFF
--- a/dom/base/nsIContentPolicy.idl
+++ b/dom/base/nsIContentPolicy.idl
@@ -523,6 +523,11 @@ interface nsIContentPolicy : nsISupports
   const short REJECT_POLICY = -5;
 
   /**
+   * Returned from shouldLoad or shouldProcess if restart is forced 
+   */
+  const short REJECT_RESTARTFORCED = -6;
+
+  /**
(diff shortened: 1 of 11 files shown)


dom/base/nsIContentPolicy.idl
```diff
@@ -522,6 +522,11 @@ interface nsIContentPolicy : nsISupports
    */
   const short REJECT_POLICY = -5;
 
+  /**
+   * Returned from shouldLoad or shouldProcess if restart is forced 
+   */
+  const short REJECT_RESTARTFORCED = -6;
```
gcp, line 528: Looks suspicious that this logic isn't shared with about:restartrequired.

gcp (approved):

dom/base/nsIContentPolicy.idl
lissyx: That's because we need to setup a content policy, it's part of the crate, and it's in charge of blocking activity and surfacing the error to the user

the `about:restartrequired` page relied on a process not reporting its proper startup so it's all tied to process abnormal exit somehow.

I think it's cleaner and safer this way (and not more complicated?)

PR #25, opened 2025-09-15
Making FeltParent a JSProcessActor

lissyx:


browser/components/BrowserContentHandler.sys.mjs
```diff
@@ -1581,6 +1581,15 @@ nsDefaultCommandLineHandler.prototype = {
       return;
     }
 
+    // Make sure that when FeltUI is requested, we do not try to open another
+    // window.
+    if (cmdLine.findFlag("feltUI", true) != -1) {
+      console.debug(`Felt: Found FeltUI in BrowserContentHandler.`);
+      Services.startup.enterLastWindowClosingSurvivalArea();
```
1rneh: Could you make sure that we also call `Services.startup.exitLastWindowClosingSurvivalArea()` once we opened a window, e.g. like this:
```
Services.obs.addObserver(function windowOpenObserver() {
    Services.startup.exitLastWindowClosingSurvivalArea();
    Services.obs.removeObserver(windowOpenObserver, "domwindowopened");
  }, "domwindowopened");
```

browser/components/BrowserGlue.sys.mjs
```diff
@@ -609,6 +609,14 @@ BrowserGlue.prototype = {
     let startTime = ChromeUtils.now();
 
     let shouldCreateWindow = isPrivateWindow => {
+      // Make sure that when FeltUI is requested, we do not try to open another
+      // window.
+      if (cmdLine.findFlag("feltUI", true) != -1) {
+        console.debug(`Felt: Found FeltUI in BrowserGlue.`);
+        Services.startup.enterLastWindowClosingSurvivalArea();
+        return false;
```
1rneh, line 617: Same as above.

browser/extensions/felt/api.js
```diff
@@ -212,7 +222,11 @@ this.felt = class extends ExtensionAPI {
         break;
 
       case "FeltParent:FirefoxStarted":
-        this._win.minimize();
+        Services.startup.enterLastWindowClosingSurvivalArea();
```
1rneh, line 225: Same as above.

browser/extensions/felt/api.js
```diff
@@ -51,6 +46,13 @@ JSWINDOWACTORS.Felt = {
   },
 };
 
+let JSPROCESSACTORS = {};
+JSPROCESSACTORS.FeltProcess = {
+  parent: {
+    esModuleURI: "chrome://felt/content/FeltParentProcess.sys.mjs",
```
1rneh: Could you rename it to `FeltProcessParent` to align with how APIs like `.getActor()` retrieve the correct classes, see the [docs](https://firefox-source-docs.mozilla.org/dom/ipc/jsactors.html#about-actor-names).

browser/extensions/felt/content/FeltChild.sys.mjs
```diff
@@ -4,10 +4,17 @@
 
 import { setTimeout } from "resource://gre/modules/Timer.sys.mjs";
 
-export class FeltChild extends JSWindowActorChild {
+console.debug(`FeltExtension: FeltChild.sys.mjs`);
+
+export class FeltWindowChild extends JSWindowActorChild {
```
1rneh, line 7: Could you rename the file to `FeltWindowChild.sys.mjs`?

browser/extensions/felt/content/FeltChild.sys.mjs
```diff
@@ -4,10 +4,17 @@
 
 import { setTimeout } from "resource://gre/modules/Timer.sys.mjs";
```
1rneh: This can be removed because it's unused.

browser/extensions/felt/content/FeltChild.sys.mjs
```diff
@@ -17,7 +24,7 @@ export class FeltChild extends JSWindowActorChild {
       return;
     }
 
-    this.sendAsyncMessage("FeltChild:StartFirefox", {});
+    this.actor.sendAsyncMessage("FeltChild:StartFirefox", {});
```
1rneh, line 25: Is sending that message the only logic that runs in `FeltProcessChild`?
If there is more to come, I'd prefer if we then move that logic to its own file `FeltProcessChild.sys.mjs`.

browser/extensions/felt/api.js
lissyx: For this I think we can just call `Services.startup.exitLastWindowClosingSurvivalArea();` when `domwindowopened` happens, and expect that we only `Services.startup.enterLastWindowClosingSurvivalArea();` when we explicitely started firefox, to make sure we control **when** we hide things and need to block closing

browser/extensions/felt/content/FeltChild.sys.mjs
lissyx: Except we are in the `JSWindowActor` here

browser/components/BrowserContentHandler.sys.mjs
lissyx: So having both this one and the one below, we trip https://searchfox.org/firefox-main/rev/938e8f38c6765875e998d5c2965ad5864f5a5ee2/toolkit/components/startup/nsAppStartup.cpp#371

browser/components/BrowserGlue.sys.mjs
lissyx: Given `BrowserGlue` seems to load first, I'll keep it only here.

1rneh (approved):
looks good to me now!

PR #26, opened 2025-09-16
Enterprise main merge 20250916

lissyx:


lissyx:
Looks like CI is failing on macOS due to a race but local runs are fine. Work in #25 will anyway change how this is done

PR #28, opened 2025-09-16
Felt Login UI

1rneh:


testing/enterprise/firefox_start.json
```diff
@@ -0,0 +1,11 @@
+{
```
lissyx: You should remove this file, it's generated from the `.in` one

browser/extensions/felt/api.js
```diff
@@ -250,15 +250,19 @@ this.felt = class extends ExtensionAPI {
...
       null,
-      "chrome://felt/content/feltui.xhtml",
+      "chrome://felt/content/felt.xhtml",
       "_blank",
       flags,
       null
     );
 
+    // for selenium to directly target the window
+    this._win.name = "felt";
```
lissyx: Do we really need ? I never felt the need

testing/enterprise/README.md
```diff
@@ -1,2 +1,6 @@
 Tests can be ran using
 $ mach enterprise-tests
+
+$ Make to build with
```
lissyx: Maybe suggest to use mozconfig enterprise specific files from under `build`?

testing/enterprise/felt_tests.py
```diff
@@ -360,6 +366,48 @@ def set_felt_window(self):
...
+    def switch_to_felt_chrome_context(
+        self, name="felt", path="chrome://felt/content/felt.xhtml"
+    ):
+
+        self._logger.info("Checking FELT pref")
+        assert self.get_felt_enabled_pref() is True, "pref should be enabled"
+
+        self._logger.info(f"Waiting for chrome window {path}.")
+        self._driver.set_context("chrome")
+        self._longwait.until(
```
lissyx: Long wait should be avoided if possible

testing/enterprise/felt_tests.py
```diff
@@ -360,6 +366,48 @@ def set_felt_window(self):
                 self._logger.info(f"Found FELT at {win}")
                 return
 
+    def switch_to_felt_chrome_context(
```
lissyx: This looks very complicated compared to what I have currently. We're you trying to go around with the dummy.xhtml window? We should assume (assert ?) that there is only ever one window

testing/enterprise/felt_tests.py
```diff
@@ -360,6 +366,48 @@ def set_felt_window(self):
...
+                }
+            }
+            return false;
+        """,
+                name,
+                path,
+            )
+        )
+
+        self._logger.info(f"Switching to chrome context of window {name}")
```
lissyx: Yes all this looks like complications from the mess of windows we don't have anymore

browser/extensions/felt/content/FeltCommon.sys.mjs
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
lissyx: Nice idea, I think we could unify at least messages names as const declared there?

testing/enterprise/README.md
1rneh: TIL that we had that!

browser/extensions/felt/content/FeltCommon.sys.mjs
1rneh: I moved this out of the scope but added two things to the list of improvements:
- use console api in FeltCommon to log across felt classes
- unify felt messaged in FeltCommon

browser/extensions/felt/api.js
```diff
@@ -250,10 +250,12 @@ this.felt = class extends ExtensionAPI {
   }
 
   showWindow() {
-    let flags = "chrome,centerscreen,titlebar";
+
```
lissyx, line 253: nit: extra line

browser/extensions/felt/api.js
```diff
@@ -278,6 +280,6 @@ this.felt = class extends ExtensionAPI {
     this.chromeHandle.destruct();
     this.chromeHandle = null;
 
-    ChromeUtils.unregisterWindowActor("FELT");
+    ChromeUtils.unregisterWindowActor("Felt");
```
lissyx, line 283: Good catch! I'm wondering if we also want to unregister the process actor?

browser/extensions/felt/content/felt.xhtml
```diff
@@ -0,0 +1,64 @@
...
+<html
+  xmlns="http://www.w3.org/1999/xhtml"
+  role="document"
+  class="system-font-size"
+  id="felt-window"
+  windowtype="navigator:browser"
+>
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="Content-Security-Policy" content="default-src chrome:" />
```
lissyx, line 15: Do we really need this? I went to some effort to sanitize CSP things when moving to 100% pure extension because some may triggers/trip assertions/csp checks, so if we can avoid taking the risk here

browser/extensions/felt/content/window.js
```diff
@@ -44,12 +50,37 @@ async function init() {
...
+function listenFormEmailSubmission() {
+  const signInBtn = document.getElementById("felt-login__form-sign-in-btn");
+  const emailInput = document.getElementById("felt-login__form-email")
+
+  function handleSubmit() {
+    load_sso_url()
   }
+
+  emailInput.addEventListener("input", () => {
+    signInBtn.disabled = emailInput.value.trim() === "";
```
lissyx, line 70: nit: maybe `emailInput.includes("@")`?

testing/enterprise/README.md
```diff
@@ -1,2 +1,5 @@
 Tests can be ran using
 $ mach enterprise-tests
+
+Build with enterprise-specific mozconfig options:
+. "$topsrcdir/build/mozconfig.common.enterprise"
```
lissyx, line 5: Sorry I should have been more precise:
 - `build/macosx/mozconfig.enterprise`
 - `build/unix/mozconfig.enterprise`
 - `build/win64/mozconfig.enterprise`

lissyx (approved):

PR #32, opened 2025-09-17
Add a dummy AddNativeForFelt

lissyx:
This makes sure that builds without --enable-felt are not broken. We need dummy since we cannot preprocess the idl file to selectively add addNativeForFelt() to the interface.


gcp:
I've found a better way to do this.

PR #35, opened 2025-09-18
FELT: Remove -feltUI and use MOZ_FELT_UI env

lissyx:


Mossop (approved):
Looks fine.

Mossop:
It just occurred to me that we can't set environment variables in Windows shortcuts, so if we need a shortcut on Windows to launch felt we probably need to handle both command line and environment variable. Might be easiest to set the environment variable in early startup if the command line is present then we can just check the environment variable elsewhere.

PR #38, opened 2025-09-19
FELT: default --enable-warnings-as-errors

lissyx:


lissyx:
Building with rust 1.90 is broken, minimum recommended version on m-c seems 1.85 so forcing this one.

PR #40, opened 2025-09-22
Remove browser.felt.enabled and cleanup felt actor registration

1rneh:


browser/extensions/felt/api.js
```diff
@@ -280,6 +195,7 @@ this.felt = class extends ExtensionAPI {
     this.chromeHandle.destruct();
     this.chromeHandle = null;
 
-    ChromeUtils.unregisterWindowActor("Felt");
+    ChromeUtils.unregisterWindowActor("FeltWindow");
```
lissyx: nit: we could make those `const` we define and reuse here and during the registration process?

lissyx (approved):

PR #42, opened 2025-09-23
Felt Frontend

1rneh:


browser/extensions/felt/api.js
```diff
@@ -27,51 +27,9 @@ this.felt = class extends ExtensionAPI {
       ["content", "felt", "content/"],
     ]);
   }
-
```
lissyx, line 30: \o/

browser/extensions/felt/api.js
```diff
@@ -27,51 +27,9 @@ this.felt = class extends ExtensionAPI {
...
-          break;
-      }
-    });
-  }
-
   registerActors() {
-    const matches = Services.prefs
-          .getStringPref("browser.felt.matches")
-          .split(",");
+    const { ConsoleClient } = ChromeUtils.importESModule("chrome://felt/content/ConsoleClient.sys.mjs")
```
lissyx, line 31: I really think we will want to expose this as an XPCOM at some point

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,152 @@
...
+
+export const ConsoleClient = {
+
+    _consoleTokenData: null,
+
+    get consoleTokenData() { return this._consoleTokenData; },
+    set consoleTokenData(val) { this._consoleTokenData = val; },
+
+    get consoleAddr() {
+        return isTesting() ? Services.prefs.getStringPref(PREFS.CONSOLE_ADDRESS, "") : "https://console.enterfox.eu";
```
lissyx, line 73: We may want to revisit this: empty string may hide what is going on here, if we dont have a pref set I think we should just directly fail?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -786,7 +786,7 @@ class RemotePoliciesProvider {
         const serverAddr = Services.prefs.getStringPref(
           "browser.policies.server"
         );
-        const response = await fetch(`${serverAddr}/policies`);
+        const response = await fetch(`${serverAddr}/api/browser/hacks/policies`);
```
lissyx, line 789: likely we want to fix that after the demo

testing/enterprise/test_felt_browser_about_config_blocked.py
```diff
@@ -45,7 +45,7 @@ def test_felt_4_change_about_config_policy(self, exp):
         self.policy_block_about_config.value = 0
         self._logger.info("Changed BlockAboutConfig policy")
 
-        url = f"http://localhost:{self.console_port}/policies"
+        url = f"http://localhost:{self.console_port}/api/browser/hacks/policies"
```
lissyx, line 48: likely we want to fix that after the demo

testing/enterprise/felt_tests.py
```diff
@@ -289,11 +272,11 @@ def setup(self):
         else:
             self._child_profile_path_value = self._child_profile_path
 
-        self.set_string_pref("browser.felt.console", console_addr)
+        # self.set_string_pref("browser.felt.console", console_addr)
```
lissyx, line 275: nit: we can clean that later?

lissyx (approved):

PR #44, opened 2025-09-23
FELT: Close window before starting Firefox

lissyx:
Somehow this makes the dock icon merging working as we would like, i.e., avoiding a duplicated icon because Felt was running while browser launched.

### DIFF
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -54,12 +54,20 @@ this.felt = class extends ExtensionAPI {
       Ci.nsIFelt
     );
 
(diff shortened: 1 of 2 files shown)


browser/extensions/felt/api.js
```diff
@@ -54,12 +54,20 @@ this.felt = class extends ExtensionAPI {
       Ci.nsIFelt
     );
 
+    // In tests if we close too early we loose the browser context
```
gcp: lose not loose

Mossop (approved):

PR #47, opened 2025-09-24
Felt private window

lissyx:


testing/enterprise/felt_tests.py
```diff
@@ -419,13 +419,32 @@ def test_felt_1_perform_sso_auth(self, exp):
...
+        # In private window this one should not reflect
+        not_expected_cookie = list(
             filter(
                 lambda x: x["name"] == self.cookie_name
                 and x["value"] == self.cookie_value,
                 self._driver.get_cookies(),
             )
         )
-        assert len(expected_cookie) == 1, f"Cookie {self.cookie_name} was properly set"
+        assert len(not_expected_cookie) == 0, f"Cookie {self.cookie_name} was not properly set"
```
1rneh: Could you help me understand this test? I understand the first part of it: The sso cookie is put into the private browsing cookie jar, we check that the cookie is accessbile in the chrome context. When switching the context to content (`self._driver.set_context("content")`), aren't we still be in the same browsing context within the same private browsing session at this moment? I would have expected that the cookie is still accessible until the private browsing session ends.  What am I missing?

lissyx: Private browsing cookies are not accessible via the default `getCookies` on the service, so I think this is what gets reflected here when we `get_cookies()` at driver level.

lissyx: confere https://searchfox.org/firefox-main/rev/054e2b072785984455b3b59acad9444ba1eeffb4/netwerk/cookie/nsICookieManager.idl#39-46 for the different

1rneh: Got it! Thanks, I didn't know that difference.

1rneh (approved):
Lgtm! Using a private window to keep a clean felt profile seems like a good idea.

lissyx:
From the design constraints on https://www.figma.com/board/29dqnGYlMoyVozr9tinfm8/Untitled?node-id=0-1&p=f&t=AYspJbxrsXr54GH4-0 I think we are fine to land this.

PR #50, opened 2025-09-25
FELT: Make FeltUI non remote and Browser remote-able

lissyx:
This is WIP and seems not to properly do what is expected at least on macOS


lissyx:
We likely dont want to do this and rather have felt handle and redirect to actual browser ; this would make sure that if the browser/felt are not running we go through SSO flow

PR #52, opened 2025-09-26
Fix jsdebugger

lissyx:


lissyx:
We need to work on a better fix, so I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1991010, but for now should make life easier

PR #61, opened 2025-09-30
Use the profile service directly to find or create the enterprise profile

Mossop:
This is quicker than re-launching Firefox. Also adds the `--foreground` argument to make sure Firefox tries to come to the foreground.

### DIFF
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -27,12 +27,12 @@ export class FeltProcessParent extends JSProcessActorParent {
         console.debug(`FeltExtension: ParentProcess: Received ${aTopic}`);
         switch (aTopic) {
           case "felt-firefox-restarting":
-	    const restartDisabled = Services.prefs.getBoolPref("browser.felt.disable_restart", false);
+            const restartDisabled = Services.prefs.getBoolPref("browser.felt.disable_restart", false);
             if (!restartDisabled) {
                 Services.ppmm.broadcastAsyncMessage("FeltParent:RestartFirefox", {});
-	    } else {
+            } else {
                 console.debug(`FeltExtension: ParentProcess: restart is disabled`);
-	    }
+            }
             break;
 
           default:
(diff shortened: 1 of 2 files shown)


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -14,6 +14,8 @@ ChromeUtils.defineESModuleGetters(lazy, {
 
 console.debug(`FeltExtension: FeltParentProcess.sys.mjs`);
 
+const ENTERPRISE_PROFILE = "enterprise-profile";
```
lissyx: Maybe make it a constant in `ConsoleClient` ?

Mossop: Like so?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -119,98 +119,83 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    const env = {
+      "MOZ_FELT_UI": null
+    };
+
+    let profilePath = Services.prefs.getStringPref(
+      "browser.felt.profile_path",
+      ""
+    );
+
+    if (!profilePath) {
```
lissyx, line 132: So the reason I relied on `try` here is because with JS I'm never 100% certain of how this will be interpreted.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -119,98 +119,83 @@ export class FeltProcessParent extends JSProcessActorParent {
...
-            return proc.exitCode === 0;
-          })
-          .catch(err => {
-            console.error(err instanceof Error ? err : err.message);
-            return false;
-          });
+    const firefoxBin = this.felt.binPath();
+
+    // Make sure we do not transmit MOZ_FELT_UI to child processes
+    const env = {
```
lissyx: this can now go away if you rebase on current `enterprise-main`

browser/extensions/felt/content/FeltProcessParent.sys.mjs
Mossop: An empty string will be considered falsey so the absence of the pref or if the pref is set to an empty string will cause us to create a new profile.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -119,98 +119,83 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      let profileService = Cc[
+        "@mozilla.org/toolkit/profile-service;1"
+      ].getService(Ci.nsIToolkitProfileService);
 
-        if (!profileCreation) {
-          console.debug(`startFirefox(): profileCreation failed`);
-          reject();
-          return;
+      let foundProfile = null;
+      for (let profile of profileService.profiles) {
```
lissyx, line 138: likely we may do:
```
let foundProfile = profileService.profiles.filter(profile => profile.name === lazy.ConsoleClient.ENTERPRISE_PROFILE)
```

but this can be improved later

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -119,98 +119,83 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      socket,
+      ...extraRunArgs,
+    ];
+
+    const firefoxRun = {
+      command: firefoxBin,
+      arguments: firefoxRunArgs,
+      stdout: "stdout",
+      stderr: "stderr",
+      environmentAppend: true,
```
lissyx: this can now go away if you rebase on current `enterprise-main`

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -119,98 +119,83 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+        "--marionette",
+        "--remote-allow-hosts",
+        "localhost",
+        "--remote-allow-system-access",
+      ];
+    }
+
+    const firefoxRunArgs = [
+      "--foreground",
+      "--profile",
```
lissyx, line 169: Why `--profile` with the path and not `"-P", lazy.ConsoleClient.ENTERPRISE_PROFILE` ?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: yes, and when one is switching a lot between js and others, I feel it can be disturbing :(

browser/extensions/felt/content/FeltProcessParent.sys.mjs
Mossop: They both do the same thing and this way we can just use the same arguments whether we're using a test profile or a real profile.

lissyx (approved):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: I always worry about passing an incorrect path and rely on `-P name` to check for me :)

PR #63, opened 2025-09-30
FELT: Make sure --jsdebugger does not spawn a felt window

lissyx:


toolkit/library/felt/ffi/lib.rs
```diff
@@ -14,9 +17,50 @@ use env_logger;
...
+        Err(_) => false,
+    };
+
+    let found_felt_ui_arg = env::args()
+            .into_iter()
+            .map(|arg| {
+                trace!("arg={}", arg);
+                return arg;
+            })
+            .any(|arg| arg == "-feltUI");
```
Mossop: Do we also want to support `--feltUI` (or `/feltUI` on Windows)?

toolkit/library/felt/ffi/lib.rs
```diff
@@ -14,9 +17,50 @@ use env_logger;
...
+
+    let is_felt_ui = found_felt_ui_arg || found_felt_ui_env;
+    trace!("felt_init(): is_felt_ui={}", is_felt_ui);
+    unsafe { felt::IS_FELT_UI.store(is_felt_ui, Ordering::Relaxed); }
```
Mossop, line 37: Why does this need to be in an unsafe block?

toolkit/library/felt/ffi/lib.rs
```diff
@@ -14,9 +17,50 @@ use env_logger;
...
+            .into_iter()
+            .map(|arg| {
+                trace!("arg={}", arg);
+                return arg;
+            })
+            .any(|arg| arg == "-felt");
+
+    trace!("felt_init(): is_felt_browser={}", is_felt_browser);
+    unsafe { felt::IS_FELT_BROWSER.store(is_felt_browser, Ordering::Relaxed); }
+
```
Mossop, line 50: Can we add a check that we don't set both `IS_FELT_UI` and `IS_FELT_BROWSER`?

Mossop (commented):
I don't understand the unsafe blocks when interacting with the `AtomicBool`. Is this something related to being called from C++?

toolkit/library/felt/ffi/lib.rs
lissyx: Yes that's likely

toolkit/library/felt/ffi/lib.rs
lissyx: Because we manipulate `static mut`. I used an `AtomicBool` to make sure it's safe enough

lissyx: We should also be safe of multiple calls to `felt_init()` because `env_logger::init();` will panic in such case

toolkit/library/felt/ffi/lib.rs
lissyx: Yes, I was wondering of doing it

Mossop (approved):
This can land as-is and the suggestions be follow-ups

toolkit/library/felt/ffi/lib.rs
lissyx: Should be taken care of

lissyx:
> This can land as-is and the suggestions be follow-ups

Already fixed them

lissyx:
I'll merge asap once a green build is there so you can rebase for #61

toolkit/library/felt/ffi/lib.rs
```diff
@@ -14,9 +17,54 @@ use env_logger;
...
+                arg
+                    .replace("-", "")
+                    .replace("/", "")
+                    .to_lowercase() == "felt"
+            });
+
+    trace!("felt_init(): is_felt_browser={}", is_felt_browser);
+    unsafe { felt::IS_FELT_BROWSER.store(is_felt_browser, Ordering::Relaxed); }
+
+    assert!((!is_felt_browser && !is_felt_ui) || ((is_felt_browser || is_felt_ui) && (is_felt_browser != is_felt_ui)), "Cannot have both -fletUI and -felt args");
```
Mossop: s/-fletUI/feltUI/

Also isn't this just `assert!(!(is_felt_browser && is_felt_ui))`?

toolkit/library/felt/ffi/lib.rs
```diff
@@ -14,9 +17,54 @@ use env_logger;
...
+        Err(_) => false,
+    };
+
+    let found_felt_ui_arg = env::args()
+            .into_iter()
+            .any(|arg| {
+                arg
+                    .replace("-", "")
+                    .replace("/", "")
+                    .to_lowercase() == "feltui"
```
Mossop, line 32: I guess this would also allow `---felt----ui` but that's probably fine for now 😆

Mossop (approved):

toolkit/library/felt/ffi/lib.rs
lissyx: Indeed: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d624fe91991b3ea0b1af1e20d14eec1b

PR #64, opened 2025-09-30
Enable transparent use of primary password

beurdouche:
This PR implement transparent use of the primary password mechanism for users.
Upon start, the browser will use the `ConsoleTokenData` to fetch a Primary Password specific to the authenticated user on the console backend. If the primary password is not set, the browser will silently set it to the fetched value, and will request it to the server for each subsequent use. As a consequence, the user is protected without needing to interact.

### DIFF
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -21,6 +21,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   BrowserUtils: "resource://gre/modules/BrowserUtils.sys.mjs",
   BrowserUsageTelemetry: "resource:///modules/BrowserUsageTelemetry.sys.mjs",
   BrowserWindowTracker: "resource:///modules/BrowserWindowTracker.sys.mjs",
+  ConsoleClient: "resource://builtin-addons/felt/content/ConsoleClient.sys.mjs",
   ContentBlockingPrefs:
     "moz-src:///browser/components/protections/ContentBlockingPrefs.sys.mjs",
   ContextualIdentityService:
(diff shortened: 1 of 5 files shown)


browser/components/BrowserGlue.sys.mjs
```diff
@@ -1067,6 +1068,82 @@ BrowserGlue.prototype = {
...
+          if (!Services.prefs.getBoolPref("browser.enterprise.encryption.enabled", false)) {
+            console.log("EnterpriseEncryption disabled, skipping initialization");
+            return;
+          }
+
+          // Get the access token from the prefs
+          const access_token = Services.prefs.getStringPref("browser.policies.access_token");
+
+          // Get the primary secret from the console backend
+          const response = lazy.ConsoleClient.getPrimarySecret();
```
Mossop: This needs an await I think.

beurdouche: Thanks, that's the only line I couldn't test because the server side patch hasn't landed yet.
The rest of the code was tested against a different endpoint doing the fetch directly instead of using the ConsoleClient.

beurdouche:
```suggestion
          const response = await lazy.ConsoleClient.getPrimarySecret();
```

toolkit/components/passwordmgr/LoginHelper.sys.mjs
```diff
@@ -1688,10 +1688,11 @@ export const LoginHelper = {
       };
     }
 
-    // So there's a primary password. But since checkPassword didn't succeed, we're logged out (per nsIPK11Token.idl).
+    // So there's a primary password. But since checkPassword intentionally
+    // failed, we're logged out (per nsIPK11Token.idl).
```
gcp: > "since checkPassword intentionally failed, we're logged out"

The code this references got removed as part of this PR. This comment may be misleading or incorrect.

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1066,6 +1067,95 @@ BrowserGlue.prototype = {
...
+
+          // Initialize the variables
+          let pk11token;
+          let isPasswordValid;
+          let isPasswordNeeded;
+
+          // Check if the PK11 token requires a password before attempting to initialize and check it
+          try {
+            pk11token = tokenDB.getInternalKeyToken();
+            isPasswordNeeded = pk11token.needsUserInit || pk11token.needsLogin();
```
gcp: Is the logic for `needsUserInit` not reversed here? On a fresh profile this is true, so we set `isPasswordNeeded`, and then skip the initialization with `primary_secret` below, even though we'll end up setting an empty `""` password, see https://searchfox.org/firefox-main/source/toolkit/components/passwordmgr/crypto-SDR.sys.mjs#54

(If the init has run, that is, we inited with as password of `""`, needsUserInit is `false`, and `needsLogin()` is `false` I think it works correctly, but that just means the code is racy, and we probably want to assert we're never in the case above, rather than check `needsUserInit`)

browser/components/aboutlogins/tests/browser/browser_osAuthDialog.js
```diff
@@ -140,6 +140,77 @@ add_task(async function test_osAuth_shown_on_reveal_password() {
...
+    );
+    // Hide it again
+    revealCheckbox.click();
+  });
+
+  // Now that the PrP is already unlocked revealing a password should trigger
+  // the OS auth prompt.
+  let osAuthDialogShown = Promise.resolve();
+  if (OSKeyStore.canReauth()) {
+    osAuthDialogShown = forceAuthTimeoutAndWaitForOSKeyStoreLogin(true);
```
gcp: Needs to be `{ loginResult: true }` (see the definition in https://searchfox.org/firefox-main/source/browser/components/aboutlogins/tests/browser/head.js#255 and the other caller below in the same file).

So uhm does this test actually work?

toolkit/components/passwordmgr/LoginHelper.sys.mjs
```diff
@@ -1688,10 +1688,11 @@ export const LoginHelper = {
...
     try {
       // Relogin and ask for the primary password.
```
gcp: This comment is also wrong now given the change below?

gcp (changes requested):

browser/components/BrowserGlue.sys.mjs
beurdouche: To make it a bit more robust, I think the solution is to check for `needsUserInit` then if true we `initPassword`, if false we check `needsLogin`. If `needsLogin` returns false it means we have an empty password so we `changePassword` to `primary_secret`. If `needsLogin` true, we `checkPassword`.

gcp:
Superseded by https://github.com/mozilla/enterprise-firefox/pull/138

PR #77, opened 2025-10-07
Resolve most eslint failures

Mossop:
The bulk of this is correcting indentation and switching to use `Services.felt` however the linting did identify some issues with the use of `new Promise` that I've corrected.


1rneh (approved):
Lgtm, thanks for fixing the promise issues!

PR #81, opened 2025-10-08
FELT: Handle startup prefs

lissyx:


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -178,6 +179,28 @@ export class FeltProcessParent extends JSProcessActorParent {
       ];
     }
 
+    const prefsJsFile = PathUtils.join(profilePath, "user.js");
+    console.debug(`Felt: provisionning prefs.js: ${prefsJsFile}`);
```
Mossop: "provisioning"

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -178,6 +179,28 @@ export class FeltProcessParent extends JSProcessActorParent {
       ];
     }
 
+    const prefsJsFile = PathUtils.join(profilePath, "user.js");
```
Mossop: I'm trying to remember how user.js works. Does it just override prefs.js when loaded on startup? Or does it impact future modifications to those prefs as well?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -178,6 +179,28 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    startupPrefs?.forEach(pref => {
+      if (typeof pref[1] == "string") {
+        prefsJsContent += `\nuser_pref("${pref[0]}", "${pref[1]}");`;
+      } else {
+        prefsJsContent += `\nuser_pref("${pref[0]}", ${pref[1]});`;
+      }
+    });
+    prefsJsContent += "\n";
+
+    const writtenPrefsJsBytes = await IOUtils.writeUTF8(
```
Mossop: Doesn't look like we need this variable.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -178,6 +179,28 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    if (await IOUtils.exists(prefsJsFile)) {
+      prefsJsContent = await IOUtils.readUTF8(prefsJsFile);
+    }
+
+    const startupPrefs = (await lazy.ConsoleClient.getStartupPrefs()).prefs;
+    startupPrefs?.forEach(pref => {
+      if (typeof pref[1] == "string") {
+        prefsJsContent += `\nuser_pref("${pref[0]}", "${pref[1]}");`;
+      } else {
+        prefsJsContent += `\nuser_pref("${pref[0]}", ${pref[1]});`;
```
Mossop: We should verify we aren't writing something bad here. So we should only write if `Number.isInteger(pref[1])` or `typeof pref[1] == "boolean"`. We should also escape the strings, `JSON.stringify` should work.

Mossop (changes requested):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: So the test were written to use `user.js` at first, and I kept it because I was unsure, but looks like `prefs.js` works as expected. Verified by stopping pushing `marionette.port=0` from the test server's `startup` prefs endpoint and as expected we get marionette listening on port 2828, confirming that the pref was not set.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -53,10 +53,30 @@ export class FeltProcessParent extends JSProcessActorParent {
     Services.obs.addObserver(this.restartObserver, "felt-firefox-restarting");
   }
 
-  startFirefox(accessToken = null) {
+  sanitizedPrefs(prefs) {
+    let sanitized = [];
+    prefs?.forEach(pref => {
+      const name = pref[0];
```
Mossop: This needs to be stringified as well

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -53,10 +53,30 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      switch (typeof pref[1]) {
+        case "string":
+          pref[1] = JSON.stringify(pref[1]);
+        case "boolean":
+        case "number":
```
Mossop, line 70: This allows non-integers which prefs do not.

Mossop (changes requested):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Non-integers? Floats? `user_pref("name", 1.5);` is incorrect?

Mossop: Yeah, prefs only support integers, see https://searchfox.org/firefox-main/rev/2d38cd5bd7d7958137dd5b4f97aef2f7185a0eb1/modules/libpref/parser/src/lib.rs#17

lissyx: weird: https://searchfox.org/firefox-main/rev/2d38cd5bd7d7958137dd5b4f97aef2f7185a0eb1/modules/libpref/nsIPrefBranch.idl#89-104

lissyx: https://searchfox.org/firefox-main/rev/2d38cd5bd7d7958137dd5b4f97aef2f7185a0eb1/modules/libpref/init/StaticPrefList.yaml#33,45-50

Mossop: Right, see https://searchfox.org/firefox-main/rev/2d38cd5bd7d7958137dd5b4f97aef2f7185a0eb1/modules/libpref/nsIPrefBranch.idl#91-92. Floats are encoded in the prefs file as a string. So I guess maybe we should do that?

lissyx: nice i can hack that

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: ok but it's messing with the `Services.felt.sendStringPreference(name, value);` calls and others

lissyx: ends up:
```
user_pref("\"browser.policies.live_polling_freq\"", 500);
user_pref("\"browser.sessionstore.restore_on_demand\"", false);
user_pref("\"browser.sessionstore.resume_from_crash\"", false);
```

Mossop (approved):

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -87,6 +87,7 @@ export const ConsoleClient = {
       SSO: `${this.consoleAddr}/sso/login?target=browser`,
       SSO_CALLBACK: `${this.consoleAddr}/sso/callback?*`,
       REDIRECT_AFTER_SSO: `${this.consoleAddr}/redirect_after_sso`,
+      STARTUP_PREFS: `${this.consoleAddr}/api/browser/hacks/startup`,
       DEFAULT_PREFS: `${this.consoleAddr}/api/browser/hacks/default`,
```
1rneh, line 91: Can you add a comment about the difference of `STARTUP_PREFS` and `DEFAULT_PREFS`?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -53,10 +53,43 @@ export class FeltProcessParent extends JSProcessActorParent {
     Services.obs.addObserver(this.restartObserver, "felt-firefox-restarting");
   }
 
-  startFirefox(accessToken = null) {
+  sanitizedPrefs(prefs) {
```
1rneh: Suggestion: Rename to `sanitizePrefs`?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -53,10 +53,43 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    let sanitized = [];
```
1rneh, line 57: We could use a `Map` here.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -53,10 +53,43 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      sanitized.push([name, value]);
+    });
+    return sanitized;
+  }
+
+  async startFirefox(accessToken = null) {
     this.restartReported = false;
     Services.cpmm.sendAsyncMessage("FeltParent:FirefoxStarting", {});
     this.firefox = this.startFirefoxProcess();
+    const defaultPrefs = (await lazy.ConsoleClient.getDefaultPrefs()).prefs;
```
1rneh: Let's also sanitize `defaultPrefs`.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -178,6 +215,20 @@ export class FeltProcessParent extends JSProcessActorParent {
       ];
     }
 
+    let prefsJsFile = PathUtils.join(profilePath, "prefs.js");
```
1rneh:
```suggestion
    const prefsJsFile = PathUtils.join(profilePath, "prefs.js");
```

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: Edit: Maybe add the comments when you call the `ConsoleClient` APIs and not in `ConsoleClient`.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -72,11 +105,13 @@ export class FeltProcessParent extends JSProcessActorParent {
             accessToken
           );
         }
-        const json = await lazy.ConsoleClient.getDefaultPrefs();
-        json.prefs.forEach(pref => {
+
+        defaultPrefs.forEach(pref => {
```
1rneh: Can we get the prefs just before they’re used, instead of defining them outside the `then` callback?
```suggestion
        const { prefs } = await lazy.ConsoleClient.getDefaultPrefs();
        prefs.forEach(pref => {
```

browser/extensions/felt/content/FeltProcessParent.sys.mjs
1rneh: **Edit:** Discard this comment. I misread what sanitizePrefs does.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -178,6 +215,20 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      prefsJsContent = await IOUtils.readUTF8(prefsJsFile);
+    }
+
+    const startupPrefs = (await lazy.ConsoleClient.getStartupPrefs()).prefs;
+    this.sanitizedPrefs(startupPrefs).forEach(pref => {
+      prefsJsContent += `\nuser_pref(${pref[0]}, ${pref[1]});`;
+    });
+    prefsJsContent += "\n";
+
+    await IOUtils.writeUTF8(prefsJsFile, prefsJsContent);
```
1rneh, line 233: Do you know if we write the latest state of prefs into `prefs.js` on shutdown?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -53,10 +53,43 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+          if (!Number.isInteger(pref[1])) {
+            value = `"${pref[1]}"`;
+          } else {
+            value = Number(pref[1]);
+          }
+          break;
+
+        default:
+          console.warn(`Pref ${pref[0]} with value '${pref[1]}' not valid`);
+          break;
```
1rneh, line 81:
```suggestion
          continue;
```
Otherwise we'll push the invalid pref to `sanitized`.

testing/enterprise/felt_tests.py
```diff
@@ -108,18 +107,27 @@ def do_GET(self):
...
                 }
             )
+        elif path == "/api/browser/hacks/startup":
+            # Browser prefs that needs to be set in the prefs.js file
+            m = json.dumps(
+                {
+                    "prefs": [
+                        ["devtools.browsertoolbox.scope", "everything"],
+                        ["marionette.port", 0],
+                        ["browser.felt.test_float", 1.5],
```
1rneh, line 127: Can you also add a pref with a boolean value? Then we have all sanitized types covered.

1rneh (approved):
Thank you! Only the `continue` change in `sanitizePrefs()` is blocking. 

But I have one more general question: Why do we distinguish between default and startup prefs? Couldn't they all be written into `prefs.js`?

lissyx:
> Thank you! Only the `continue` change in `sanitizePrefs()` is blocking.
> 
> But I have one more general question: Why do we distinguish between default and startup prefs? Couldn't they all be written into `prefs.js`?

I'd say there is no real need in the current form, but likely we want to be able to send live pref changes if required for things that cannot be covered by policies?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: @1rneh you mean on browser shutdown? I am unsure but I can verify, what are you worried about?

1rneh: I was worried that we accumulate duplicates long-term and grow the file on every startup, if `prefs.js` doesn't get overridden on shutdown.

lissyx: oh its rewritten for sure, you can verify with the `marionette.port` value after running the tests, it's not `0` anymore

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: What do we win?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: The idea here was to move requests out of the startup code to make things more resilient

testing/enterprise/felt_tests.py
lissyx: Sure

PR #83, opened 2025-10-08
Add lint checks for files changed in a PR

Mossop:


Mossop:
I split the lints into three jobs because clippy and cargo audit seem to take a bunch of time. We can probably improve this later with some caching (`~/.mozbuild` and `node_modules` are good first candidates).

.github/workflows/enterprise.yml
```diff
@@ -663,3 +663,47 @@ jobs:
...
+          fetch-depth: 10
+
+      - name: "Run lint checks on PR changes"
+        run: |
+          # This is needed for linting anyway but it has the side effect of making mach create its
+          # state directories so when we list the linters we don't get log messages for that.
+          ./mach eslint --setup
+          # List all the linters except those we run in separate jobs.
+          LINTERS=$(./mach lint --list | grep -E -v 'clang-tidy|cargo-audit|clippy' | awk 'NF {printf "-l %s ", $0}')
+          ./mach lint --outgoing ${{ github.event.pull_request.base.sha }} $LINTERS
```
lissyx: `--warnings` as well?

.github/workflows/enterprise.yml
```diff
@@ -663,3 +663,47 @@ jobs:
...
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 10
+
+      - name: "Run clippy on PR changes"
+        run: |
+          ./mach lint --outgoing ${{ github.event.pull_request.base.sha }} -l clippy
```
lissyx: `--warnings` as well?

.github/workflows/enterprise.yml
```diff
@@ -663,3 +663,47 @@ jobs:
...
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 10
+
+      - name: "Run cargo-audit on PR changes"
+        run: |
+          ./mach lint --outgoing ${{ github.event.pull_request.base.sha }} -l cargo-audit
```
lissyx: `--warnings` as well?

lissyx (approved):
It looks good, but I really think we should at least make use of github caching on `~/.mozbuild` (and/or put it in some other place). The only thing that may be blocking would be the size of the directory?

Mossop:
Ok yeah with caching there doesn't seem any need to separate out the lint checks.

.github/workflows/enterprise.yml
```diff
@@ -663,3 +663,28 @@ jobs:
...
+
+      - name: Cache lint dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo
+            ~/.npm
+            ~/.mozbuild
+            node_modules
+          key: lint-${{ runner.os }}-${{ hashFiles('package-lock.json', 'Cargo.lock') }}
```
lissyx, line 684: TIL we have `package-lock.json` in tree

lissyx: Overall cache is 1.3GB

lissyx (approved):

lissyx:
> Ok yeah with caching there doesn't seem any need to separate out the lint checks.

I think it's still nice to have the per-linter separation, but it's up to you

Mossop:
> > Ok yeah with caching there doesn't seem any need to separate out the lint checks.
> 
> I think it's still nice to have the per-linter separation, but it's up to you

The bulk of the workflow is now spent in checking out the tree so if you split up the linters it really increases the amount of runner time we use.

PR #86, opened 2025-10-09
Default to a different MOZ_APP_BASENAME

Mossop:
This moves where the profiles and profiles database are stored giving separation between the enterprise profiles and the normal Firefox profiles.

### DIFF
--- a/.github/workflows/enterprise.yml
+++ b/.github/workflows/enterprise.yml
@@ -193,7 +193,7 @@ jobs:
 
       - name: "test firefox start"
         run: |
-          RUNTIME_VERSION=$(${{ steps.unpack.outputs.runtime_path }} --version | awk '{ print $3 }')
+          RUNTIME_VERSION=$(${{ steps.unpack.outputs.runtime_path }} --version | awk '{ print $4 }')
           sed -e "s/#RUNTIME_VERSION#/${RUNTIME_VERSION}/g" < firefox_start.json.in > firefox_start.json
           mkdir -p ./profiles/
           . venv_tests/bin/activate
@@ -617,7 +617,7 @@ jobs:
 
       - name: "test firefox start"
         run: |
-          RUNTIME_VERSION=$("${{ steps.attach.outputs.runtime_path }}" --version | awk '{ print $3 }')
+          RUNTIME_VERSION=$("${{ steps.attach.outputs.runtime_path }}" --version | awk '{ print $4 }')
           sed -e "s/#RUNTIME_VERSION#/${RUNTIME_VERSION}/g" < firefox_start.json.in > firefox_start.json
           mkdir -p ./profiles/
           . venv_tests/bin/activate
--- a/browser/extensions/moz.build
+++ b/browser/extensions/moz.build
@@ -14,7 +14,7 @@ DIRS += [
     "newtab",
 ]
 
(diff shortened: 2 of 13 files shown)


build/mozconfig.common.enterprise
```diff
@@ -3,6 +3,8 @@ ac_add_options --enable-bootstrap
 ac_add_options --enable-application=browser
 ac_add_options --enable-warnings-as-errors
 ac_add_options --enable-felt
+ac_add_options --with-app-basename="Firefox Enterprise"
+ac_add_options --with-app-name=firefox
```
lissyx: Shouldn't we rather handle that by `browser/moz.configure` and be dependant on `--enable-felt` (and/or introduce a `--enable-enterprise` like we discussed a few times with @gcp ?)

Mossop: We can do yes

Mossop:
Wasn't sure if you also wanted to switch the `MOZ_WIDGET_FELT` define but I think it makes sense to.

browser/extensions/moz.build
```diff
@@ -14,7 +14,7 @@ DIRS += [
     "newtab",
 ]
 
-if CONFIG["MOZ_WIDGET_FELT"]:
+if CONFIG["MOZ_ENTERPRISE"]:
```
lissyx, line 17: @gcp What do you think? We discussed about renaming `MOZ_WIDGET_FELT` to `MOZ_ENTERPRISE` and you had opinions that having both would make sense?

gcp: It's easy to make ``--enable-enterprise`` default `MOZ_WIDGET_FELT` on. I don't really remember what cases I had in mind where it makes sense to separate them. Maybe for testing features that can be tested without console?

Mossop: `MOZ_WIDGET_FELT` is a bit of an odd naming since this isn't anything to do with the widget code which is partly why I figured I would change it when moving from `--enable-felt` to `--enable-enterprise`. Unless we have a use case in mind now I would suggest we just stick with a single define, if we need to separate later I hope we won't have many instances of it in the tree anyway.

.github/workflows/enterprise.yml
```diff
@@ -193,7 +193,7 @@ jobs:
 
       - name: "test firefox start"
         run: |
-          RUNTIME_VERSION=$(${{ steps.unpack.outputs.runtime_path }} --version | awk '{ print $3 }')
+          RUNTIME_VERSION=$(${{ steps.unpack.outputs.runtime_path }} --version | awk '{ print $4 }')
```
lissyx, line 196: Why do we need this change?

Mossop: The version string previously printed as "Mozilla Firefox 145.0a1" but with the change to the basename it now prints as "Mozilla Firefox Enterprise 145.0a1" so the version number is the 4th string.

browser/extensions/moz.build
lissyx: both are fine for me, I'll just have to rebase my changes to use `MOZ_ENTERPRISE` if we want to move to that. Do we nuke `MOZ_WIDGET_FELT` then?

Mossop: I think so

lissyx (approved):

PR #91, opened 2025-10-13
FELT: Split IPC client and thread

lissyx:


toolkit/library/felt/ffi/lib.rs
```diff
@@ -58,14 +58,39 @@ pub extern "C" fn is_felt_browser() -> bool {
     felt::IS_FELT_BROWSER.load(Ordering::Relaxed)
 }
 
+pub static FELT_CLIENT: Mutex<Option<felt::FeltClientThread>> = Mutex::new(None);
```
lissyx, line 61: a bit meh, but how worse is it compared to sharing the pointer to C++ side and passing it around back? especially since we'd end up in some global var at some point ... ?

lissyx:
Rationale for doing that is being able to avoid racy behavior between our startup code and storage requirements when doing https://github.com/mozilla/enterprise-firefox/pull/88

toolkit/library/felt/src/lib.rs
```diff
@@ -128,114 +129,124 @@ impl FeltIpcClient {
...
+            Err(())
         }
     }
 
     pub fn start_thread(&self) -> nserror::nsresult {
         trace!("FeltClientThread::start_thread()");
+        trace!("FeltClientThread::start_thread(): creating thread");
+        if let Ok(thread) = moz_task::create_thread("felt_client").map_err(|_| {
+            trace!("FeltClientThread::start_thread(): felt_client thread error");
+        }) {
```
Mossop: Suggested alternative:

```rust
let Ok(thread) = moz_task::create_thread("felt_client") else {
    trace!("FeltClientThread::start_thread(): felt_client thread error");
    return NS_ERROR_FAILURE;
}

/// ...

NS_OK
```

toolkit/library/felt/ffi/lib.rs
```diff
@@ -58,14 +58,39 @@ pub extern "C" fn is_felt_browser() -> bool {
...
+}
+
+#[no_mangle]
+pub extern "C" fn firefox_felt_connection_start_thread() -> () {
+    let mut guard = FELT_CLIENT.lock().expect("Could not get lock");
+    match guard.take() {
+        Some(client) => {
+            trace!("firefox_connect_to_felt(): connected, starting thread");
+            client.start_thread();
+            // client.wait_for_startup_requirements();
```
Mossop, line 88: Is this intentionally skipped?

Mossop (approved):

toolkit/library/felt/ffi/lib.rs
lissyx: Indeed, and to be removed when I'm certain there is no race.

PR #94, opened 2025-10-15
FELT: Create a contentPrincipal with privateBrowsingID for SSO

lissyx:


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -267,7 +267,13 @@ export class FeltProcessParent extends JSProcessActorParent {
     );
     switch (message.name) {
       case "FeltChild:StartFirefox":
-        this.startFirefox(message.data?.access_token ?? null);
+        const cookies = this.getAllCookies();
+        console.debug(`Collected cookies:`, cookies);
```
lissyx: To be removed when merging

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -267,7 +267,13 @@ export class FeltProcessParent extends JSProcessActorParent {
     );
     switch (message.name) {
       case "FeltChild:StartFirefox":
-        this.startFirefox(message.data?.access_token ?? null);
+        const cookies = this.getAllCookies();
```
lissyx: we collect there because I ran into races where sometimes it was already too late and cookies were not here

browser/extensions/felt/content/window.js
```diff
@@ -38,8 +38,12 @@ function connectToConsole(email) {
   );
 
   console.debug("Load SSO Page: ", SOURCE_URI);
+  const uri = Services.io.newURI(lazy.ConsoleClient.ssoUri);
+  console.debug(`FeltExtension: creating contentPrincipal with privateBrowsingId=${lazy.ConsoleClient.PRIVATE_BROWSING_ID}`);
+  const contentPrincipal = Services.scriptSecurityManager.createContentPrincipal(uri, { privateBrowsingId: lazy.ConsoleClient.PRIVATE_BROWSING_ID });
```
lissyx: SystemPrincipal was bad whatever we want to think about private browsing enabling

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -66,6 +66,7 @@ class ConsoleTokenData {
 
 export const ConsoleClient = {
   ENTERPRISE_PROFILE: "enterprise-profile",
+  PRIVATE_BROWSING_ID: 1,
```
1rneh: The `ConsoleClient` shouldn't keep any profile or private browsing information. It's simply handles requests to the console and token lifecycle. 
Could you move them to e.g. `FeltProcessParent` or create a `FeltCommon` file to collect these properties?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
1rneh: The cookies get removed once we close the window on `FeltParent:FirefoxStarting` (for non test environment). It's racy because we send it out before retrieving the cookies. Couldn't we simply retrieve them in `startFirefox` before sending `FeltParent:FirefoxStarting` instead of passing them to `startFirefox`?

browser/extensions/felt/content/window.js
```diff
@@ -38,8 +38,12 @@ function connectToConsole(email) {
   );
 
   console.debug("Load SSO Page: ", SOURCE_URI);
+  const uri = Services.io.newURI(lazy.ConsoleClient.ssoUri);
```
1rneh: Use `SOURCE_URI` here since it contains the email in the query string.

browser/extensions/felt/content/window.js
1rneh: Nice catch! This shouldn't have been a system principal in the first place...

1rneh (changes requested):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: We might, I need to verify if it's not racy enough though.

browser/extensions/felt/content/window.js
lissyx: Well that's something I am not sure we want to do? You prefer to have a unique URI there? I need to check if the contentPrincipal will be OK/consider the query string

browser/extensions/felt/content/window.js
lissyx: Well there was a moment it made sense, but we should have fixed it earlier indeed. My bad for missing it during reviews especially since I remember it crossed my mind.

browser/extensions/felt/content/ConsoleClient.sys.mjs
lissyx: I was looking for an easy way to share the information between the `window.js` code and `FeltProcessParent` without hardcoding `1` everywhere, but I can introduce a `FeltCommon`

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -90,6 +91,13 @@ export class FeltProcessParent extends JSProcessActorParent {
 
   async startFirefox(accessToken = null) {
     this.restartReported = false;
+    const ssoCollectedCookies = this.getAllCookies();
+    console.debug(`Collected cookies: ${ssoCollectedCookies.length}`);
+    // When a restart was reported we assume cookies were stored properly on the
+    // browser side?
+    if (!ssoCollectedCookies.length && !this.restartReported) {
```
lissyx: @1rneh I need your opinion there, because when the browser performs a restarts handled with felt (so technically not a restart, a close that felt will handle and perform a second start) then the cookies would have disappeared. With this I assume it's fine and we can live with existing saved cookies, but ... ?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -90,6 +91,13 @@ export class FeltProcessParent extends JSProcessActorParent {
 
   async startFirefox(accessToken = null) {
     this.restartReported = false;
+    const ssoCollectedCookies = this.getAllCookies();
```
lissyx: hm i can still hit a race on opt build for example, as on CI

1rneh (approved):
Thank you!

PR #97, opened 2025-10-16
CI: Add GCP-EU repackage

lissyx:


browser/branding/enterprise/pref/firefox-felt.js
```diff
@@ -0,0 +1,8 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global pref */
+
+// This file contains felt-specific prefs.
+pref("browser.felt.console", "https://console.enterfox.eu/");
```
julienw: Could be good to have a comment for the pref itself

lissyx: What would you want to see there?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -98,9 +105,7 @@ export const ConsoleClient = {
   },
 
   get ssoCallbackUri() {
-    return isTesting()
-      ? Services.prefs.getStringPref(PREFS.SSO_CALLBACK, "")
-      : this.ENDPOINTS.SSO_CALLBACK;
+    return this.ENDPOINTS.SSO_CALLBACK;
```
1rneh, line 111: I think the `browser.felt.*` pref names are outdated and no longer tied to Felt. Can we rename them to something like e.g., `enterprise.console.address`, `enterprise.is_testing`)

browser/branding/branding-common.mozbuild
```diff
@@ -35,3 +35,8 @@ def FirefoxBranding():
             "default48.png",
             "default64.png",
         ]
+
+    if CONFIG["MOZ_ENTERPRISE"]:
+        JS_PREFERENCE_FILES += [
+            "pref/firefox-felt.js",
```
1rneh: Same here: It's not really felt related, so maybe `firefox-enterprise.js` suits the use case better?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -77,15 +76,23 @@ export const ConsoleClient = {
...
-      : "https://console.enterfox.eu";
+    return Services.prefs.getStringPref(
+      PREFS.CONSOLE_ADDRESS,
+      "https://console.enterfox.eu"
+    );
+  },
+
+  // The matcher https://searchfox.org/firefox-main/source/toolkit/components/extensions/MatchPattern.cpp#370-384
+  // expects no port
+  get consoleAddrWithoutPort() {
```
1rneh: Could we call the function `consoleAddrMatchingCallback` and then explain inside the function that we're dropping the port because the matcher doesn't expect a port and would otherwise not spin up our actors? I don't think this is otherwise clear why we are not simply using `this.consoleAddr`.

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -11,10 +11,9 @@
 const PREFS = {
   CONSOLE_ADDRESS: "browser.felt.console",
   IS_TESTING_ENVIRONMENT: "browser.felt.is_testing",
-  SSO_CALLBACK: "browser.felt.matches",
```
1rneh, line 14: Nice! Happy about every static pref we get rid of!

1rneh (approved):

browser/extensions/felt/content/ConsoleClient.sys.mjs
lissyx: Should be better now, let me know

PR #101, opened 2025-10-20
ConsoleClient

1rneh:
Felt injects the refresh token into Firefox. Once Firefox is launched it retrieves the refresh token to refresh the session and obtain a valid access token. For now Firefox' ConsoleClient handles all communication with the remote console and the token lifecycle and doesn't sync the token data back to Felt.


toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -927,15 +922,14 @@ class WindowsGPOPoliciesProvider {
       if (Cu.isInAutomation || isXpcshell) {
         try {
           regLocation = Services.prefs.getStringPref(PREF_ALTERNATE_GPO);
-        } catch (e) {}
+        } catch (e) { }
```
lissyx: Can we split that kind of lint changes in its own patch and land it on m-c ?

toolkit/components/enterprise/nsIConsoleClient.idl
```diff
@@ -0,0 +1,80 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "nsISupports.idl"
+#include "nsIURI.idl"
+
+[scriptable, uuid(6a568972-cc91-4bf5-963e-3768f3319b8a)]
+interface nsIEnterprisePolicies : nsISupports
```
lissyx: I dont see that used anywhere? Also, it should be named `nsIEnterprisePolicies.idl` if the interface name is `nsIEnterprisePolicies`

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -136,41 +303,138 @@ export const ConsoleClient = {
...
-  sendDevicePosture() {},
-
-  onConsoleTokenDataReceived(tokenData) {
-    console.debug("ConsoleClient: onReceivedConsoleToken");
+  /**
+   * Populates in-memory token state at the moment that we receive the 
+   * token data on initial authentication with the console. 
+   * @param {Object} tokenData { access_token, refresh_token, expires_in, token_type }
+   */
+  applyTokenData(tokenData) {
```
lissyx: nit: I dont really like the `apply` and would rather expect `ensureTokenData` ? But it's really personnal preference

browser/extensions/felt/api.js
```diff
@@ -50,22 +59,29 @@ this.felt = class extends ExtensionAPI {
...
 
-    if (Services.felt.isFeltUI()) {
-      this.registerChrome();
-      this.registerActors();
-      this.showWindow();
-      Services.ppmm.addMessageListener("FeltParent:FirefoxNormalExit", this);
-      Services.ppmm.addMessageListener("FeltParent:FirefoxAbnormalExit", this);
-      Services.ppmm.addMessageListener("FeltParent:FirefoxStarting", this);
-      Services.ppmm.addMessageListener("FeltParent:FirefoxStarted", this);
+    if (!Services.felt.isFeltUI()) {
```
lissyx: nit: We have `isFeltBrowser()` that would be more readable than a negation

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -3,30 +3,55 @@
...
+ * ConsoleClient takes care of all communication with the remote enterprise console.
  */
 
-const PREFS = {
+export const PREFS = {
   CONSOLE_ADDRESS: "browser.felt.console",
   IS_TESTING_ENVIRONMENT: "browser.felt.is_testing",
   SSO_CALLBACK: "browser.felt.matches",
+  // Temporary pref to share refresh token between Felt and Firefox
+  REFRESH_TOKEN: "enterprise.console.refresh_token",
```
lissyx: nit: we should make sure we verify that we dont leak those data to child processes

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -3,30 +3,55 @@
...
 };
 
+const TOKEN_EXPIRY_SKEW_MS = 5 * 60_000;
```
lissyx: Any reason this is not on FeltCommon?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -3,30 +3,55 @@
...
+    accessToken,
+    refreshToken,
+    expiresInS,
+    tokenType,
+    issuedAtMs = Date.now()
+  ) {
     this._accessToken = accessToken;
     this._refreshToken = refreshToken;
-    this._expiresIn = expiresIn;
+    this._expiresInMs = expiresInS * 1000;
```
lissyx: I'm wondering how much relevant that is, expiring in seconds is likely good enough right?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+  }
+
+  isExpiringSoon() {
+    return Date.now() + TOKEN_EXPIRY_SKEW_MS >= this.expiresAtMs;
+  }
 }
 
 export const ConsoleClient = {
-  ENTERPRISE_PROFILE: "enterprise-profile",
+  SHARED_READY_KEY: "console-client-ready",
```
lissyx: Since for now we share that information with EnterprisePoliciesParent (we should really decouple as fast as we can but that's not for this PR), it might be worth a comment to say it there (it might be mroe obvious there becaue we `lazy.ConsoleClient`) and here.

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+  CONSOLE_CLIENT_READY: "ConsoleClient:Ready",
 
-  _consoleTokenData: null,
+  /**
+   * Ensures ConsoleClient is all setup for the communication with the remote console.
+   * - If no refresh token is present, registers an observer
```
lissyx: I would have assumed from the name of the method that when we exit it, then we have everything ready. Here it sounds like some other step is required ?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+
+    console.debug("ConsoleClient:Ready");
+    Services.ppmm.sharedData.set(this.SHARED_READY_KEY, true);
+    Services.obs.notifyObservers(null, this.CONSOLE_CLIENT_READY);
+    Services.prefs.removeObserver(PREFS.REFRESH_TOKEN, this);
+  },
+
+  observe(subject, topic, data) {
+    if (topic === "nsPref:changed" && data === PREFS.REFRESH_TOKEN) {
+      (async () => {
```
lissyx: is this just to avoid leaking `async` requirement that we call it this way?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+  /**
+   * Ensures ConsoleClient is all setup for the communication with the remote console.
+   * - If no refresh token is present, registers an observer
+   * - If present, refreshes the session, and notifies ConsoleClient consumers
+   * @throws {ReauthRequiredError} If the session refresh with the refresh token from Felt fails.
+   */
+  async ensureReadiness() {
+    console.debug("ConsoleClient: ensureReadiness");
+    if (!Services.prefs.getStringPref(PREFS.REFRESH_TOKEN, "")) {
+      Services.prefs.addObserver(PREFS.REFRESH_TOKEN, this);
```
lissyx: Maybe we want to keep track of the calls here to ensure we have a balanced number of calls to add/remove observers, otherwise we risk leaking/keeping references that may be problematic.

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+      await this._refreshSession(refresh_token);
+    } catch (e) {
+      throw new ReauthRequiredError(
+        "ConsoleClient: Failed to refresh initial session with the refresh token from the prefs.",
+        "REFRESH_FAILED", { cause: e }
+      );
+    }
+
+    console.debug("ConsoleClient:Ready");
+    Services.ppmm.sharedData.set(this.SHARED_READY_KEY, true);
```
lissyx: This gives me ideas :)

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+    const u = new URL(url, this.consoleAddr);
+    return u.origin === this.consoleOrigin;
+  },
+
+  /**
+   * Performs an authenticated fetch against console endpoints.
+   * Attaches an Authorization header when targeting the console origin.
+   * On 401/403, attempts a single silent refresh and one retry.
+   * 
+   * @param {string} url (from ENDPOINTS)
```
lissyx: Since we control everything, why having full url and not just path/query string components? This way we dont have to check the origin etc.

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+   * @throws {ReauthRequiredError} If authorization remains invalid after refresh.
+   * @returns {Promise<Response>} Fetch response 
+   */
+  async _authorizedFetch(url) {
+    console.debug("ConsoleClient: _authorizedFetch", url);
+
+    await this._ensureValidAccessToken();
```
lissyx: So this may throw and we rethrow and it's up to the caller to handle?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+
+    let res;
+    try {
+      res = await fetch(url, { headers });
+    } catch (e) {
+      console.error(`ConsoleClient.fetch: Request failed for ${url}`, e);
+      throw e;
+    }
+
+    if (res?.status === 401 || res?.status === 403) {
```
lissyx: We should handle other cases, e.g., name resolution failure? other error codes? (5xx, 4xx)?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -45,70 +70,212 @@ class ConsoleTokenData {
...
+      }
+      res = await fetch(url, { headers: retryHeaders });
+    }
+
+    if (res?.status === 401 || res?.status === 403) {
+      console.warn(
+        `ConsoleClient: ${res.status} — Still unauthorized after token refresh attempt`
+      );
+      this._clearTokenData();
+      throw new ReauthRequiredError(
```
lissyx: I am wondering if it's right to use the same class even with `STILL_UNAUTHORIZED`. If we are still unauthorized after trying to auth, we would have to assume that either credentials are plain wrong OR there's something very bad. But I think it would deserve a distinct error, because we are past the "re-auth required" point here.

Maybe `InvalidAuthError` ?

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -136,41 +303,138 @@ export const ConsoleClient = {
...
+    if (!this.tokenData?.refreshToken && !token) {
+      this._clearTokenData();
+      throw new ReauthRequiredError("No refresh token available", "MISSING_REFRESH_TOKEN");
+    }
+
+    if (this._refreshPromise) {
+      return this._refreshPromise;
+    }
+
+    this._refreshPromise = (async () => {
```
lissyx: Is this really a promove we are returning this way? Constructing it this way is a bit unclear.

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -136,41 +303,138 @@ export const ConsoleClient = {
...
+        // Sync token data with felt?
+      } catch (e) {
+        this._clearTokenData();
+        throw new ReauthRequiredError((e?.message || "Token refresh failed"), "REFRESH_FAILED", { cause: e });
+      } finally {
+        this._refreshPromise = null;
+      }
+    })();
+
+    return this._refreshPromise;
```
lissyx: It might be equivalent (but I have a doubt) but at least I think it's more clear this way, especially on how to handle resolution or rejection of the promise

```
return new Promise((resolve, reject) => {
 ... body of this._refreshPromise ...
});

File: browser/extensions/felt/content/ConsoleClient.sys.mjs [hunk 19]
```diff
@@ -136,41 +303,138 @@ export const ConsoleClient = {
...
+  },
+
+  /**
+   * Clears persisted and in-memory token data and readiness state.
+   */
+  _clearTokenData() {
```
lissyx: I'm wondering if this should be tied to a shutdown observer as well, out of safety.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -745,43 +740,44 @@ class RemotePoliciesProvider {
...
+            return
+          }
+          this._restartPolling()
         }
         break;
-
       case "xpcom-shutdown":
         this._stopPolling();
+        try {
+          Services.prefs.removeObserver(this.POLLING_FREQUENCY_PREF, this);
```
lissyx: Thanks!

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -745,43 +740,44 @@ class RemotePoliciesProvider {
...
+          Services.prefs.removeObserver(this.CONSOLE_CLIENT_READY, this);
```
lissyx: This is wrong, `Services.obs.removeObserver` should be and not `Services.prefs`. And while it may be a good idea to cleanup in xpcom shutdown, I think it should be removed earlier.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -834,7 +827,27 @@ class RemotePoliciesProvider {
       });
   }
 
+  _startPollingOnceReady() {
+    console.debug("EnterprisePoliciesParent: _startPollingOnceReady")
+    const shared = Services.ppmm.sharedData;
+    const isReadyToPollPolicies = shared.has(this.SHARED_READY_KEY) && shared.get(this.SHARED_READY_KEY) === true
+    if (!isReadyToPollPolicies) {
+      console.debug("EnterprisePoliciesParent: Adding observer for ", this.CONSOLE_CLIENT_READY);
+      Services.obs.addObserver(this, this.CONSOLE_CLIENT_READY);
```
lissyx: We should remove this observer as soon as console is ready in fact, so when we start the poller below in that method?

browser/extensions/felt/api.js
lissyx: Can we use `Services.felt.isFeltBrowser()` ? Maybe we need to rename in case it looks a bit weird.

browser/extensions/felt/api.js
```diff
@@ -50,22 +59,29 @@ this.felt = class extends ExtensionAPI {
...
+      // Load ConsoleClient to ensure the refresh token observer is registered, 
+      // if the token is not already present.
+      const { ConsoleClient } = ChromeUtils.importESModule(
+        "chrome://felt/content/ConsoleClient.sys.mjs"
+      );
+      (async () => {
```
lissyx: Any specific reason for making this construction? I assume it's for avoiding leaking the async requirement?

lissyx (changes requested):
It looks good but there's a few ironing required I think.

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -136,41 +303,138 @@ export const ConsoleClient = {
...
+          );
+        }
+
+        const t = await res.json();
+        this.tokenData = new ConsoleTokenData(
+          t.access_token,
+          t.refresh_token,
+          t.expires_in,
+          t.token_type,
+          Date.now()
```
lissyx: We need to be careful as this is dependant on system's date, if we have any kind of changes there this would break us

browser/extensions/felt/api.js
```diff
@@ -50,22 +59,29 @@ this.felt = class extends ExtensionAPI {
...
+        try {
+          await ConsoleClient.ensureReadiness();
+        } catch (e) {
+          console.error("ConsoleClient ensureReadiness failed", e);
+        }
+      })();
+      return;
```
fiji-flo: That should do the same.
```suggestion
      ConsoleClient.ensureReadiness().catch(e =>
        console.error("ConsoleClient ensureReadiness failed", e)
      );
      return;
```

browser/extensions/felt/content/ConsoleClient.sys.mjs
```diff
@@ -136,41 +303,138 @@ export const ConsoleClient = {
...
+
+  /**
+   * Refreshes the session using a refresh token.
+   * Uses the provided token if given; otherwise the stored token.
+   * Serializes concurrent refreshes via an internal promise.
+   * @param {string|null} token refresh token
+   * @throws {ReauthRequiredError} If the refresh fails.
+   * @returns {Promise}
+   */
+  async _refreshSession(token = null) {
```
fiji-flo: I gave feedback in person. The gist is we return `Promise<void> | null` and that should be changed to `Promise<void>` which needs a new property to manage the state.

fiji-flo (commented):
Part 1.

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: I'll move it inside the `ConsoleTokenData` class since it's only used there.

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: Let me convert all ms values to seconds.

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: Such as? :)

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: For now yes. The idea was that the ConsoleClient shouldn't be responsible to show any reauthentication UI. Also we have no plan yet how to do reauthentication, so this is a temporary state.

1rneh: **Update:** I'll change this to only log the error for now and call some reauthentication method which we'll then need to implement once we decide what we do/show.

browser/extensions/felt/content/ConsoleClient.sys.mjs
lissyx: And I dont see any reason to pass the full URL anyway?

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: I think there is very little we can do about that... Do you have an alternative approach?

lissyx: Something monotonic, I know padenot fixed a lot of issues related to that but mostly on the C++ side. Maybe `Performance.now()` would work?

```
16:21:28,413 Date(performance.timeOrigin + performance.now())
16:21:28,426 "Wed Oct 22 2025 16:21:28 GMT+0200 (heure d’été d’Europe centrale)" 
```

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: It does the same job, but I’d prefer not to use `new Promise` here - async/await is more modern and avoids extra nesting.

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: TIL about all our [shutdown phases](https://firefox-source-docs.mozilla.org/ipc/proc_shutdown.html#shutdown-flavors). So which of the 13 phases is best for us? :D

lissyx: I'd say at least `xpcom-shutdown`. Maybe there are others we should listen to for extension-specific?

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: Good point, I think [`ChromeUtils.now()`](https://searchfox.org/firefox-main/rev/48f8a74041be2973d7df8ded3ad5e64c2c00e8dd/dom/chrome-webidl/ChromeUtils.webidl#329-334) is what we need.

lissyx: Nice one !

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: Thanks for the hint! TIL about `sRestrictFromWebContentProcesses` list in `Preferences.cpp`. 
I added `enterprise.console.` and `browser_policies.live_polling.` prefs to the list.

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: We now call `ConsoleClient.uninit()` in `BrowserGlue.sys.mjs` on shutdown.

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: As discussed, the extra property is not needed.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -747,41 +751,29 @@ class RemotePoliciesProvider {
...
+          if (p === this.isPollingEnabled) {
+            return;
+          }
+          this._startPolling()
         }
         break;
-
       case "xpcom-shutdown":
         this._stopPolling();
+        Services.prefs.removeObserver(this.POLLING_FREQUENCY_PREF, this);
```
1rneh, line 790: Note: Adding `Services.prefs.removeObserver(this.POLLING_ENABLED_PREF, this);` on next update.

1rneh:
Changes:
- Moved ConsoleClient from `extensions/felt` to `browser/components/enterprise` to make sure the client is not loaded with a delay
- Token handling: Felt injects the refresh token into Firefox as pref (temporary workaround). The pref is only accessible in the parent process. On the first request to the console, Firefox retrieves the refresh token from the prefs, refreshed the session and receives a valid access token. For now only the launched Firefox (not Felt) communicates with the console and handles the token lifecycle.
- Removed racy test `test_felt_2_ensure_sso_cookie_stored` because we already test that the sso cookie made it to firefox in `test_felt_3_browser_started`.
- Added `browser.policies.live_polling.enabled` to make sure we generally guard the live polling

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: I added a todo. We first need to define what to do with network issues and offline support.

browser/extensions/felt/content/ConsoleClient.sys.mjs
1rneh: Hm, this is a way to self-invoke the async function and also make `this._refreshPromise` hold the promise that is returned...

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+
+  /**
+   * Base url of the enterprise console
+   */
+  get consoleAddr() {
+    return Services.prefs.getStringPref(
+      PREFS.CONSOLE_ADDRESS,
+      "https://console.enterfox.eu"
+    );
+  },
```
fiji-flo:
```suggestion
  get consoleAddr() {
    return new URL(Services.prefs.getStringPref(
      PREFS.CONSOLE_ADDRESS,
      "https://console.enterfox.eu"
    ));
  },
```

modules/libpref/Preferences.cpp
```diff
@@ -6286,7 +6285,12 @@ static const PrefListEntry sRestrictFromWebContentProcesses[] = {
     PREF_LIST_ENTRY("extensions.lastAppBuildId"),
     PREF_LIST_ENTRY("media.gmp-manager.buildID"),
     PREF_LIST_ENTRY("toolkit.telemetry.previousBuildID"),
-};
+
+    // Remove access to console settings
+    PREF_LIST_ENTRY("enterprise.console."),
+
+    // Remove access to settings for live policy polling
+    PREF_LIST_ENTRY("browser.policies.live_polling.")};
```
Mossop: Curious to know why this needs to be restricted?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+      SSO: "/sso/login?target=browser",
+      STARTUP_PREFS: "/api/browser/hacks/startup",
+      DEFAULT_PREFS: "/api/browser/hacks/default",
+      REMOTE_POLICIES: "/api/browser/policies",
+      TOKEN: "/sso/token",
+    };
+  },
+
+  get ssoUri() {
+    return this.consoleAddr + this._paths.SSO;
```
fiji-flo:
```suggestion
    let url = this.consoleAddr;
    url.pathname = this._paths.SSO;
    return url;
```

There's more places. We saw issues with `https://console-gcp-eu.enterfox.eu//api/browser/policies` behaving different than `https://console.enterfox.eu//api/browser/policies` because of the cloud provider, so we want to _properly_ construct urls.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+  get consoleAddr() {
+    return Services.prefs.getStringPref(
+      PREFS.CONSOLE_ADDRESS,
+      "https://console.enterfox.eu"
+    );
+  },
+
+  get _paths() {
+    return {
+      SSO: "/sso/login?target=browser",
```
fiji-flo:
```suggestion
      SSO: "/sso/login?target=browser",
      SSO_CALLBACK: "/sso/callback",
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+  get ssoCallbackUri() {
+    // Dropping the port is required here because the matcher being used by
+    // JSActors code relies on WebExtensions MatchPattern
+    // https://searchfox.org/firefox-main/source/toolkit/components/extensions/MatchPattern.cpp#370-384
+    // The match pattern should then NOT use any port otherwise matching would
+    // not happen.
+    const uri = Services.io.newURI(this.consoleAddr);
+    const consoleAddrExcludingPort = `${uri.scheme}://${uri.host}`;
+    return consoleAddrExcludingPort + "/sso/callback?*";
+  },
```
fiji-flo: I'd change the name to `ssoCallbackUriNoPort`
```suggestion
  get ssoCallbackUri() {
    // Dropping the port is required here because the matcher being used by
    // JSActors code relies on WebExtensions MatchPattern
    // https://searchfox.org/firefox-main/source/toolkit/components/extensions/MatchPattern.cpp#370-384
    // The match pattern should then NOT use any port otherwise matching would
    // not happen.
    const url = this.consoleAddr;
    url.pathname = this._paths.SSO_CALLBACK;
    url.port = "";
    return url.href + "?*";
  },
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+    headers.set("Authorization", `${tokenType} ${accessToken}`);
+
+    const url = this.consoleAddr + path;
+    const res = await fetch(url, { headers });
+
+    if (res.ok) {
+      return await res.json();
+    }
+
+    if ((res.status === 404 || res.status === 401) && !_didRefresh) {
```
fiji-flo: `401` and `403` should be enough if any route is returning a `404` for not authorized I need to fix this in the console.
```suggestion
    if ((res.status === 403 || res.status === 401) && !_didRefresh) {
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+            "Content-Type": "application/json",
+            Accept: "application/json",
+          },
+          body: JSON.stringify({
+            grant_type: "refresh_token",
+            refresh_token: refreshToken,
+          }),
+        });
+
+        if (res.status === 401) {
```
fiji-flo:
```suggestion
        if (res.status === 401 || res.status === 403) {
```
We might want to differentiate that in a future but not for now.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+
+        const t = await res.json();
+        this.tokenData = new ConsoleTokenData(
+          t.access_token,
+          t.refresh_token,
+          t.expires_in,
+          t.token_type
+        );
+
+        Services.prefs.setStringPref(
```
fiji-flo: maybe we move this to `ensureTokenData`, too.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+    );
+  },
+
+  /**
+   * Clears persisted and in-memory token data.
+   */
+  clearTokenData() {
+    try {
+      // Only exists in Firefox, not in Felt.
+      Services.prefs.clearUserPref(PREFS.REFRESH_TOKEN);
```
Mossop: This doesn't throw even if the pref doesn't exist so there is no need to wrap in a try...catch.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+      REMOTE_POLICIES: "/api/browser/policies",
+      TOKEN: "/sso/token",
+    };
+  },
+
+  get ssoUri() {
+    return this.consoleAddr + this._paths.SSO;
+  },
+
+  get ssoCallbackUri() {
```
Mossop: As this isn't the actual URI, but the match pattern for the URI let's call this property `ssoCallbackUriMatchPattern`

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,387 @@
...
+          // TODO: Handle network issues, offline support, etc.
+        }
+
+        const t = await res.json();
+        this.tokenData = new ConsoleTokenData(
+          t.access_token,
+          t.refresh_token,
+          t.expires_in,
+          t.token_type
+        );
```
fiji-flo: `this.ensureTokenData(t)` would do the same.

Mossop (approved):

fiji-flo (commented):
nit / question. I see some places where we use [private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) in FF is there a guideline on this?

The other question would be to maybe add some more [JSDoc](https://jsdoc.app/).

modules/libpref/Preferences.cpp
1rneh: I thought it was best to mitigate that a web content process can enable live policy polling in the parent or change the url that the policies are polled from. Do you think it's not necessary?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Just to confirm - are we also not expecting a `404` response here?

modules/libpref/Preferences.cpp
Mossop: Content processes cannot directly modify prefs at all so there is no risk of that. This list hides prefs entirely from content processes where they might contain user identifiable information or other pseudo-identifiers that could enable fingerprinting.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: I left out any refresh token preference changes, since we're very likely to move away from this temporary workaround now.

fiji-flo (approved):
r+ for the console client part

1rneh:
Rebased.

browser/components/BrowserGlue.sys.mjs
```diff
@@ -376,6 +377,10 @@ BrowserGlue.prototype = {
     }
 
     lazy.ContentBlockingPrefs.uninit();
+
+    if (AppConstants.MOZ_ENTERPRISE) {
```
lissyx: I think it would be better if we directly installed the observer in the `ConsoleClient` and react on it there, `_dispose()` is called on `profile-after-change` it seems: https://searchfox.org/firefox-main/rev/0b21972a78f8915f73ce5579eeee2aa8c9c7d67e/browser/components/BrowserGlue.sys.mjs#265-269

testing/enterprise/felt_tests.py
```diff
@@ -33,8 +34,7 @@ def do_POST(self):
         if self.path == "/:shutdown":
             print("Shutting down as requested")
             self.reply("OK")
-            setattr(self.server, "_BaseServer__shutdown_request", True)
-            self.server.server_close()
+            threading.Thread(target=self.server.shutdown, daemon=True).start()
```
lissyx: It's already running on a different thread, why manually doing it?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -702,35 +703,42 @@ class JSONPoliciesProvider {
...
-    this._serverAddr = Services.prefs.getStringPref(
-      "browser.policies.server",
-      ""
+      this.POLLING_FREQUENCY_PREF,
+      this.POLLING_FREQUENCY_FALLBACK
     );
-    this._accessToken = Services.prefs.getStringPref(
-      "browser.policies.access_token",
-      ""
+    this._isPollingEnabled = Services.prefs.getBoolPref(
```
lissyx, line 727: Why do we need to introduce another pref here?

lissyx: Ah we need before we remove both the server and the access token and the frequency.

lissyx: @1rneh Do you think we need a default polling value or could we remove the `.enabled` pref if we define that by default, 0 polling means not enabled?

I had the default value to make this more landable on m-c quickly, but for now since it's still coupled to ConsoleClient maybe it's not worth keeping the default 60 secs polling and we should just make felt set the polling frequency, and fallback that no pref value means disabled.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -802,40 +808,40 @@ class RemotePoliciesProvider {
     return this._failed;
   }
 
+  _restartPolling() {
+    if (this._poller) {
```
lissyx: We already check for that in `_stopPolling()`, is it really needed here?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -802,40 +808,40 @@ class RemotePoliciesProvider {
     return this._failed;
   }
 
+  _restartPolling() {
```
lissyx: there's only one place where we call it I'm not sure it's really useful to have such method (but it's also not a big deal to keep it)

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -802,40 +808,40 @@ class RemotePoliciesProvider {
...
         console.warn(
-          `RemotePoliciesProvider: performPolling(): ${this._pollingFrequency}: error ${error}`
+          `RemotePoliciesProvider performPolling() with frequency ${this._pollingFrequency} caused error ${error}`
         );
         this._hasRemoteConnection = false;
       });
   }
 
   _startPolling() {
-    Services.obs.addObserver(this, "xpcom-shutdown");
```
lissyx, line 838: I'm a bit worried we dont register this anymore?

lissyx: nevermind we do it in the constructor. I'm not sure this is a good thing, I've seen we are creating a lot of instances of the provider and I think our cleanup path is bad, but it's also all things I'm fixing in #104

testing/enterprise/felt_tests.py
```diff
@@ -208,10 +231,13 @@ def serve(
         httpd.policy_block_about_config = policy_block_about_config
     if policy_access_token:
         httpd.policy_access_token = policy_access_token
+    if policy_refresh_token:
+        httpd.policy_refresh_token = policy_refresh_token
     print(
         f"Serving localhost:{port} SSO={sso_port} CONSOLE={console_port} with {classname}"
     )
     httpd.serve_forever()
+    httpd.server_close()
```
lissyx: That looks very weird, can we elaborate?

browser/components/BrowserGlue.sys.mjs
1rneh: You're right, this is the wrong spot then. ConsoleClient is a process singleton `const` hence we'll need to call either `.init()` on startup to register the observer or `.uninit()` on shutdown. I could call `.init()` in [`BrowserComponents.manifest`](https://searchfox.org/firefox-main/rev/0b21972a78f8915f73ce5579eeee2aa8c9c7d67e/browser/components/BrowserComponents.manifest#39)?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Yes, we can have felt set the frequency. But I'd prefer an explicit `.enabled` pref. I'm not too keen on having the absence of one pref implicitly meaning another thing as well.

browser/components/BrowserGlue.sys.mjs
1rneh: Nevermind, I'll export the `ConsoleClient` and immediately invoke `.init()`, like it's done for the [`SyncTabsErrorHandler`](https://searchfox.org/firefox-main/rev/0b21972a78f8915f73ce5579eeee2aa8c9c7d67e/browser/components/firefoxview/firefox-view-synced-tabs-error-handler.sys.mjs#187). This way it's self-contained in `ConsoleClient.sys.mjs`.

testing/enterprise/felt_tests.py
lissyx: This is not needed, it should be reverted

testing/enterprise/felt_tests.py
lissyx: This is not needed, it should be reverted

testing/enterprise/felt_tests.py
```diff
@@ -332,8 +360,10 @@ def teardown(self):
         print("Shutting down SSO")
         requests.post(f"http://localhost:{self.sso_port}/:shutdown", timeout=2)
         print("Stopping process console")
+        self.console_httpd.terminate()
```
lissyx: This is not needed, it should be reverted

testing/enterprise/felt_tests.py
```diff
@@ -332,8 +360,10 @@ def teardown(self):
...
         self.console_httpd.join()
         print("Stopping process SSO")
+        self.sso_httpd.terminate()
```
lissyx: This is not needed, it should be reverted

testing/enterprise/felt_tests.py
```diff
@@ -186,6 +186,28 @@ def do_GET(self):
         else:
             self.not_found(path)
 
+    def do_POST(self):
```
lissyx, line 189: We need a call to `super().do_POST()` here ot handle the shutdown path

lissyx: cf https://github.com/mozilla/enterprise-firefox/pull/117

testing/enterprise/felt_tests.py
lissyx: cf https://github.com/mozilla/enterprise-firefox/pull/117

testing/enterprise/felt_tests.py
lissyx: cf https://github.com/mozilla/enterprise-firefox/pull/117

testing/enterprise/felt_tests.py
lissyx: cf https://github.com/mozilla/enterprise-firefox/pull/117

testing/enterprise/felt_tests.py
lissyx: cf https://github.com/mozilla/enterprise-firefox/pull/117

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: What do you mean that our cleanup path is bad?

lissyx: It's not a problem in your patch

1rneh: Could you elaborate on the problem and how this relates to adding the `xpcom-shutdown` observer in the constructor?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,466 @@
...
+    Services.obs.addObserver(this, "xpcom-shutdown");
+    return this;
+  },
+
+  observe(_, topic) {
+    switch (topic) {
+      case "xpcom-shutdown": {
+        this.clearTokenData();
+        this._refreshPromise = null;
+        Services.obs.removeObserver(this, "xpcom-shutdown");
```
lissyx: nit: I think we prefer to remove the observer as early as possible

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -0,0 +1,466 @@
...
+      }
+    }
+  },
+}.init();
```
lissyx, line 469: if I follow this right, each time we load the `ConsoleClient.sys.mjs` we will perform an `init()` call? I'm not 100% sure of that, but if it's the case it might be risky? And so we'll add each time an `xpcom-shutdown` observer that we will remove only once?

browser/components/enterprise/tests/moz.build
lissyx, line 1: I dont think we need an empty `moz.build` file

browser/extensions/felt/content/window.js
```diff
@@ -39,12 +39,12 @@ function connectToConsole(email) {
   );
 
   console.debug("Load SSO Page: ", SOURCE_URI);
-  const uri = Services.io.newURI(SOURCE_URI);
+  const ssoLoginURI = Services.io.newURI(SOURCE_URI);
```
lissyx: nit: I think it should be part of `constructSsoLoginURI()` since that's the only user of it

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: many `RemotePoliciesProvider` object being created, each adding a new observer will leak observers which will later create other problems

lissyx (approved):
Let's make sure the `ConsoleClient.init()` is sound and we're good?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Good question! But nothing to worry about. ES modules are evaluated once and then cached per module instance. Importing it in other files reuses the cached export, so `init` does not run again, unless we explicitly call `ConsoleClient.init()`.

Mossop: Yes, system modules are singletons so this will only ever run once (per process if it is ever loaded in multiple processes)

PR #104, opened 2025-10-21
Correctly apply/unapply policies

lissyx:


toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -173,11 +173,36 @@ EnterprisePoliciesManager.prototype = {
...
-    let { schema } = ChromeUtils.importESModule(
+    const { schema } = ChromeUtils.importESModule(
+      // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
       "resource:///modules/policies/schema.sys.mjs"
     );
 
-    for (let policyName of Object.keys(unparsedPolicies)) {
+    this._previousPolicies = {};
+    try {
+      this._previousPolicies = JSON.parse(JSON.stringify(this._parsedPolicies));
```
mkaply: Aren't the parse and stringify cancelling each other out?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -173,11 +173,36 @@ EnterprisePoliciesManager.prototype = {
...
+    } catch {
+      ///
+    }
+
+    this._parsedPolicies = {};
+
+    const policyNames = Object.keys(unparsedPolicies);
+
+    const previousNames = Object.keys(this._previousPolicies)
```
mkaply: This isn't necessarily going to work because of subpolicies. Not everything is toplevel.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -173,11 +173,36 @@ EnterprisePoliciesManager.prototype = {
...
+      .filter(policyName => {
+        let policyImpl = lazy.Policies[policyName];
```
mkaply: This all seems unnecessary since we know that previousNames was valid policy?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -213,12 +237,41 @@ EnterprisePoliciesManager.prototype = {
...
+        lazy.log.info(`${policyName} has been deprecated.`);
+        continue;
+      }
+
+      const liveable = "liveable" in policyImpl && policyImpl.liveable;
+      if (!liveable) {
+        continue;
+      }
+
+      this._maybeCallbackPolicy(
```
mkaply: I think we're going to have to think about how this works. 

If a policy has gone away, we can't do a callback with the previous policy. We need some way to say "undo this policy". But if the policy isn't true/false, we won't know how to undo it?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: They are but I'm just doing them as a way to deep copy objects

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: Right, leftover of previous attempts

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: Good to know, which ones should I look at for this?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: That's why I shared my WIP :)

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
mkaply: Unfortunately it's a lot of them. Let me do some thinking on how to do this best.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -359,8 +364,19 @@ export var Policies = {
...
         blockAboutPage(manager, "about:config");
         setAndLockPref("devtools.chrome.enabled", false);
-      } else {
+      }
+    },
+    onRemove(manager, oldParams) {
+      if (oldParams) {
+        // if it was block, just unblock
         unblockAboutPage(manager, "about:config");
+        Services.prefs.clearUserPref("devtools.chrome.enabled");
```
Mossop: This isn't quite right. The policy sets the default pref not the user pref. So we need to reset the default pref back to what it was previously.

lissyx: that's why it's WIP  :)

Mossop: Just generally my concern is that policies will need to cache some state in order to unapply, and for some policies that will need to persist across a restart (knowing to remove an added bookmark for example). It would be good to figure out the mechanism for doing that.

lissyx: Maybe it's something we should address when we identify policies where it's required? I dont feel like it's going a big issue

mkaply: Yeah, the default pref stuff will be tricky just because of the way it's designed in Firefox.

One thought I had was just storing the original value of the pref when we set the default so we can undo it.

But some of the prefs are set on default for the first time and you can't clear a default pref.

It might be the case that we just ignore some of the default pref setting.

One other thought I had was updating the lockPref concept in preferences so it accepted a value and had a separate branch. So instead of setting a default and then locking, you lock it to a value and when you unlock, that value goes away. That would be a pretty significant change to prefs, though (and would only cover cases where you lock the pref)

lissyx: Ok, but the current PR is about enabling the proper behavior on how we apply policies. Fixing the pref usage is something to do after (right after, yes).

Can I get both @mkaply @Mossop  your opinion on the current implementation wrt the needs we have? If so, we can start by landing this current PR (after any change you think justified) that:
 - implement what we want to track previous policies and ensure we apply what we need,
 - properly handled application and removal for the `about:config` example

I know @mkaply has pending changes with the implementation of a few things, so it would be nice we get them after as well.

Once it's in place we can:
 - convert all policies that are trivial,
 - work on policies that manipulate prefs,

Mossop: This doesn't support undoing policies that change across a browser restart. Are there any where that is an issue?

lissyx: I dont think it was ever mentionned as a requirement?

lissyx: Well if we ever hit this then we'll deal with it? We could serialize the ~~states~~ parameters of previous policies, but I'd like to avoid it if we can for now.

mkaply: > This doesn't support undoing policies that change across a browser restart. Are there any where that is an issue?

I'm not sure what you mean.

Pretty much all of the policies stuff goes away when the browser closes (exceptions are Bookmarks, Handlers, and a couple others).

Mossop: It's that "pretty much" it would be good to understand at some point. I guess we can cross that bridge when we come to it though.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -359,8 +364,20 @@ export var Policies = {
...
         blockAboutPage(manager, "about:config");
         setAndLockPref("devtools.chrome.enabled", false);
-      } else {
+      }
+    },
+    onRemove(manager, oldParams) {
+      if (oldParams) {
+        // if it was block, just unblock
         unblockAboutPage(manager, "about:config");
+        unsetAndUnlockPref("devtools.chrome.enabled");
```
lissyx, line 373: @Mossop @mkaply Now we should be able to do things correctly. There's a test for that behavior in `browser_policies_diffing.js` that should make the flow clear.

toolkit/components/enterprisepolicies/tests/browser/browser_policies_diffing.js
```diff
@@ -0,0 +1,282 @@
...
+    customSchema
+  );
+
+  // Policy being removed it means the blocking should get lifted
+  ok(!blockAnotherPageApplied, "BlockAnotherPage disabled");
+
+  delete lazy.Policies.BlockAnotherPage;
+});
+
+add_task(async function test_simple_policy_pref_setAndLock() {
```
lissyx: This should clearly assert the flow for prefs. I hope I'm not mistaken, I dont have so much used branches on prefs.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -359,8 +364,20 @@ export var Policies = {
...
+        /*
+      } else {
+        // TODO: should we care about that???
+	// if it was not blocked, just block
+        blockAboutPage(manager, "about:config");
+        setAndLockPref("devtools.chrome.enabled", false);
+      */
```
Mossop: I think we just don't need to do anything here.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3067,6 +3168,24 @@ export var PoliciesUtils = {
...
+   * @param {string} prefName
+   *        The pref to be changed
+   */
+  unsetDefaultPref(prefName) {
+    let prefWasLocked = Services.prefs.prefIsLocked(prefName);
+    if (prefWasLocked) {
+      Services.prefs.unlockPref(prefName);
+    }
+
+    Services.prefs.deleteBranch(prefName);
```
Mossop: `restoreDefaultPref` will restore the pref value anyway so I don't think we need this. It also has the risk that it will delete any sub-prefs of the branch. We don't do that much but it's a bit of a footgun.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -173,11 +176,26 @@ EnterprisePoliciesManager.prototype = {
...
+    const { schema } = ChromeUtils.importESModule(
+      // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
       "resource:///modules/policies/schema.sys.mjs"
     );
 
-    for (let policyName of Object.keys(unparsedPolicies)) {
+    // Make a deep copy that will be trimmed later
+    const previousPolicies = JSON.parse(
+      JSON.stringify(this._parsedPolicies || {})
+    );
```
Mossop: Can you just use `structuredClone` for this?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -212,14 +229,86 @@ EnterprisePoliciesManager.prototype = {
...
+        if (previousParameters == JSON.stringify(parsedParameters)) {
+          VERBOSE_DEBUG &&
+            console.debug(
+              `EnterprisePoliciesParent: policyName=${policyName} not changed, dont schedule for running`
+            );
+          shouldSchedule = false;
+        }
+      }
+
       this._parsedPolicies[policyName] = parsedParameters;
```
Mossop: If you move this above the if block then you can just use `continue` rather than needing the `shouldSchedule` variable.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -362,22 +499,47 @@ EnterprisePoliciesManager.prototype = {
...
               return cb;
             });
           })
-          .filter(cbName =>
-            this.observersReceived.includes(policiesCallbackMapping[cbName])
+          .filter(
+            cbName =>
+              cbName !== "onRemove" &&
+              this.observersReceived.includes(policiesCallbackMapping[cbName])
           );
```
Mossop: This will generate a list of callbacks names but each callback name can be included in the list multiple times (which is sort of ok since `_runPoliciesCallbacks` clears the list) but they won't necessarily be in the correct order. I don't know if that breaks anything but it might. I suggest we remove this block entirely and do my suggestion for calling the callbacks below.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -362,22 +499,47 @@ EnterprisePoliciesManager.prototype = {
...
+          );
+        VERBOSE_DEBUG &&
+          console.debug(
+            `EnterprisePoliciesParent:Activate running now: callbacksToRun=${callbacksToRun}`
+          );
+        VERBOSE_DEBUG &&
+          console.debug(
+            `EnterprisePoliciesParent:Activate running now: onRemoveToRun=${onRemoveToRun}`
+          );
         callbacksToRun.map(cb => this._runPoliciesCallbacks(cb));
```
Mossop:
```suggestion
        Object.keys(this._callbacks)
          .filter(cbName => this.observersReceived.includes(policiesCallbackMapping[cbName))
          .map(cb => this._runPoliciesCallbacks(cb));
```

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -710,6 +872,12 @@ class JSONPoliciesProvider {
 
 class RemotePoliciesProvider {
   constructor() {
+    if (RemotePoliciesProvider._instance) {
+      return RemotePoliciesProvider._instance;
+    }
+
+    RemotePoliciesProvider._instance = this;
+
```
Mossop, line 786:
```suggestion
  static #instance = null;
  static createInstance() {
    if (!RemotePoliciesProvider.#instance) {
      RemotePoliciesProvider.#instance = new RemotePoliciesProvider();
    }

    return RemotePoliciesProvider.#instance;
  }

  constructor() {
```

Returning from the constructor is an odd pattern in JavaScript. Please use a static creator method instead.

Mossop (changes requested):
I think this is a little broken in the case of multiple policy providers being active at the same time but I'm not sure whether we need to support this case.

Also there is an awful lot of logging in this patch, can we slim it down some?

lissyx:
> I think this is a little broken in the case of multiple policy providers being active at the same time but I'm not sure whether we need to support this case.

Can you elaborate? It should not break them

> 
> Also there is an awful lot of logging in this patch, can we slim it down some?

Yep, that's on purpose for now. I think we may need better logging capabilities in the code itself, but it's likely for another time.

browser/components/enterprisepolicies/Policies.sys.mjs
lissyx: Then I'll remove

browser/components/enterprisepolicies/Policies.sys.mjs
lissyx: Isn't it required so that the notion of default value is correct ? I'll verify

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: We have access to that? it's a Window object method right? https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone

I'll see if we have an implementation we can use

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: Sure

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
Mossop: I see it used in many existing modules so I assume so: https://searchfox.org/firefox-main/source/toolkit/components/search/SearchEngine.sys.mjs#1139

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3007,27 +3037,98 @@ function setPrefIfPresentAndLock(param, paramKey, prefName) {
...
+   *    "prefName": {
+   *      "defaultValue": ...,
+   *      "userValue": ...,
+   *    }
+   *  }
+   *
+   */
+  _savedPrefs: {},
+
+  saveDefaultPref(prefName) {
```
Mossop, line 3042: This needs to bail out early if we already have defaults saved or if a policy changes the pref multiple times we won't be able to restore the original default. This isn't possible with the `BlockAboutConfig` that is implemented for now but looks possible with some of the others we may implement later.

browser/components/enterprisepolicies/Policies.sys.mjs
lissyx: Ok it seems to work properly without

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: Thanks!

browser/components/enterprisepolicies/Policies.sys.mjs
lissyx: So we just exit early if we already have a key with the pref name

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: I need to check that more deeply

lissyx: We probably still want to filter out `onRemove` ?

Mossop: The check I suggested below (`this.observersReceived.includes(policiesCallbackMapping[cbName]))`) should handle that since there is no policiesCallbackMapping entry for onRemove. No harm in doing it here as well though.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3007,27 +3030,103 @@ function setPrefIfPresentAndLock(param, paramKey, prefName) {
...
+    let values = {
+      defaultValue: undefined,
+      userValue: undefined,
+    };
+
+    // This was already changed, do not overwrite.
+    if (prefName in Object.keys(this._savedPrefs)) {
```
Mossop:
```suggestion
    if (prefName in this._savedPrefs) {
```

Mossop:
> > I think this is a little broken in the case of multiple policy providers being active at the same time but I'm not sure whether we need to support this case.
> 
> Can you elaborate? It should not break them

So my assumption is if one provider sets a value for the policy and another provider does not then we still apply the policy (that seems to match [this code comment](https://github.com/mozilla/enterprise-firefox/blob/f71b2882f2994b6ead07efd1a675c1be970b1645/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L1081)).

But let's say that you have a JSON provider that sets `BlockAboutConfig` and then the remote provider does not. My reading of the code is that the json policies [will be activated first](https://github.com/mozilla/enterprise-firefox/blob/f71b2882f2994b6ead07efd1a675c1be970b1645/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L158) and so the policy will be added to `_parsedPolicies`. Then later the remote provider will retrieve its policies which will not include `BlockAboutConfig`. And so when the remote provider calls to activate its policies we will call `onRemove` for `BlockAboutConfig`.

I think we should be using the `CombinedProvider` to resolve the final set of policies, but `CombinedProvider` doesn't get given the `onPoliciesChanges` callback to call when the set of policies changes, and besides the `RemotePoliciesProvider` activates its policies directly via the `EnterprisePolicies:Activate` notification.

lissyx:
> > > I think this is a little broken in the case of multiple policy providers being active at the same time but I'm not sure whether we need to support this case.
> > 
> > 
> > Can you elaborate? It should not break them
> 
> So my assumption is if one provider sets a value for the policy and another provider does not then we still apply the policy (that seems to match [this code comment](https://github.com/mozilla/enterprise-firefox/blob/f71b2882f2994b6ead07efd1a675c1be970b1645/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L1081)).
> 
> But let's say that you have a JSON provider that sets `BlockAboutConfig` and then the remote provider does not. My reading of the code is that the json policies [will be activated first](https://github.com/mozilla/enterprise-firefox/blob/f71b2882f2994b6ead07efd1a675c1be970b1645/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L158) and so the policy will be added to `_parsedPolicies`. Then later the remote provider will retrieve its policies which will not include `BlockAboutConfig`. And so when the remote provider calls to activate its policies we will call `onRemove` for `BlockAboutConfig`.
> 
> I think we should be using the `CombinedProvider` to resolve the final set of policies, but `CombinedProvider` doesn't get given the `onPoliciesChanges` callback to call when the set of policies changes, and besides the `RemotePoliciesProvider` activates its policies directly via the `EnterprisePolicies:Activate` notification.

Yes, that's also why we should decouple polling from the provider itself, then we should be able to properly handle that

Mossop:
> > > > I think this is a little broken in the case of multiple policy providers being active at the same time but I'm not sure whether we need to support this case.
> > > 
> > > 
> > > Can you elaborate? It should not break them
> > 
> > 
> > So my assumption is if one provider sets a value for the policy and another provider does not then we still apply the policy (that seems to match [this code comment](https://github.com/mozilla/enterprise-firefox/blob/f71b2882f2994b6ead07efd1a675c1be970b1645/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L1081)).
> > But let's say that you have a JSON provider that sets `BlockAboutConfig` and then the remote provider does not. My reading of the code is that the json policies [will be activated first](https://github.com/mozilla/enterprise-firefox/blob/f71b2882f2994b6ead07efd1a675c1be970b1645/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L158) and so the policy will be added to `_parsedPolicies`. Then later the remote provider will retrieve its policies which will not include `BlockAboutConfig`. And so when the remote provider calls to activate its policies we will call `onRemove` for `BlockAboutConfig`.
> > I think we should be using the `CombinedProvider` to resolve the final set of policies, but `CombinedProvider` doesn't get given the `onPoliciesChanges` callback to call when the set of policies changes, and besides the `RemotePoliciesProvider` activates its policies directly via the `EnterprisePolicies:Activate` notification.
> 
> Yes, that's also why we should decouple polling from the provider itself, then we should be able to properly handle that

Either that or if all the policies flow through onPoliciesChanges then the combined provider can pass on the correctly combined set.

lissyx:
> > > > > I think this is a little broken in the case of multiple policy providers being active at the same time but I'm not sure whether we need to support this case.
> > > > 
> > > > 
> > > > Can you elaborate? It should not break them
> > > 
> > > 
> > > So my assumption is if one provider sets a value for the policy and another provider does not then we still apply the policy (that seems to match [this code comment](https://github.com/mozilla/enterprise-firefox/blob/f71b2882f2994b6ead07efd1a675c1be970b1645/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L1081)).
> > > But let's say that you have a JSON provider that sets `BlockAboutConfig` and then the remote provider does not. My reading of the code is that the json policies [will be activated first](https://github.com/mozilla/enterprise-firefox/blob/f71b2882f2994b6ead07efd1a675c1be970b1645/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L158) and so the policy will be added to `_parsedPolicies`. Then later the remote provider will retrieve its policies which will not include `BlockAboutConfig`. And so when the remote provider calls to activate its policies we will call `onRemove` for `BlockAboutConfig`.
> > > I think we should be using the `CombinedProvider` to resolve the final set of policies, but `CombinedProvider` doesn't get given the `onPoliciesChanges` callback to call when the set of policies changes, and besides the `RemotePoliciesProvider` activates its policies directly via the `EnterprisePolicies:Activate` notification.
> > 
> > 
> > Yes, that's also why we should decouple polling from the provider itself, then we should be able to properly handle that
> 
> Either that or if all the policies flow through onPoliciesChanges then the combined provider can pass on the correctly combined set.

Let's make sure we address that when we change polling behavior? It's going to take a bit more of work and I'd like to do that after `ConsoleClient` gets merged

lissyx:
@Mossop @mkaply I think the current scope of changes is sound enough to be in a state we can (assuming we agree on those changes here) merge and then iterate from by starting to move more policies. In parallel we can address how polling is done.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -362,22 +424,25 @@ EnterprisePoliciesManager.prototype = {
...
+              cbName !== "onRemove" &&
+              this.observersReceived.includes(policiesCallbackMapping[cbName])
+          )
+          .map(cb => this._runPoliciesCallbacks(cb));
+
+        Object.keys(this._previousPolicies)
+          .filter(
+            name => name in lazy.Policies && "onRemove" in lazy.Policies[name]
+          )
+          .map(name => lazy.Policies[name].onRemove.name)
```
Mossop: It's not clear to me that this does anything?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -362,22 +424,25 @@ EnterprisePoliciesManager.prototype = {
...
+          .map(cb => this._runPoliciesCallbacks(cb));
```
Mossop: Is there a reason to not just call `this._runPoliciesCallback("onRemove")`? `this._callbacks.onRemove` should only contain callbacks ready to be run.

Mossop (approved):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: This gets the `onRemove` name's from the callback. Maybe `policyName` instead of `name` would be less confusing?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: You might be right I need to check

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
Mossop: Isn't it just always "onRemove"?

lissyx: yeah but if i can express it in a much more complex way it's more fun :p

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: There's no reason it would not work, I think it's all leftover from how it was written at first and I missed the fact ...

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3063,6 +3151,23 @@ export var PoliciesUtils = {
...
+  /**
+   * unsetDefaultPref
+   *
+   * Unsets the _default_ value of a pref and unlock it if it was locked.
+   *
+   * @param {string} prefName
+   *        The pref to be changed
+   */
+  unsetDefaultPref(prefName) {
+    let prefWasLocked = Services.prefs.prefIsLocked(prefName);
```
mkaply, line 3164: I think there's no harm in just unlocking the pref without the check?

mkaply (approved):
Overall this looks reasonable to me.

I think once we have this and start implementing the onRemoves, we'll see how well things work :)

PR #116, opened 2025-10-31
FELT: Remove rust leftovers

lissyx:


toolkit/library/rust/gkrust-features.mozbuild
```diff
@@ -85,8 +85,5 @@ if CONFIG["MOZ_ICU4X"]:
 if CONFIG["USE_LIBZ_RS"]:
     gkrust_features += ["libz-rs-sys"]
 
-if CONFIG["MOZ_ENTERPRISE"]:
-    gkrust_features += ["felt"]
-
```
Mossop, line 90: Isn't this still required to enable [this](https://github.com/mozilla/enterprise-firefox/blob/enterprise-main/toolkit/library/rust/shared/Cargo.toml#L165)?

Mossop (changes requested):
This doesn't build for me because the felt crate is no longer included.

toolkit/library/rust/gkrust-features.mozbuild
lissyx: Exactly what I was looking for but got changed and I failed to find!

toolkit/library/moz.build
```diff
@@ -377,11 +377,6 @@ if CONFIG["OS_ARCH"] == "WINNT":
             "uiautomationcore",
         ]
 
-    if CONFIG["MOZ_ENTERPRISE"]:
```
lissyx, line 380: We still need it but likely we want to do that somewhere else ... ?

Mossop (approved):

browser/extensions/felt/rust/moz.build
```diff
@@ -8,6 +8,11 @@ EXPORTS.mozilla.browser.extensions.felt += [
     "felt.h",
 ]
 
+if CONFIG["OS_ARCH"] == "WINNT":
+    OS_LIBS += [
```
lissyx, line 12: this may not apply to the correct place.

PR #121, opened 2025-11-03
FELT: Adding Device Posture for SSO URL

lissyx:


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -416,6 +429,33 @@ export const ConsoleClient = {
...
+
+  _getDevicePosture() {
+    const devicePosturePayload = {
+      os: lazy.TelemetryEnvironment.currentEnvironment.system.os,
+      security: lazy.TelemetryEnvironment.currentEnvironment.system.sec,
+      build: lazy.TelemetryEnvironment.currentEnvironment.build,
+      // TODO: Client posture IP addr is P2, to be filled later.
+      network: {
+        ipv4: "",
+        ipv6: "",
```
fiji-flo:
```suggestion
        ipv4: null
        ipv6: null
```

I'll make those typed `Option<Ipv4Addr>` so null is better.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -416,6 +429,33 @@ export const ConsoleClient = {
...
+        Accept: "application/json",
+      },
+      body: JSON.stringify(payload),
+    });
+    return await req.json();
+  },
+
+  _getDevicePosture() {
+    const devicePosturePayload = {
+      os: lazy.TelemetryEnvironment.currentEnvironment.system.os,
```
fiji-flo, line 475: I think I followed this a bit and this is rather low level. So this would be `Darwin` for macOS. I think we keep it like that but we should discuss what we want to send here.

fiji-flo (approved):
Looks good (look at my comments though)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: It can be seen in `about:telemetry` if you want. I'm not sure there's a lot of value in diverging from what telemetry collects here

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Thanks, I fixed it

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -284,6 +292,11 @@ export const ConsoleClient = {
     return payload;
   },
 
+  async sendDevicePosture() {
+    const payload = this._getDevicePosture();
+    return this._devicePosture(payload);
+  },
```
1rneh, line 319: Let's unify `_devicePosture` and `sendDevicePosture` to one method and add some JSDoc.
Also should we rename `_getDevicePosture` to `_collectDevicePosture` so that it's not confused with initiating a GET request?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -416,6 +429,33 @@ export const ConsoleClient = {
     return this._refreshPromise;
   },
 
+  async _devicePosture(payload) {
+    const url = this.constructURI(this._paths.DEVICE_POSTURE);
+    const req = await fetch(url, {
```
```suggestion
    const res = await fetch(url, {
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -416,6 +429,33 @@ export const ConsoleClient = {
...
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Accept: "application/json",
+      },
+      body: JSON.stringify(payload),
+    });
+    return await req.json();
```
1rneh: Let's throw an error if the response !== 200.
```suggestion
    if (res.ok) {
          return await res.json();
    }
    const text = await res.text().catch(() => "");
    throw new Error(`Post failed (${res.status}): ${text}`);
```

browser/extensions/felt/content/felt.xhtml
```diff
@@ -51,6 +51,10 @@
               ></moz-button>
             </div>
           </form>
+          <h1
+            class="felt-login__forbidden is-hidden"
+            data-l10n-id="felt-login-forbidden"
+          ></h1>
```
1rneh: This UI requirement is new to me. Is there any design for it yet?

testing/enterprise/felt_tests.py
```diff
@@ -65,15 +74,20 @@ def do_GET(self):
...
 
         elif path == "/auth":
+            query = urllib.parse.parse_qs(parsed.query)
+            if not "devicePostureToken" in query.keys():
+                self.forbidden()
+
             expires = datetime.datetime.utcnow() + datetime.timedelta(hours=8)
             cookie_expiry = expires.strftime("%a, %d %b %Y %H:%M:%S GMT")
-            location = f"http://localhost:{self.server.console_port}/sso/callback?foo"
+            location = f"http://localhost:{self.server.console_port}/sso/callback?foo&devicePostureToken={query['devicePostureToken'][0]}"
```
1rneh: Maybe I am missing something here.. Why are we expecting that the device posture is forwarded to the sso provider and then part of the `query` when redirected to `sso/callback`? 
I don't think we need to test this.

browser/extensions/felt/content/FeltWindowChild.sys.mjs
```diff
@@ -18,14 +18,16 @@ export class FeltWindowChild extends JSWindowActorChild {
...
-
-    console.debug(
-      "FeltWindowChild: Sending token data to ConsoleClient and starting Firefox"
-    );
-    this.actor.sendAsyncMessage("FeltChild:StartFirefox", consoleTokenData);
+    const tokenData = event.target.querySelector("#token_data");
+    if (tokenData) {
+      console.debug("FeltWindowChild: Extracting token data");
+      const consoleTokenData = JSON.parse(tokenData.textContent);
+      if (consoleTokenData.success) {
```
1rneh: Getting redirected to the `sso/callback` + the token data in the script tag seems to indicate success. Do we need an explicit `success` property?

1rneh (changes requested):
I'm not fully up to speed on the device posture details - how much has been discussed so far?

testing/enterprise/felt_tests.py
```diff
@@ -160,13 +179,34 @@ def do_GET(self):
...
+            policy_refresh_token = self.server.policy_refresh_token.value
+            with self.server.device_posture_reply_forbidden.get_lock():
+                if self.server.device_posture_reply_forbidden.value == 1:
+                    callback_success = False
+                    policy_access_token = ""
+                    policy_refresh_token = ""
+
+            obj = json.dumps(
+                {
+                    "success": callback_success,
```
1rneh: @fiji-flo Is the success callback related to device posture or to user authorization?

browser/extensions/felt/content/FeltWindowChild.sys.mjs
lissyx: I prefer an explicit state, but I'll change it

browser/extensions/felt/content/felt.xhtml
lissyx: No, I hacked that as a way to show the status

lissyx: I'll remove

testing/enterprise/felt_tests.py
lissyx: I added that and it was here in my mind to reflect rather device posture check. But it's true one can assume we get the tokens when the device posture is correct. It was out of lazyness for the forbidden test that I made it it his way

testing/enterprise/felt_tests.py
lissyx: I'll need to recheck but we need that to make sure we pass around the device token. Console/SSO redirects FELT all the way so I dont see when we can from Felt send the token again so that the check happens at the SSO callback redirect step. It's likely the real console has better session handling than this code though ...

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -416,6 +449,29 @@ export const ConsoleClient = {
...
+      security: lazy.TelemetryEnvironment.currentEnvironment.system.sec,
+      build: lazy.TelemetryEnvironment.currentEnvironment.build,
+      // TODO: Client posture IP addr is P2, to be filled later.
+      network: {
+        ipv4: null,
+        ipv6: null,
+      },
+    };
+    return devicePosturePayload;
+  },
```
1rneh, line 485:
```suggestion
  /**
   * @typedef {Object} DeviceNetwork
   * @property {null} ipv4 IPv4 address, TBD
   * @property {null} ipv6 IPv6 address, TBD
   */

  /**
   * @typedef {Object} DevicePosture
   * @property {object} os Telemetry-reported os information.
   * @property {object|undefined} security Telemetry-reported security software info (windows only)
   * @property {object} build Telemetry-reported build info info
   * @property {DeviceNetwork} network Network posture (placeholders for now).
   */

  /**
   * Collects the device posture from TelemetryEnvironment.currentEnvironment.
   *
   * @returns {DevicePosture} devicePosture
   */
  _collectDevicePosture() {
    const devicePosture = {
      os: lazy.TelemetryEnvironment.currentEnvironment.system.os,
      security: lazy.TelemetryEnvironment.currentEnvironment.system.sec,
      build: lazy.TelemetryEnvironment.currentEnvironment.build,
      // TODO: Client posture IP addr is P2, to be filled later.
      network: {
        ipv4: null,
        ipv6: null,
      },
    };
    return devicePosture;
  },
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -284,6 +291,32 @@ export const ConsoleClient = {
     return payload;
   },
 
+  /**
+   * Collect the device posture data and send them to the console.
+   *
+   * @returns {Promise<{token: string}>} Token reported by console.
+   */
+  async sendDevicePosture() {
+    const payload = this._collectDevicePosture();
```
```suggestion
    const devicePosture = this._collectDevicePosture();
```

testing/enterprise/felt_tests.py
```diff
@@ -160,13 +179,31 @@ def do_GET(self):
                 m = json.dumps({"policies": {}})
 
         elif path == "/sso/callback":
+            query = urllib.parse.parse_qs(parsed.query)
+            if not "devicePostureToken" in query.keys():
+                self.forbidden()
```
1rneh: If I understood the flow correctly, the device posture token won't be in this uri. We get the posture token back on POST `sso/device_posture` and then include it as a query parameter to `sso/login`.

1rneh (commented):
Thank you - a lot of this already looks good! 
I'd like to understand what the console returns when the device posture fails the check. Imo, it should be more explicit than just an empty `#tokenData` element in the dom.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: It's not all going to be from TelemetryEnvironment, given there's currently nothing that reports the network info we may want

lissyx:
> Thank you - a lot of this already looks good! I'd like to understand what the console returns when the device posture fails the check. Imo, it should be more explicit than just an empty `#tokenData` element in the dom.

I am not sure I follow you there. `#tokenData` will not be empty, the object will be there just the token values would be empty (at least that is what I implemented in the test).

browser/extensions/felt/content/window.js
```diff
@@ -18,13 +18,19 @@ ChromeUtils.defineESModuleGetters(lazy, {
 // Will at least make move forward marionette
 Services.obs.notifyObservers(window, "browser-delayed-startup-finished");
 
-function connectToConsole(email) {
+async function connectToConsole(email) {
+  const devicePostureToken = await lazy.ConsoleClient.sendDevicePosture();
```
1rneh:
```suggestion
const { posture } = await lazy.ConsoleClient.sendDevicePosture();
if (!posture) {
    // TODO: Currently we don't check the posture yet. In the future we need to handle rejected device posture
    return;
}
```

1rneh (approved):
Nit: After our conversation in which we clarified that as of now there is no device posture check, no device posture is rejected, can we remove all forbidden device posture code for now? And iterate over it, once we start handling rejected device postures.

lissyx:
> Nit: After our conversation in which we clarified that as of now there is no device posture check, no device posture is rejected, can we remove all forbidden device posture code for now? And iterate over it, once we start handling rejected device postures.

I've kept some device posture checking and commented what was about blocking access based on the posture

PR #130, opened 2025-11-06
Bug 1998475 - Move felt connection thread start later r?mossop

lissyx:
Starting the thread too early hits the cookie service at a moment it is inappropriate leading to connections leaking.

### DIFF
--- a/browser/extensions/felt/rust/src/client.rs
+++ b/browser/extensions/felt/rust/src/client.rs
@@ -150,6 +150,7 @@ impl FeltClientThread {
         // Define an observer
         #[xpcom(implement(nsIObserver), nonatomic)]
         struct Observer {
+            profile_ready: Arc<AtomicBool>,
             thread_stop: Arc<AtomicBool>,
             notify_restart: Arc<AtomicBool>,
         }
@@ -163,6 +164,10 @@ impl FeltClientThread {
                 data: *const u16,
             ) -> nsresult {
                 match unsafe { CStr::from_ptr(topic).to_str() } {
+                    Ok("profile-after-change") => {
+                        trace!("FeltClientThread::start_thread::observe() profile-after-change");
+                        self.profile_ready.store(true, Ordering::Relaxed);
+                    }
                     Ok("xpcom-shutdown-threads") => {
                         trace!("FeltClientThread::start_thread::observe() xpcom-shutdown-threads");
                         self.thread_stop.store(true, Ordering::Relaxed);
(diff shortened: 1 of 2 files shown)


browser/extensions/felt/rust/src/client.rs
```diff
@@ -250,17 +268,33 @@ impl FeltClientThread {
...
                         if notify_restart_internal.load(Ordering::Relaxed) {
                             notify_restart_internal.store(false, Ordering::Relaxed);
                             trace!("FeltClientThread::felt_client::ipc_loop(): restart notification required!");
                             felt_client.notify_restart();
                         }
 
+                        if pending_cookies.len() > 0 && profile_ready_internal.load(Ordering::Relaxed) {
+                            trace!("FeltClientThread::felt_client::ipc_loop(): Profile ready! Start cookies injection!");
+                            while pending_cookies.len() > 0 {
+                                if let Some(one_cookie) = pending_cookies.pop() {
```
Mossop:
```suggestion
                            while let Some(one_cookie) = pending_cookies.pop() {
```

browser/extensions/felt/rust/src/client.rs
```diff
@@ -250,17 +268,33 @@ impl FeltClientThread {
...
+                    let mut pending_cookies: Vec<String> = Vec::new();
+
                     loop {
                         if notify_restart_internal.load(Ordering::Relaxed) {
                             notify_restart_internal.store(false, Ordering::Relaxed);
                             trace!("FeltClientThread::felt_client::ipc_loop(): restart notification required!");
                             felt_client.notify_restart();
                         }
 
+                        if pending_cookies.len() > 0 && profile_ready_internal.load(Ordering::Relaxed) {
```
Mossop, line 280: This relies on us receiving an IPC message after profile_ready_internal is set but before the cookie is needed. Do we have something that ensures this?

An alternative could be to put `pending_cookies` behind a mutex and have the observer inject the cookies immediately

lissyx: `profile_ready_internal` is set when `profile-after-change` happens, so I am unsure where there is a risk of a race here. If we receive the ipc cookie message before, it's queued and when the observer sets the value this code here will dispatch and the next ipc message will directly dispatch to main thread as well

Mossop: Right, the sequence is:

1. We receive various cookies to set
2. We observe `profile-after-change`
3. On the next IPC message we add the cookies.

I don't know how frequently we are sending IPC messages in general. What IPC messages are we sending after `StartupReady` that will trigger the cookie insertion? If the user attempts to load a page between 2 and 3 then they won't have the right cookies.

lissyx: This block is in a loop and there's a 250ms timeout on the `rx.try_recv()` so it will be evaluated early?

Mossop: Ah ok I didn't know there was a timeout.

Mossop (approved):

PR #140, opened 2025-11-12
manage console tokens and expiration in Rust

fiji-flo:


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -117,6 +117,11 @@ export class FeltProcessParent extends JSProcessActorParent {
     this.firefox
       .then(async () => {
         this.sendPrefsToFirefox();
+        Services.felt.sendTokens(
```
lissyx: If we already have set the tokens (we do so in `ensureTokenData()`, why would we to get them again from `ConsoleClient` here?

browser/extensions/felt/rust/nsIFelt.idl
```diff
@@ -43,4 +43,19 @@ interface nsIFelt : nsISupports {
 
   [must_use]
   void sendRestartForced();
+
+  [must_use]
+  void sendTokens(in ACString access_token, in ACString refresh_token, in unsigned long long expires_at);
```
lissyx: I dont think this should be the signature, we should rather have `sendTokens()` and send what we have set via `setTokens()`. Is there any usecase where it makes sense to pass different tokens than the ones we have set?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -171,7 +140,7 @@ export const ConsoleClient = {
    * @returns {string}
    */
   get refreshTokenBackup() {
-    return Services.prefs.getStringPref(PREFS.REFRESH_TOKEN, "");
+    return Services.felt.getRefreshToken();
```
lissyx: This and `_ensureValidSession()` reflecting to the Rust XPCOM for checking lifetime makes it like the xpcom is taking ownership of the token, which is not a bad idea. But if we go down this route, then we should remove storing the values in the `ConsoleClient` and forward the calls to the XPCOM?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: We might want to send that before the prefs from above, but it's not really important for this message I think

browser/extensions/felt/rust/src/components.rs
```diff
@@ -106,6 +108,81 @@ impl FeltXPCOM {
...
+    ) -> nserror::nsresult {
+        let access_token = unsafe { (*access_token).to_string() };
+        let refresh_token = unsafe { (*refresh_token).to_string() };
+        trace!(
+            "FeltXPCOM::SendTokens ({} {} {})",
+            access_token,
+            refresh_token,
+            expires_at
+        );
+        self.send(FeltMessage::Tokens(Tokens {
```
lissyx: This might be a bit risky future wise, from talking with nika there was no guarantee in the serialization format, so if we have to support multiple versions we might be shooting ourselves in the foot. That's also why the cookie are serialized as string.

browser/extensions/felt/rust/src/message.rs
```diff
@@ -14,6 +16,7 @@ pub enum FeltMessage {
     StringPreference((String, String)),
     IntPreference((String, i32)),
     StartupReady,
+    Tokens(Tokens),
```
lissyx: We might want to rather serialize `Tokens` to a set of strings we have control over, `(String, String, u64)`?

browser/extensions/felt/rust/src/utils.rs
```diff
@@ -5,13 +5,27 @@
...
+use serde::{Deserialize, Serialize};
+use std::sync::{Arc, LazyLock, RwLock};
 use std::{ffi::CString, future::Future};
 use time::OffsetDateTime;
 use xpcom::interfaces::{nsICookie, nsICookieManager, nsIObserverService, nsIPrefBranch};
 use xpcom::RefPtr;
 
 use log::trace;
 
+#[derive(Default, Debug, Serialize, Deserialize)]
```
lissyx, line 17: Or I'm never sure here of the behavior, is serde going to serialize properly to `(String, String, u64)` in a future-proof way?

lissyx (changes requested):
Mostly I'd like we simplify the XPCOM interface or we explain why we have set and send both with the token when we store them in so many places :)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -73,32 +73,17 @@ class InvalidAuthError extends Error {
...
-    issuedAtSec = Math.floor(ChromeUtils.now() / 1000)
-  ) {
+  constructor(accessToken, refreshToken, expiresAtSec, tokenType = "Bearer") {
     this._accessToken = accessToken;
     this._refreshToken = refreshToken;
-    this._expiresInSec = expiresInSec;
-    this._tokenType = tokenType;
-    this._issuedAtSec = issuedAtSec;
+    this._expiresAtSec = expiresAtSec;
+    this.tokenType = tokenType;
```
```suggestion
    this._tokenType = tokenType;
```

1rneh (commented):
We've discussed the feedback offline and will provide a more thorough review in the next iteration.

lissyx (approved):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -350,11 +236,9 @@ export const ConsoleClient = {
    * @returns {Promise<any>} Parsed JSON response body.
    */
   async _get(path, { _didRefresh = false } = {}) {
-    await this._ensureValidSession();
-
     const headers = new Headers({});
-    const { tokenType, accessToken } = this.tokenData;
-    headers.set("Authorization", `${tokenType} ${accessToken}`);
+    const accessToken = await this.getAccessToken();
+    headers.set("Authorization", `Bearer ${accessToken}`);
```
1rneh, line 244: The console returns a token type. But we hardcode `Bearer` here to avoid having another xpcom call such as `getTokenType`?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -375,20 +259,22 @@ export const ConsoleClient = {
...
+   * @returns {Promise<string>}
    */
-  async _ensureValidSession() {
-    const td = this.tokenData;
-    if (!td?.accessToken || td.isExpiringSoon()) {
+  async getAccessToken() {
+    let accessToken = Services.felt.getAccessTokenIfValid();
+    if (!accessToken) {
       await this._refreshSession();
+      accessToken = Services.felt.getAccessTokenIfValid();
```
1rneh, line 282: I had to think for a moment why we are calling `Services.felt.getAccessTokenIfValid();` again. Maybe add a small comment, such as:
```suggestion
    // ConsoleClient does not cache the access token since felt owns the token state.
    // So after refreshing we must retrieve the updated token from felt.
    accessToken = Services.felt.getAccessTokenIfValid();
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -460,12 +345,8 @@ export const ConsoleClient = {
       }
 
       const t = await res.json();
-      this.ensureTokenData(t);
-
-      Services.prefs.setStringPref(
-        PREFS.REFRESH_TOKEN,
-        this.tokenData.refreshToken
-      );
+      const { access_token, refresh_token, expires_in } = t;
```
1rneh: Nit: `const { access_token, refresh_token, expires_in } = await res.json();`

1rneh (approved):
r+ for the JS side! 

Woop, no more tokens in prefs! 🎉

browser/components/enterprise/modules/ConsoleClient.sys.mjs
fiji-flo: Yes, I return the token type because it's in a standard response but we don't want to support anything else so I'd rather hard code it.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
fiji-flo: I think that's more distracting post merge when you read the new code without remembering the old way.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
fiji-flo: 👍

PR #150, opened 2025-11-17
Add enterprise badge and panel

1rneh:


browser/base/content/browser-init.js
```diff
@@ -185,6 +185,10 @@ var gBrowserInit = {
       }
     });
 
+    if (AppConstants.MOZ_ENTERPRISE) {
+      gEnterpriseHandler.init();
+    }
+
```
Mossop: Please initialize this using the `browser-window-domcontentloaded` category registration

browser/base/content/browser-sets.inc
```diff
@@ -405,4 +408,6 @@
          modifiers="accel,alt"
          internal="true"/>
 #endif
+#ifdef MOZ_ENTERPRISE
+
```
Mossop: This looks like a mistake? The ifdef never ends and will remove the closing tag in non-enterprise builds.

browser/components/enterprise/content/EnterpriseHandler.js
```diff
@@ -0,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+// Declare var to avoid redeclaration syntax error
```
Mossop: I'm interested, where were you seeing this error?

browser/components/enterprise/content/EnterpriseHandler.js
```diff
@@ -0,0 +1,58 @@
...
+    await this.initUser();
+    this.updateBadge();
+  },
+
+  async initUser() {
+    const { ConsoleClient } = ChromeUtils.importESModule(
+      "resource:///modules/enterprise/ConsoleClient.sys.mjs"
+    );
+    const { name, email, picture } = await ConsoleClient.getLoggedInUserInfo();
+    this._signedInUser = { name, email, pictureURL: picture };
```
Mossop: Seems potentially confusing to use different property keys here, do we need to?

browser/components/enterprise/content/EnterpriseHandler.js
```diff
@@ -0,0 +1,58 @@
...
+
+    await this.initUser();
+    this.updateBadge();
+  },
+
+  async initUser() {
+    const { ConsoleClient } = ChromeUtils.importESModule(
+      "resource:///modules/enterprise/ConsoleClient.sys.mjs"
+    );
+    const { name, email, picture } = await ConsoleClient.getLoggedInUserInfo();
```
Mossop: What should we do if this call fails?

browser/components/enterprise/content/enterprise-badge.inc.xhtml
```diff
@@ -0,0 +1,15 @@
...
+<html:link rel="stylesheet" href="chrome://browser/content/enterprise/enterprise-badge.css" />
+<html:link rel="localization" href="preview/enterprise.ftl" />
+
+<toolbarbutton id="enterprise-badge-toolbar-button" class="toolbarbutton-1" data-l10n-id="enterprise-toolbar-button" removable="false">
+    <hbox>
+        <div class="enterprise-badge">
+            <image id="enterprise-company-logo" />
+            <image id="enterprise-user-icon" />
+        </div>
+    </hbox>
```
Mossop: Use 2-space indent please

browser/themes/shared/customizableui/panelUI-shared.css
Mossop, line 1: Looks like you've reformatted this somehow. Can you remove the irrelevant changes?

Mossop (changes requested):
I'd like to look again after you've fixed these and the lint errors

browser/base/content/browser-init.js
1rneh: In order to register it this way, I also need to convert `EnterpriseHandler.js` from a global script to a system module, right?

browser/components/enterprise/content/EnterpriseHandler.js
1rneh: We could either change the key in the payload or simply keep `picture` although it's an URL. I don't have a strong preference.

@fiji-flo, does the console simply forward this from the sso provider?

1rneh: Update: 
Yes, the console forwards a subset of the json from the sso provider. So likely this key might be different across sso providers, right? 
@fiji-flo Could the console make sure it always sends back the same keys and change the `picture` key to `pictureURL`?

browser/components/enterprise/content/EnterpriseHandler.js
1rneh: Nice catch! This is a workaround leftover in which I didn't use a global script. The issue was that `DOMContentLoaded` was dispatched twice for the browser and `const EnterpriseHandler` (not a system module) was getting declared twice throwing a declaration error.

1rneh: Edit: If you take the patch as it is and declare `EnterpriseHandler` with `const` you still get a redeclaration error.

browser/components/enterprise/content/EnterpriseHandler.js
1rneh: We could also display a fallback icon in the badge or minimize badge to just displaying the company logo, and a placeholder text in place of the email address in the panel.

browser/base/content/browser-init.js
Mossop: Oh that's true. Ok, you can keep it this way if you prefer. Though a system module would have the benefit of only needing to initialize and do the API call once per browsing session.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -0,0 +1,75 @@
...
+  updateBadge() {
+    const userIcon = this._document.querySelector("#enterprise-user-icon");
+    userIcon.setProperty(
+      "list-style-image",
+      `url(${this._signedInUser.pictureURL})`
+    );
+  },
+
+  openPanel(element, event) {
+    this._document.ownerGlobal.PanelUI.showSubView(
```
Mossop: `this._document` is just the most recently opened browser window, not necessarily the window that the panel is in. Use `element.ownerDocument` instead.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -0,0 +1,75 @@
...
+   * @typedef {object} User
+   * @property {string} name name
+   * @property {string} email email
+   * @property {string} pictureUrl picture url
+   */
+  _signedInUser: null,
+  _document: null,
+
+  async init(window) {
+    this._document = window.document;
```
Mossop: Since this is a global singleton now it would be better not to store the most recently opened browser window here. Instead we will need to pass the relevant window/document to the functions that need one.

Mossop (changes requested):

browser/base/content/browser-init.js
1rneh: Converted 👍

browser/base/content/browser-sets.inc
```diff
@@ -405,4 +408,5 @@
          modifiers="accel,alt"
          internal="true"/>
 #endif
+
```
Mossop: Let's not add unrelated changes

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -0,0 +1,67 @@
...
+      );
+    }
+  },
+
+  updateBadge(window) {
+    const userIcon = window.document.querySelector("#enterprise-user-icon");
+    userIcon.style.setProperty(
+      "list-style-image",
+      `url(${this._signedInUser.pictureUrl})`
+    );
```
Mossop, line 51: Can you either test that `this._signedInUser` was set or file a follow-up to handle the error case.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -0,0 +1,67 @@
...
+  openPanel(element, event) {
+    element.ownerGlobal.PanelUI.showSubView(
+      "panelUI-enterprise",
+      element,
+      event
+    );
+    const document = element.ownerDocument;
+    const email = document.querySelector(".panelUI-enterprise__email");
+    if (!email.textContent) {
+      email.textContent = this._signedInUser.email;
```
Mossop, line 73: Again either handle the error case or file the follow-up.

Mossop (approved):

browser/components/enterprise/content/EnterpriseHandler.js
fiji-flo: I'd prefer to keep picture. This comes from https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse

browser/components/enterprise/content/EnterpriseHandler.js
1rneh: Going with user icon and email hiding for now.

PR #163, opened 2025-11-19
Email addresses must be URL encoded.

gcp:
And probably the device posture is also not guaranteed to be URL safe?

### DIFF
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -247,7 +247,9 @@ export const ConsoleClient = {
   constructSsoLoginURI(email, devicePostureToken) {
     const url = this.consoleBaseURI;
     url.pathname = this._paths.SSO;
-    url.search = `target=browser&email=${email}&devicePostureToken=${devicePostureToken}`;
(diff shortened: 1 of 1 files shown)


lissyx (approved):
I'm not sure there's a real need for device posture token, but it cannot be bad?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -246,7 +246,9 @@ export const ConsoleClient = {
   constructSsoLoginURI(email, devicePostureToken) {
     const url = this.consoleBaseURI;
     url.pathname = this._paths.SSO;
-    url.search = `target=browser&email=${email}&devicePostureToken=${devicePostureToken}`;
+    url.search = `target=browser&email=${encodeURIComponent(
+      email
+    )}&devicePostureToken=${encodeURIComponent(devicePostureToken)}`;
```
1rneh: `URLSearchParams` could do the encoding for us.

```suggestion
    url.searchParams.set("target", "browser");
    url.searchParams.set("email", email);
    url.searchParams.set("devicePostureToken", devicePostureToken);
```

1rneh (approved):
Good catch!

PR #164, opened 2025-11-19
Fxci enterprise tests

lissyx:


taskcluster/gecko_taskgraph/decision.py
```diff
@@ -116,7 +116,8 @@
         "target_tasks_method": "default",
     },
     "enterprise-firefox": {
-        "target_tasks_method": "enterprise_firefox_tasks",
+        # "target_tasks_method": "enterprise_firefox_tasks",
```
gcp: Remove stale stuff.

taskcluster/gecko_taskgraph/transforms/enterprise_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+            this_task["label"] = f"enterprise-test-{test_type}-{this_name}"
+            dep = get_primary_dependency(config, this_task)
+            assert dep
+
+            inherit_treeherder_from_dep(this_task, dep)
+
+            th_group = dep.task["extra"]["treeherder"]["symbol"].replace("Bent", "Tent")
+            th_symbol = f"{test_type}-{dep.task['extra']['treeherder']['symbol']}"
+            task["treeherder"]["symbol"] = f"{th_group}({th_symbol})"
```
gcp: I think this should have been `this_task`, i.e. what you actually yield back? Not the template?

taskcluster/kinds/enterprise-test/kind.yml
```diff
@@ -0,0 +1,170 @@
...
+                      type: directory
+                      path: build/src/artifacts/
+                macosx.*:
+                    - name: public/build
+                      type: directory
+                      path: artifacts/
+                linux.*:
+                    - name: public/build
+                      type: directory
+                      path: builds/worker/artifacts/
```
gcp: `builds` or `/builds`?

third_party/python/taskcluster_taskgraph/taskgraph/run-task/run-task
```diff
@@ -1341,21 +1341,14 @@ def main(args):
...
+    # Install Python requirements after fetches in case tasks want to use
+    # fetches to grab dependencies.
+    install_pip_requirements(repositories)
 
-        return run_command(b"task", task_args, cwd=args.task_cwd)
-    finally:
-        fetches_dir = os.environ.get("MOZ_FETCHES_DIR")
-        if fetches_dir and os.path.isdir(fetches_dir):
-            print_line(b"fetches", b"removing %s\n" % fetches_dir.encode("utf-8"))
-            remove(fetches_dir)
```
gcp, line 1357: I think this keeps all fetched stuff around? Is that intended?

gcp (commented):
I think you need a real reviewer for this, who knows about taskcluster/treeherder

third_party/python/taskcluster_taskgraph/taskgraph/run-task/run-task
lissyx: this one was provided by @ahal

taskcluster/kinds/enterprise-test/kind.yml
lissyx: both works but I'll fix :)

lissyx:
> I think you need a real reviewer for this, who knows about taskcluster/treeherder

yep but i wanted that some people on our team had a look as well

taskcluster/gecko_taskgraph/transforms/enterprise_test.py
lissyx: that's right

.taskcluster.yml
```diff
@@ -354,7 +354,7 @@ tasks:
               ownTaskId: {$eval: as_slugid("decision_task")}
           in:
               $if: >
-                tasks_for == "github-push" && head_branch == "refs/heads/enterprise-main"
+                tasks_for == "github-push" && head_branch == "refs/heads/enterprise-main-tests"
```
lissyx: This will not be part of the final merge

taskcluster/gecko_taskgraph/transforms/enterprise_test.py
```diff
@@ -0,0 +1,101 @@
...
+                item_name=this_task["name"],
+                **{"test-platform": platform},
+            )
+
+            resolve_keyed_by(
+                item=this_task,
+                field="run.command",
+                item_name=this_task["name"],
+                **{"test-platform": platform},
+            )
```
ahal, line 87: suggestion: Loop over all the field names

taskcluster/kinds/enterprise-test/kind.yml
```diff
@@ -0,0 +1,162 @@
...
+        description: Testing FELT device posture behavior
+        attributes:
+            enterprise_test_file: test_felt_device_posture.py
+            enterprise_test_slug: feltDevicePosture
+
+    felt-browser-signout:
+        description: Testing signout from browser and FELT takes over
+        attributes:
+            enterprise_test_file: test_felt_browser_signout.py
+            enterprise_test_slug: feltSignout
```
ahal, line 162: Is it worth having a single task per test file? Seems like it would be a lot more efficient to run them all in the same task. Or do they take a really long time or something?

taskcluster/kinds/enterprise-test/kind.yml
```diff
@@ -0,0 +1,162 @@
...
+    run-on-projects: ["enterprise-firefox"]
+    run-on-repo-type: ["git"]
+    always-target: true
+    from-deps:
+        group-by: single-with-filters
+    attributes:
+        retrigger: true
+    treeherder:
+        kind: test
+        tier: 2
```
ahal, line 36: You're missing treeherder symbols

taskcluster/kinds/enterprise-test/kind.yml
```diff
@@ -0,0 +1,162 @@
...
+                    APP_PATH="$(dirname "$(dirname "$(dirname "$(find /Volumes/ -type f -name "firefox")")")")" && APP_NAME=$(basename "$APP_PATH") &&
+                    cp -a "$APP_PATH" $MOZ_FETCHES_DIR &&
+                    hdiutil unmount -force -whole "$MOUNT_DIR" &&
+                    RUNTIME_PATH=$(find "$MOZ_FETCHES_DIR/$APP_NAME" -type f -name "firefox") &&
+                    ./mach enterprise-tests --firefox-bin="$RUNTIME_PATH" --geckodriver-bin=$MOZ_FETCHES_DIR/geckodriver testing/enterprise/{enterprise_test_file}
+
+                linux.*: >-
+                    source /builds/worker/scripts/xvfb.sh &&
+                    start_xvfb '1600x1200x24' 0 &&
+                    ./mach enterprise-tests --firefox-bin=$MOZ_FETCHES_DIR/firefox/firefox --geckodriver-bin=$MOZ_FETCHES_DIR/geckodriver testing/enterprise/{enterprise_test_file}
```
ahal, line 59: Instead of interpolating `{enterprise_test_file}` in the transforms, you could use `taskgraph.transforms.task_context`:
https://taskcluster-taskgraph.readthedocs.io/en/stable/reference/transforms/task_context.html#from-object

It's mostly personal preference, so consider this optional. But I typically like to keep the transforms as simple as possible, I find it makes things easier to understand in the long run.

taskcluster/kinds/enterprise-test/kind.yml
```diff
@@ -0,0 +1,162 @@
...
+                      type: directory
+                      path: build/src/artifacts/
+                macosx.*:
+                    - name: public/build
+                      type: directory
+                      path: artifacts/
+                linux.*:
+                    - name: public/build
+                      type: directory
+                      path: /builds/worker/artifacts/
```
ahal, line 113: I think rather than using `by-test-platform` for nearly almost every config, it's better to shift the definitions such that you define three tasks, one each for linux, windows and mac. Then you can use the [matrix](https://taskcluster-taskgraph.readthedocs.io/en/stable/reference/transforms/matrix.html) transforms to handle the duplication across test files. You might need to do a bit of normalization so e.g the slugs, task names and test files are more consistent.

I say this because it seems tasks are much more similar across test files than they are across platforms.

taskcluster/gecko_taskgraph/transforms/enterprise_test.py
ahal: Though if you follow my suggestion in a later comment, you'll be able to remove all these calls entirely.

ahal (changes requested):

taskcluster/kinds/enterprise-test/kind.yml
ahal: Actually, if you use the `matrix` transforms like I suggested in a comment below, those could handle the substitution as well, so no need for `task_context`.

taskcluster/kinds/enterprise-test/kind.yml
lissyx: They are computed in the transform. I tried setting them here but it was painful and I needed to compute then in the transform anyway

taskcluster/kinds/enterprise-test/kind.yml
lissyx: Right, but this transform is just interim work until we get proper CI finished and start moving those tests to marionette.

taskcluster/kinds/enterprise-test/kind.yml
lissyx: No but I prefer to avoid the current limitation that one failing test breaks keeps us from seeing the others. The amount of tests is not going to grow much and the goal is that in a couple of weeks we can start moving to marionette. Pretty sure the extra cost here will be absorbed by decommissionning github actions

taskcluster/kinds/enterprise-test/kind.yml
lissyx: I tried the matrix option at first and it was a bit less clear IMHO

PR #165, opened 2025-11-19
Enterprise fix signout console payload

lissyx:


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -568,10 +568,17 @@ export const ConsoleClient = {
     // TODO: Assert or force-enable session restore?
 
     const res = await this._post(this._paths.SIGNOUT);
-    if (res) {
+    // Server should maybe return better JSON?
+    if (res == null) {
```
1rneh, line 572: In `_get` we only return if `res.ok` and otherwise we throw anyway.
```suggestion
    await this._post(this._paths.SIGNOUT);
```

testing/enterprise/felt_tests.py
```diff
@@ -279,7 +279,7 @@ def do_POST(self):
 
         elif path == "/sso/logout":
             self.check_auth()
-            m = json.dumps({})
+            m = json.dumps(None)
```
1rneh, line 282: No need to change this if we only check for `res.ok` (see above)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -568,10 +568,17 @@ export const ConsoleClient = {
...
       // After successful server-side logout clear local state and notify FELT.
       this.clearTokenData();
 
+      // Make sure we signal early enough to the system that FELT should take
+      // over. Relevant at least for macOS dock icon. Not having this would
+      // at least intermittently result in missing dock icon for FELT after
+      // signout.
+      Services.felt.makeBackgroundProcess(true);
```
1rneh: This sounds racy.

1rneh (changes requested):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: In this case `res.ok` is indeed `true`, so I'm unsure why we would do this

testing/enterprise/felt_tests.py
lissyx: Until the server side returns a proper JSON payload, we should return the same thing, otherwise we risk bugs like this one.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Maybe, but so far for me it's fixing all issues.

gcp: I'm confused that the comment says Felt should take over yet we're telling Felt to go to the background.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: This way we don't reflect the broken response of the server in the client, if all the client needs to know is if the request was successful. We likely receive an empty body anyway. But if you want to keep it, that's fine, we'll update it once the server response format changes.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: This code is ran on the browser side, but we rely on the `felt` XPCOM to do the work

1rneh: Can you pull this out of this PR? I'd like to understand the issue here instead of landing something that might break longterm.

1rneh (approved):
I'm approving the response format changes.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: I can but I 100% repro'd the issue before this change and I dont repo at all after.

PR #166, opened 2025-11-19
FELT: Ensure to be in background when signout

lissyx:


gcp:
This sort of explicitly relinquishes ownership of the icon by backgrounding Firefox first, then doing the logout, so Felt foregrounding itself succeeds. It sounds like the race potential comes from the process exit being an implicit backgrounding, which means that in theory this isn't needed, but we did in fact find that macOS may itself have bugs here: https://bugzilla.mozilla.org/show_bug.cgi?id=1952338#c2...where we solved a similar issue by inserting arbitrary sleeps. Yikes.

Based on this, I don't think we have anything better here for now.

gcp (approved):

PR #167, opened 2025-11-19
Enterprise: Use correct branding assets for macOS

lissyx:


lissyx:
@1rneh is already preparing this in a branding-related PR

PR #168, opened 2025-11-19
Bug 1998944 - Disable legacy Telemetry submission in Enterprise builds.

gcp:
As per discussion with @chutten


jonathanmendez (approved):
lgtm

PR #170, opened 2025-11-20
Connect ConsoleClient.signout to sign-out button

1rneh:


browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -81,6 +84,7 @@ export const EnterpriseHandler = {
     fxaBtn.hidden = true;
   },
 
-  // TODO: Open signout dialog
-  onSignOut() {},
+  onSignOut() {
+    lazy.ConsoleClient.signout().catch((e) => console.error(`Unable to signout the user: ${e}`));
```
lissyx: Do we want to surface this to higher levels? I dont think we have UX yet but might be worth filing a bug so we dont forget

testing/enterprise/test_felt_browser_signout.py
```diff
@@ -59,19 +55,45 @@ def get_private_cookies(self):
         )
         self._driver.set_context("content")
         return private_cookies
+    
+    def test_felt_3_browser_ui_state_when_user_is_logged_in(self, exp):
+        self.connect_child_browser()
+
+        whoami = self.felt_whoami()
```
lissyx, line 62: nit: maybe worth asserting what we receive here since we trust it later for assertions

lissyx (approved):

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: Yes, we'll display a signout dialog . I have a WIP ready that just needs a few more small adjustments. I add the `// TODO: Open signout dialog` comment back until it lands.

PR #173, opened 2025-11-20
Encrypt SQLite databases from mozStorageService

Not-Nik:
Encrypt SQLite database whose connections are created in mozStorageService. Cloned and `ATTACH`ed connections are also affected. The mechanism is behind a new `security.storage.keystore.enabled`. The keys for the connections are stored, encrypted by the SDR, in a new `bikeshed/keystore.enc`.

### DIFF
--- a/modules/libpref/init/StaticPrefList.yaml
+++ b/modules/libpref/init/StaticPrefList.yaml
@@ -18224,6 +18224,12 @@
   value: false
   mirror: always
 
+# Enable sqlite database encryption
+- name: security.storage.encryption.sqlite.enabled
+  type: bool
+  value: true
+  mirror: once
+
 - name: security.tls13.aes_128_gcm_sha256
   type: RelaxedAtomicBool
   value: true
--- a/security/keystore/KeyStorage.cpp
+++ b/security/keystore/KeyStorage.cpp
@@ -1,0 +1,60 @@
+/* -*- Mode: C++; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*-
+ * vim: sw=2 ts=2 et lcs=trail\:.,tab\:>~ :
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "KeyStorage.h"
+
+#include "Persist.h"
+
+#include "mozilla/Logging.h"
+#include "GMPUtils.h"
+#include "nsLocalFile.h"
+#include "ScopedNSSTypes.h"
+
+namespace mozilla::storage::key {
+void Shutdown() { ShutdownStorage(); }
+mozilla::LogModule* GetKeyStorageLog() {
+  static mozilla::LazyLogModule sLog("KeyStorage");
+
+  return sLog;
+}
+
+nsresult GetKeyByPath(const char* aPath, nsCString& aKey) {
+  nsCOMPtr<nsIFile> file = new nsLocalFile();
+  nsresult rv = file->InitWithNativePath(nsCString(aPath));
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  return GetKeyByFile(*file, aKey);
+}
+
+nsresult GetKeyByFile(nsIFile& aFile, nsCString& aKeyString) {
+  nsAutoString profile_path;
+  nsresult rv = GetCurrentProfilePath(profile_path);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  nsAutoString realFile;
+  rv = aFile.GetPath(realFile);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  nsCOMPtr<nsIFile> profile = new nsLocalFile();
+  rv = profile->InitWithPath(profile_path);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  nsAutoCString filename;
+  rv = aFile.GetRelativePath(profile, filename);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  MOZ_LOG(GetKeyStorageLog(), LogLevel::Debug,
+          ("Fetching key for %s", filename.get()));
+
+  UniqueSECItem key(::SECITEM_AllocItem(nullptr, nullptr, 0));
+  rv = FetchOrCreateKey(filename, key.get());
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  aKeyString = mozilla::ToHexString(key->data, key->len);
+
+  return NS_OK;
+}
+}  // namespace mozilla::storage::key
--- a/security/keystore/Persist.cpp
+++ b/security/keystore/Persist.cpp
@@ -1,0 +1,466 @@
+/* -*- Mode: C++; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*-
+ * vim: sw=2 ts=2 et lcs=trail\:.,tab\:>~ :
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "Persist.h"
+
+#include "mozilla/Base64.h"
+#include "mozilla/dom/quota/QuotaManager.h"
+#include "mozilla/MozPromise.h"
+#include "mozilla/SyncRunnable.h"
+#include "nsAppDirectoryServiceDefs.h"
+#include "nsCOMPtr.h"
+#include "nsCRTGlue.h"
+#include "nsTHashMap.h"
+#include "nsThreadUtils.h"
+#include "nsHashtablesFwd.h"
+#include "nsIToolkitProfileService.h"
+#include "nsIFile.h"
+#include "nsIProperties.h"
+#include "nsNetUtil.h"
+#include "nsNSSHelper.h"
+#include "NSSErrorsService.h"
+#include "nsServiceManagerUtils.h"
+#include "nsStreamUtils.h"
+#include "pk11sdr.h"
+#include "prerror.h"
+#include "prio.h"
+#include "ScopedNSSTypes.h"
+
+#include "KeyStorage.h"
+
+#define KEYSTORE_MAGIC "# mozilla secure key storage\n"
+#define KEYSTORE_PATH FILE_PATH_SEPARATOR "keystore.db"
+#define SYSTEM_KEY_NAME "system"
+
+namespace mozilla::storage::key {
+
+struct Key {
+  mozilla::UniqueSECItem key;
+  mozilla::UniqueSECItem iv;
+
+  Key() = default;
+
+  Key(mozilla::UniqueSECItem key, mozilla::UniqueSECItem iv)
+      : key(std::move(key)), iv(std::move(iv)) {}
+
+  Key(const Key& other)
+      : key(SECITEM_DupItem(other.key.get())),
+        iv(SECITEM_DupItem(other.iv.get())) {}
+
+  Key(Key&&) = default;
+
+  Key& operator=(Key&&) = default;
+};
+
+mozilla::StaticMutex sKeyMutex;
+MOZ_RUNINIT static nsTHashMap<nsCString, Key> sKeyMap;
+constinit static mozilla::UniquePK11SymKey sSystemKey;
+constinit nsString sProfilePath;
+
+void ShutdownStorage() {
+  mozilla::StaticMutexAutoLock lock(sKeyMutex);
+  sSystemKey.reset(nullptr);
+  sKeyMap.Clear();
+  sProfilePath.Truncate();
+}
+
+void SetCurrentProfilePath(const nsAString& aPath) { sProfilePath = aPath; }
+
+nsresult GetCurrentProfilePath(nsAString& aPath) {
+  if (NS_WARN_IF(sProfilePath.IsEmpty())) {
+    sProfilePath = NS_ConvertUTF8toUTF16(PR_GetEnv("XPCSHELL_TEST_PROFILE_DIR"));
+  }
+  aPath = sProfilePath;
+  return NS_OK;
+}
+
+/// Create path directory for keystore file
+nsresult GetKeyStorePath(nsAString& aPath) {
+  nsresult rv = GetCurrentProfilePath(aPath);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  aPath.Append(NS_LITERAL_STRING_FROM_CSTRING(KEYSTORE_PATH));
+
+  return NS_OK;
+}
+
+/// Load file contents into string
+nsresult LoadFileToString(const nsCOMPtr<nsIFile>& aFile,
+                          nsACString& aContents) {
+  nsCOMPtr<nsIInputStream> stream;
+  nsresult rv = NS_NewLocalFileInputStream(getter_AddRefs(stream), aFile.get());
+  // Manually return for fnf error, to avoid having a misleading error message
+  // in the logs. This is an expected case, that is handled elsewhere
+  if (rv == NS_ERROR_FILE_NOT_FOUND) return NS_ERROR_FILE_NOT_FOUND;
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  return NS_ConsumeStream(stream, UINT32_MAX, aContents);
+}
+
+/// Open file for appending and write string to it
+nsresult AppendStringToFile(const nsCOMPtr<nsIFile>& aFile,
+                            nsACString& aContents) {
+  nsCOMPtr<nsIOutputStream> stream;
+  nsresult rv = NS_NewLocalFileOutputStream(
+      getter_AddRefs(stream), aFile.get(),
+      PR_WRONLY | PR_CREATE_FILE | PR_APPEND, PR_IRUSR | PR_IWUSR);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  uint32_t count;
+  rv = stream->Write(aContents.Data(), aContents.Length(), &count);
+  NS_ENSURE_SUCCESS(rv, rv);
+  if (count != aContents.Length()) return NS_ERROR_FAILURE;
+
+  return NS_OK;
+}
+
+/// Import a single, encoded and encrypted key, and encoded IV as `path`
+nsresult ImportKey(const nsCString& aPath, const nsCString& aEncodedKey,
+                   const nsCString& aEncodedIV) {
+  sKeyMutex.AssertCurrentThreadOwns();
+
+  // Key and IV are encoded Base64 strings
+  nsCString encryptedKey, stringIV;
+
+  nsresult rv = mozilla::Base64Decode(aEncodedKey, encryptedKey);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  rv = mozilla::Base64Decode(aEncodedIV, stringIV);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  // Key and IV need to go into heap-allocated SECItems, because they may be
+  // stored into sKeyMap, which outlives encryptedKey/stringIV
+  mozilla::UniqueSECItem key(::SECITEM_AllocItem(nullptr, nullptr, 0));
+  SECStatus stat =
+      SECITEM_MakeItem(nullptr, key.get(), (unsigned char*)encryptedKey.Data(),
+                       encryptedKey.Length());
+  if (stat != SECSuccess) return MapSECStatus(stat);
+
+  mozilla::UniqueSECItem iv(::SECITEM_AllocItem(nullptr, nullptr, 0));
+  stat = SECITEM_MakeItem(nullptr, iv.get(), (unsigned char*)stringIV.Data(),
+                          stringIV.Length());
+  if (stat != SECSuccess) return MapSECStatus(stat);
+
+  // Keystore contains one (1) "system" key which encrypts all other keys
+  if (aPath == SYSTEM_KEY_NAME) {
+    mozilla::UniqueSECItem result(::SECITEM_AllocItem(nullptr, nullptr, 0));
+
+    // "System" key is encrypted through the SDR
+    stat = PK11SDR_Decrypt(key.get(), result.get(), nullptr);
+    if (stat != SECSuccess) return MapSECStatus(stat);
+
+    mozilla::UniquePK11SlotInfo slot(PK11_GetInternalSlot());
+    if (!slot) {
+      return NS_ERROR_FAILURE;
+    }
+
+    sSystemKey = mozilla::UniquePK11SymKey(
+        PK11_ImportSymKey(slot.get(), CKM_AES_GCM, PK11_OriginUnwrap,
+                          CKA_ENCRYPT | CKA_DECRYPT, result.get(), nullptr));
+
+    if (!sSystemKey) {
+      return NS_ERROR_FAILURE;
+    }
+  } else {
+    sKeyMap.InsertOrUpdate(aPath, Key{std::move(key), std::move(iv)});
+  }
+
+  return NS_OK;
+}
+
+/// Load keys from keystore file to memory
+nsresult LoadKeysFromDisk() {
+  sKeyMutex.AssertCurrentThreadOwns();
+
+  nsCString fileContents;
+
+  // Get the file to the key store in the profile and turn it into a file ref
+  nsAutoString filePath;
+  nsresult rv = GetKeyStorePath(filePath);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  if (!sKeyMap.IsEmpty()) {
+    // Keys were loaded on another thread during GetKeyStorePath
+    return NS_OK;
+  }
+
+  nsCOMPtr<nsIFile> file;
+  rv = NS_NewLocalFile(filePath, getter_AddRefs(file));
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  // Load all the file contents into one string
+  rv = LoadFileToString(file, fileContents);
+  if (rv == nsresult::NS_ERROR_FILE_NOT_FOUND) {
+    // Non existent keystore is OK and will be handled outside of this
+    // function
+    return NS_OK;
+  }
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  // Verify keystore file
+  if (fileContents.Find(KEYSTORE_MAGIC) != 0) {
+    return NS_ERROR_INVALID_SIGNATURE;
+  }
+
+  // Go through each line
+  for (const auto& line : fileContents.Split('\n')) {
+    int32_t delimiter1 = line.Find(":"_ns.View());
+    int32_t delimiter2 = line.RFind(":"_ns.View());
+    // Ignore invalid or incomplete lines
+    if (delimiter1 == kNotFound || delimiter1 == delimiter2) continue;
+
+    // Each line holds a path/identifier, a key and an IV
+    nsCString path, encodedKey, encodedIV;
+    path.Assign(line.Data(), delimiter1);
+
+    encodedKey.Assign(line.Data() + delimiter1 + 1,
+                      delimiter2 - delimiter1 - 1);
+
+    encodedIV.Assign(line.Data() + delimiter2 + 1,
+                     line.Length() - delimiter2 - 1);
+
+    rv = ImportKey(path, encodedKey, encodedIV);
+    NS_ENSURE_SUCCESS(rv, rv);
+  }
+  return NS_OK;
+}
+
+/// Create "system" key
+/// `keyOut` will contain the SDR encrypted key bytes
+/// `SYSTEM_KEY` will the a AES-GCM PK11SymKey created from those bytes
+nsresult CreateSystemKey(Key& aKeyOut) {
+  sKeyMutex.AssertCurrentThreadOwns();
+
+  // Every key is 32 bytes long
+  mozilla::UniqueSECItem key =
+      mozilla::UniqueSECItem(::SECITEM_AllocItem(nullptr, nullptr, 32));
+  if (!key) return NS_ERROR_FAILURE;
+
+  // Key data is random
+  SECStatus stat = PK11_GenerateRandom(key->data, key->len);
+  if (stat != SECSuccess) return MapSECStatus(stat);
+
+  mozilla::UniqueSECItem encryptedKey(nullptr);
+
+  mozilla::UniquePK11SlotInfo slot(PK11_GetInternalSlot());
+  if (!slot) {
+    return NS_ERROR_FAILURE;
+  }
+
+  sSystemKey = mozilla::UniquePK11SymKey(
+      PK11_ImportSymKey(slot.get(), CKM_AES_GCM, PK11_OriginUnwrap,
+                        CKA_ENCRYPT | CKA_DECRYPT, key.get(), nullptr));
+
+  if (!sSystemKey) {
+    return NS_ERROR_FAILURE;
+  }
+
+  // Use the default SDR key
+  SECItem keyid = {siBuffer, nullptr, 0};
+
+  // PK11SDR_EncryptWithMechanism will allocate the needed buffer in SECItem
+  encryptedKey.reset(::SECITEM_AllocItem(nullptr, nullptr, 0));
+
+  stat = PK11SDR_EncryptWithMechanism(nullptr, &keyid, CKM_AES_CBC, key.get(),
+                                      encryptedKey.get(), nullptr);
+  if (stat != SECSuccess) return MapSECStatus(stat);
+
+  aKeyOut =
+      Key{std::move(encryptedKey),
+          mozilla::UniqueSECItem(::SECITEM_AllocItem(nullptr, nullptr, 0))};
+
+  return NS_OK;
+}
+
+/// Create a new key
+nsresult CreateKey(nsAutoCString& aIdentifier, Key& aKeyOut) {
+  // Every key is 32 bytes long
+  mozilla::UniqueSECItem key =
+      mozilla::UniqueSECItem(::SECITEM_AllocItem(nullptr, nullptr, 32));
+  if (!key) return NS_ERROR_FAILURE;
+
+  // Key data is random
+  SECStatus stat = PK11_GenerateRandom(key->data, key->len);
+  if (stat != SECSuccess) return MapSECStatus(stat);
+
+  mozilla::UniqueSECItem encryptedKey(nullptr);
+  // IV exists regardless of wether the key is "system"
+  mozilla::UniqueSECItem iv(::SECITEM_AllocItem(nullptr, nullptr, 12));
+  if (!iv) return NS_ERROR_FAILURE;
+
+  // IV data is also random
+  stat = PK11_GenerateRandom(iv->data, iv->len);
+  if (stat != SECSuccess) return MapSECStatus(stat);
+
+  CK_GCM_PARAMS gcm_params;
+  gcm_params.pIv = (CK_BYTE_PTR)iv->data;
+  gcm_params.ulIvLen = iv->len;
+  gcm_params.ulIvBits = iv->len * 8;
+  gcm_params.pAAD = (CK_BYTE_PTR)aIdentifier.get();
+  gcm_params.ulAADLen = aIdentifier.Length();
+  gcm_params.ulTagBits = 128;
+
+  SECItem gcm_item;
+  gcm_item.type = siBuffer;
+  gcm_item.data = (unsigned char*)&gcm_params;
+  gcm_item.len = sizeof(gcm_params);
+
+  // PK11_Encrypt needs an existing buffer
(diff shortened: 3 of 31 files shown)


security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,109 @@
...
+    nsCString encryptedKey;
+    rv = mozilla::Base64Decode(encodedKey, encryptedKey);
+    NS_ENSURE_SUCCESS(rv, rv);
+
+    SECItem data = {}, result = {};
+    data.data = (unsigned char*)encryptedKey.Data();
+    data.len = encryptedKey.Length();
+    data.type = siBuffer;
+
+    PK11SDR_Decrypt(&data, &result, nullptr);
```
gcp: This leaks memory.

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,109 @@
...
+    PK11_GenerateRandom(key, 32);
+
+    keyString.AssignASCII(ArrayToHexString(key, 32));
+
+    SECItem data = {}, result = {}, keyid = {siBuffer, nullptr, 0};
+    data.data = key;
+    data.len = 32;
+    data.type = siBuffer;
+
+    PK11SDR_Encrypt(&keyid, &data, &result, nullptr);
```
gcp: Same, this leaks memory.

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,109 @@
...
+nsresult GetKeyForPath(const char* aPath, nsCString& key) {
+  nsIFile* file = MakeAndAddRef<nsLocalFile>().take();
+  nsresult rv = file->InitWithNativePath(nsCString(aPath));
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  return GetKeyForFile(file, key);
+}
+
+nsresult GetKeyForFile(nsIFile* aFile, nsCString& keyString) {
+  nsAutoCString telemetryFilename;
```
gcp: Weird naming here.

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,109 @@
...
+  nsresult rv = aFile->GetNativeLeafName(telemetryFilename);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  rv = WaitForTokenDB();
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  nsCString encodedKey;
+  rv = LoadKeyFromDisk(telemetryFilename, encodedKey, 112);
```
gcp: We have multiple `metadata.sqlite` and `data.sqlite` in the profile so this won't work, needs keying with the full absolute path.

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,133 @@
...
+  nsCOMPtr<nsIFile> file;
+  rv = NS_NewLocalFile(filePath, getter_AddRefs(file));
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  bool needs_magic;
+  file->Exists(&needs_magic);
+
+  nsCString data;
+  if (!needs_magic) {
+    data.Append(KEYSTORE_MAGIC);
```
gcp: This makes no sense (if we don't need the magic, write the magic), but I think the code does the right thing and the bool is named the wrong way around.

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,109 @@
...
+  nsCOMPtr<nsIPK11TokenDB> tokenDB =
+      do_GetService("@mozilla.org/security/pk11tokendb;1", &rv);
+
+  nsIPK11Token* internalToken;
+  tokenDB->GetInternalKeyToken(&internalToken);
+
+  bool logged_in, needs_login;
+  rv = internalToken->NeedsLogin(&needs_login);
+  NS_ENSURE_SUCCESS(rv, rv);
+  while (needs_login && (rv = internalToken->IsLoggedIn(&logged_in)) == NS_OK &&
```
gcp: This at the very least looks like a potential shutdown hang.

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,109 @@
...
+    data.data = (unsigned char*)encryptedKey.Data();
+    data.len = encryptedKey.Length();
+    data.type = siBuffer;
+
+    PK11SDR_Decrypt(&data, &result, nullptr);
+
+    keyString.AssignASCII(ArrayToHexString(result.data, result.len));
+  } else {
+    unsigned char key[32];
+    PK11_GenerateRandom(key, 32);
```
gcp: No error checks? `CVE-2008-0166`

modules/libpref/init/StaticPrefList.yaml
```diff
@@ -17637,6 +17637,12 @@
   value: true
   mirror: always
 
+# Enable keystore file for enterprise storage encryption.
+- name: security.storage.keystore.enabled
```
gcp: What happens if this pref is flipped on an existing profile?

What happens if the pref is flipped while Firefox is running?

security/keystore/KeyStorage.sys.mjs
```diff
@@ -0,0 +1,28 @@
...
+    let pk11token;
+    try {
+      pk11token = tokenDB.getInternalKeyToken();
+    } catch (e) {
+      console.error("KeyStorage.init: Error getting PK11 token: " + e);
+      return;
+    }
+
+    try {
+      pk11token.login(true);
```
gcp: I guess we don't actually want this (force pp login dialog?), so this couldn't land without @beurdouche 's follow up patches.

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+  return NS_ERROR_FAILURE;
+}
+
+nsresult GetOrCreateKeyFilePath(nsAString& path) {
+  nsresult rv = GetCurrentProfilePath(path);
+  NS_ENSURE_SUCCESS(rv, rv);
+  path.Append(NS_LITERAL_STRING_FROM_CSTRING("/bikeshed"));
+
+  PR_MakeDir(NS_ConvertUTF16toUTF8(path).get(), 0777);
```
gcp: Error checks missing.

storage/mozStorageConnection.cpp
```diff
@@ -1127,6 +1130,71 @@ nsresult Connection::initialize(nsIFile* aDatabaseFile) {
...
+    nsCOMPtr<nsIPrefService> prefs =
+        do_GetService(NS_PREFSERVICE_CONTRACTID, &rv);
+    if (NS_FAILED(rv)) return rv;
+
+    rv = prefs->GetBranch(nullptr, getter_AddRefs(mPrefBranch));
+    if (NS_FAILED(rv)) return rv;
+  }
+
+  bool encryptionEnabled;
+  rv = mPrefBranch->GetBoolPref("security.storage.keystore.enabled",
```
gcp: This isn't allowed (off main thread prefs access), and will crash on startup:
```

[64381] Hit MOZ_CRASH(nsPrefBranch not thread-safe) at /Users/morbo/git/firefox/xpcom/base/nsISupportsImpl.cpp:43
#01: nsAutoOwningThread::AssertCurrentThreadOwnsMe(char const*) const[/Users/morbo/git/firefox/obj-aarch64-apple-darwin25.1.0/dist/Firefox EnterpriseDebug.app/Contents/MacOS/XUL +0x4014bc]
#02: nsPrefBranch::AddRef()[/Users/morbo/git/firefox/obj-aarch64-apple-darwin25.1.0/dist/Firefox EnterpriseDebug.app/Contents/MacOS/XUL +0x583cf8]
#03: mozilla::Preferences::GetBranch(char const*, nsIPrefBranch**)[/Users/morbo/git/firefox/obj-aarch64-apple-darwin25.1.0/dist/Firefox EnterpriseDebug.app/Contents/MacOS/XUL +0x5d320c]
#04: mozilla::storage::Connection::initializeSecure(nsIFile*)[/Users/morbo/git/firefox/obj-aarch64-apple-darwin25.1.0/dist/Firefox EnterpriseDebug.app/Contents/MacOS/XUL +0x113a27c]
```

storage/mozStorageConnection.cpp
```diff
@@ -2588,6 +2673,50 @@ Connection::CreateTable(const char* aTableName, const char* aTableSchema) {
...
+    rv = prefs->GetBranch(nullptr, getter_AddRefs(mPrefBranch));
+    NS_ENSURE_SUCCESS(rv, rv);
+  }
+
+  bool encryptionEnabled;
+  rv = mPrefBranch->GetBoolPref("security.storage.keystore.enabled",
+                                &encryptionEnabled);
+  if (NS_SUCCEEDED(rv) && encryptionEnabled) {
+    nsCString aDBKey;
+    key::GetKeyForPath(aPath, aDBKey);
```
gcp: Your IDL claims `aPath` may be in URI form, and if we drill down, you'll see this will cause `GetKeyForPath` to fail.

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,109 @@
...
+  while (needs_login && (rv = internalToken->IsLoggedIn(&logged_in)) == NS_OK &&
+         !logged_in);
+  NS_ENSURE_SUCCESS(rv, rv);
+  return NS_OK;
+}
+
+namespace mozilla::storage::key {
+nsresult GetKeyForPath(const char* aPath, nsCString& key) {
+  nsIFile* file = MakeAndAddRef<nsLocalFile>().take();
+  nsresult rv = file->InitWithNativePath(nsCString(aPath));
```
gcp, line 26: ...see remarks about what your own API spec says about aPath!

storage/mozStorageConnection.cpp
gcp: What makes this worse is that there is no error checking here, so we may end up encrypting with an empty key.

That would be bad.

gcp (changes requested):

gcp:
I think this wants tests as well. For Foxfooding we can deal with some flakyness but causing all SQLite databases to get corrupted will probably get old soon for the poor testers 😄

security/keystore/KeyStorage.cpp
Not-Nik: I'm hesitant to use a full absolute path, because there is a good chance someone will move their Firefox profile for whatever reason, and then we couldn't find the keys anymore.

We could turn the absolute paths into paths relative to the profile directory, but that could have a bit of a performance impact. Alternatively, is it really that bad if multiple databases share a key?

security/keystore/KeyStorage.sys.mjs
Not-Nik: Not on enterprise, no, but this wouldn't do anything here, because the token is already unlocked. I can remove it for this PR, but release needs some way to unlock the token once, for fetching the keys, and then possibly re-locking it once they are loaded.

security/keystore/KeyStorage.cpp
gcp: >I'm hesitant to use a full absolute path, because there is a good chance someone will move their Firefox profile for whatever reason

_whistles innocently_
https://bugzilla.mozilla.org/show_bug.cgi?id=259356#c230

Using a path anchored in the profile sounds fine.

>Alternatively, is it really that bad if multiple databases share a key?

Probably you should ask @simon-friedberger about the impact of this :-) If this is a pure key, probably not, but then why do we have this mapping in the first place. If this "key" is really an "IV+key" or similar, then I think it would be insecure, but I'm not the expert on this.

storage/mozStorageConnection.cpp
Not-Nik: Any idea how to resolve this? We could possibly have a global variable that's set once on startup, and only do proper switches between encrypted/unencrypted SQLite between sessions.

gcp: Yes, I think if you switch encryption mid-session that's going to be a pain to deal with anyway.

gcp: Alternatively you need to monitor the pref on the main thread and broadcast that encryption has to be switched on to any worker threads (via an atomic bool). If you can sensibly enable the encryption in the middle of processing that's fine, and it's definitely better if this kind of setting takes immediately without having to force a restart.

But it's clearly harder.

Not-Nik: Is there an easy way to fetch a pref on startup and set a global variable from it? The only thing I could think of is dispatching a runner to the main thread the first time we need the pref.

gcp: Look into `StaticPrefs::` and/or `mirror: once` plus `StaticPrefs::*AtStartup()` which do what you want.

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,349 @@
...
+        PK11_ImportSymKey(slot.get(), CKM_AES_GCM, PK11_OriginUnwrap,
+                          CKA_ENCRYPT | CKA_DECRYPT, key.get(), nullptr));
+
+    // Use the default SDR key
+    SECItem keyid = {siBuffer, nullptr, 0};
+
+    // PK11SDR_Encrypt will allocate the needed buffer in SECItem
+    encryptedKey.reset(new SECItem());
+
+    stat = PK11SDR_Encrypt(&keyid, key.get(), encryptedKey.get(), nullptr);
```
mozkeeler: This uses 3DES. You want `PK11SDR_EncryptWithMechanism`.

security/keystore/KeyStorage.cpp
mozkeeler, line 1: Is there a particular reason to have this in its own directory? These kinds of files tend to belong in `security/manager` (or at least a subdirectory therein).

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,349 @@
...
+    NS_ENSURE_SUCCESS(rv, rv);
+
+    rv = mozilla::Base64Decode(encodedIV, stringIV);
+    NS_ENSURE_SUCCESS(rv, rv);
+
+    // Key and IV need to go into SECItems to be useful
+    mozilla::UniqueSECItem key = mozilla::UniqueSECItem(new SECItem());
+    SECStatus stat = SECITEM_MakeItem(nullptr, key.get(),
+                                      (unsigned char*)encryptedKey.Data(),
+                                      encryptedKey.Length());
```
mozkeeler: Generally speaking, you probably don't need to do the whole `new SECItem() ... SECITEM_MakeItem(...)` - you can do something like `SECItem key = { siBuffer, (unsigned char*)encryptedKey.Data(), encryptedKey.Length() };` (assuming the NSS functions you call either copy the data or only borrow it for the length of the call, which is most of them).

security/keystore/Persist.cpp
mozkeeler, line 1: Some of this code seems very similar to `NSSKeyStore` - is there a reason to not re-use that? (currently it's only used as a fallback on Linux, but you could use it unconditionally here)

Not-Nik: For one the custom solution allows us some flexibility right away, only requiring the token to be unlocked once for loading all the keys. Later database open operations can then rely on the master key to do their work (create new key, decrypt existing key). Additionally, there is a plan to turn the keystore into the lockstore module which Benjamin has been working on.

security/keystore/Persist.cpp
Not-Nik: That particular example needs a "proper" SECItem (on the heap), because it is later stored in `KEY_MAP`, which is static and thus outlives `encryptedKey`. I already tried to reduce buffer allocation, and I'll move to a stack `SECItem` in the `path == SYSTEM_KEY_NAME` path, but if I missed anything else do let me know.

security/keystore/KeyStorage.cpp
beurdouche: I think the key management component will be more visible one ladder up in its own directory. Also it is neither PKI or TLS specific, so I would prefer to keep it there for now... : )

storage/mozStorageService.cpp
Not-Nik, line 585: Let's check for the pref here, instead of calling a `initializeSecure` which won't always open a secure database.

storage/mozStorageConnection.cpp
Not-Nik: This is no longer needed.

Not-Nik: This is no longer needed.

Not-Nik: Document why there is a new function, instead of just a boolean passed to `initialize` and why this doesn't use "readonly-immutable-nolock".

Not-Nik: No longer needed as well.

Not-Nik: No longer needed as well.

Not-Nik: Check the pref before calling `initializeSecure`.

Not-Nik: No longer needed.

Not-Nik: Add a comment on why this path is structured this way.

Not-Nik: Create a separate function for extracting URI path and query.

Not-Nik: This needs to go into all other connection initializers as well.

toolkit/components/places/Database.cpp
Not-Nik: This is no longer needed.

Not-Nik: Check if the path passed into this function is always just a path, and not a URI like in `Connection::AttachDatabase`.

security/keystore/KeyStorage.cpp
Not-Nik: Rename `telemetryFilename` to `filename`.

security/keystore/Persist.cpp
Not-Nik: Rename `LoadKeyFromDisk` to `FetchOrCreateKey`.

Not-Nik: Rename `InitialiseKeys` to `InitializeKeys`.

Not-Nik: Check the return value from `InitialiseKeys`.

Not-Nik: Create a function for key decryption.

Not-Nik: Is this the correct way of getting the profile directory outside of the main thread?

cc @mozkeeler @gcp

Not-Nik: Don't put `keystore.enc` into a separate directory.

Not-Nik: Rename `keystore.enc` to `keystore.db` since not all of the contents are encrypted.

Not-Nik: Extract the storage part from `CreateKey` and put it into `KEY_MAP` or `SYSTEM_KEY` at its call sites.

Not-Nik: Storing keys to files should go into a new, separate function.

Not-Nik: Rename `keyLine` to `keyEntry`.

Not-Nik (commented):
Several notes from Benjamin and me

toolkit/components/places/Database.cpp
Not-Nik: All the calls are 'pure' paths.

modules/libpref/Preferences.cpp
```diff
@@ -5902,7 +5902,7 @@ void MaybeInitOncePrefs() {
     // value.
     return;
   }
-  StaticMutexAutoLock lock(sOncePrefMutex);
+  // StaticMutexAutoLock lock(sOncePrefMutex);
```
Not-Nik: @gcp this lock is causing a deadlock. It seems like its not needed. [Bug 2014243](https://bugzilla.mozilla.org/show_bug.cgi?id=2014243)

storage/mozStorageConnection.cpp
```diff
@@ -2588,6 +2698,67 @@ Connection::CreateTable(const char* aTableName, const char* aTableSchema) {
...
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  rv = stmt->BindUTF8StringByName("path"_ns, uri);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  nsCOMPtr<mozIStoragePendingStatement> pendingStatement;
+  rv = stmt->ExecuteAsync(aCallback, getter_AddRefs(pendingStatement));
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  pendingStatement.forget(_handle);
```
gcp, line 2766: `_handle` to me implies this argument is unused (maybe written too much Rust). But I see the surrounding code uses it too 🤷

modules/libpref/init/StaticPrefList.yaml
```diff
@@ -17922,6 +17906,12 @@
   value: false
   mirror: always
 
+# Enable keystore file for enterprise storage encryption.
```
gcp: The reference to 'keystore' here is weird. We enable storage encryption. What's the keyfile? Why doesn't it have a separate setting? Can we have encryption without it? (That all just comes to mind reading that comment)

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "nsLocalFile.h"
+#include "ScopedNSSTypes.h"
+
+#include <iomanip>
+#include <sstream>
+#include <string>
+
+static std::string ArrayToHexString(const unsigned char* array,
+                                    unsigned int length) {
+  std::ostringstream oss;
```
gcp: We have C++20 now (as of 2 months ago 🤣 ), which means you can simplify this with https://en.cppreference.com/w/cpp/utility/format/format.html and https://en.cppreference.com/w/cpp/utility/format/format_to.html

gcp: That said, we have at least 6 implementations of this (e.g. https://searchfox.org/firefox-main/source/third_party/abseil-cpp/absl/strings/escaping.h#181) so maybe it's worth checking if one of them can be reused or it this should go into `/mfbt` so we have 7 but at least the next person can use one of them 😁

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+nsresult GetKeyForPath(const char* aPath, nsCString& key) {
+  nsIFile* file = MakeAndAddRef<nsLocalFile>().take();
+  nsresult rv = file->InitWithNativePath(nsCString(aPath));
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  return GetKeyForFile(file, key);
+}
+
+nsresult GetKeyForFile(nsIFile* aFile, nsCString& keyString) {
```
gcp: nit: we know aFile won't be null so it could be a ref

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+#define KEYSTORE_MAGIC "# mozilla secure key storage\n"
+#define SYSTEM_KEY_NAME "system"
+
+struct Key {
+  mozilla::UniqueSECItem key;
+  mozilla::UniqueSECItem iv;
+};
+
+static mozilla::StaticMutex KEY_MUTEX;
```
gcp: bit weird naming, `sKeyMutex` would be more standard?

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+  }
+  return NS_ERROR_FAILURE;
+}
+
+/// Create path directory for keystore file
+nsresult GetKeyStorePath(nsAString& path) {
+  nsresult rv = GetCurrentProfilePath(path);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  path.Append(NS_LITERAL_STRING_FROM_CSTRING("/keystore.db"));
```
gcp: nit: I was expecting this to be a constant defined at the top of the file.

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+/// Load file contents into string
+nsresult LoadFileToString(const nsCOMPtr<nsIFile>& file, nsACString& contents) {
+  PRFileDesc* desc;
+  nsresult rv = file->OpenNSPRFileDesc(PR_RDONLY, PR_IRUSR | PR_IWUSR, &desc);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  unsigned char buf[1024];
+  int count = 0;
+  while ((count = PR_Read(desc, buf, 1024)) > 0) {
+    contents.Append(mozilla::Span<unsigned char>(buf, count));
```
gcp: Yee gods, do we have nothing more recent than this?

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+}
+
+/// Load file contents into string
+nsresult LoadFileToString(const nsCOMPtr<nsIFile>& file, nsACString& contents) {
+  PRFileDesc* desc;
+  nsresult rv = file->OpenNSPRFileDesc(PR_RDONLY, PR_IRUSR | PR_IWUSR, &desc);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  unsigned char buf[1024];
+  int count = 0;
```
gcp: `int32_t` 

Probably not an issue here, but that avoids buggy code like: https://searchfox.org/firefox-main/source/widget/nsTransferable.cpp#168 because note https://searchfox.org/firefox-main/source/nsprpub/pr/src/io/priometh.c#98 and https://searchfox.org/firefox-main/source/nsprpub/pr/src/io/priometh.c#50 so that could have been a nasty bug if that code reads a 4GB file.

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+  nsresult rv = mozilla::Base64Decode(encodedKey, encryptedKey);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  rv = mozilla::Base64Decode(encodedIV, stringIV);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  // Key and IV need to go into SECItems to be useful
+  mozilla::UniqueSECItem key = mozilla::UniqueSECItem(new SECItem());
+  SECStatus stat =
+      SECITEM_MakeItem(nullptr, key.get(), (unsigned char*)encryptedKey.Data(),
```
gcp, line 138: Can't you do `get()` here on `encryptedKey`?

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+  rv = GetKeyStorePath(filePath);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  nsCOMPtr<nsIFile> file;
+  rv = NS_NewLocalFile(filePath, getter_AddRefs(file));
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  // If the file doesn't exist, we need to write the magic before anything else
+  bool has_magic;
+  file->Exists(&has_magic);
```
gcp: Wishful thinking? 😁 

If the magic is not there, would we want to rewrite the file? (If we don't do it, the profile could be broken forever after a power loss?)

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+/// Create "system" key
+nsresult InitializeKeys() {
+  nsAutoCString system(SYSTEM_KEY_NAME);
+
+  Key key = {0, 0};
+
+  nsresult rv = CreateSystemKey(key);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  WriteKeyToDisk(system, key);
```
gcp: No error check? Is it intentional?

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+  SECITEM_AllocItem(nullptr, data, 32);
+
+  unsigned int decrypted_len = 0;
+
+  return PK11_Decrypt(SYSTEM_KEY.get(), CKM_AES_GCM, &gcm_item, data->data,
+                      &decrypted_len, data->len, key.key->data, key.key->len);
+}
+
+/// Obtain requested key assuming owned mutex
+nsresult FetchOrCreateKeyInner(nsAutoCString& identifier, SECItem* data) {
```
gcp: Given the comment, `AssertCurrentThreadOwns`?

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+      });
+  if (stat != SECSuccess) {
+    return NS_ERROR_FAILURE;
+  }
+  return NS_OK;
+}
+
+nsresult FetchOrCreateKey(nsAutoCString& identifier, SECItem* data) {
+  // Lock mutex here, so we can return the rv easily
+  KEY_MUTEX.Lock();
```
gcp: We probably have some `AutoLock` thing suitable here?

storage/mozStorageConnection.cpp
```diff
@@ -1127,6 +1164,66 @@ nsresult Connection::initialize(nsIFile* aDatabaseFile) {
...
+  MOZ_LOG(key::GetKeyStorageLog(), LogLevel::Debug,
+          ("Fetched database encryption key"));
+
+  // Do not set mFileURL here since this is database does not have an associated
+  // URL.
+  mDatabaseFile = aDatabaseFile;
+  mDatabaseEncrypted = true;
+
+  nsAutoString dbPath;
+  aDatabaseFile->GetPath(dbPath);
```
gcp: no error checks

storage/mozStorageConnection.cpp
```diff
@@ -1978,6 +2075,19 @@ nsresult Connection::initializeClone(Connection* aClone, bool aReadOnly) {
           rv = aClone->CreateStatement("ATTACH DATABASE :path AS "_ns + name,
                                        getter_AddRefs(attachStmt));
           NS_ENSURE_SUCCESS(rv, rv);
+
+          if (mDatabaseEncrypted) {
+            nsCString aDBKey;
+            rv = key::GetKeyForPath(path.get(), aDBKey);
+            NS_ENSURE_SUCCESS(rv, rv);
+#ifdef _WIN32
```
gcp: Could be a helper, it's repeated a few times.

storage/mozIStorageAsyncConnection.idl
gcp, line 24: You're changing the contract, so this needs a bump.

toolkit/modules/Sqlite.sys.mjs
```diff
@@ -1218,6 +1218,10 @@ ConnectionData.prototype = Object.freeze({
       );
     });
   },
+
+  attachDatabase(path, name) {
+    return this._dbConn.attachDatabase(path, name);
```
gcp: You await the result, so this needs to be a Promise?

storage/mozStorageConnection.cpp
```diff
@@ -253,6 +257,29 @@ void basicFunctionHelper(sqlite3_context* aCtx, int aArgc,
...
 }
 
+nsresult ExtractURIPathAndQuery(const char* uri, nsCString& path,
+                                nsCString& query) {
+  if (strstr(uri, "file:") != uri) {
+    return NS_ERROR_FAILURE;
+  }
+  const char* queryDelim = strstr(uri, "?");
+  // strstr returns nullptr if it can't find "?" or aPath if it is empty
+  if (!queryDelim || queryDelim == uri) {
```
gcp, line 278: This suggests we should never get here for unencrypted files (as they wouldn't have a query string with the key), but it's not clear to me that's the case from the rest of the code.

toolkit/components/places/Database.cpp
```diff
@@ -334,12 +338,31 @@ nsresult SetupDurability(nsCOMPtr<mozIStorageConnection>& aDBConn,
...
+  aName.GetData(&aNameBuf);
+
+  nsresult rv;
+  nsCString path;
+
+  bool encryptionEnabled =
+      StaticPrefs::security_storage_encryption_sqlite_enabled_AtStartup();
+  if (encryptionEnabled) {
+    nsCString aDBKey;
+    storage::key::GetKeyForPath(aPathBuf, aDBKey);
```
gcp: Error checks?

toolkit/components/places/Database.cpp
gcp: Doesn't this need to append the `.enc`?

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  unsigned char buf[1024];
+  int count = 0;
+  while ((count = PR_Read(desc, buf, 1024)) > 0) {
+    contents.Append(mozilla::Span<unsigned char>(buf, count));
+  }
+
+  if (count < 0) return mozilla::psm::GetXPCOMFromNSSError(PR_GetError());
+  return NS_OK;
```
gcp: I stand by the previous comment, as this leaks memory. (`PR_Close()` is needed)

security/keystore/Persist.cpp
```diff
@@ -0,0 +1,407 @@
...
+  nsCString fileContents;
+
+  // Get the file to the key store in the profile and turn it into a file ref
+  nsAutoString filePath;
+  nsresult rv = GetKeyStorePath(filePath);
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  nsCOMPtr<nsIFile> file;
+  rv = NS_NewLocalFile(filePath, getter_AddRefs(file));
+  NS_ENSURE_SUCCESS(rv, rv);
```
gcp, line 192: This is the right way to do it.

security/keystore/KeyStorage.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+namespace mozilla::storage::key {
+mozilla::LogModule* GetKeyStorageLog() {
+  static mozilla::LazyLogModule sLog("KeyStorage");
+
+  return sLog;
+}
+
+nsresult GetKeyForPath(const char* aPath, nsCString& key) {
+  nsIFile* file = MakeAndAddRef<nsLocalFile>().take();
```
gcp: This leaks? It's now a raw pointer and it's not released anywhere.

You have the correct pattern with `nsCOMPtr<nsIFile>` elsewhere in the patch.

security/keystore/Persist.cpp
gcp: Running in debug mode will answer this question: https://searchfox.org/firefox-main/rev/286e75a82e5c489d020d18121674809ebf114ae4/xpcom/io/nsDirectoryService.cpp#184 😀 😀 😀 😀

security/keystore/Persist.cpp
gcp: 100% untested:

```
  nsresult LoadFileToString(nsIFile& aFile, nsACString& aContents) {
    nsCOMPtr<nsIInputStream> stream;
    nsresult rv = NS_NewLocalFileInputStream(getter_AddRefs(stream), &aFile);
    NS_ENSURE_SUCCESS(rv, rv);

    return NS_ConsumeStream(stream, UINT32_MAX, aContents);
  }
```

storage/mozStorageConnection.cpp
gcp, line 2766: This naming with underscore looks so weird if you're used to Rust code but it seems common in this code 🤷 .

gcp (changes requested):

modules/libpref/Preferences.cpp
gcp: Yes, but you need the extra checks that your patch in that bug has, see my remarks there.

modules/libpref/init/StaticPrefList.yaml
gcp: What I'm saying is that I would have expected this to just read _"Enable enterprise storage encryption"_.

storage/mozStorageConnection.cpp
Not-Nik: Seems to be the standard name for `mozIStoragePendingStatement**`

security/keystore/Persist.cpp
Not-Nik: Dana brought up the same issue, I'll add a comment

> That particular example needs a "proper" SECItem (on the heap), because it is later stored in KEY_MAP, which is static and thus outlives encryptedKey. I already tried to reduce buffer allocation, and I'll move to a stack SECItem in the path == SYSTEM_KEY_NAME path, but if I missed anything else do let me know.

storage/mozStorageConnection.cpp
Not-Nik: There may be unencrypted databases that still use URIs.

security/keystore/Persist.cpp
Not-Nik: Confusing variable naming here. The actual check for the magic is earlier in the load function. The writer doesn't check for the magic at all, but if the key store doesn't exist, it is initialised with the magic.

security/keystore/Persist.cpp
Not-Nik: Did you mistype? Otherwise, what does this comment indicate?

gcp: This is not the right way to do it: https://github.com/mozilla/enterprise-firefox/pull/173#discussion_r2771385774

The comments are reordered, maybe that's why this is confusing? The code here is fine, the other code is not.

Not-Nik: Ah right, I replaced the raw pointer with a nsCOMPtr already, so I can resolve this, thanks!

security/keystore/Persist.cpp
gcp: My thinking was that the writer assumes that either the file is there and this implies the magic is there, or it isn't, and then it will write it.

But it's also possible for the file to be there and not have the magic (crash, power failure).

PR #182, opened 2025-11-21
fix clippy lints

fiji-flo:


lissyx (approved):
Why isn't clippy part of the default linting 😭

PR #192, opened 2025-11-25
Update felt login

1rneh:


browser/extensions/felt/content/styles/felt.css
```diff
@@ -113,14 +128,18 @@ moz-input-text::part(label) {
 
 .felt-powered-by {
   position: absolute;
+  /* stylelint-disable-next-line */
   bottom: 8px;
+  /* stylelint-disable-next-line */
```
Mossop, line 133: Can we list the specific rule we have to disable here?

Mossop (approved):

browser/extensions/felt/content/styles/felt.css
1rneh: We agreed offline not to add an explicit list, because I'd have to disable two rules with the following line:
`/* stylelint-disable-next-line stylelint-plugin-mozilla/use-space-tokens */ /* stylelint-disable-next-line stylelint-plugin-mozilla/use-size-tokens */`

PR #198, opened 2025-11-25
Correct offsetY in document.icns

1rneh:


lissyx (approved):
This is why the image was misplaced?

1rneh:
> This is why the image was misplaced?

No, this was just generally a misaligned asset that I still needed to clarify the requirements for.

PR #199, opened 2025-11-25
Fix lint errors

gcp:


.prettierignore
```diff
@@ -1239,6 +1239,12 @@ toolkit/components/translations/cld2/
 toolkit/mozapps/update/tests/data/xpcshellConstantsPP.js
 toolkit/modules/AppConstants.sys.mjs
 
+# Files with MOZ_ENTERPRISE preprocessor directives
```
gcp: It kind of sucks to completely eliminate these from linting, but it's this or not having the lint in treeherder be of that much use?

lissyx (approved):

PR #202, opened 2025-11-26
Include the enterprise theme in the package and make it the default theme

Mossop:
For the time being I have copied the preview and icon files from the light theme, we can switch these out once Matt can generate the correct versions.


browser/locales/en-US/browser/appExtensionFields.ftl
```diff
@@ -16,3 +16,6 @@ extension-firefox-compact-dark-description=A theme with a dark color scheme.
 
 extension-firefox-alpenglow-name=Firefox Alpenglow
 extension-firefox-alpenglow-description=Use a colorful appearance for buttons, menus, and windows.
+
+extension-firefox-enterprise-name=Enterprise
+extension-firefox-enterprise-description=Firefox for Enterprise
```
1rneh: We are not calling it "Firefox for Enterprise" anymore but just "Firefox Enterprise".

1rneh (approved):

browser/locales/en-US/browser/appExtensionFields.ftl
Mossop: Ok changed that. We probably want to add something a bit more descriptive later.

PR #203, opened 2025-11-26
Bug 2002482: Disable the new profiles UI in enterprise

Mossop:


browser/app/profile/firefox.js
```diff
@@ -3473,7 +3473,7 @@ pref("browser.backup.scheduled.user-disabled", false);
 // backup process.
 pref("browser.backup.tab-flush-timeout", 5000);
 
-#ifdef NIGHTLY_BUILD
+#if defined(NIGHTLY_BUILD) && !defined(MOZ_ENTERPRISE)
```
lissyx: Shouldn't we maybe introduce a dedicated `firefox_enterprise.js` pref file?

lissyx (approved):

browser/app/profile/firefox.js
Mossop: Turns out we already have one!

PR #208, opened 2025-11-26
GHA: MSI gcpEU and fixes for version

lissyx:


lissyx:
Confirmed to be OK by erkan

PR #211, opened 2025-11-27
Bug 1578489 - Update msi installer icon

1rneh:


1rneh:
This is not changing the MSI installer.

lissyx (approved):
Let's see if it does anything, but you should rename the commit and the PR since it's not the MSI but the SFX installer

PR #214, opened 2025-11-27
Allow disabling the JIT for specific domains

Mossop:
@mkaply What do you think about the proposed policy schema change here?


mkaply:
Overall it looks good, but I think "Isolation" is probably a little too generic.

Also you need to add the description to:

https://searchfox.org/firefox-main/source/browser/locales/en-US/browser/policies/policies-descriptions.ftl

PR #215, opened 2025-11-27
When the felt process purges its startup cache instruct the Firefox process to do the same

Mossop:
Some changes in development builds only take effect if the profile's startup cache is purged on the next launch. mach takes care of this by creating a special flag file. But for enterprise the first process launched is felt which then deletes the flag file so when the Firefox process starts it doesn't know to purge the cache. We can pass this through the command line.


1rneh (approved):
Thank you!

PR #216, opened 2025-11-27
Enterprise: Test for external link handling

lissyx:


testing/enterprise/test_felt_browser_external_link.py
```diff
@@ -0,0 +1,89 @@
...
+            if not display and not wayland:
+                self._logger.warning(
+                    f"Opening URL cannot work in headless mode, but DISPLAY={display} and WAYLAND_DISPLAY={wayland}. xvfb-run will be used."
+                )
+
+        return True
+
+    def test_felt_4_verify_prefs(self, exp):
+        """
+        No care.
```
gcp: nit: Don't care

testing/enterprise/test_felt_browser_external_link.py
```diff
@@ -0,0 +1,89 @@
...
+        assert not has_tab, f"Should not have {self._external_link} opened"
+
+        # At least on linux, one cannot ues --headless for this, it ends up in
+        # ProfileLockedDialog() because --headless disables remote client
+        args = [f"{sys.argv[1]}", "-profile", self._profile_path, self._external_link]
+        subprocess.check_call(args, shell=False)
+
+        self._child_driver.switch_to.new_window("tab")
+        self._child_wait.until(EC.new_window_is_opened(tabs))
+        new_tabs = self._child_driver.window_handles
```
gcp: Selenium docs have this as `property window_handles: list[str]` so I don't think the loop below does what you think it does. (It's not a handle, it returns the actual list at call time)

gcp (changes requested):
Tab open detection loop looks broken to me.

testing/enterprise/test_felt_browser_external_link.py
lissyx: You think we should rather get the list of windows handle in the `while` ? That may be less risky indeed

gcp: I mean the loop wasn't doing anything before 🤷

lissyx: More or less yes :)

gcp (approved):

PR #218, opened 2025-11-27
Switch to 'enterprise' trust domain

ahal:


lissyx (approved):

taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -447,7 +447,12 @@ def filter(task):
 
         return False
 
-    return [l for l in filtered_for_project if filter(full_task_graph[l])]
+    tasks = [l for l in filtered_for_project if filter(full_task_graph[l])]
+    tasks.extend([
+        "fetch-wix-3.14.0",
+        "toolchain-linux64-mkbom",
```
lissyx: i saw failures on macOS because of missing `toolchain-linux64-xar` shouldn't we add it here?

taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -490,6 +495,9 @@ def filter(task):
         if task.kind == "build" and "all" in task.attributes.get("run_on_projects"):
             return False
 
+        if task.kind in ("upload-symbols",):
```
lissyx, line 493: Do we really need this? I'm fine with treeherder being broken. Is breaking this mean it's going to break the graph and not run everything?

taskcluster/kinds/enterprise-repack-repackage/kind.yml
```diff
@@ -23,7 +23,7 @@ kind-dependencies:
 
 task-defaults:
     run-on-repo-type: [git]
-    run-on-projects: ["enterprise-firefox"]
+    run-on-projects: []
```
lissyx, line 26: Same, is this really required ?

taskcluster/kinds/enterprise-repack/kind.yml
```diff
@@ -24,8 +24,7 @@ kind-dependencies:
 #    - shippable-l10n-mac-notarization
 
 task-defaults:
-    run-on-repo-type: [git]
-    run-on-projects: ["enterprise-firefox"]
+    run-on-projects: []
```
lissyx, line 27: Same, is this really required ?

taskcluster/kinds/upload-symbols/kind.yml
```diff
@@ -28,8 +28,8 @@ not-for-build-platforms:
     - win64-aarch64-geckodriver/opt
 
 task-defaults:
-    run-on-repo-type: ["git"]
-    run-on-projects: ["enterprise-firefox"]
+    # Scopes aren't setup for enterprise yet
```
lissyx, line 31: Same, is this really required ?

taskcluster/gecko_taskgraph/target_tasks.py
ahal: Yes, and yes for all the others as well.

The Decision task will fail with scope errors if we have them running. We can revert these changes easily enough once we sort out the scopes.

taskcluster/gecko_taskgraph/target_tasks.py
ahal: Been playing some whack-a-mole, but it's green now.

PR #220, opened 2025-11-28
Persist user email in felt.json and pre-fill it into felt login

1rneh:


Mossop:
Can we use `JSONFile` here? It has a lot of functionality for handling backups, and asynchronous persistence and we shouldn't re-invent that.

1rneh:
> Can we use `JSONFile` here? It has a lot of functionality for handling backups, and asynchronous persistence and we shouldn't re-invent that.

Thanks, we should definitely use `JSONFile`.

browser/extensions/felt/FeltStorage.sys.mjs
```diff
@@ -0,0 +1,56 @@
...
+    if (this._feltStorage.data.lastSignedInUserEmail === email) {
+      // Nothing changed.
+      return;
+    }
+    this._feltStorage.data.lastSignedInUserEmail = email;
+    this._feltStorage.saveSoon();
+  },
+
+  async uninit() {
+    await this._feltStorage.finalize();
```
Mossop: JSONFile automatically finalizes during shutdown so is this necessary?

browser/extensions/felt/content/window.js
```diff
@@ -68,10 +69,16 @@ async function connectToConsole(email) {
   ssoBrowsingContext.focus();
 }
 
-function listenFormEmailSubmission() {
+async function listenFormEmailSubmission() {
   const signInBtn = document.getElementById("felt-form__sign-in-btn");
   const emailInput = document.getElementById("felt-form__email");
 
+  const lastUsedUserEmail = await lazy.FeltStorage.getLastSignedInUser();
```
Mossop: You don't need to await this, it isn't a promise.

Mossop (approved):

PR #231, opened 2025-12-02
Bug 2003628 - Support Github pull requests in Enterprise Firefox

ahal:


.github/workflows/enterprise.yml
lissyx, line 1: I would keep GitHub Actions around a couple of days, until we know that things are all good on TaskCluster

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -1883,6 +1889,8 @@ def add_generic_index_routes(config, task):
     except KeyError:
         pass
 
+    subs["project"] = get_project_alias(config)
```
lissyx, line 1892: do we need routes at all for PR ?

.github/workflows/enterprise.yml
lissyx: Especially since we dont have yet `sccache` back on TaskCluster, so builds are longer (and yes it's important)

ahal: I was just getting annoyed by all the wasted builds on my re-pushes, that's why I added DO NOT LAND ;)

taskcluster/gecko_taskgraph/transforms/task.py
ahal: Doesn't hurt, and just as easy to do this as it is to remove them

lissyx: ok, i was worried about index pollution, I'm not sure there's a good usecase here

lissyx (approved):

PR #236, opened 2025-12-03
Enterprise re enable symbols upload

lissyx:


taskcluster/gecko_taskgraph/transforms/build.py
```diff
@@ -321,4 +321,19 @@ def add_enterprise_secret_scopes(config, jobs):
                     f"secrets:get:project/releng/gecko/build/level-{level}/mozilla-desktop-geoloc-api.key",
                 ]
             )
+
+        if "enterprise" in job["name"] and "symbol-upload" in job["name"]:
```
ahal: nit: `config.kind == "upload-symbols"`

taskcluster/gecko_taskgraph/transforms/build.py
```diff
@@ -321,4 +321,19 @@ def add_enterprise_secret_scopes(config, jobs):
...
+                ]
+            )
+
+        if "enterprise" in job["name"] and "sign" in job["name"]:
+            job.setdefault("scopes", []).extend(
+                [
+                    f"project:enterprise:releng:signing:cert:dep-signing",
+                ]
+            )
+
```
ahal: Actually `upload-symbols` and `signing` tasks don't use these transforms, so this is all a no-op afaict.

ahal (changes requested):

taskcluster/gecko_taskgraph/transforms/build.py
lissyx: I was running into scope issues without, I'll try again

PR #240, opened 2025-12-04
Telemetry gaps downloads

jonathanmendez:
Add file path and file hash to download telemetry for enterprise.

Also fix download_attempted telemetry to obey url/file telemetry prefs and use same 'source_url' key.

### DIFF
--- a/browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
+++ b/browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
@@ -234,8 +234,8 @@ export const DownloadsTelemetryEnterprise = {
         file_path: fileInfo.file_path,
         extension: fileInfo.extension,
         mime_type: fileInfo.mime_type,
-        sha256_hash: download.saver.getSha256Hash(),
-        size_bytes: sizeBytes,
+        sha256_hash: download.saver.getSha256Hash() || "",
+        size_bytes: sizeBytes || 0,
         source_url: sourceUrl || "",
         is_private: download.source?.isPrivate || false,
       };
--- a/browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
+++ b/browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
@@ -226,7 +226,7 @@ export const DownloadsTelemetryEnterprise = {
       // Get file size
       let sizeBytes = download.target?.size;
       if (typeof sizeBytes !== "number" || sizeBytes < 0) {
-        sizeBytes = null;
+        sizeBytes = 0;
       }
 
       const telemetryData = {
@@ -235,7 +235,7 @@ export const DownloadsTelemetryEnterprise = {
         extension: fileInfo.extension,
         mime_type: fileInfo.mime_type,
         sha256_hash: download.saver.getSha256Hash() || "",
-        size_bytes: sizeBytes || 0,
+        size_bytes: sizeBytes,
         source_url: sourceUrl || "",
         is_private: download.source?.isPrivate || false,
       };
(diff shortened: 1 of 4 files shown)


browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
```diff
@@ -221,8 +231,10 @@ export const DownloadsTelemetryEnterprise = {
 
       const telemetryData = {
         filename: fileInfo.filename,
+        file_path: fileInfo.file_path,
         extension: fileInfo.extension,
         mime_type: fileInfo.mime_type,
+        sha256_hash: download.saver.getSha256Hash(),
```
gcp: Ok that was easy :)

toolkit/components/downloads/DownloadCore.sys.mjs
```diff
@@ -720,23 +831,24 @@ Download.prototype = {
...
       Glean.downloads.downloadAttempt.record({
-        filename,
-        extension: fileExtension,
-        mime_type: this.contentType || "",
-        source_url_domain: sourceDomain,
+        filename: fileInfo.filename,
+        file_path: fileInfo.file_path,
+        extension: fileInfo.extension,
+        mime_type: fileInfo.mime_type,
+        source_url: sourceUrl,
```
gcp: You wrote `processSourceUrl` so that this can be `null`. Does that work for Glean? (I would suspect not because all other stuff is defaulted to `""`?)

browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
gcp: Is this guaranteed to exist? (If not, same remark as with `source_url` below)

From digging, one case would be external protocol handlers, but then we also don't really download anything, so I guess that's fine?

gcp (changes requested):
Probably can't send `null` in the indicated case.

toolkit/components/downloads/DownloadCore.sys.mjs
jonathanmendez: Good catch, forgot to copy the default over from the downloadcompleted code.

browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
jonathanmendez: From my own digging, it seemed like if we did reach the success case it should exist? But it doesn't hurt to check against null.

FWIW, since I noticed that sizeBytes could also technically be null, I went ahead and tested what happens if you pass null; the event still got sent and received (at least by my local http-edge), and the event was just missing that key from the data. That being said, I'll go ahead and default it to 0 just in case.

browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
```diff
@@ -221,9 +231,11 @@ export const DownloadsTelemetryEnterprise = {
 
       const telemetryData = {
         filename: fileInfo.filename,
+        file_path: fileInfo.file_path,
         extension: fileInfo.extension,
         mime_type: fileInfo.mime_type,
-        size_bytes: sizeBytes,
+        sha256_hash: download.saver.getSha256Hash() || "",
+        size_bytes: sizeBytes || 0,
```
gcp: Why not fix the code right above? (...which sets `size_bytes` to null to begin with!)

gcp (approved):
OK to land but one more cleanup I'd like to see.

browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs
jonathanmendez: Done!

PR #247, opened 2025-12-08
FxCI: Allow release/nightly signing

lissyx:


taskcluster/gecko_taskgraph/transforms/signing.py
```diff
@@ -173,7 +173,7 @@ def make_task_description(config, jobs):
         signing_type = get_signing_type_per_platform(
             build_platform, is_shippable, config
         )
-        worker_type_alias = "linux-signing" if is_shippable else "linux-depsigning"
+        worker_type_alias = "linux-signing" if is_shippable or "enterprise" in label else "linux-depsigning"
```
ahal, line 176: We should set `shippable: true` on the build task's attributes rather than doing this.

ahal (changes requested):

taskcluster/gecko_taskgraph/transforms/signing.py
ahal: And only if `config.params["level"] == 3`

lissyx: TBH I'm afraid of the consequences

lissyx: @ahal I'm not sure exactly where it needs to be done ?

ahal: > And only if config.params["level"] == 3

Actually, shippable should be set regardless of level, it's just a signal that "these are builds that are acceptable to ship". Chain of trust / scriptworker will deal with the security implications if we try to actually use them from a level 1 context.

> TBH I'm afraid of the consequences

These are builds that you intend to ship onto other people's devices, therefore  by definition they are shippable :)

> I'm not sure exactly where it needs to be done ?

You can simply set `shippable: true` in e.g `taskcluster/kinds/build/linux.yml` under the attributes section. Maybe we should add "shippable" to the label as well to make it clear? I'm not sure if you also intend to have non-shippable builds like we do for Firefox. I guess we'll need this eventually, so maybe it's best to add `shippable` to the label.

lissyx: Ok i made that except I'm not sure where to check for the level value to change `shippable` ?

PR #252, opened 2025-12-09
Disable installer telemetry by default for enterprise builds.

gcp:


browser/installer/windows/nsis/stub.nsh
```diff
@@ -997,12 +997,19 @@ Function SendPing
     SetAutoClose true
     StrCpy $R9 "2"
     Call RelativeGotoPage
+!else
+!ifdef DISABLE_INSTALLER_TELEMETRY
+    ; Installer telemetry is disabled; skip sending the ping and exit
```
cpdupuis: The telemetry in this file is only used by the stub installer. Are we planning to have a version of the stub installer for enterprise?

cpdupuis: If you do want stub installer support for enterprise, then the better way to handle this would be (on line 1000):

```
!else if DISABLE_INSTALLER_TELEMETRY
    ; Installer telemetry is disabled; skip sending the ping and exit
!else
```

Then you won't need to copy the other bits.

gcp: Maybe because it's late I don't see a way to keep the logic the same this way, but `ifdef x or y` should work?

cpdupuis (approved):
Looks good!

PR #259, opened 2025-12-10
ci: run Taskcluster tasks on all branches

ahal:


hneiva (approved):

lissyx:
Looks fine but any push to any branch will trigger build, that makes me uneasy, can we keep those branches but allow PR events? Or are things not ready yet?

ahal:
I'm not sure why pushes to branches make you uneasy but events on PRs doesn't.. I assume they would both generate roughly the same number of tasks (pushes probably even less because many people are likely pushing to a fork).

Anyway, PRs are close.. maybe I can wrap them up today. I agree we don't really need this once we have PR events, so I can just abandon this.

lissyx:
I dislike the practice of directly pushing to the repo. We did it for now because private repo and credentials management on GitHub actions forced us to, but I wish we can soon all send PRs from our forks instead and sanctuarize the main repo and it's branch.

PR #261, opened 2025-12-10
Encode download hash in telemetry as hex string

jonathanmendez:
The string returned from `getSha256Hash()` was a cstring byte array. We need to encode it as hex to be valid/readable json.


gcp (changes requested):
Is it possible to add tests for this?

gcp (approved):

PR #268, opened 2025-12-12
FxCI: Grant some scopes for build tasks on enterprise-firefox

lissyx:
We will want PGO at some point so I think we can enable for any build ?

### DIFF
--- a/taskcluster/gecko_taskgraph/transforms/build.py
+++ b/taskcluster/gecko_taskgraph/transforms/build.py
@@ -312,8 +312,7 @@ def add_enterprise_secret_scopes(config, jobs):
     """Enterprise builds re-use some secrets from the Gecko trust domain."""
     level = config.params["level"]
     for job in jobs:
-        attributes = job.get("attributes")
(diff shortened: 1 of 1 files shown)


taskcluster/gecko_taskgraph/transforms/build.py
```diff
@@ -313,7 +313,7 @@ def add_enterprise_secret_scopes(config, jobs):
     level = config.params["level"]
     for job in jobs:
         attributes = job.get("attributes")
```
gcp: nit: dead code?

gcp (approved):

taskcluster/gecko_taskgraph/transforms/build.py
lissyx: yes

lissyx:
Group with the changes: https://firefox-ci-tc.services.mozilla.com/tasks/groups/XHv1mLW3T2akNOvjir8uYA

ahal (approved):

PR #269, opened 2025-12-12
FxCI: Some enterprise_repacks cleanup

lissyx:
It's not perfect, but it should already be an improvement

### DIFF
--- a/taskcluster/gecko_taskgraph/decision.py
+++ b/taskcluster/gecko_taskgraph/decision.py
@@ -121,6 +121,26 @@ PER_PROJECT_PARAMETERS = {
         "release_product": "firefox-enterprise",
         "release_partners": ["sample"],
         "release_partner_config": {
+            "repackage-deb": {
+                "sample": {
+                    "gcpEU": {
+                        "locales": ["en-US"],
+                        "platforms": [
+                            "linux64-enterprise-shippable",
+                        ],
+                    },
+                },
+            },
+            "repackage-msi": {
+                "sample": {
+                    "gcpEU": {
+                        "locales": ["en-US"],
+                        "platforms": [
+                            "win64-enterprise-shippable",
+                        ],
+                    },
+                },
+            },
             "enterprise-repack-repackage": {
                 "sample": {
                     "gcpEU": {
--- a/taskcluster/gecko_taskgraph/transforms/repackage.py
+++ b/taskcluster/gecko_taskgraph/transforms/repackage.py
@@ -4,6 +4,7 @@
 """
 Transform the repackage task into an actual task description.
 """
+import copy
 
 from taskgraph.transforms.base import TransformSequence
 from taskgraph.util.copy import deepcopy
@@ -808,15 +809,7 @@ def make_job_description(config, jobs):
             "extra": job.get("extra", {}),
             "worker": worker,
             "run": run,
-            "fetches": _generate_download_config(
-                config,
-                dep_job,
-                build_platform,
-                signing_task,
-                repackage_signing_task,
-                locale=locale,
-                existing_fetch=job.get("fetches"),
-            ),
+            "fetches": job.get("fetches"),
         }
 
         if build_platform.startswith("macosx"):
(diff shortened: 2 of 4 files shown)


lissyx:
Feel free to redirect if you think someone else should review

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -836,7 +829,51 @@ def make_job_description(config, jobs):
...
+        repacks = []
+        if "enterprise-repack" == dep_job.kind or "enterprise-repack-repackage" == dep_job.kind:
+            this_repack = config.params["release_partner_config"].get(config.kind)
+            if this_repack:
+                for enterprise_name in this_repack.keys():
+                    for repack_name in this_repack[enterprise_name]:
+                        for platform_name in this_repack[enterprise_name][repack_name][
+                            "platforms"
+                        ]:
+                            for locale in this_repack[enterprise_name][repack_name][
```
gcp: Variable name clash here?

gcp: Is that the reason for the TODO comment? 🤔

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -836,7 +829,51 @@ def make_job_description(config, jobs):
...
+        if len(repacks) == 0:
+            repacks = [None]
+
+        for repack in repacks:
+            if repack:
+                repack_label = repack.replace("/", "-")
+                build_platform_with_repack = f"{build_platform}-{repack_label}"
+                task["label"] = task["label"].replace(
+                    build_platform, build_platform_with_repack
+                )
```
gcp, line 865: Don't you want to copy `task` before you substitute? Else iteration 2 of this loop isn't going to do much.

gcp (changes requested):

taskcluster/gecko_taskgraph/transforms/repackage.py
lissyx: no the comment is for something else :)

taskcluster/gecko_taskgraph/transforms/repackage.py
lissyx: yep, not visible bug because only one repack for now

taskcluster/gecko_taskgraph/decision.py
```diff
@@ -121,6 +121,26 @@
...
+            "repackage-msi": {
+                "sample": {
+                    "gcpEU": {
+                        "locales": ["en-US"],
+                        "platforms": [
+                            "win64-enterprise-shippable",
+                        ],
+                    },
+                },
+            },
```
ahal, line 143: Yeah, this is definitely not stuff we want to be hardcoding into the default parameters. But I guess that ship sailed so this isn't really making things worse. But we'll want to remove everything other than `target_tasks_method` here in the future.

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -836,7 +829,53 @@ def make_job_description(config, jobs):
...
+                                    ]
+            else:
+                previous_repack = dep_job.task.get("extra").get("repack_id")
+                if previous_repack:
+                    repacks += [previous_repack]
+
+        if len(repacks) == 0:
+            repacks = [None]
+
+        for repack in repacks:
```
ahal, line 857: I'm confused about why we need to iterate over repacks and create a new task for each one. Afaict, the tasks should be using the [from_deps](https://taskcluster-taskgraph.readthedocs.io/en/stable/reference/transforms/from_deps.html) transforms which is supposed to create a task per dependency implicitly.

Can we follow the model that Firefox is using?

ahal (approved):
I won't block you on review comments, but this feels wrong :p

taskcluster/gecko_taskgraph/decision.py
lissyx: I think it's something we can do when we have some proper release pipeline/promotion ?

taskcluster/gecko_taskgraph/transforms/repackage.py
lissyx: @ahal that's where I say it's an improvement because it should give us support for more than one repack, but it's still not correct, and it's a place where I will likely not be able to be efficient

PR #270, opened 2025-12-12
Bug 2001078 - Use generic _fetch to initiate post and get requests in ConsoleClient

1rneh:


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -245,43 +245,64 @@ export const ConsoleClient = {
    * a registered console endpoint. If we get a 401 or 403 refresh and retry once.
    *
    * @param {string} path - Console API to request
-   * @param {string} method - Console API method to use, GET or POST
-   * @param {{_didRefresh?: boolean}} [options]
+   * @param {"GET"|"POST"} method - Console API method to use
```
lissyx, line 243: nit: if we state GET/POST we should reject I think

lissyx (approved):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp, line 471: This is now invalid, as you changed the signature.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -245,43 +245,64 @@ export const ConsoleClient = {
...
+   *
+   * @throws {InvalidAuthError|Error}
+   *
+   * @returns {Promise<any>} Promise which resolves to a parsed JSON response body.
    */
-  async _post(path) {
-    // TODO: Bug 2001078 - ConsoleClient shouldn't use the
-    // _get function to initiate POST requests
-    return this._get(path, "POST");
+  async _post(path, jsonBody) {
```
gcp: Note this is changing the signature, so old callers are passing in `undefined`.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -245,43 +245,64 @@ export const ConsoleClient = {
...
+    if (jsonBody !== null) {
+      headers.set("Content-Type", "application/json");
+    }
 
     const url = this.constructURI(path);
-    const res = await fetch(url, { method, headers });
+    const res = await fetch(url, {
+      method,
+      headers,
+      body: jsonBody === null ? undefined : JSON.stringify(jsonBody),
```
gcp, line 266: I think here you will try to stringify `undefined`.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -245,43 +245,64 @@ export const ConsoleClient = {
...
+   * @param {{ _didRefresh?: boolean, jsonBody?: object }} [options]
    * @throws {InvalidAuthError|Error}
    * @returns {Promise<any>} Parsed JSON response body.
    */
-  async _get(path, method = "GET", { _didRefresh = false } = {}) {
+  async _fetch(path, method, { _didRefresh = false, jsonBody = null } = {}) {
     const headers = new Headers({});
     const accessToken = await this.getAccessToken();
     headers.set("Authorization", `Bearer ${accessToken}`);
+    if (jsonBody !== null) {
```
gcp, line 258: This will trigger even if `jsonBody` is undefined.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -339,8 +360,8 @@ export const ConsoleClient = {
             Accept: "application/json",
           },
           body: JSON.stringify({
-            grant_type: "refresh_token",
-            refresh_token: refreshToken,
+          grant_type: "refresh_token",
```
gcp: nit: noise.

Also your formatter is broken I think ;-)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -245,43 +245,64 @@ export const ConsoleClient = {
...
-   * @returns {Promise<any>} Parsed JSON response body.
+   *
+   * @returns {Promise<any>} Promise which resolves to a parsed JSON response body.
+   */
+  async _get(path) {
+    return this._fetch(path, "GET");
+  },
+
+  /**
+   * Initiates a POST request again a registered console endpoint.
```
gcp: nit: spelling

gcp (changes requested):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: I guess the simplest is just to give this a default and then the other code can stay as it is.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: I just re-checked the behavior of [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) because I wasn't sure anymore neither:
> Default function parameters allow named parameters to be initialized with default values if no value or undefined is passed.

So if `undefined` is passed to `_fetch` for `jsonBody`, then it's initialized with the default value `null` and we won't pass is to `JSON.stringify`.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: `jsonBody` can't be `undefined`, see explanation below.

gcp (approved):

PR #271, opened 2025-12-12
Enterprise add routes

lissyx:


taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -1911,6 +1895,7 @@ def add_generic_index_routes(config, task):
 
 @index_builder("shippable")
 def add_shippable_index_routes(config, task):
+    print("add_shippable_index_routes", config.kind, task["label"])
```
ahal: leftover debug statement

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -1927,18 +1912,9 @@ def add_shippable_index_routes(config, task):
     subs["product"] = index["product"]
     subs["trust-domain"] = config.graph_config["trust-domain"]
     subs["branch_rev"] = get_branch_rev(config)
-    try:
-        subs["branch_git_rev"] = get_branch_git_rev(config)
-    except KeyError:
-        pass
-    subs["project"] = get_project_alias(config)
```
ahal, line 1934: I think we need to keep `get_project_alias` here, otherwise if you try to run a task with one of these routes from a PR it'll be a scope error.

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -280,12 +278,6 @@ def get_branch_rev(config):
     ]
 
 
-def get_branch_git_rev(config):
-    return config.params[
-        "{}head_git_rev".format(config.graph_config["project-repo-param-prefix"])
-    ]
```
ahal, line 286: Removing this is fine here, though it'll be nice to solve this on `firefox-main` in a way that works for both Enterprise and normal Firefox. That way other forks can benefit too and there will be fewer merge conflicts you need to deal with :).

I can look into landing something on firefox-main though.

ahal (changes requested):

taskcluster/gecko_taskgraph/transforms/task.py
lissyx: yeah it's not finished :)

ahal: my bad, misunderstood

PR #272, opened 2025-12-12
Bug 2005817 - Use a separate shipping-product for Enterprise Firefox

ahal:


lissyx (approved):

taskcluster/config.yml
```diff
@@ -371,6 +371,7 @@ index:
         - 'source'
         - 'system-symbols'
         - 'geckodriver'
+        - 'enterprise'
```
lissyx, line 374: or `firefoxenterprise` ? this may be more aligned with what I saw in how e.g. debian repackage was expecting based on the product name

ahal: I'm not really sure tbh.. I think it's mainly important that it be separate for now, we can change it later

PR #277, opened 2025-12-15
FxA and Sync Support

1rneh:
Note: FxA and Sync integration tests will follow once we can spin up the test services (console, fxa, sync) in TaskCluster to avoid mocking every single endpoint.


.github/workflows/enterprise.yml
```diff
@@ -260,6 +260,11 @@ jobs:
           . venv_tests/bin/activate
           MOZ_LOG=console:5 xvfb-run python3 test_felt_browser_external_link.py ${{ steps.unpack.outputs.runtime_path }} $PWD/geckodriver $PWD/profiles/
 
+      - name: "test browser fxa"
+        run: |
+          . venv_tests/bin/activate
+          MOZ_LOG=console:5 python3 test_felt_browser_fxa.py ${{ steps.unpack.outputs.runtime_path }} $PWD/geckodriver $PWD/profiles/
```
lissyx, line 266: Those tests needs to be added to `taskcluster/kinds/enterprise-test/kind.yml`

1rneh: I add the test [here](https://github.com/mozilla/enterprise-firefox/pull/277/files/e0858099bf9e9015b1cb858525ee56ab39b48745#diff-54e80c1fee2c3087dddd0363165ee8394de76784f07c3c681c423c7d82ae07b9).

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -11,21 +11,87 @@ ChromeUtils.defineLazyGetter(lazy, "localization", () => {
...
+    } else {
+      // Disconnect sync
+      window.gSync.disconnect({ confirm: false, disconnectAccount: false });
+    }
+  },
+
+  observe(_, topic) {
+    switch (topic) {
+      case lazy.UIState.ON_UPDATE:
+        this.setUpSync();
```
jonathanmendez: Does this need to pass a window?

services/fxaccounts/moz.build
```diff
@@ -15,8 +15,7 @@ XPCSHELL_TESTS_MANIFESTS += ["tests/xpcshell/xpcshell.toml"]
 
 EXTRA_JS_MODULES += [
     "Credentials.sys.mjs",
-    "FxAccounts.sys.mjs",
-    "FxAccountsClient.sys.mjs",
+    "EnterpriseAccountsStorage.sys.mjs",
```
jonathanmendez: Should this only be added for enterprise builds? e.g.
```
if CONFIG["MOZ_ENTERPRISE"]:
    EXTRA_JS_MODULES += ["EnterpriseAccountsStorage.sys.mjs"]
```

.github/workflows/enterprise.yml
lissyx: Sorry, I was not seeing it in the list on the left

browser/components/preferences/sync.js
```diff
@@ -427,6 +428,10 @@ var gSyncPane = {
     this.setupEnginesUI();
     this.updateSyncUI();
 
+    if (lazy.AppConstants.MOZ_ENTERPRISE) {
+      this.restrictEnterpriseView();
```
lissyx, line 432: Ok that's nicer than what I did

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
```diff
@@ -0,0 +1,112 @@
...
+   * @param {Array<string>|string} [fieldNames=null]
+   *
+   * @returns {object} account data
+   */
+  async getAccountData(fieldNames = null) {
+    lazy.log.debug("getAccountData");
+
+    const data = await this.#getAccountDataPromise;
+
+    if (fieldNames) {
```
lissyx, line 64: Having had to scratch my head around all of that, I'm tempted to welcome a bit more of comments around why we need this, it was kinda working well (despite the Sync server issue, or was it related?) without this

browser/components/enterprise/EnterpriseHandler.sys.mjs
lissyx: Agreed

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
```diff
@@ -0,0 +1,112 @@
...
+        data[name] = value;
+        if (name === "device") {
+          const deviceId = data[name]?.id;
+          if (deviceId) {
+            Services.prefs.setStringPref(
+              lazy.EnterpriseCommon.ENTERPRISE_DEVICE_ID_PREF,
+              deviceId
+            );
+          }
+        }
```
lissyx, line 118: Is this the part that was making us bad and break Sync? If so, might be worth a bit of comments to make sure it's obvious in the future

services/fxaccounts/moz.build
lissyx: @jonathanmendez is right, it should only be included on `MOZ_ENTERPRISE` builds

lissyx (changes requested):
`setUpSync()` needs a window object? Otherwise looks fine.

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
1rneh: The `fieldNames` filtering matches what `FxAccountsStorage` already does, and avoids returning the full data object when only specific fields (e.g. `["device"]`) are requested.

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: Good catch, thanks!

services/fxaccounts/moz.build
1rneh: I moved `EnterpriseAccountsStorage.sys.mjs` to `browser/components/enterprise`. The component is only added in enterprise builds.

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
1rneh: This is to support device management. 
But I just noticed that something is not yet working in syncing tabs per device. I'll add more comments/descriptions once it's fixed.

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
1rneh: Let me know if anything could use more description.

fiji-flo (approved):
Good from my side. I'd land this.

The `ERROR	outgoing ClientRemoteTabs` comes from Nightly and most likely is just bad logging.

One thing we need to look at is whether we want to remove sending tabs. But I think that's a fine limitation for now.

browser/branding/enterprise/pref/firefox-enterprise.js
```diff
@@ -4,7 +4,18 @@
...
 pref("enterprise.console.address", "https://console.enterfox.eu");
+
+// Endpoint will be provided by the console.
+pref(
+  "identity.sync.tokenserver.uri",
+  "https://ent-dev-tokenserver.sync.nonprod.webservices.mozgcp.net/1.0/sync/1.5"
+);
+
+// Sync enabled by default.
+pref("enterprise.sync.enabledByDefault", true);
```
lissyx: nit: since this will also get into FELT, I'd feel safer if we verified in the tests we have that the pref is turned off later in felt.

browser/branding/enterprise/pref/firefox-enterprise.js
```diff
@@ -4,7 +4,18 @@
...
+
 pref("browser.profiles.enabled", false);
 pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
+
+pref("enterprise.loglevel", "Debug");
```
lissyx: Do we want to leave this ?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -11,21 +11,93 @@ ChromeUtils.defineLazyGetter(lazy, "localization", () => {
...
+  /**
+   * Customize Firefox Enterprise on startup
+   *    - Gets user information from the console and customizes the enterprise badge
+   *    - Sets up Sync
+   *    - Hides fxa toolbar button and app menu item (hamburger menu)
+   *
+   * @param {Window} window chrome window
+   */
   async init(window) {
+    this._chromeWindow = window;
```
lissyx: I tend to remember that `EnterpriseHandler` would be a singleton, dont we risk having multiple windows overwriting themselves?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -99,14 +173,21 @@ export const EnterpriseHandler = {
...
+
+    // Hides fxa item and separator in main view (hamburg menu)
+    this._chromeWindow.PanelUI.mainView.setAttribute(
+      "restricted-enterprise-view",
+      true
+    );
   },
 
-  async onSignOut(window) {
+  async onSignOut() {
```
lissyx: Since we already have it being passed from `await EnterpriseHandler.onSignOut(window);` in `browser/base/content/browser-sets.js`, why changing this ?

browser/components/enterprise/EnterpriseHandler.sys.mjs
lissyx: So it's from `browser/base/content/browser-init.js` and more precisely:
```
   155     BrowserUtils.callModulesFromCategory(
   156       { categoryName: "browser-window-domcontentloaded-before-tabbrowser" },
   157       window
   158     );
```

Code around mentions `// All of this needs setting up before we create the first remote browser.`

What would happen if e.g., I open a second window, then close the first one, and we have state changes of Sync that triggers `UIState.ON_UPDATE` ?

I'm not sure exactly what would happen there and if we have a risk that the `this._chromeWindow` has been lost in the meantime.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -11,21 +11,93 @@ ChromeUtils.defineLazyGetter(lazy, "localization", () => {
...
-    this.hideFxaToolbarButton(window);
-    this.updateBadge(window);
+  /**
+   * Check if the FxA state is initialised yet.
+   *    - If the state is still undefined, listen for a state update
+   *      and set up once the state update occurs.
+   *    - If the state is initialized, set up sync immediately.
+   */
+  setupSyncOnceInitialized() {
+    if (lazy.UIState.get().syncEnabled === undefined) {
```
lissyx: nit: could it be `null` ?

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
lissyx: Well that's a bit why I was asking, from the memory I had and I just checked, description in `FxAccountsStorage` is a bit vague and mentions storage/ in memory. Do those applies to us?

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
```diff
@@ -0,0 +1,133 @@
...
+export class EnterpriseStorageManager {
+  #getAccountDataPromise = Promise.reject(
+    "EnterpriseStorageManager: Initialize not called"
+  );
+
+  /**
+   * Initialize the storage by fetching account data from the remote console.
+   * This method swaps the internal #getAccountDataPromise with the promise
+   * returned by lazy.ConsoleClient.getFxAccountData(). Callers need to call
+   * this before invoking getAccountData() or updateAccountData().
```
lissyx: Is this a requirement from us or from how storage manager works?

lissyx (approved):
Just worried on the `_chromeWindow` thing, but maybe it's fine?

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: Oh, you pointing this out made me realize quite a flaw in how we initialize the `EnterpriseHandler`.

> I tend to remember that EnterpriseHandler would be a singleton, dont we risk having multiple windows overwriting themselves?

`EnterpriseHandler` is only a **per-process** singleton. It's initialized on `browser-window-domcontentloaded`. That means a new `EnterpriseHandler` instance is initialized for each new browser window, so the window passed will be the correct one. 

But.. what that also means is that some of the setup code, such as setting up the initial sync state, should really only run once on startup instead of for each new browser window. 

> What would happen if e.g., I open a second window, then close the first one, and we have state changes of Sync that triggers UIState.ON_UPDATE ?

Currently we only set up a listener for `UIState.ON_UPDATE` if `lazy.UIState.get().syncEnabled === undefined` which can only be the case on startup . But this logic still doesn't hold, because then we still check whether the current state equals the preference `enterprise.sync.enabledByDefault` and otherwise update the sync state. This is a problem, because the user might have changed the sync state in the meantime and we would override that when opening a new browser window!

I'll need to iterate over this again.

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
```diff
@@ -0,0 +1,133 @@
...
+   * Retrieve account data (or a subset of fields) from the cached console data.
+   *
+   * @param {Array<string>|string} [fieldNames=null]
+   *
+   * @returns {Promise<object>} Promise which resolves to the cached account data.
+   */
+  async getAccountData(fieldNames = null) {
+    lazy.log.debug("getAccountData");
+
+    const data = await this.#getAccountDataPromise;
```
gcp, line 62: Can this fail? `updateAccountData` has a guard here.

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
```diff
@@ -0,0 +1,133 @@
...
+    if (!newFields || "uid" in newFields) {
+      // Prevent callers from changing uid.
+      throw new Error("Can't change uid");
+    }
+
+    lazy.log.debug("_updateAccountData with items", Object.keys(newFields));
+
+    for (let [name, value] of Object.entries(newFields)) {
+      if (value == null) {
+        delete data[name];
```
gcp, line 104: Does this have special behavior? It looks superfluous.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -189,6 +221,21 @@ export const ConsoleClient = {
...
+  /**
+   * Fetch the account data used for fxa and sync.
+   *
+   * @returns {Promise<object>}
+   */
+  async getFxAccountData() {
+    const payload = await this._post(this._paths.SYNC_ACCOUNT, {
+      device_id: Services.prefs.getStringPref(
+        lazy.EnterpriseCommon.ENTERPRISE_DEVICE_ID_PREF,
+        undefined
```
gcp: I think this eventually becomes `{"device_id":null}` after passing through IDL and C++ code. Maybe that's what we want but I think it's better if that is explicit.

fiji-flo: What a footgun 🤯.

so
```JS
device_id: Services.prefs.getStringPref(
        lazy.EnterpriseCommon.ENTERPRISE_DEVICE_ID_PREF,
        null) ?? undefined
```
would produce what I meant. However @1rneh the API accepts a `null` so let's make it explicit and do:
```JS
device_id: Services.prefs.getStringPref(
        lazy.EnterpriseCommon.ENTERPRISE_DEVICE_ID_PREF,
        null)
```

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
fiji-flo: Core of JS: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: Note that according to IDL `getStringPref` really expects a string default, so it's probably better to default `""` and then check for that and return `null` instead (if needed).

`JSON.stringify` will turn `{device_id:undefined}` into `{}`. So I'm not clear if that was what you intended but it did seem to be sensible.

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
1rneh: This is how `FxAccountsStorage` works and I'm simply mirroring its behavior here.

lissyx: Well since it's far from being obvious when reading their code, I'd say it would be useful to write it in the comment, that you're just copying its behavior.

1rneh: Is adding this class description enough?
```
/**
 * EnterpriseStorageManager maintains and updates the FxA and Sync account data. 
 * 
 * Its implementation is mirroring the one from FxAccountsStorage, hence it surfaces the same 
 * APIs that are necessary to sign the user into FxA and to enable Sync.
 * Unlike FxAccountsStorage, which reads the storage data from the profile, EnterpriseStorageManager 
 * requests the required data once from the console and stores and udpates it in memory only.
 */
```

browser/branding/enterprise/pref/firefox-enterprise.js
1rneh: I'll send the preference in `FeltProcessParent` instead.

browser/branding/enterprise/pref/firefox-enterprise.js
1rneh: No :)

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
1rneh: Yes, if `.initialize` wasn't called in the first place or if `lazy.ConsoleClient.getFxAccountData()` returns a rejected promise. It's uncaught in FxAccountStorage as well.

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: I undid the caching of the chrome window.

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: I changed this to check for the `status` instead of the `syncEnabled` property.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: I'm defaulting to `""` and including `device_id: null` in the request body.

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
lissyx: Perfect!

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
gcp: I mean:

```
delete data[name];
data[name] = null;
```

The delete seems totally redundant?

gcp: In fact, the entire branch seems redundant, since you can just fall through to the `data[name] = value` branch, maybe just skipping the device handling if value happens to be `null`.

fiji-flo: I was just looking at the comment. I think that comes from my initial PoC and is inspired by https://searchfox.org/firefox-main/rev/268969d4a8096d6572447e0b03893ea7146797a2/services/fxaccounts/FxAccountsStorage.sys.mjs#210 but we only have one _cache_. So it should be just the `delete data[name]`

1rneh: Good catch! I removed `data[name] = null`.
 
> In fact, the entire branch seems redundant, since you can just fall through to the data[name] = value branch, maybe just skipping the device handling if value happens to be null.

We can't fall through though, because if `value === null` we still want to remove the property.

browser/branding/enterprise/pref/firefox-enterprise.js
```diff
@@ -14,3 +14,5 @@ pref(
 
 pref("browser.profiles.enabled", false);
 pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
+
+pref("enterprise.loglevel", "Error");
```
1rneh, line 18: @lissyx I had to put it back, because the pref needs to exist before I pass it to `console.createInstance`.

PR #283, opened 2025-12-16
FxCI: Shippable

lissyx:


taskcluster/gecko_taskgraph/util/verify.py
```diff
@@ -396,6 +398,31 @@ def has_one_route_with(what):
             )
 
 
+@verifications.add("full_task_graph")
+def verify_no_shippable_enterprise(
```
ahal: Just noting that there are probably times you'd want to schedule shippable builds on L1. You can totally schedule them for Firefox on try as well, we just don't do it by default. So while this verification is useful for testing, I'm not sure we'd actually want to land it.

taskcluster/kinds/build/linux.yml
```diff
@@ -52,16 +52,69 @@ linux64/opt:
...
+        using: mozharness
+        actions: [get-secrets, build]
+        config:
+            - builds/releng_base_firefox.py
+            - builds/releng_base_linux_64_builds.py
+            - taskcluster_nightly.py
+        script: "mozharness/scripts/fx_desktop_build.py"
+        secrets: true
+        tooltool-downloads: public
+    run-on-projects: ["enterprise-firefox"]
```
ahal, line 90: Upstream Taskgraph has a `run-on-git-branches` key we might be able to use:
https://github.com/taskcluster/taskgraph/blob/51941c5ceb9ad58c986497ba2a5779eaebf51a4c/src/taskgraph/transforms/task.py#L303

It's not currently available in `gecko_taskgraph`, but I'll make a quick patch to port it over.

taskcluster/gecko_taskgraph/util/verify.py
```diff
@@ -396,6 +398,31 @@ def has_one_route_with(what):
             )
 
 
+@verifications.add("full_task_graph")
```
ahal: The problem is that you're registering the verification at the `full_task_graph` phase, which means it runs before any of the code in `target_tasks.py` ever executes. You want it at the `target_task_graph` phase.

ahal (changes requested):

taskcluster/gecko_taskgraph/util/verify.py
lissyx: Maybe, but when I did pushed earlier it ended up scheduling them anyway

taskcluster/kinds/build/linux.yml
lissyx: But filtering them out in the target task function is not good enough?

taskcluster/gecko_taskgraph/util/verify.py
lissyx: We dont have try handling for now, but once this filtering works I can remove this check for sure

taskcluster/gecko_taskgraph/util/verify.py
lissyx:
```
Exception: Unexpected shippable status for task of level `1` `instrumented-build-linux64-enterprise-shippable/opt`: False
```

Even doing the verification on `target_task_graph`

taskcluster/kinds/build/linux.yml
ahal: Yeah, this would just be another way of accomplishing the same thing. No need to block, I was mostly just musing to myself.

taskcluster/gecko_taskgraph/transforms/enterprise_test.py
```diff
@@ -32,6 +32,10 @@ def fill_template(config, tasks):
 
             assert "enterprise" in this_task.get("name")
 
+            level = int(config.params["level"])
+            if level < 3 and "shippable" in task.get("name"):
+                continue
+
```
ahal: You should leave this filtering to the `target_tasks`. Filtering in the transforms is a bit of an anti-pattern.

taskcluster/gecko_taskgraph/target_tasks.py
ahal: Just move this block below the shippable one you created below and tests should get filtered out as well.

ahal (changes requested):

ahal (approved):
This looks reasonable.

PR #291, opened 2025-12-17
Enterprise: Handle browser crash

lissyx:
UX is open to discussion.

### DIFF
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -144,13 +144,12 @@ this.felt = class extends ExtensionAPI {
         );
         break;
 
-      case "FeltParent:FirefoxAbnormalExit":
-        Services.ppmm.removeMessageListener(
-          "FeltParent:FirefoxAbnormalExit",
-          this
-        );
-        // TODO: What should we do, restart Firefox?
+      case "FeltParent:FirefoxAbnormalExit": {
+        const success = Services.felt.makeBackgroundProcess(false);
+        console.debug(`FeltExtension: makeBackgroundProcess? ${success}`);
+        this.showWindow("felt-browser-error-multiple-crashes");
         break;
+      }
 
       case "FeltParent:FirefoxLogoutExit": {
         const success = Services.felt.makeBackgroundProcess(false);
@@ -188,13 +187,17 @@ this.felt = class extends ExtensionAPI {
     }
   }
 
-  showWindow() {
+  showWindow(errorMessage = "") {
     // Height and width are for now set to fit the sso.mozilla.com without the need to resize the window
     let flags =
       "chrome,private,centerscreen,titlebar,resizable,width=727,height=744";
+    const queryString = errorMessage
+      ? `?error=${encodeURIComponent(errorMessage)}`
+      : "";
+
     this._win = Services.ww.openWindow(
       null,
-      "chrome://felt/content/felt.xhtml",
+      `chrome://felt/content/felt.xhtml${queryString}`,
       "_blank",
       flags,
       null
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -55,6 +55,26 @@ export class FeltProcessParent extends JSProcessActorParent {
     // Track extension ready state (extension must register its observer)
     this.extensionReady = false;
 
+    this.abnormalExitCounter = 0;
+
+    // Amount of abnormal exit to allow over abnormal_exit_period
+    this.abnormalExitLimit = Services.prefs.getIntPref(
+      "enterprise.browser.abnormal_exit_limit",
+      3
+    );
+
+    /* Time period (in seconds) considered for checking the amount of abnormal
+     * exits. Hitting the limit defined above within this period will stop
+     * automatic restart and show user an error.
+     *
+     * confere shouldAbortRestarting()
+     */
+    this.abnormalExitPeriod = Services.prefs.getIntPref(
+      "enterprise.browser.abnormal_exit_period",
+      120
+    );
+    this.abnormalExitFirstTime = 0;
+
     this.restartObserver = {
       observe(aSubject, aTopic) {
         console.debug(`FeltExtension: ParentProcess: Received ${aTopic}`);
@@ -253,19 +273,63 @@ export class FeltProcessParent extends JSProcessActorParent {
           );
           if (!this.restartReported && !this.logoutReported) {
             if (this.proc.exitCode === 0) {
+              this.abnormalExitCounter = 0;
+              this.abnormalExitFirstTime = 0;
               Services.cpmm.sendAsyncMessage(
                 "FeltParent:FirefoxNormalExit",
                 {}
               );
             } else {
-              Services.cpmm.sendAsyncMessage(
-                "FeltParent:FirefoxAbnormalExit",
-                {}
-              );
+              this.handleRestartAfterAbnormalExit();
             }
           }
         });
       });
+  }
+
+  /**
+   * Handles the abnormal exit and decides whether to restart the Firefox
+   * again or to inform the user of the set of crashes.
+   */
+  handleRestartAfterAbnormalExit() {
+    if (this.abnormalExitCounter === 0) {
+      this.abnormalExitFirstTime =
+        Services.telemetry.msSinceProcessStart() / 1000;
+    }
+    this.abnormalExitCounter += 1;
+    if (this.shouldAbortRestarting()) {
+      console.debug(
+        "Abort restarting Firefox and inform the user of the crashes."
+      );
+      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxAbnormalExit", {});
+    } else {
+      console.debug("Trying to restart Firefox again.");
+      this.startFirefox([]);
+    }
+  }
+  /**
+   * Checks the state of the recent abnormal exits, meaning whether the crashes
+   * counter exceeds a pre-set counter limit within a pre-set time period.
+   *
+   * @returns {boolean} Whether these "abnormal" thresholds are exceeded.
+   */
+  shouldAbortRestarting() {
+    console.debug(
+      `Firefox AbnormalExit abnormalExitLimit=${this.abnormalExitLimit} abnormalExitCounter=${this.abnormalExitCounter} ; firstTime=${this.abnormalExitFirstTime} abnormalExitPeriod=${this.abnormalExitPeriod}`
+    );
+    // Have we reached the limit of allowed crashes ?
+    const isExceedingCrashCounterLimit =
+      this.abnormalExitCounter >= this.abnormalExitLimit;
+    // How much time since the first crash we recorded in this session ?
+    const timeSinceFirstCrash =
+      Services.telemetry.msSinceProcessStart() / 1000 -
+      this.abnormalExitFirstTime;
+    // Is the time since first crash too recent ?
+    const isWithinCrashPeriod = timeSinceFirstCrash <= this.abnormalExitPeriod;
+    console.debug(
+      `Firefox AbnormalExit crashLimitHit=${isExceedingCrashCounterLimit} timeSinceFirstCrash=${timeSinceFirstCrash} crashedNotLongAgoEnough=${isWithinCrashPeriod}`
+    );
+    return isExceedingCrashCounterLimit && isWithinCrashPeriod;
   }
 
   async startFirefoxProcess() {
--- a/browser/extensions/felt/content/felt.xhtml
+++ b/browser/extensions/felt/content/felt.xhtml
@@ -81,6 +81,13 @@
             </div>
           </form>
         </div>
+        <div class="felt-browser-error is-hidden">
+          <span
+            class="felt-browser-error-multiple-crashes is-hidden"
+            data-l10n-id="felt-browser-error-multiple-crashes"
+          >
+          </span>
+        </div>
         <span
           class="felt-powered-by text-deemphasized"
           data-l10n-id="felt-powered-by"
--- a/browser/extensions/felt/content/styles/felt.css
+++ b/browser/extensions/felt/content/styles/felt.css
@@ -160,3 +160,15 @@ body:has(#notification-popup-box[open]) .felt:not(#notification-popup-box .felt)
   opacity: 0.5;
   pointer-events: none;
 }
+
+.felt-browser-error {
+  position: absolute;
+  /* stylelint-disable-next-line */
+  bottom: 32px;
+  /* stylelint-disable-next-line */
+  left: 8px;
+  font-size: medium;
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
+  color: var(--icon-color-critical);
+  font-weight: var(--font-weight-bold);
+}
--- a/browser/extensions/felt/content/window.js
+++ b/browser/extensions/felt/content/window.js
@@ -97,6 +97,18 @@ async function listenFormEmailSubmission() {
   });
 }
 
+function informAboutPotentialStartupFailure() {
+  if (window.location.search) {
+    const errorClass = new URLSearchParams(window.location.search).get("error");
+    if (errorClass) {
+      document
+        .querySelector(".felt-browser-error")
+        .classList.remove("is-hidden");
+      document.querySelector(`.${errorClass}`).classList.remove("is-hidden");
+    }
+  }
+}
+
 function setupMarionetteEnvironment() {
   window.fullScreen = false;
 
@@ -186,6 +198,7 @@ window.addEventListener(
     setupMarionetteEnvironment();
     setupPopupNotifications();
     listenFormEmailSubmission();
+    informAboutPotentialStartupFailure();
   },
   true
 );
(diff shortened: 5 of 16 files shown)


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -256,10 +261,21 @@ export class FeltProcessParent extends JSProcessActorParent {
...
-              Services.cpmm.sendAsyncMessage(
-                "FeltParent:FirefoxAbnormalExit",
-                {}
+              this.abnormalExitCounter += 1;
+              console.debug(
+                `Firefox AbornmalExit amounts: ${this.abnormalExitCounter}`
               );
+              if (this.abnormalExitCounter < this.abnormalExitLimit) {
+                // Cookies would be out of private browsing anyway at that point.
+                // TODO: Assume they have not expired and just restart things or
```
gcp: I don't see why we care about them being valid at this particular point compared to anywhere else.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -256,10 +261,21 @@ export class FeltProcessParent extends JSProcessActorParent {
                 {}
               );
             } else {
-              Services.cpmm.sendAsyncMessage(
-                "FeltParent:FirefoxAbnormalExit",
-                {}
+              this.abnormalExitCounter += 1;
```
gcp: We don't ever seem to reset this. I think it makes sense to consider the last crash timing. If we crash say twice in 2 weeks (and Felt isn't restarted), it's probably not worth escalating. However if we crash twice immediately in succession, that's another matter.

browser/extensions/felt/api.js
```diff
@@ -188,13 +187,14 @@ this.felt = class extends ExtensionAPI {
...
-  showWindow() {
+  showWindow(errorMessage = "") {
     // Height and width are for now set to fit the sso.mozilla.com without the need to resize the window
     let flags =
       "chrome,private,centerscreen,titlebar,resizable,width=727,height=744";
+    const queryString = errorMessage ? `?error=${errorMessage}` : "";
     this._win = Services.ww.openWindow(
       null,
-      "chrome://felt/content/felt.xhtml",
+      `chrome://felt/content/felt.xhtml${queryString}`,
```
gcp, line 200: Needs `encodeURIComponent`

browser/extensions/felt/content/window.js
```diff
@@ -94,6 +94,20 @@ async function listenFormEmailSubmission() {
       connectToConsole(emailInput.value);
     }
   });
+
+  if (window.location.search) {
+    const params = Array.from(new URL(window.location).searchParams);
```
gcp: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get 

`const errorClass = new URLSearchParams(window.location.search).get("error");`

gcp (changes requested):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: I was thinking of the case where we would handle a crash after long running period, but maybe we should just try to refresh token there and handle it this way ?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Good point, I'll change it

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -253,21 +272,49 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    );
+    const crashLimitHit = this.abnormalExitCounter <= this.abnormalExitLimit;
+    const timeSinceFirstCrash =
+      Services.telemetry.msSinceProcessStart() / 1000 -
+      this.abnormalExit_firstTime;
+    const crashedLongAgoEnough = timeSinceFirstCrash >= this.abnormalExitPeriod;
+    console.debug(
+      `Firefox AbnormalExit crashLimitHit=${crashLimitHit} timeSinceFirstCrash=${timeSinceFirstCrash} crashedLongAgoEnough=${crashedLongAgoEnough}`
+    );
+    if (!(crashLimitHit && crashedLongAgoEnough)) {
```
gcp: Uhhh, this looks completely wrong to me?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -253,21 +272,49 @@ export class FeltProcessParent extends JSProcessActorParent {
...
 
+  /*
+   * Verifies the status of abnormal exit counter and limits to decide if it is
+   * OK to restart Firefox or should the user be warned of a problem.
+   */
+  shouldRestartAfterAbnormalExit() {
+    console.debug(
+      `Firefox AbnormalExit amounts: ${this.abnormalExitCounter} ; firstTime=${this.abnormalExit_firstTime} abnormalExitPeriod=${this.abnormalExitPeriod}`
+    );
+    const crashLimitHit = this.abnormalExitCounter <= this.abnormalExitLimit;
```
gcp: `crashLimitNotHit` or `belowCrashLimit`? It's possible my other remark is wrong because this name is the wrong way around.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -55,6 +55,25 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    /* Time period (in seconds) to allow for abnormal exit before stopping
+     * restarting the browser and showing an error.
+     *
+     * confere shouldRestartAfterAbnormalExit()
+     */
+    this.abnormalExitPeriod = Services.prefs.getIntPref(
+      "enterprise.browser.abnormal_exit_period",
+      120
+    );
+    this.abnormalExit_firstTime = 0;
```
gcp: Do we have lints for naming consistency? 😛

browser/extensions/felt/content/styles/felt.css
```diff
@@ -160,3 +160,14 @@ body:has(#notification-popup-box[open]) .felt:not(#notification-popup-box .felt)
...
 }
+
+.felt-browser-error {
+  position: absolute;
+  /* stylelint-disable-next-line */
+  bottom: 32px;
+  /* stylelint-disable-next-line */
+  left: 8px;
+  font-size: medium;
+  color: #ff0000;
```
gcp: Does this work correctly in dark mode and all? @1rneh

testing/enterprise/felt_browser_crashes.py
```diff
@@ -0,0 +1,45 @@
...
+        self._browser_pid = self._child_driver.capabilities["moz:processID"]
+        self._logger.info(f"Crashing browser at {self._browser_pid}")
+
+        try:
+            # This is going to trigger exception for sure
+            self._logger.info("Crashing main process")
+            self.open_tab_child("about:crashparent")
+        except Exception as ex:
+            self._logger.info(f"Caught exception {ex}")
+            pass
```
gcp, line 38: Why no `return True` like the other tests?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
gcp: I think the logic still isn't correct even accounting for this, which mean you should check the tests too, because they shouldn't have passed.

gcp (changes requested):

browser/extensions/felt/content/styles/felt.css
lissyx: no, this needs UX from matthew anyway

testing/enterprise/felt_browser_crashes.py
lissyx: because it's not a test method, just a helper

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: I dont think it is wrong, but maybe the way the condition is check can be confusing?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Fortunately no :)

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: I've reworked it, I hope it's less confusing

browser/extensions/felt/content/styles/felt.css
gcp: I mean you can use some var here that takes the right color in dark vs light mode.

testing/enterprise/felt_browser_crashes.py
gcp: But then `connect_and_crash` should have it?

lissyx: it does not need since it's not `test_` method

lissyx:
I'll fix the lint issues here, and keep this UX for now. We will improve later.

1rneh:
> I'll fix the lint issues here, and keep this UX for now. We will improve later.

As discussed last week, please get the UX approved by Matt and add the correct designs first.

browser/extensions/felt/content/styles/felt.css
1rneh: We could use `color: var(--icon-color-critical);` for now.

browser/extensions/felt/content/felt.xhtml
```diff
@@ -81,6 +81,13 @@
             </div>
           </form>
         </div>
+        <div class="felt-browser-error is-hidden">
+          <span
+            class="felt-browser-error-multiple-crashes is-hidden"
```
1rneh, line 86: Since the `.felt-browser-error` parent has the `is-hidden` class, the ancestor element `.felt-browser-error-multiple-crashes` already inherits that.

1rneh: Or is the idea to have a `<span>` element for each error category?

browser/extensions/felt/content/window.js
```diff
@@ -95,6 +95,16 @@ async function listenFormEmailSubmission() {
...
+
+  if (window.location.search) {
+    const errorClass = new URLSearchParams(window.location.search).get("error");
+    if (errorClass) {
+      document
+        .querySelector(".felt-browser-error")
+        .classList.remove("is-hidden");
+      document.querySelector(`.${errorClass}`).classList.remove("is-hidden");
+    }
+  }
```
1rneh, line 109: This is out of place in `listenFormEmailSubmissions`. Maybe add another function `informAboutPotentialStartupFailure` and call it once the window is loaded.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -253,21 +272,56 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      `Firefox AbnormalExit crashLimitHit=${crashLimitHit} timeSinceFirstCrash=${timeSinceFirstCrash} crashedNotLongAgoEnough=${crashedNotLongAgoEnough}`
+    );
+    // We have crashed too much for too little time, do not restart
+    // automatically and rather inform the user
+    if (crashLimitHit && crashedNotLongAgoEnough) {
+      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxAbnormalExit", {});
+    } else {
+      this.startFirefox([]);
+    }
+  }
```
1rneh, line 333: `shouldRestartAfterAbnormalExit` implies that it returns a `boolean`. I think it would help to split everything into multiple functions, because I think `startFirefox` already packs enough.
```suggestion
            } else {
              this.handleRestartAfterAbnormalExit()
            }
          }
        });
      });
  }

  /**
   * Handles the abnormal exit and decides whether to restart the Firefox 
   * again or to inform the user of the set of crashes.
   */
  handleRestartAfterAbnormalExit() {
    if (this.abnormalExitCounter === 0) {
      this.abnormalExitFirstTime =
        Services.telemetry.msSinceProcessStart() / 1000;
    }

    this.abnormalExitCounter += 1;

    if (this.shouldAbortRestarting()) {
      console.debug("Abort restarting Firefox and inform the user of the crashes.")
      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxAbnormalExit", {});
    } else {
      console.debug("Trying to restart Firefox again.")
      this.startFirefox([]);
    }
  }

  /**
   * Checks the state of the recent abnormal exits, meaning whether the crashes 
   * counter exceeds a pre-set counter limit within a pre-set time period.
   * 
   * @returns {boolean} Whether these "abnormal" thresholds are exceeded.
   */
  shouldAbortRestarting() {
    console.debug(
      `Firefox AbnormalExit abnormalExitLimit=${this.abnormalExitLimit} abnormalExitCounter=${this.abnormalExitCounter} ; firstTime=${this.abnormalExitFirstTime} abnormalExitPeriod=${this.abnormalExitPeriod}`
    );

    // Have we reached the limit of allowed crashes ?
    const isExceedingCrashCounterLimit = this.abnormalExitCounter >= this.abnormalExitLimit;

    // How much time since the first crash we recorded in this session ?
    const timeSinceFirstCrash =
      Services.telemetry.msSinceProcessStart() / 1000 -
      this.abnormalExitFirstTime;

    // Is the time since first crash too recent ?
    const isWithinCrashPeriod =
      timeSinceFirstCrash <= this.abnormalExitPeriod;
    console.debug(
      `Firefox AbnormalExit crashLimitHit=${isExceedingCrashCounterLimit} timeSinceFirstCrash=${timeSinceFirstCrash} crashedNotLongAgoEnough=${isWithinCrashPeriod}`
    );
    return isExceedingCrashCounterLimit && isWithinCrashPeriod
  }
```

1rneh (changes requested):
As re-discussed, if this is urgent to land, because QA needs the visible feedback about the crashes, then I'm ok with landing this now and follow-up on the UX once that is defined.

Additionally I have some suggestions about the code structure.

browser/extensions/felt/content/felt.xhtml
lissyx: that was the idea, but since UX is not clear I dont think we need to think too much about it

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: That's a nice improvement, I'm taking it

browser/extensions/felt/content/styles/felt.css
1rneh: Could you still add the design token here?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -55,6 +55,25 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    this.abnormalExitLimit = Services.prefs.getIntPref(
+      "enterprise.browser.abnormal_exit_limit",
+      3
+    );
+
+    /* Time period (in seconds) to allow for abnormal exit before stopping
+     * restarting the browser and showing an error.
+     *
+     * confere shouldRestartAfterAbnormalExit()
+     */
```
1rneh: I find the description confusing. What I get from the code is:
- The limit of exits is reached within the time period --> abort restarting and inform the user
- The limit of exits is reached but not within the time period --> restart again

The description reads as if the exits within the period always result in a restart and outside of the time period we will stop restarting. Could you clarify what is intended?

Also `shouldRestartAfterAbnormalExit` doesn't exist anymore.

testing/enterprise/felt_browser_crash_abort_restart.json
```diff
@@ -0,0 +1,14 @@
+{
```
1rneh, line 1: Could you rename this test suite to something more informative such as `felt_browser_crash_abort_restart.json`, we also "handle" the crash when we decide to restart.

1rneh (changes requested):
Thank you! Just some smaller things... Only important thing left is to clarify the intention of the time period.

browser/extensions/felt/content/styles/felt.css
lissyx: Sure, done

1rneh (approved):
Lgtm now, thanks!

browser/extensions/felt/content/styles/felt.css
```diff
@@ -160,3 +160,15 @@ body:has(#notification-popup-box[open]) .felt:not(#notification-popup-box .felt)
...
 }
+
+.felt-browser-error {
+  position: absolute;
+  /* stylelint-disable-next-line */
+  bottom: 32px;
+  /* stylelint-disable-next-line */
+  left: 8px;
+  font-size: medium;
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
```
gcp, line 171: Was this lint exception actually needed? Didn't you forget to remove this?

1rneh: Yes, https://treeherder.mozilla.org/logviewer?job_id=543562202&repo=enterprise-firefox-pr&task=DGk37lvMQKqsKjfUW9OEYA.0&lineNumber=212. We agreed that once we have a well defined crash ux, this will be revamped anyhow.

PR #293, opened 2025-12-17
FxCI: Add Windows Aarch64

lissyx:


python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -3062,7 +3062,7 @@ def repackage_msi(
 @CommandArgument(
     "--channel",
     type=str,
-    choices=["official", "beta", "esr", "aurora", "nightly", "unofficial"],
+    choices=["official", "beta", "esr", "aurora", "nightly", "unofficial", "FirefoxEnterprise"],
```
ahal, line 3065: I feel like "enterprise" is more in-line with the other options, but this is firmly your call

taskcluster/gecko_taskgraph/transforms/repackage.py
ahal, line 528: Normally we would accomplish this by copying attributes over from the dependency task, this can be done via the `from_deps` transforms.. but we can clean this up as follow-up.

ahal (approved):

PR #300, opened 2025-12-19
Enterprise network informations device posture

lissyx:


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -473,15 +473,17 @@ export const ConsoleClient = {
    * @returns {DevicePosture} devicePosture
    */
   _collectDevicePosture() {
+    Cc["@mozilla.org/network/network-link-service;1"]
+      .createInstance()
+      .QueryInterface(Ci.nsINetworkLinkService);
```
lissyx: I have seen cases locally while hacking where the instance may not be started, but I'm not sure this is required

netwerk/base/nsNetworkInterface.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+void NetworkInterface::setMAC(const uint8_t* aAddr) {
+  mMAC = nsPrintfCString("%02x:%02x:%02x:%02x:%02x:%02x", aAddr[0], aAddr[1],
+                         aAddr[2], aAddr[3], aAddr[4], aAddr[5]);
+}
+
+#if defined(XP_WIN)
+NetworkInterface::NetworkInterface(PIP_ADAPTER_ADDRESSES aAdapter)
+    : mName(NS_ConvertUTF16toUTF8(aAdapter->FriendlyName)) {
+  uint8_t macAddress[6] = {0, 0, 0, 0, 0, 0};
```
gcp: This needs to be `MAX_ADAPTER_ADDRESS_LENGTH`. Which is not 6 😄

netwerk/base/nsNetworkInterface.cpp
```diff
@@ -0,0 +1,279 @@
...
+  memcpy(&macAddress, aAdapter->PhysicalAddress,
+         aAdapter->PhysicalAddressLength);
+  setMAC(macAddress);
```
gcp, line 52: What if `PhysicalAddressLength` was not 6? What if it is 0?

netwerk/base/nsNetworkInterface.cpp
```diff
@@ -0,0 +1,279 @@
...
+}
+#endif
+
+#if defined(XP_MACOSX)
+NetworkInterface::NetworkInterface(
+    const char* aName,
+    nsTHashMap<nsCString, nsTArray<std::pair<int, nsCString>>>& aRoutes,
+    struct ifaddrs* aIfap)
+    : mName(aName) {
+  // setMAC(aMac);
```
gcp: Yes? No? Needed or not? I guess not because it's done explicitly below?

netwerk/base/nsNetworkInterface.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+void NetworkInterface::AddGWv6(const nsACString& aIp) {
+  mGwv6.AppendElement(std::move(aIp));
+}
+
+void NetworkInterface::setMAC(struct sockaddr_dl* aLink) {
+  if (aLink) {
+    uint8_t mac_addr[aLink->sdl_alen];
+    memcpy(mac_addr, aLink->sdl_data + aLink->sdl_nlen, aLink->sdl_alen);
+    setMAC(mac_addr);
```
gcp: Probably same issue that this can't cope with unexpected lengths.

netwerk/build/nsNetCID.h
```diff
@@ -12,260 +12,260 @@
...
 // service implementing nsIIOService
 #define NS_IOSERVICE_CONTRACTID "@mozilla.org/network/io-service;1"
-#define NS_IOSERVICE_CID                      \
-  {/* 9ac9e770-18bc-11d3-9337-00104ba0fd40 */ \
-   0x9ac9e770,                                \
-   0x18bc,                                    \
-   0x11d3,                                    \
-   {0x93, 0x37, 0x00, 0x10, 0x4b, 0xa0, 0xfd, 0x40}}
+#define NS_IOSERVICE_CID                             \
+  { /* 9ac9e770-18bc-11d3-9337-00104ba0fd40 */       \
```
gcp: Don't do this, this will make merging a pain.

netwerk/system/mac/nsNetworkLinkService.mm
```diff
@@ -988,6 +993,49 @@ nsAutoCString newString(reinterpret_cast<char*>(digest),
...
+  std::vector<const char*> ifNames;
+  struct ifaddrs* ifa;
+  for (ifa = ifap; ifa; ifa = ifa->ifa_next) {
+    if (ifa->ifa_addr == NULL) {
+      continue;
+    }
+
+    if (std::find(ifNames.begin(), ifNames.end(), ifa->ifa_name) ==
+        ifNames.end()) {
+      ifNames.emplace_back(ifa->ifa_name);
```
gcp: `std::set`?

This works too, it's up to you.

netwerk/system/netlink/NetlinkService.cpp
```diff
@@ -500,17 +527,17 @@ class NetlinkMsg {
   bool SendRequest(int aFD, void* aRequest, uint32_t aRequestLength) {
     MOZ_ASSERT(!mIsPending, "Request has been already sent!");
 
-    struct sockaddr_nl kernel{};
+    struct sockaddr_nl kernel {};
```
gcp: Meeh, don't do this.

netwerk/system/mac/nsNetworkLinkService.mm
gcp: Given that `ifa_name` is a `char*` and not a `std::string`, don't you need a string comparison here?

netwerk/system/netlink/NetlinkService.cpp
```diff
@@ -1908,6 +1935,52 @@ nsresult NetlinkService::GetResolvers(nsTArray<NetAddr>& aResolvers) {
 #endif
 }
 
+#if defined(MOZ_ENTERPRISE)
+nsresult NetlinkService::GetNetworkInterfaces(
+    nsTArray<NetworkInterface>& aNetworkInterfaces) {
+  nsTArray<NetworkInterface> networkInterfaces;
+  for (const auto& linkInfo : mLinks.Values()) {
```
gcp, line 1946: Doesn't this need `MutexAutoLock lock(mMutex);` ?

gcp (changes requested):
This is very complicated and big, probably needs another set of eyes.

netwerk/build/nsNetCID.h
lissyx: Right, removed

netwerk/system/netlink/NetlinkService.cpp
lissyx: linters :(

netwerk/base/nsNetworkInterface.cpp
lissyx: Yes, leftover sorry.

netwerk/system/mac/nsNetworkLinkService.mm
lissyx: Maybe. We might be lucky here because `ifa->ifa_name` seems to always point to the same `char *` but I have no idea if that's really pure luck or by design

netwerk/base/nsNetworkInterface.cpp
lissyx: MSDN checks for PhysicalAddressLength != 0 :)

netwerk/system/netlink/NetlinkService.cpp
lissyx: I have seen the mutex only used on a subset of members, like  `mListener`, `mDNSSuffixList` or `mDNSResolvers`, not none of the users of `mLinks` use it

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: `getServices()` in the line below should be enough?

netwerk/system/linux/nsNetworkLinkService.cpp
```diff
@@ -215,3 +215,31 @@ void nsNetworkLinkService::NotifyObservers(const char* aTopic,
...
+  for (const auto& intf : interfaces) {
+    aNetworkInterfaces.AppendElement(MakeRefPtr<nsNetworkInterface>(&intf));
+  }
+
+  return NS_OK;
+}
+#else
+NS_IMETHODIMP
+nsNetworkLinkService::GetNetworkInterfaces(
+    nsTArray<nsINetworkInterface>& aNetworkInterfaces) {
```
gcp: Signature doesn't match the other branch. Make sure you test non-Enterprise builds or this won't be mergable back.

netwerk/base/nsNetworkInterface.cpp
```diff
@@ -0,0 +1,285 @@
...
+  mIpv4.AppendElement(std::move(ip));
+}
+
+void NetworkInterface::AddIP(const struct sockaddr_in6* aAddr) {
+  nsCString ip;
+  GetIP(aAddr, ip);
+  mIpv6.AppendElement(std::move(ip));
+}
+
+void NetworkInterface::AddGWv4(const nsACString& aIp) {
```
gcp: `nsACString&&` else the `std::move` won't achieve anything?

netwerk/base/nsNetworkInterface.h
```diff
@@ -0,0 +1,126 @@
...
+  NetworkInterface(const nsAutoCString aName, const uint8_t* aMac);
+  void AddIP(const struct in_addr* aAddr);
+  void AddIP(const struct in6_addr* aAddr);
+  void AddGW(const struct in_addr* aAddr);
+  void AddGW(const struct in6_addr* aAddr);
+#endif
+
+  NetworkInterface(const NetworkInterface& aIntf);
+
+  nsCString& Name() { return mName; }
```
gcp: `const`?

netwerk/base/nsNetworkInterface.h
```diff
@@ -0,0 +1,126 @@
...
+  nsCString& Mac() { return mMAC; }
```
gcp: Same.

netwerk/system/win32/nsNotifyAddrListener.cpp
```diff
@@ -734,6 +740,24 @@ void nsNotifyAddrListener::CheckLinkStatus(void) {
...
+  MutexAutoLock lock(mMutex);
+  for (const auto& intf : mNetworkInterfaces) {
+    aNetworkInterfaces.AppendElement(MakeRefPtr<nsNetworkInterface>(&intf));
+  }
+  return NS_OK;
+}
+#else
+NS_IMETHODIMP
+nsNotifyAddrListener::GetNetworkInterfaces(
+    nsTArray<nsINetworkInterface>& aNetworkInterfaces) {
```
gcp: Same, different signature.

gcp (changes requested):
Do we want a Necko peer to at least briefly look at this?

lissyx:
> Do we want a Necko peer to at least briefly look at this?

Likely, that was my plan before the holidays, but I wanted to clean things first

netwerk/base/nsNetworkInterface.h
lissyx: it's ok if move from `nsCString&` to `nsCString`

valenting:
Looks good generally. My main question is whether we need to guard the new code on MOZ_ENTERPRISE ?

netwerk/system/netlink/NetlinkService.cpp
valenting: mLinks was previously only called on the [netlink thread](https://searchfox.org/firefox-main/query/default?q=calls-to%3A%27mozilla%3A%3Anet%3A%3ANetlinkService%3A%3AmLinks%27%20depth%3A4) so a lock wasn't needed.
It now seems getNetworkInterfaces can be called from JS (ie main thread, so there should be a lock here, otherwise we have a data race.

lissyx: Right, I'll add it everywhere then.

lissyx:
> Looks good generally. My main question is whether we need to guard the new code on MOZ_ENTERPRISE ?

In the light that we plan to merge this back to mozilla-central as soon as feasible, I think we want

valenting:
> > Looks good generally. My main question is whether we need to guard the new code on MOZ_ENTERPRISE ?
> 
> In the light that we plan to merge this back to mozilla-central as soon as feasible, I think we want

My main concern with guarding new code by MOZ_ENTERPRISE is that if there's a bug in the MOZ_ENTERPRISE code, we might not see it in regular build. So thread safety and logic bugs might be missed by CI runs.
I would suggest the added capabilities and tests run in all Firefox flavors, even if nsINetworkLinkService.networkInterfaces is only used in enterprise builds.

lissyx:
> > > Looks good generally. My main question is whether we need to guard the new code on MOZ_ENTERPRISE ?
> > 
> > 
> > In the light that we plan to merge this back to mozilla-central as soon as feasible, I think we want
> 
> My main concern with guarding new code by MOZ_ENTERPRISE is that if there's a bug in the MOZ_ENTERPRISE code, we might not see it in regular build. So thread safety and logic bugs might be missed by CI runs. I would suggest the added capabilities and tests run in all Firefox flavors, even if nsINetworkLinkService.networkInterfaces is only used in enterprise builds.

We have testing https://treeherder.mozilla.org/jobs?repo=enterprise-firefox and I'm making progress so that classic test suites are running as well. Do you think it's not enough?

valenting:
> We have testing https://treeherder.mozilla.org/jobs?repo=enterprise-firefox and I'm making progress so that classic test suites are running as well. Do you think it's not enough?

It seems to me that enterprise specific features can be tested on the enterprise branch, but general platform implementations should run and be tested under all branches. That would make me feel more confident that enterprise specific code isn't introducing issues that are difficult to reproduce.

testing/enterprise/test_felt_device_posture.py
```diff
@@ -40,9 +40,24 @@ def test_felt_2_device_posture_content(self, exp):
...
+        )
+        device_posture_network = device_posture["network"][0]
+
+        assert len(device_posture_network["ipv4"]) >= 1, (
+            "Device posture reports network interfaces (IPv4)"
+        )
+        assert len(device_posture_network["ipv6"]) >= 1, (
+            "Device posture reports network interfaces (IPv6)"
+        )
+        assert len(device_posture_network["mac"]) == 17, (
```
gcp: This reminds me, are you filtering the loopback interface in this code?

(generally, not just the tests)

lissyx: No, there is no filtering yet, since the initial plan was to collect all interfaces. But we can change that for sure.

gcp: I mention this here because this test assumes it won't get the loopback.

lissyx: Weird, last time I checked the output it was including it

lissyx: @gcp Do you think we want to include loopback or rather avoid? I could augment the test for that as well

gcp: I mean this test assumes the first interface has an Ethernet MAC, which will fail if it's the loopback (at least I think?), even if the second interface has one.

I don't see the point of including the loopback in posture, it doesn't give us any information.

lissyx: Interfaces without a mac should report `00:00:00:00:00:00`, it's the behavior on linux, but it looks like macOS is different here. I'll skip the loopback interfaces then.

netwerk/system/netlink/NetlinkService.cpp
```diff
@@ -1908,6 +1937,53 @@ nsresult NetlinkService::GetResolvers(nsTArray<NetAddr>& aResolvers) {
...
 }
 
+#if defined(MOZ_ENTERPRISE)
+nsresult NetlinkService::GetNetworkInterfaces(
+    nsTArray<NetworkInterface>& aNetworkInterfaces) {
+  MutexAutoLock lock(mMutex);
+
+  nsTArray<NetworkInterface> networkInterfaces;
+  for (const auto& linkInfo : mLinks.Values()) {
+    if (linkInfo->mIsUp) {
```
lissyx, line 1947: https://searchfox.org/firefox-main/rev/5917a9f2af3294b27a325371c5c499e7dd9554fd/netwerk/system/netlink/NetlinkService.cpp#258-261 so we dont report loopback on linux @gcp

netwerk/system/win32/nsNotifyAddrListener.cpp
```diff
@@ -497,6 +500,11 @@ nsNotifyAddrListener::CheckAdaptersAddresses(void) {
         continue;
       }
 
+#if defined(MOZ_ENTERPRISE)
+      // Skip non up interfaces as well as loopback
+      networkInterfaces.AppendElement(NetworkInterface(adapter));
```
lissyx, line 505: @gcp moved this so we skip loopback interfaces

netwerk/system/mac/nsNetworkLinkService.mm
```diff
@@ -988,6 +999,82 @@ nsAutoCString newString(reinterpret_cast<char*>(digest),
...
+
+    if (ioctl(s, SIOCGIFXMEDIA, (caddr_t)&ifmr) < 0) {
+      close(s);
+      continue;
+    }
+    
+    if (!(ifmr.ifm_status & IFM_ACTIVE)) {
+      close(s);
+      continue;
+    }
```
lissyx, line 1055: @gcp this allows up to skip interfaces that are:
 - not up or not running
 - loopback
 - or dont report as active on the media level

netwerk/system/mac/nsNetworkLinkService.mm
```diff
@@ -988,6 +999,85 @@ nsAutoCString newString(reinterpret_cast<char*>(digest),
...
+
+    if (!(ifmr.ifm_status & IFM_ACTIVE)) {
+      close(s);
+      continue;
+    }
+
+    std::string ifa_name(ifa->ifa_name);
+    if (std::find(ifNames.begin(), ifNames.end(), ifa_name) == ifNames.end()) {
```
lissyx, line 1058: I still need to fix this, but I'm working on it

taskcluster/config.yml
```diff
@@ -840,11 +840,16 @@ workers:
...
+        win11-64-24h2(|-gpu|-large|-webgpu):
             provisioner: '{trust-domain}-t'
             implementation: generic-worker
             os: windows
             worker-type: '{alias}'
+        win11-64-24h2-source:
+            provisioner: '{trust-domain}-t'
+            implementation: generic-worker
+            os: windows
+            worker-type: '{alias}-alpha'
```
jwmossmoz, line 852: @lissyx You can remove the `-alpha` as this worker pool now supports the ipv6 adapter after merging https://github.com/mozilla-releng/fxci-config/pull/741

lissyx: Thanks, updated!

netwerk/base/nsNetworkInterface.cpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+void NetworkInterface::AddIP(const struct sockaddr_in6* aAddr) {
+  nsCString ip;
+  GetIP(aAddr, ip);
+  mIpv6.AppendElement(std::move(ip));
+}
+
+void NetworkInterface::AddGWv4(const nsACString&& aIp) {
+  mGwv4.AppendElement(std::move(aIp));
```
gcp: Can't move `const &&` (would lose the constness in almost all cases)

netwerk/system/mac/nsNetworkLinkService.mm
```diff
@@ -988,6 +999,85 @@ nsAutoCString newString(reinterpret_cast<char*>(digest),
...
+      ifNames.emplace_back(ifa_name);
+    }
```
gcp, line 1060: `close(s)`?

gcp (changes requested):

PR #305, opened 2025-12-22
Include "Enterprise" in Firefox wordmark

1rneh:


gcp:
Screenshot 2025-12-23 at 16 37 14

There's an extremely subtle bug here: your new version adds `fill="black"` to the end of the SVG path. This breaks the mark in private browsing mode, and probably dark mode.

(I would have never noticed this, but Codex pointed it out and I verified it is correct)

gcp (changes requested):
See above.

1rneh:
@gcp Nice catch!

gcp (approved):

PR #307, opened 2025-12-22
Blocked domain telemetry: Add referrer and original URL for redirects

jonathanmendez:
1. Add referrer to telemetry.
  - Move telemetry from WebsiteFilter's `shouldLoad` to the caller in C++ that has access to the IHttpChannel with referrer information
  - Add code to obtain original referrer to the telemetry
  - For redirect responses, add code to observer in WebsiteFilter to include original referrer in telemetry
2. Since URLs that are blocked after redirect responses (3xx) do not go through the same code path, send telemetry in WebsiteFilter's `observe` callback when it blocks a redirected URL.
  - In the case of a redirect, include an `original_url` field.
  - This will require changes to the console ingestion (in PR to console repo)
3. Add a test for the telemetry


browser/components/enterprisepolicies/tests/browser/browser_policy_websitefilter_telemetry.js
```diff
@@ -0,0 +1,149 @@
...
+    referrerURL,
+  });
+  await checkBlockedPageTelemetry(
+    "view-source:" + SUPPORT_FILES_PATH + BLOCKED_PAGE
+  );
+  await checkBlockedPageTelemetry(
+    "about:reader?url=" + SUPPORT_FILES_PATH + BLOCKED_PAGE
+  );
+  await checkBlockedPageTelemetry(
+    "about:READER?url=" + SUPPORT_FILES_PATH + BLOCKED_PAGE
```
gcp: This is the only occurence of `about:READER` in the codebase?  Maybe it's a matter of just normalizing the URL in the test function?

browser/components/enterprisepolicies/tests/browser/browser_policy_websitefilter_telemetry.js
```diff
@@ -0,0 +1,149 @@
...
+    }
+    await promise;
+
+    let events =
+      Glean.contentPolicy.blocklistDomainBrowsed.testGetValue("enterprise");
+    Assert.ok(events?.length, "Should have recorded events");
+    if (!events?.length) {
+      return;
+    }
+    Assert.greaterOrEqual(events.length, 1, "Should record at least one event"); // TODO this should eventually be exactly 1
```
gcp, line 118: TODO?

browser/components/enterprisepolicies/metrics.yaml
```diff
@@ -32,6 +32,15 @@ content_policy:
     extra_keys:
       url:
         description: >
-          The blocked url that was browsed.
+          The url that was blocked by policy.
+        type: string
+      original_url:
+        description: >
+          The original url, prior to redirects, that was requested resulting in a block,
```
gcp, line 39: FWIW, for myself I set a block on `morbo.be`, then navigate to `https://sjeng.org/referrer-hop.html`, which is:

```  
  document.getElementById("go").addEventListener("click", () => {
    location.href = "http://morbo.org/";
  });
```

`morbo.org` has a redirect to www.morbo.be

This records

```
  "events": [                                                                                                
    {                                                                                                        
      "category": "content_policy",                                                                          
      "extra": {                                                                                             
        "glean_timestamp": "1766755855522",                                                                  
        "original_url": "http://morbo.org/",                                                                 
        "referrer": "",                                                                                      
        "url": "http://www.morbo.be/"                                                                        
      },                                                                                                     
      "name": "blocklist_domain_browsed",                                                                    
      "timestamp": 0                                                                                         
    }                                                                                                        
  ],
```

a previous version of this patch was still getting the original referrer right:

```
  "events": [
    {
      "category": "content_policy",
      "extra": {
        "glean_timestamp": "1766060860297",
        "referrer": "https://sjeng.org/referrer-hop.html",
        "url": "http://www.morbo.be/"
      },
      "name": "blocklist_domain_browsed",
      "timestamp": 0
    }
  ],
```

browser/components/enterprisepolicies/tests/browser/browser_policy_websitefilter_telemetry.js
gcp: Ah this refers to the debouncing/deduplication which this is still missing.

browser/components/enterprisepolicies/metrics.yaml
gcp: I've added https://sjeng.org/referrer-hop-direct.html, which is

```
  document.getElementById("go").addEventListener("click", () => {
    location.href = "http://www.morbo.be/";
  });
```

this records

```
  "events": [
    {
      "category": "content_policy",
      "extra": {
        "glean_timestamp": "1766756680769",
        "referrer": "",
        "url": "http://www.morbo.be/"
      },
      "name": "blocklist_domain_browsed",
      "timestamp": 0
    }
  ],
```

So the referrer isn't detected correctly (...any more).

gcp (changes requested):

browser/components/enterprisepolicies/tests/browser/browser_policy_websitefilter_telemetry.js
jonathanmendez: I copied these test cases from the [policy filter tests](https://searchfox.org/firefox-main/source/browser/components/enterprisepolicies/tests/browser/browser_policy_websitefilter.js#42). I can potentially see caring about testing the different casing there (making sure that using unusual casing doesn't avoid the block), but here the testing is more for "when we block, we get telemetry" so the extra casing isn't necessary. I'll remove it from this test.

browser/components/enterprisepolicies/tests/browser/browser_policy_websitefilter_telemetry.js
jonathanmendez: Yes, I planned to implement the debouncing/deduplication in a separate PR. I don't know if we have a policy around landing TODOs into the code base and/or tagging them with links to bugs to track, etc.?

browser/components/enterprisepolicies/metrics.yaml
jonathanmendez: In an offline discussion we clarified that the lack of referrer in this case is a result of the choice to use the "computed referrer", i.e. matching what is in the "Referer" header after applying the [referrer-policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Referrer-Policy). In these scenarios, the block is happening while executing an http request from an https referrer, so the policy has stripped the referrer.

It is technically easy to instead use the original referrer (pre-policy), so we're left with a judgment call of which is the correct data to include. I'll pursue that question and update here accordingly.

browser/components/enterprisepolicies/tests/browser/browser_policy_websitefilter_telemetry.js
gcp: Just make sure the bug is on file, code comments don't necessarily guarantee stuff gets done so those are optional IMHO.

gcp (approved):

browser/components/enterprisepolicies/tests/browser/browser_policy_websitefilter_telemetry.js
jonathanmendez: https://bugzilla.mozilla.org/show_bug.cgi?id=2009442
Once this is merged I'll update the bug with a link to this code to make sure it gets updated as part of the fix.

browser/components/enterprisepolicies/metrics.yaml
jonathanmendez: I've gone ahead and moved to the original referrer after discussion, as the purpose of the referrer-policy does not apply here. That means that referrer is now successfully included in the above test scenarios.

PR #310, opened 2025-12-26
Forward Felt new window requests via IPC.

gcp:
Extend the URL opening IPC with a flag indicating where to open things.


browser/components/BrowserContentHandler.sys.mjs
```diff
@@ -1211,9 +1243,23 @@ nsBrowserContentHandler.prototype = {
 var gBrowserContentHandler = new nsBrowserContentHandler();
 
 // Module-level queue for Felt external link handling
-// URLs are stored here when they arrive via command line (before Felt extension loads)
+// URL requests are stored here when they arrive via command line (before Felt extension loads)
 // FeltProcessParent imports this module and manages forwarding from this queue
 export let gFeltPendingURLs = [];
+const FELT_OPEN_DISPOSITION_DEFAULT = 0;
+const FELT_OPEN_DISPOSITION_NEW_WINDOW = 1;
+const FELT_OPEN_DISPOSITION_NEW_PRIVATE_WINDOW = 2;
```
1rneh: Let's export these flags here and import them wherever needed (e.g. in `browser/base/content/nonbrowser-mac.js` and `browser/extensions/felt/api.js`). Maybe also wrap them in an enum?
```
export const FELT_OPEN_WINDOW_DISPOSITION = {
  DEFAULT : 0,
  NEW_WINDOW : 1,
  NEW_PRIVATE_WINDOW : 2,
}
```

browser/components/BrowserContentHandler.sys.mjs
```diff
@@ -1595,7 +1642,7 @@ nsDefaultCommandLineHandler.prototype = {
 
     // Make sure that when FeltUI is requested, we do not try to open another
     // window. Instead, forward any URLs to be opened in the real Firefox.
-    if (Services.felt.isFeltUI()) {
+    if (isFeltUI) {
```
1rneh: Could we delcare `isFeltUI` closer to its usage? Or actually, since it's only used once, we could generally drop the variable and call `Services.felt.isFeltUI();` directly.

browser/components/BrowserContentHandler.sys.mjs
```diff
@@ -1211,9 +1243,23 @@ nsBrowserContentHandler.prototype = {
...
+const FELT_OPEN_DISPOSITION_NEW_PRIVATE_WINDOW = 2;
+
+export function queueFeltURL(payload) {
+  try {
+    const { queueURL } = ChromeUtils.importESModule(
+      "chrome://felt/content/FeltProcessParent.sys.mjs"
+    );
+    queueURL(payload);
+  } catch (err) {
+    gFeltPendingURLs.push(payload);
```
1rneh: Maybe also leave a log message ```console.warn(`Retrying to queue url ${payload.url} after initial failure.`);```

1rneh (changes requested):
Thanks, overall the changes on the js side look good!

Two things I noticed:
- Most of these changes still need some `#ifdef MOZ_ENTERPRISE`-ing.
- The code that handles external links and requests new windows is very distributed. I think a new module that centralizes all that logic would help.

browser/base/content/nonbrowser-mac.js
```diff
@@ -3,11 +3,34 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var { AppConstants } = ChromeUtils.importESModule(
+  "resource://gre/modules/AppConstants.sys.mjs"
+);
```
lissyx: Since we rely on `Services.felt?`, maybe we can avoid importing `AppConstants` and just use it?

browser/base/content/nonbrowser-mac.js
```diff
@@ -3,11 +3,34 @@
...
+
+function queueFeltDockAction(isPrivate) {
+  if (!AppConstants.MOZ_ENTERPRISE) {
```
lissyx: I would also guard on `Services.felt.isBrowserUI()` to make sure the browser does not get a way to interact and all flows through FELT ?

browser/base/content/nonbrowser-mac.js
```diff
@@ -3,11 +3,34 @@
...
+  };
+  queueFeltURL(payload);
+}
+
 var NonBrowserWindow = {
   delayedStartupTimeoutId: null,
   MAC_HIDDEN_WINDOW: "chrome://browser/content/hiddenWindowMac.xhtml",
 
   openBrowserWindowFromDockMenu(options = {}) {
+    if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI()) {
```
lissyx: Ok we check here

browser/base/content/nonbrowser-mac.js
```diff
@@ -108,11 +131,47 @@ var NonBrowserWindow = {
...
+      // In Felt mode, disable dock menu items until Firefox is ready
+      if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI()) {
+        this.setupFeltDockMenuState();
+      }
     }
 
     this.delayedStartupTimeoutId = setTimeout(() => this.delayedStartup(), 0);
   },
 
+  setupFeltDockMenuState() {
```
lissyx, line 138: Is this controlling appearance of the icon in the dock? This might be a better handling than the racy things we did in the past to make sure FELT and browser would not overlap each other in the dock?

browser/components/BrowserContentHandler.sys.mjs
```diff
@@ -469,7 +472,14 @@ nsBrowserContentHandler.prototype = {
       Services.prefs.lockPref("browser.gesture.pinch.out.shift");
     }
     if (cmdLine.handleFlag("browser", false)) {
-      openBrowserWindow(cmdLine, lazy.gSystemPrincipal);
+      if (isFeltUI) {
+        queueFeltURL({
+          url: "",
```
lissyx: It's not top obvious here why we pass an empty URL. Is this because we want to just open a new window ?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -15,28 +15,57 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
+    gFeltPendingURLs.push(payload);
   }
 }
 
 let gFeltProcessParentInstance = null;
 
+function extractURLPayload(payload) {
+  return {
+    url: payload.url ?? "",
+    disposition: payload.disposition ?? 0,
```
lissyx: nit: `0` ? Should we use `FELT_OPEN_WINDOW_DISPOSITION.DEFAULT` or you wanted to avoid the extra inclusion?

testing/enterprise/test_felt_browser_new_window_from_cli.py
```diff
@@ -0,0 +1,98 @@
...
+              results.push({
+                url: win.gBrowser?.currentURI?.spec ?? "",
+                isPrivate: PrivateBrowsingUtils.isWindowPrivate(win),
+              });
+            }
+            return results;
+            """
+        )
+        self._child_driver.set_context("content")
+        return windows
```
lissyx, line 37: Isn't the window exposed by Selenium and Marionette? Why do it like that?

lissyx: They also offer way to get the URL

lissyx (approved):
LGTM but I'd like to make sure we have re-implemented get windows for a good reason (and that it is made obvious in the code why)

browser/base/content/nonbrowser-mac.js
gcp: It's about disabling the items in the dock menu, so I don't think it will help simplifying the "how owns the dock" problem.

testing/enterprise/test_felt_browser_new_window_from_cli.py
gcp: They, however, do not offer a way to get private window status AFAIK.

browser/components/BrowserContentHandler.sys.mjs
gcp: Yes, but you're right it doesn't have to go in struct as we can just default it.

browser/base/content/nonbrowser-mac.js
```diff
@@ -108,11 +125,47 @@ var NonBrowserWindow = {
...
+    const { isFeltFirefoxWindowReady, waitForFeltFirefoxWindowReady } =
+      ChromeUtils.importESModule("resource:///modules/FeltURLHandler.sys.mjs");
+
+    // Check if Firefox is already ready (e.g., after restart)
+    if (isFeltFirefoxWindowReady()) {
+      return;
+    }
+
+    // Disable dock menu items until Firefox is ready
+    if (newWindowItem && !newWindowItem.hidden) {
```
Mossop: I'm curious why bother with the hidden check?

browser/base/content/nonbrowser-mac.js
```diff
@@ -108,11 +125,47 @@ var NonBrowserWindow = {
...
     }
 
     this.delayedStartupTimeoutId = setTimeout(() => this.delayedStartup(), 0);
   },
 
+  setupFeltDockMenuState() {
+    let newWindowItem = document.getElementById("macDockMenuNewWindow");
+    let privateWindowItem = document.getElementById(
+      "macDockMenuNewPrivateWindow"
+    );
```
Mossop: Minor nit: These aren't needed until after the `isFeltFirefoxWindowReady` check fails

browser/components/FeltURLHandler.sys.mjs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    let alert = Cc["@mozilla.org/alert-notification;1"].createInstance(
+      Ci.nsIAlertNotification
+    );
+    alert.init(
+      "felt-pending-action",
+      "chrome://branding/content/icon64.png",
+      "Firefox",
+      "Please wait while Firefox starts...",
```
Mossop: This needs to be localized

lissyx: For L10N it looks like it will simplify handling if we add any new strings to files scopes to enterprise, and not existing firefox ones

gcp:
>LGTM but I'd like to make sure we have re-implemented get windows for a good reason (and that it is made obvious in the code why)

Added comments to explain (it's to get private window status).

browser/locales/en-US/browser/enterprise/felt.ftl
```diff
@@ -0,0 +1,5 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+felt-pending-action-notification = Please wait while { -brand-short-name } starts…
```
lissyx, line 5: Please move to `browser/locales/en-US/browser/enterprise/felt.ftl` as it will be there in #398

lissyx (changes requested):
Move FTL file please

browser/components/BrowserContentHandler.sys.mjs
```diff
@@ -448,6 +448,9 @@ nsBrowserContentHandler.prototype = {
 
   /* nsICommandLineHandler */
   handle: function bch_handle(cmdLine) {
+    const isFeltUI = AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI();
+    // XXX: The following flags are not forwarded to Firefox in Felt mode:
```
lissyx, line 452: nit: do we want a follow up?

lissyx (approved):

browser/components/BrowserContentHandler.sys.mjs
gcp: https://bugzilla.mozilla.org/show_bug.cgi?id=2014879

PR #313, opened 2025-12-30
Introduce policy Sync to customize sync state

1rneh:
Exemplary policy setting:
```
{
  "policies": {
    "SyncSettings": {
      "SyncEnabled": true,
      "TypesEnabled": [
        "bookmarks",
        "passwords"
      ]
    }
  }
}
```


toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -204,6 +204,13 @@ EnterprisePoliciesManager.prototype = {
     const policyNames = Object.keys(unparsedPolicies || {});
 
     for (let policyName of policyNames) {
+
+#ifdef MOZ_ENTERPRISE
+      if (policyName in ["DisableAccounts", "DisableFirefoxAccounts"]) {
```
gcp:
```
policyName = "DisableAccounts";
console.log(policyName in ["DisableAccounts", "DisableFirefoxAccounts"])
```

Return `false`.

browser/locales/en-US/browser/policies/policies-descriptions.ftl
```diff
@@ -249,6 +249,8 @@ policy-StartDownloadsInTempDirectory = Force downloads to start off in a local,
 
 policy-SupportMenu = Add a custom support menu item to the help menu.
 
+policy-SyncSettings = Enable or disabled Sync and additionally define which data to sync.
```
gcp: typo: disable

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -1693,6 +1693,24 @@
       "required": ["Title", "URL"]
     },
 
+#ifdef MOZ_ENTERPRISE
+    "SyncSettings": {
+      "type": "object",
+      "properties": {
+        "SyncEnabled": {
```
gcp: Make this required?

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -1693,6 +1693,24 @@
       "required": ["Title", "URL"]
     },
 
+#ifdef MOZ_ENTERPRISE
```
gcp: Huh, does this work without marking this somewhere in a `moz.build`?

browser/components/enterprisepolicies/helpers/SyncSettingsPolicy.sys.mjs
```diff
@@ -0,0 +1,148 @@
...
+    if (!param.SyncEnabled) {
+      lazy.log.debug("Force-disable Sync");
+      if (isSyncEnabled) {
+        await this.disconnectSync(manager);
+      }
+      return;
+    }
+
+    lazy.log.debug("Force-enable Sync");
+    await this.connectSync(manager);
```
gcp: Move this after the prefs setting?

gcp (changes requested):
See comment.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Right! Just writing it down here again, because I thought it was an interesting one to think about.
The "in" operator checks for the presence of a specific property in an object. An Array such as `["DisableAccounts", "DisableFirefoxAccounts")` is also just an "indexed object" and therefore it compiles but doesn't check for what we wanted. `["DisableAccounts", "DisableFirefoxAccounts"].includes(policyName)` will do the job.

browser/components/enterprisepolicies/schemas/policies-schema.json
1rneh: Nope, you're right. I added [Bug 2012084](https://bugzilla.mozilla.org/show_bug.cgi?id=2012084) to put enterprise-only policies behind a build flag.

1rneh:
Updated policy sample for SyncSettings:
```
{
  "policies": {
    "SyncSettings": {
      "State": "enabled",
      "TypesEnabled": [
        "bookmarks",
        "passwords"
      ]
    }
  }
}
```

browser/locales/en-US/browser/policies/policies-descriptions.ftl
```diff
@@ -251,6 +251,8 @@ policy-StartDownloadsInTempDirectory = Force downloads to start off in a local,
 
 policy-SupportMenu = Add a custom support menu item to the help menu.
 
+policy-SyncSettings = Enable or disable Sync and additionally define which data to sync.
```
flodolo:
```suggestion
policy-SyncSettings = Enable or disable sync and define which data to include.
```
See https://acorn.firefox.com/latest/content/word-list/s-t-D8DB7Wjd#section-sync-16

browser/components/enterprisepolicies/moz.build
```diff
@@ -16,7 +16,7 @@ DIRS += [
 
 TEST_DIRS += ["tests"]
 
-EXTRA_JS_MODULES.policies += [
+EXTRA_PP_JS_MODULES.policies += [
```
Mossop: We generally want to avoid preprocessing JavaScript files when possible. Can you use `AppConstants.MOZ_ENTERPRISE` instead of preprocessing instructions?

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -1697,6 +1697,33 @@
...
+              "passwords",
+              "paymentMethods",
+              "addons",
+              "settings"
+            ]
+          }
+        },
+        "required": ["State"]
+      }
+    },
```
Mossop: It looks like policies-schema.json is actually preprocessed, can we use an ifdef and include an `enterprise-policies-schema.json`?

services/sync/moz.build
```diff
@@ -39,6 +38,10 @@ EXTRA_JS_MODULES["services-sync"] += [
     "Weave.sys.mjs",
 ]
 
+EXTRA_PP_JS_MODULES["services-sync"] += [
+    "modules/service.sys.mjs",
+]
+
```
Mossop: Again we want to avoid preprocessing JS files.

toolkit/components/enterprisepolicies/moz.build
```diff
@@ -19,6 +19,9 @@ if CONFIG["MOZ_WIDGET_TOOLKIT"] != "android":
     EXTRA_JS_MODULES += [
         "EnterprisePolicies.sys.mjs",
         "EnterprisePoliciesContent.sys.mjs",
+    ]
+
+    EXTRA_PP_JS_MODULES += [
```
Mossop: Same again

Mossop (changes requested):

browser/components/enterprisepolicies/moz.build
1rneh: A couple of weeks ago we said we'd prefer to preprocess code at build time instead of including code at runtime. I guess this has changed?

Mossop: I'm not sure who said this, but the policy for Firefox code is that we don't preprocess JS unless there is no way to avoid it. So for our code to be eventually landed in firefox-main it can't be preprocessed. Easier to not do it now than to try to undo it later I think.

1rneh: Agreed.

browser/components/enterprisepolicies/schemas/policies-schema.json
1rneh: Should you point me to where it gets pre-processed?

I added a follow-up [Bug 2012084](https://bugzilla.mozilla.org/show_bug.cgi?id=2012084) to generate an enterprise policies schema. Since we not only need to put enterprise-only policies behind a build flag, but also exclude policies that are not applicable in enterprise builds.

Mossop: `policies-schema.json` is only included in the build via the preprocessor: https://searchfox.org/firefox-main/rev/81cd862538be1cf95842a2ddfb7a5dd22c11ab25/browser/components/enterprisepolicies/schemas/schema.sys.mjs#6

Mossop: Though I guess maybe we need to check with @mkaply whether this file is used elsewhere that wouldn't like preprocessing

browser/locales/en-US/browser/policies/policies-descriptions.ftl
```diff
@@ -251,6 +251,8 @@ policy-StartDownloadsInTempDirectory = Force downloads to start off in a local,
 
 policy-SupportMenu = Add a custom support menu item to the help menu.
 
+policy-SyncSettings = Enable or disable sync and define which data to include.
```
lissyx: @1rneh to make sure we dont land a string in an existing file subtlely until #398 lands, do you mind moving this string to (newly created) `browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl` ? This way, if you land first, I rebase and it's in the new file expected by L10n repo, and if I land first it's here so you just add new string to it.

Having it land on `enterprise-main` in the wrong file may prompt issues on l10n repo

mkaply:
It's funny you should pick this one, as folks have requested having this in the core Firefox as a policy, so I'd like to make sure we build it in a way where it can live in both and I would rather just do a patch and implement in nightly.

https://github.com/mozilla/policy-templates/issues/621

We already have the very poorly named DisableFirefoxAccounts policy (https://firefox-admin-docs.mozilla.org/reference/policies/disablefirefoxaccounts/) which controls whether or not sync is enabled.

I like the idea of moving this all under one Sync policy, but the style here is inconsistent with other policies. When I'm designing I tend to look at Chrome and Edge to see what they do. Here's Chrome's policy (they don't do nested JSON like we do)

https://chromeenterprise.google/policies/#SyncTypesListDisabled

Although I'm not a fan of theirs (it requires manually typing the list)

So a couple thoughts.

1. We shouldn't have a default concept in general because default is what happens if there is no policy; it should be enabled or disabled.
2. Status should just be a Enabled as a boolean (true/false). We want to avoid enums because in the UI we end up having to have dropdowns for everything. It also makes the GPOs easiers
3. I think that the list of types should be a list of things that are booleans. Again, if you use an enum, then the UI ends up having to be a multi select list or a user having to type things.

1rneh:
> It's funny you should pick this one, as folks have requested having this in the core Firefox as a policy, so I'd like to make sure we build it in a way where it can live in both and I would rather just do a patch and implement in nightly.
> 
> [mozilla/policy-templates#621](https://github.com/mozilla/policy-templates/issues/621)
> 
> We already have the very poorly named DisableFirefoxAccounts policy (https://firefox-admin-docs.mozilla.org/reference/policies/disablefirefoxaccounts/) which controls whether or not sync is enabled.
> 
> I like the idea of moving this all under one Sync policy, but the style here is inconsistent with other policies. When I'm designing I tend to look at Chrome and Edge to see what they do. Here's Chrome's policy (they don't do nested JSON like we do)
> 
> https://chromeenterprise.google/policies/#SyncTypesListDisabled
> 
> Although I'm not a fan of theirs (it requires manually typing the list)
> 
> So a couple thoughts.
> 
>     1. We shouldn't have a default concept in general because default is what happens if there is no policy; it should be enabled or disabled.
> 
>     2. Status should just be a Enabled as a boolean (true/false). We want to avoid enums because in the UI we end up having to have dropdowns for everything. It also makes the GPOs easiers
> 
>     3. I think that the list of types should be a list of things that are booleans. Again, if you use an enum, then the UI ends up having to be a multi select list or a user having to type things.

Thanks @mkaply!

Chrome's policy seems to be limited to force disabling Sync types, while we were planning to force enable types as well.

Regarding your other thoughts:

1. Addressed in point 2.

2. My understanding is that we need at least three states. We want to force enable or disable Sync + what to (not) sync. In these two states the users can't modify any Sync settings anymore. Additionally we'd need a third state, a default state, that lifts that modification restriction and restores a default state. From what you suggest, if I set Enabled=false that could be two things: Force disable Sync or restore the default.

3. I'll change that.

mkaply:
The default state is just no policy though, isn't it?

This is something we've been struggling with all along. The concept of "Default" in my mind just means there is no policy applied at all. An admin shouldn't have to force a default.

For instance, here's how we describe the existing Firefox Accounts policy:

If this policy is enabled, Firefox Accounts is disabled which includes disabling Sync.

If this policy is disabled or not configured, Firefox Accounts and Sync are available.

Policies have three states - on, off or not configured and "not configured" should be the default case.

It also might be that we want to have case where an admin sets the default but the user can still change it. In that case, I typically add a new parameter "Locked" versus always disabling everything.

1rneh:
Ok, after skimming through the policies schema I assume this is what you have in mind?

```
    "Sync": {
      "type": "object",
      "properties": {
        "Enabled": {
          "type": "boolean"
        },
        "addresses": {
          "type": "boolean"
        },
        "bookmarks": {
          "type": "boolean"
        },
        "history": {
          "type": "boolean"
        },
        "openTabs": {
          "type": "boolean"
        },
        "passwords": {
          "type": "boolean"
        },
        "paymentMethods": {
          "type": "boolean"
        },
        "addons": {
          "type": "boolean"
        },
        "settings": {
          "type": "boolean"
        },
        "Locked": {
          "type": "boolean"
        },
        "required": ["Enabled", "Locked"]
      }
```

1rneh:
> This is something we've been struggling with all along. The concept of "Default" in my mind just means there is no policy applied at all. An admin shouldn't have to force a default.

I prefer the second schema. But no matter the schema for `Sync`, the admin would still need to force a default if we are trying to remove a local policy when fetching remote policies. 
Let's say the policy `Sync` is applied locally. We fetch remote policies and `Sync` is not in the fetched policy set. That wouldn't remove the policy because remote and local policies are merged. So it remains applied due to local set of policies. So the only way we can "remove" the policy is by force setting a default when fetching remote policies next time.

mkaply:
> Let's say the policy `Sync` is applied locally. We fetch remote policies and `Sync` is not in the fetched policy set. That wouldn't remove the policy because remote and local policies are merged. So it remains applied due to local set of policies. So the only way we can "remove" the policy is by force setting a default when fetching remote policies next time.

I thought we weren't having local policies anymore?

But won't we reread the local policies and remerge so it would go away?

I do like that version better, but neither Enabled or Locked should be required. Someone could just use that policy to set the default things that were synced without changing enabled or locked.

And locked is never required.

I'd also use camel case.

1rneh:
> I thought we weren't having local policies anymore?

I thought the stretch goal was to keep supporting both and have local policies be reflected in the console though. And discarding local policies is just a temporary solution until we have the former.

> But won't we reread the local policies and remerge so it would go away?

Yes, we would re-merge the policy and therefore it would not go away. If `Sync` is in the local policies but not in the fetched remote policies, the policy would get re-merged and remains applied.

mkaply:
> Yes, we would re-merge the policy and therefore it would not go away. If `Sync` is in the local policies but not in the fetched remote policies, the policy would get re-merged and remains applied.

Which would be the same if the policy were set to default?

Does default mean "Firefox default" or "use whatever policy is already applied"

1rneh:
> Which would be the same if the policy were set to default?

Depends on the setting of the local policy. E.g. if we have a local policy that enables Sync and locks it, then we'd fallback to that state and that's not "the default". We'd need to override it to whatever we think "the default" is.

> Does default mean "Firefox default" or "use whatever policy is already applied"

We should define what default means.

Example: 
Starting point: No policy applied. User is syncing passwords and bookmarks. We fetch remote policies and apply the `Sync` policy and override the types of data to sync. Then we remove the policy. Now what is "the default"? Is the default whatever the user had prior to applying the policy? What if it was not locked and the user had modified which data types to sync while the policy was applied. Do we need to persist what the settings were prior to any policy so that that information survives even if Firefox was shutdown in the meantime.

gcp:
>Someone could just use that policy to set the default things that were synced without changing enabled or locked.

Simple to understand use case here is:

```
    "Sync": {
        "paymentMethods": false
      }
```

If I don't want anyone in my company to store credit card info on my server. They can still choose whether to enable or disable sync. So either a) `enable` is optional or b) is required but has a `don't change` setting (and I expressly don't call it `default` here).

mkaply:
I want to be clear that I don't know what "the right thing" is here. We have a lot of inconsistency in the existing policy and I have moments where I wish we could just start over. (I didn't do a lot of the original design).

But I think in the long run, if we can get away from always using prefs for policies and starting to explicitly call function in policies to check for things, we're better off (I'm checking with the sync team to see if I can figure out things).

1rneh:
**New schema, new try!** 

Follow ups:
- If we want the Sync policy to be a regular m-c policy, we'll need to follow up on this. It's out of scope for Alpha.
- I filed [Bug 2012084](https://bugzilla.mozilla.org/show_bug.cgi?id=2012084) to generate an enterprise policies schema. Since we not only need to put enterprise-only policies behind a build flag, but also exclude policies that are not applicable in enterprise builds.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -0,0 +1,151 @@
...
+   */
+  async restoreSettings(manager) {
+    if (!Services.policies.isAllowed(SYNC_FEATURE)) {
+      manager.allowFeature(SYNC_FEATURE);
+    }
+    for (const pref of Object.values(ENGINE_PREFS)) {
+      lazy.log.debug(`Unsetting ${pref}`);
+      unsetAndUnlockPref(pref);
+    }
+    // TODO: Restoring to sync being enabled or disabled?
```
1rneh: What sync state do we fall back to when removing the policy ?

gcp: From discussion: we currently don't have anything that stores the pref, but from trying it, we enable sync for everything as soon as you make an account. So recommendation is to fall back to enabled and then add an explicit pref later on in a follow-up bug.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -0,0 +1,151 @@
...
+    } = params;
+
+    const isSyncEnabled = this.isSyncEnabled();
+
+    if (isEnableSync === true) {
+      lazy.log.debug("Enable Sync");
+      if (!isSyncEnabled) {
+        await this.connectSync(manager);
+      }
+    } else if (isEnableSync === false) {
```
lissyx: nit: do we care about something else than true/false?

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -0,0 +1,151 @@
...
+import {
+  PREF_LOGLEVEL,
+  setAndLockPref,
+  unsetAndUnlockPref,
+  PoliciesUtils,
+} from "resource:///modules/policies/Policies.sys.mjs";
+
+const lazy = {};
+ChromeUtils.defineESModuleGetters(lazy, {
+  STATUS_OK: "resource://services-sync/constants.sys.mjs",
```
lissyx: nit: `SYNC_STATUS_OK` otherwise it reads weird

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
```diff
@@ -2,7 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+
```
lissyx: empty space, remove?

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
```diff
@@ -2,7 +2,9 @@
...
 policy-BlocklistDomainBrowsedTelemetry = Enable and configure security logging/telemetry when { -brand-short-name } blocks a visit to a blocklisted domain.
 policy-DownloadTelemetry = Enable and configure security logging/telemetry when a download is triggered.
 policy-EnterpriseStorageEncryption = Enable enterprise-managed primary password for encrypted storage.
 policy-PrintPageTelemetry = Enable and configure security logging/telemetry when a page is printed.
+policy-Sync = Enable or disable sync and define which data to include.
```
lissyx, line 9: Do we intend to land Sync Setting on m-c? If so, I'd like we make sure with l10n team how this string is going to be handled? Or we may have to remember/keep note here that when this happens we will have to change its IDs

services/sync/modules/service.sys.mjs
```diff
@@ -970,6 +970,10 @@ Sync11Service.prototype = {
 
   // configures/enabled/turns-on sync. There must be an FxA user signed in.
   async configure() {
+    if (AppConstants.MOZ_ENTERPRISE && !Services.policies.isAllowed("change-sync-state")) {
```
lissyx: nit: maybe this should be factored with the `startOver()` usage as well

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -204,6 +204,13 @@ EnterprisePoliciesManager.prototype = {
     const policyNames = Object.keys(unparsedPolicies || {});
 
     for (let policyName of policyNames) {
+
+      if (AppConstants.MOZ_ENTERPRISE) {
+        if (["DisableAccounts", "DisableFirefoxAccounts"].includes(policyName)) {
```
lissyx: Hard coding policies like that makes me uncomfortable. Let's make sure we file a followup on that and we define a better mechanism to handle that kind of usecase. Having had a look at update service, there may be other usecases where we want to not consider a policy for Enterprise

.prettierignore
```diff
@@ -1246,6 +1246,9 @@ toolkit/components/downloads/DownloadCore.sys.mjs
 toolkit/components/printing/content/print.js
 toolkit/mozapps/extensions/AddonManager.sys.mjs
 toolkit/mozapps/extensions/internal/XPIDatabase.sys.mjs
+toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
+browser/components/enterprisepolicies/Policies.sys.mjs
+services/sync/modules/service.sys.mjs
```
lissyx: Do we really need to ignore all?

lissyx (approved):

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -0,0 +1,151 @@
...
+      Locked: isLockSettings,
+      ...typeSettings
+    } = params;
+
+    const isSyncEnabled = this.isSyncEnabled();
+
+    if (isEnableSync === true) {
+      lazy.log.debug("Enable Sync");
+      if (!isSyncEnabled) {
+        await this.connectSync(manager);
```
gcp, line 96: This can throw from here: https://searchfox.org/firefox-main/rev/150ca809c7243bd6e994f28cc4a3750430783364/services/sync/modules/service.sys.mjs#977

Not sure if it can reasonably happen in our setup, but it would mean all the other policy/settings enforcement code below is skipped. We might want to eat the error and rethrow after we did that?

services/sync/modules/service.sys.mjs
```diff
@@ -991,6 +995,10 @@ Sync11Service.prototype = {
 
   // resets/turns-off sync.
   async startOver() {
```
gcp, line 1003: Looking at the callers, this seems to be more of an internal method that's called as part of a larger state change, so I think you want to do this block higher up to ensure consistency?

https://searchfox.org/firefox-main/rev/150ca809c7243bd6e994f28cc4a3750430783364/browser/base/content/browser-sync.js#2473 (okay, here it would only be false telemetry, so not that big of a deal)

https://searchfox.org/firefox-main/rev/150ca809c7243bd6e994f28cc4a3750430783364/services/sync/modules/SyncDisconnect.sys.mjs#135 (here you would still sign out, enable the service, ...so this usage worries me a bit)

https://searchfox.org/firefox-main/source/services/sync/modules/sync_auth.sys.mjs#213 (this seems harmless, i.e. dead code)

gcp (commented):
I'm not sure we should block `startOver()` but maybe rather modify its callers to no-op?

You could then maybe change this to assert that this is never called in the faulty state, that is, in Enterprise with changes blocked.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
1rneh: `undefined`, because the property `Enabled` might not be set.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: I already filed a follow up bug for this: https://bugzilla.mozilla.org/show_bug.cgi?id=2012084.

.prettierignore
1rneh: No, they were still left over from when we pre-proccesed at build time.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
1rneh: You mean something like this to make it clear when used in code?
`const SYNC_STATUS_OK = lazy.STATUS_OK`

lissyx: yes

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
lissyx: I'm curious how `undefined` we can get, since `return lazy.Weave.Status.checkSetup() == lazy.STATUS_OK;` we directly return this so it's going to be a boolean for sure?

1rneh: `isEnableSync` !== `isSyncEnabled`

Ehhh, now that I am seeing the two variables names next to each other, it's really hard to see the difference.
`isSyncEnabled` is the current sync state (always a boolean). `isEnableSync` is parsed from the policy parameters and might be `undefined`.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
1rneh: I filed [Bug 2017719](https://bugzilla.mozilla.org/show_bug.cgi?id=2017719) to add e.g. a default preference for the sync state.

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
1rneh: I think that is the plan, but not part of this PR's scope. I filed [Bug 2017722](https://bugzilla.mozilla.org/show_bug.cgi?id=2017722).

services/sync/modules/service.sys.mjs
1rneh: Not sure I understand, could you elaborate?

lissyx: nothing really big it's just the same code in both, maybe it should be factorized somewhere as a function and called from those places

services/sync/modules/service.sys.mjs
1rneh: We are actually guarding all entry points that trigger the callers:
- We hide `fxa-menu-account-signout-button`
- We hide `fxaUnlinkButton` in about:preferences#sync
- We hide the `Disconnect...` button in about:preferences#sync if sync state is enabled and locked by Sync policy
- We hide the `Turn on syncing...` button in about:preferences#sync if sync state is disabled locked by Sync policy

So if the sync state is locked, an update to the Sync policy is the only caller of `Weave.startOver()` and `Weave.configure()`.

1rneh: I added a comment in `startOver` and `configure` to inform that the callers are guarded.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
1rneh: I hope the naming is less confusing now!

services/sync/modules/service.sys.mjs
1rneh: Hm, not sure if it's worth refactoring for this. The log messages differ as well.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
1rneh: Would you suggest doing anything in `Policies.sys.mjs` with that error besides logging the failure? Logging we could also just do here. We could also send back an event inconsistency event to the console, but then Sync would be the only policy to inform the console about such a failure.

1rneh: As a first step I wrapped `connectSync` and `disconnectSync` in a try-catch block.

gcp (approved):

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -0,0 +1,166 @@
...
+import {
+  PREF_LOGLEVEL,
+  setAndLockPref,
+  unsetAndUnlockPref,
+  PoliciesUtils,
+} from "resource:///modules/policies/Policies.sys.mjs";
+
+const SYNC_STATUS_OK = ChromeUtils.importESModule(
+  "resource://services-sync/constants.sys.mjs"
+).STATUS_OK
```
Mossop:
```suggestion
const { STATUS_OK: SYNC_STATUS_OK } = ChromeUtils.importESModule(
  "resource://services-sync/constants.sys.mjs"
)
```

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3921,3 +3926,17 @@ function processMIMEInfo(mimeInfo, realMIMEInfo) {
...
+  });
+  Policies.Sync = {
+    async onBeforeAddons(manager, param) {
+      await lazy.SyncPolicy.applySettings(manager, param);
+    },
+    async onRemove(manager, _) {
+      await lazy.SyncPolicy.restoreSettings(manager);
+    }
+  }
+}
```
Mossop: An alternative here is to define this as normal and do the `MOZ_ENTERPRISE` check within the policy to make it non-functional in the non-enterprise case. That might make it easier to find this as it then appears in the normal alphabetical list of policy declarations. But I don't have a strong opinion, leave this as it is if you like

Mossop (approved):
From what I can tell this looks good but my knowledge of the sync service isn't great. Do we have some plans to add automated tests for this?

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
1rneh: Eslint won't let me: `Please use static import instead of ChromeUtils.importESModule`

So I'm going with a static import.

browser/components/enterprisepolicies/Policies.sys.mjs
1rneh: The stretch goal is land this in m-c ([Bug 2017722](https://bugzilla.mozilla.org/show_bug.cgi?id=2017722)), so this should be temporary anyway.

1rneh:
> From what I can tell this looks good but my knowledge of the sync service isn't great. Do we have some plans to add automated tests for this?

I filed [Bug 2017837](https://bugzilla.mozilla.org/show_bug.cgi?id=2017837) to add an end-to-end test for the SyncPolicy.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
1rneh: I filed [Bug 2017848](https://bugzilla.mozilla.org/show_bug.cgi?id=2017848) for the Eslint inconsistency.

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -1697,6 +1697,42 @@
       "required": ["Title", "URL"]
     },
 
+    "Sync": {
+      "type": "object",
+      "properties": {
+        "Enabled": {
+          "type": "boolean"
+        },
+        "addresses": {
```
mkaply: I know it's a nit, but could you make these start with uppercase so they match other policies?

mkaply: Or maybe I'm overthinking this. Never mind.

1rneh: Done :)

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3926,3 +3926,17 @@ function processMIMEInfo(mimeInfo, realMIMEInfo) {
...
+  });
+  Policies.Sync = {
+    async onBeforeAddons(manager, param) {
+      await lazy.SyncPolicy.applySettings(manager, param);
+    },
+    async onRemove(manager, _) {
+      await lazy.SyncPolicy.restoreSettings(manager);
+    },
+  };
+}
```
lissyx, line 3942: https://bugzilla.mozilla.org/show_bug.cgi?id=2027744

PR #318, opened 2026-01-07
Support Taskcluster actions in Github repositories

ahal:


ahal:
I'm not really sure how to test this because the action hooks contain a hash of the `.taskcluster.yml` in their names, so attempting to run an action from this PR results in "No such hook" errors (as expected).

But this shouldn't break anything so I propose we land this and then I'll do a follow-up PR to fix anything I missed.

lissyx (approved):

lissyx:
> I'm not really sure how to test this because the action hooks contain a hash of the `.taskcluster.yml` in their names, so attempting to run an action from this PR results in "No such hook" errors (as expected).
> 
> But this shouldn't break anything so I propose we land this and then I'll do a follow-up PR to fix anything I missed.

Yes, as long as we are not breaking GitHub Actions or treeherder badly :). But maybe fix the lint errors?

PR #321, opened 2026-01-08
Refactor updating enterprise policies

1rneh:
- Refactor updating remote policies
- Fix merging policies of a combined provider when remote policies are updated


toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -381,6 +482,8 @@ EnterprisePoliciesManager.prototype = {
 
     this._status = Ci.nsIEnterprisePolicies.UNINITIALIZED;
     this._parsedPolicies = undefined;
+    this._provider = null;
+    this._topicsObserved = null;
```
gcp: `new Set()`

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -401,24 +504,24 @@ EnterprisePoliciesManager.prototype = {
...
+  async observe(aSubject, aTopic) {
+    this._topicsObserved.add(aTopic);
 
-    switch (topic) {
+    switch (aTopic) {
       case "policies-startup":
         // Before the first set of policy callbacks runs, we must
         // initialize the service.
-        this._initialize();
+        await this._initialize();
```
gcp: Don't the callers need to know about this? They would assume things are already initialized when this returns, but it's not? So you'd get new race conditions?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -91,102 +92,142 @@ EnterprisePoliciesManager.prototype = {
...
+    for (const [policyName, policyParams] of Object.entries(
+      this._provider.policies || {}
+    )) {
+      const { isValid, parsedParams } = this._validatePolicyParams(
+        policyName,
+        policyParams
+      );
+
+      if (!isValid) {
+        continue;
```
gcp, line 247: This will cause the policy to be removed because it doesn't go in `parsedPolicies`, unclear if that is what you want.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -381,6 +482,8 @@ EnterprisePoliciesManager.prototype = {
 
     this._status = Ci.nsIEnterprisePolicies.UNINITIALIZED;
     this._parsedPolicies = undefined;
+    this._provider = null;
```
gcp, line 537: I think we can't actually do this unless we stop the `RemotePoliciesProvider` polling first, because that sends the `Update` message, and processing that assumes that `_provider` is non-null.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Good point! I didn't change that behavior. But let's re-think this :)
 
We have two options when encountering invalid parameters:
- Policy is currently active: We can remove the policy or keep it active with unchanged parameters. I think I prefer the latter. It's seems more intentional. In both ways we can warn in the logs about invalid parameters. 
- Policy is currently inactive: Do nothing.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Good catch! As temporary work around we are now spinning the event loop until the remote policies are fetched.

1rneh:
**Update:** This patch includes:
- Refactor updating remote policies
- Fix merging policies of a combined provider when remote policies are updated
- Test merging local and remote policies

Question: 
What should we do with the tests under `browser/components/enterprisepolicies`? Is it enough to run them with local policies?

toolkit/modules/Preferences.sys.mjs
```diff
@@ -246,7 +246,7 @@ Preferences.lock = function (prefName) {
  * Unlock a pref so it can be changed.
  *
  * @param   prefName  {String|Array}
- *          the pref to lock, or an array of prefs to lock
+ *          the pref to unlock, or an array of prefs to unlock
```
Mossop, line 249: We could just land this in firefox-main

.prettierignore
```diff
@@ -1240,6 +1240,7 @@ toolkit/modules/AppConstants.sys.mjs
 
 # Files with MOZ_ENTERPRISE preprocessor directives
 browser/components/enterprisepolicies/helpers/WebsiteFilter.sys.mjs
+browser/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```
Mossop: Why are we doing this?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -641,6 +766,41 @@ EnterprisePoliciesManager.prototype = {
...
+      .catch(e => {
+        error = e;
+      })
+      .then(r => {
+        result = r;
+        done = true;
+      });
+
+    Services.tm.spinEventLoopUntil(
+      "BrowserContentHandler.sys.mjs:BCH_spinResolve",
```
Mossop: You should use a more relevant name here

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -105,88 +120,135 @@ EnterprisePoliciesManager.prototype = {
...
 
-    let provider = this._chooseProvider(changesHandler);
-    if (provider.failed) {
+    const localProvider = this._chooseProvider();
+    if (this.isRemotePoliciesSupported()) {
+      const remoteProvider = RemotePoliciesProvider.getInstance();
+      try {
+        // Poll and ingest initial set of policies
+        const remotePoliciesPromise = remoteProvider.ingestPolicies();
+        this.spinResolve(remotePoliciesPromise);
```
Mossop: I would prefer to make `_initialize` async so it is clear that it does async work and then call spinResolve from the observer

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -105,88 +120,135 @@ EnterprisePoliciesManager.prototype = {
...
+      this._provider = localProvider;
+    }
+
+    if (this._provider.failed) {
       this._status = Ci.nsIEnterprisePolicies.FAILED;
+      return;
+    }
+
+    if (!this._provider.hasPolicies) {
+      this._status = Ci.nsIEnterprisePolicies.INACTIVE;
```
Mossop: Status is set to INACTIVE above so is this needed?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -105,88 +120,135 @@ EnterprisePoliciesManager.prototype = {
...
+    // from the registry.
+    const policies = this._provider.policies;
+    if (
+      Object.keys(policies).length === 1 &&
+      policies.Certificates &&
+      Object.keys(policies.Certificates).length === 1 &&
+      (policies.Certificates.ImportEnterpriseRoots === true ||
+        policies.Certificates.ImportEnterpriseRoots === 1)
+    ) {
+      this._status = Ci.nsIEnterprisePolicies.INACTIVE;
```
Mossop: Status is set to INACTIVE above so is this needed?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -105,88 +120,135 @@ EnterprisePoliciesManager.prototype = {
...
+  /**
+   * Parses, validates and applies any policy changes by comparing
+   * the previously parsed set of policies with the updated set
+   * from the remote provider.
+   *
+   * - Apply no changes to a policy if it remains unchanged.
+   * - Re-apply a policy if the parameters changed.
+   * - Remove a policy if it's missing in the updated set.
+   */
+  _updatePolicies() {
```
Mossop, line 268: This function probably needs to update `_status` to `ACTIVE` or `INACTIVE` depending on whether there are policies present or not

Mossop (changes requested):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -373,17 +495,26 @@ EnterprisePoliciesManager.prototype = {
...
+  async _resetEngine() {
     DisallowedFeatures = {};
 
     Services.ppmm.sharedData.delete("EnterprisePolicies:Status");
     Services.ppmm.sharedData.delete("EnterprisePolicies:DisallowedFeatures");
 
     this._status = Ci.nsIEnterprisePolicies.UNINITIALIZED;
-    this._parsedPolicies = undefined;
+    this._parsedPolicies = {};
+    if (this.isRemotePoliciesSupported) {
```
gcp: `this.isRemotePoliciesSupported()`

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
gcp, line 665: Given how error prone updating `_status` is, maybe we can add some sanity checks here for inconsistent status?

gcp (changes requested):

lissyx:
> * 
> 
> 
> Question: What should we do with the tests under `browser/components/enterprisepolicies`? Is it enough to run them with local policies?

(I have not had a look at the PR yet so this may be invalid) From past experience I ran several times in regressions I introduced with remote fetching that broke some policies. So I think we should try and run with both, but maybe not right now, and maybe we'd like to scope them in a flagged suite, maybe `enterprise` subsuite for mochtest browser and xpcshell ?

toolkit/modules/Preferences.sys.mjs
1rneh: Agreed! https://phabricator.services.mozilla.com/D281445.

.prettierignore
1rneh: This is was leftover when I still pre-processed something in `EnterprisePoliciesParent`, which has been removed.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Why is `_status` error prone?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: You're right, it's not needed. I removed it.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Same here.

toolkit/modules/Preferences.sys.mjs
1rneh: Fix landed in firefox/main. Just need rebasing.

1rneh:
> What should we do with the tests under browser/components/enterprisepolicies? Is it enough to run them with local policies?

From an offline conversation with @mkaply: 

Tests under `/browser/components/enterprisepolicies/` cover each policy logic. Tests under `/toolkit/components/enterprisepolicies/` test the policy engine. We should test whether each policy is applied correctly when provided **remotely after startup** (not during engine initialization).

1rneh:
This PR is growing and entails quite a lot of entangled changes and therefore difficult to split into multiple patches:
- [Bug 2015809 - Fix merging local and remote policies](https://bugzilla.mozilla.org/show_bug.cgi?id=2015809)
- Add preference to control policy provision for [Bug 2012440 - Repack disables local policies.json and GPO](https://bugzilla.mozilla.org/show_bug.cgi?id=2012440)
- [Bug 2015833 - Fetch remote policies during policy engine initialisation](https://bugzilla.mozilla.org/show_bug.cgi?id=2015833)
- Removing test suite `/browser/components/enterprisepolicies/**/live` because it's not testing the policies in a different way except that they are getting applied remotely. That irrelevant for the policy logic. We should aim for test that test the policies when applied **remotely and after startup**
- General refactoring

**Note:** 
Leave out the tests in your review for now. I just noticed that I have removed `Testing.servePolicyWithJson` but it's still used in the `/browser/components/enterprisepolicies/` tests. I will have to fix this first.

toolkit/components/enterprisepolicies/tests/browser/remote/browser_remote_policies_diffing.js
```diff
@@ -0,0 +1,209 @@
...
+
+const customSchema = {
+  properties: {
+    TestPolicy: {
+      type: "string",
+    },
+  },
+};
+
+let policyValue = POLICY_PARAM_STATE.DEFAULT;
```
Mossop, line 14: Throughout this test you verify the state of the active policies but you should sprinkle in checks that `policyValue` has the expected state after updates.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -1026,81 +1161,45 @@ class WindowsGPOPoliciesProvider {
...
-    for (let policyName of Object.keys(this._secondary.policies)) {
+    this._policies = structuredClone(this._primaryProvider.policies);
+    for (let [policyName, policyParams] of Object.entries(
+      this._secondaryProvider.policies || {}
+    )) {
       if (!(policyName in this._policies)) {
-        this._policies[policyName] = this._secondary.policies[policyName];
+        this._policies[policyName] = policyParams;
       }
     }
```
Mossop:
```suggestion
    // Combine policies with primaryProvider taking precedence.
    // We only do this for top level policies.
    return Object.assign({}, this._secondaryProvider.policies ?? {}, this._primaryProvider.policies);
```

Mossop: I think this is just a manual implementation of `Object.assign`

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -91,102 +95,185 @@ EnterprisePoliciesManager.prototype = {
...
-        );
+        return new CombinedProvider(platformProvider, jsonProvider);
       }
-      return new CombinedProvider(remoteProvider, platformProvider);
+      return platformProvider;
     }
-    if (jsonProvider.hasPolicies) {
-      return new CombinedProvider(remoteProvider, jsonProvider);
+    return jsonProvider;
+  },
```
Mossop, line 219: You don't have to do it here but I think things would make a lot more sense if instead of having `CombinedProvider` merge two policies it could instead merge any number of policies. Then instead of a binary tree of policy providers there would just be a single `CombinedProfiler` holding all the real providers and it would be able to merge them all in a single operation.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
Mossop: This function is only called during initialization though so the policy cannot already be active here. The question is what happens later when the policies are updated if this invalid policy isn't listed in `_parsedPolicies`. It looks to me like you don't want it in `_parsedPolicies` because that would then cause us to try to call an `onRemove` handler for the policy even though we haven't actually applied it.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -91,102 +95,185 @@ EnterprisePoliciesManager.prototype = {
...
+    for (const [policyName, policyParams] of Object.entries(
+      this._provider.policies || {}
+    )) {
+      const { isValid, parsedParams } = this._validatePolicyParams(
+        policyName,
+        policyParams
+      );
+
+      if (!isValid) {
+        console.warn(`Parameters for policy ${policyName} are invalid`);
```
Mossop: `_validatePolicyParams` already logs warnings in the event of invalid data so we don't need to do it here.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -435,26 +593,20 @@ EnterprisePoliciesManager.prototype = {
         this._runPoliciesCallbacks("onAllWindowsRestored");
         break;
 
+      case "EnterprisePolicies:Reset":
+        this._resetEngine().then(null, console.error);
```
Mossop:
```suggestion
        this._resetEngine().catch(console.error);
```

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -91,102 +95,185 @@ EnterprisePoliciesManager.prototype = {
...
     }
-    if (jsonProvider.hasPolicies) {
-      return new CombinedProvider(remoteProvider, jsonProvider);
+    return jsonProvider;
+  },
+
+  /**
+   * Update engine status after parsing a new set of policies
+   */
+  _updateStatus() {
```
Mossop, line 224: This should set the status to `FAILED` if `this._provider.failed`

Mossop (changes requested):

toolkit/components/enterprisepolicies/tests/EnterprisePolicyTesting.sys.mjs
```diff
@@ -208,7 +243,9 @@ export var PoliciesPrefTracker = {
       // If a pref was used through setDefaultPref instead
       // of setAndLockPref, it wasn't locked, but calling
       // unlockPref is harmless
-      Preferences.unlock(prefName);
+      if (Services.prefs.prefIsLocked(prefName)) {
```
mkaply, line 250: Why was this added? As said above, Preference.unlock is harmless?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -11,6 +11,10 @@ ChromeUtils.defineESModuleGetters(lazy, {
   isTesting: "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
   FeltCommon: "chrome://felt/content/FeltCommon.sys.mjs",
   FeltStorage: "resource:///modules/FeltStorage.sys.mjs",
+  PREF_REMOTE_POLICIES_ENABLED:
```
mkaply, line 14: What's the reason the same JS is loaded into two different identifiers?

toolkit/components/enterprisepolicies/tests/browser/remote/browser_remote_policies_missing_live.js
```diff
@@ -32,6 +32,7 @@ add_task(async function check_all_policies_are_live() {
     "DisableDefaultBrowserAgent",
     "DisableForgetButton",
     "DisableFormHistory",
+    "DisableRemoteImprovements",
```
mkaply, line 35: This is unrelated to this patch.

mkaply (approved):
I've asked some questions and made a few suggestions, but assuming this pasts the tests (and you've addressed Mossop's questions), this seems good.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Yes, it's also to avoid calling the `onRemove` callback when it's not applied. Additionally we probably want to avoid removing a valid policy if a remote policy update includes that policy with invalid parameters.

Instead we could add another property `valid : true/false` to each policy entry in `_parsedPolicies` and take that into account when activating/updating policies.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: We do that already in `_initialize` but I'll add since the remote provider could fail when fetching remote policies (no `policies` in response).

I noticed that I'm setting `this._policies = {}` when the remote provider fails, which I shouldn't because this cause us to remove all active remote policies.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
1rneh: I removed the unused import of `PREF_LOCAL_POLICIES_ENABLED`

toolkit/components/enterprisepolicies/tests/EnterprisePolicyTesting.sys.mjs
1rneh: In a previous version of `tests/remote/browser_remote_policies_unsetAndUnlockPref.js` the preference that was set via a test policy [was removed at the end of the test](https://github.com/mozilla/enterprise-firefox/pull/321/changes/8761391b06b00ea140c5da530f79568e17615943#diff-e07719108316eccda1cc8b0928dad1aca368aee726e520cada9c8f24cb8065b2L150) which caused `Preferences.unlock(prefName)` to throw. So I'm verifying the existence of the pref now before unlocking.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: I filed [Bug 2017105](https://bugzilla.mozilla.org/show_bug.cgi?id=2017105) to follow up on this.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
gcp: IIRC part of the review pointed out mistakes in updating `_status`. So the idea is to add some checks/assertions here to make catching any bugs easier, by checking if `_status` matches with the state of the rest of the object?

gcp: The trigger was this: https://github.com/mozilla/enterprise-firefox/pull/321#discussion_r2720634231

gcp: Also this: https://github.com/mozilla/enterprise-firefox/pull/321#discussion_r2798488538

1rneh:
Continued in #1069.

PR #334, opened 2026-01-12
Enterprise: Check -felt for profile selection only on startup

lissyx:
Doing it in SelectStartupProfile() was somehow hitting race condition leading to a segfault while running some tests, the stack including XPCOM version of it.

### DIFF
--- a/toolkit/xre/nsAppRunner.cpp
+++ b/toolkit/xre/nsAppRunner.cpp
@@ -46,6 +46,7 @@
 
 #if defined(MOZ_ENTERPRISE)
 #  include "mozilla/browser/extensions/felt/felt.h"
+#  include "SpecialSystemDirectory.h"
 #endif
 
 #include "nsAppRunner.h"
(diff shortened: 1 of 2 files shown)


toolkit/xre/nsAppRunner.cpp
```diff
@@ -3118,6 +3119,39 @@ static nsresult SelectProfile(nsToolkitProfileService* aProfileSvc,
...
+        // many (thousands) of existing directories, which is unlikely to
+        // happen.
+        MOZ_TRY(file->CreateUnique(nsIFile::DIRECTORY_TYPE, 0700));
+      }
+
+      nsCOMPtr<nsIFile> localDir = file;
+      file.forget(aRootDir);
+      localDir.forget(aLocalDir);
+      // Background tasks never use profiles known to the profile service.
+      aProfile = nullptr;
```
gcp: ``nsIToolkitProfile** aProfile`` so this probably doesn't do what you want. But how did we get here....given that this should have been copypasted?

toolkit/xre/nsAppRunner.cpp
gcp, line 3167: You're now bypassing this logic and the checks that follow.

gcp (changes requested):

toolkit/xre/nsAppRunner.cpp
lissyx: I may have mistakenly broke it

toolkit/xre/nsAppRunner.cpp
lissyx: We were never changing either `aDidCreate` nor `aIsResetting`, so I dont think we regress anything here

gcp (approved):

PR #335, opened 2026-01-12
Enterprise: Add secure boot status to device posture

lissyx:


lissyx:
(not yet ready but opened the PR to run tests)

lissyx:
Should be good to review now

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1421,6 +1421,59 @@ BOOL WINAPI IsUserCetAvailableInEnvironment(_In_ DWORD UserCetEnvironment);
 #  define USER_CET_ENVIRONMENT_WIN32_PROCESS 0x00000000
 #endif
 
+#if defined(MOZ_ENTERPRISE) && defined(XP_LINUX)
+bool GetSecureBootStatus_Linux() {
+  std::ifstream input(
+      "/sys/firmware/efi/efivars/"
+      "SecureBoot-8be4df61-93ca-11d2-aa0d-00e098032b8c",
+      std::ios::binary);
```
gcp, line 1429: note this

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1421,6 +1421,59 @@ BOOL WINAPI IsUserCetAvailableInEnvironment(_In_ DWORD UserCetEnvironment);
...
+  if (input.fail()) {
+    return false;
+  }
+
+  unsigned char bytes[5] = {0};
+  for (int i = 0; i < 5; ++i) {
+    unsigned char byte;
+    input >> byte;
```
gcp: https://en.cppreference.com/w/cpp/io/basic_istream/operator_gtgt.html this skips whitespace, even if the stream is set to binary

Why not simply
```  
  input.read(reinterpret_cast<char*>(bytes), sizeof(bytes));
  /* Check for read failure incl. short read */
  if (!input) {
    return false;
  }
```

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1421,6 +1421,59 @@ BOOL WINAPI IsUserCetAvailableInEnvironment(_In_ DWORD UserCetEnvironment);
...
+  }
+
+  nsAutoRegKey secureBootStateKey(secureBootStateHKey);
+
+  DWORD data, len;
+  len = sizeof(data);
+
+  if (RegQueryValueEx(secureBootStateHKey, L"UEFISecureBootEnabled", 0, 0,
+                      reinterpret_cast<LPBYTE>(&data), &len) == ERROR_SUCCESS) {
+    RegCloseKey(secureBootStateHKey);
```
gcp: Don't need this, you're using RAII.

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1421,6 +1421,59 @@ BOOL WINAPI IsUserCetAvailableInEnvironment(_In_ DWORD UserCetEnvironment);
...
+    return static_cast<bool>(data);
+  }
+
+  RegCloseKey(secureBootStateHKey);
```
gcp: Same.

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1421,6 +1421,59 @@ BOOL WINAPI IsUserCetAvailableInEnvironment(_In_ DWORD UserCetEnvironment);
...
+    bytes[i] = byte;
+  }
+
+  return bytes[4] == 0x01;
+}
+#endif
+
+#if defined(MOZ_ENTERPRISE) && defined(XP_WIN)
+bool GetSecureBootStatus_Windows() {
+  NS_WARNING(nsPrintfCString("%s", __PRETTY_FUNCTION__).get());
```
gcp: Blah

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -482,6 +482,7 @@ export const ConsoleClient = {
         ipv4: null,
         ipv6: null,
       },
+      secureBootEnabled: Services.sysinfo.get("secureBootEnabled"),
```
gcp: ```getPropertyAsBool```

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1421,6 +1421,59 @@ BOOL WINAPI IsUserCetAvailableInEnvironment(_In_ DWORD UserCetEnvironment);
...
+  HKEY secureBootStateHKey;
+  LONG status =
+      RegOpenKeyEx(HKEY_LOCAL_MACHINE,
```
gcp, line 1448: Is it possible we fail to get this? Maybe we should report that state back then? Or do we assume can't get = false?

gcp (changes requested):

xpcom/base/nsSystemInfo.cpp
lissyx: Are we? I was not aware

xpcom/base/nsSystemInfo.cpp
lissyx: So far I have assumed that failure mean false yes. I dont know what kind of failures we could hit outside of the key not being existent? And if it's not existent, it means it's not enabled.

gcp: I was thinking permission issues from some hardening attempt or whatever.

xpcom/base/nsSystemInfo.cpp
gcp: It's the `Auto` in `nsAutoRegKey` .

gcp (approved):

xpcom/base/nsSystemInfo.cpp
lissyx: Should we report a telemetry event to the console maybe? There's not so much on MSDN about that registry key

PR #337, opened 2026-01-13
Show detailed network errors in Felt login UI

gcp:
When connecting to the enterprise console fails (e.g., due to an expired certificate), the error was a generic "NetworkError when attempting to fetch resource" which made debugging difficult.

This change:
- Adds _xhrFetch() wrapper in ConsoleClient that uses XMLHttpRequest internally to access channel.status (a Gecko extension) on network errors
- Converts status codes to readable names (e.g., SEC_ERROR_EXPIRED_CERTIFICATE) using nsINSSErrorsService
- Displays the specific error in the Felt login UI
- Replaces all fetch() calls with _xhrFetch() for consistent error handling (we could argue about this one)

### DIFF
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -240,6 +240,98 @@ export const ConsoleClient = {
   },
 
   /**
+   * Gets the error name for a channel status code.
+   *
+   * @param {number} status - The channel status code
+   * @returns {string} Human-readable error name
+   */
+  _getErrorNameForStatus(status) {
+    try {
+      const nssErrorsService = Cc[
+        "@mozilla.org/nss_errors_service;1"
+      ].getService(Ci.nsINSSErrorsService);
+      return nssErrorsService.getErrorName(status);
+    } catch {
+      // Not an NSS error, check common network error codes
+      const networkErrors = {
+        [Cr.NS_ERROR_UNKNOWN_HOST]: "unknown-host",
+        [Cr.NS_ERROR_CONNECTION_REFUSED]: "connection-refused",
+        [Cr.NS_ERROR_NET_TIMEOUT]: "net-timeout",
+        [Cr.NS_ERROR_NET_RESET]: "net-reset",
+        [Cr.NS_ERROR_NET_INTERRUPT]: "net-interrupt",
+        [Cr.NS_ERROR_OFFLINE]: "offline",
+      };
+      return networkErrors[status] || `error-code-${status}`;
+    }
+  },
+
+  /**
+   * Fetch-like wrapper that exposes detailed network errors.
+   * Uses XMLHttpRequest internally to access channel.status on error,
+   * which native fetch() does not expose.
+   *
+   * Limitations compared to native fetch():
+   * - Response object only has: ok, status, json(), text()
+   * - Missing: statusText, headers, url, redirected, clone(),
+   *   arrayBuffer(), blob(), formData()
+   * - json()/text() can be called multiple times (no body consumption)
+   *
+   * @param {string} url - The URL to fetch
+   * @param {object} options - Fetch-like options
+   * @param {string} [options.method="GET"] - HTTP method
+   * @param {object} [options.headers={}] - Request headers
+   * @param {string|null} [options.body=null] - Request body
+   * @returns {Promise<{ok: boolean, status: number, json: Function, text: Function}>}
+   */
+  _xhrFetch(url, { method = "GET", headers = {}, body = null } = {}) {
+    return new Promise((resolve, reject) => {
+      const xhr = new XMLHttpRequest();
+      xhr.open(method, url, true);
+
+      // Handle both plain objects and Headers instances
+      const headerEntries = Headers.isInstance(headers)
+        ? headers.entries()
+        : Object.entries(headers);
+      for (const [key, value] of headerEntries) {
+        xhr.setRequestHeader(key, value);
+      }
+
+      xhr.onload = () => {
+        const response = {
+          ok: xhr.status >= 200 && xhr.status < 300,
+          status: xhr.status,
+          json() {
+            try {
+              return Promise.resolve(JSON.parse(xhr.responseText));
+            } catch (e) {
+              return Promise.reject(e);
+            }
+          },
+          text: () => Promise.resolve(xhr.responseText),
+        };
+        resolve(response);
+      };
+
+      xhr.onerror = () => {
+        const errorName = xhr.channel
+          ? this._getErrorNameForStatus(xhr.channel.status)
+          : "NetworkError";
+        reject(new TypeError(errorName));
+      };
+
+      xhr.ontimeout = () => {
+        reject(new TypeError("NS_ERROR_NET_TIMEOUT"));
+      };
+
+      xhr.onabort = () => {
+        reject(new TypeError("NS_BINDING_ABORTED"));
+      };
+
+      xhr.send(body);
+    });
+  },
+
+  /**
    * Collect the device posture data and send them to the console.
    *
    * @returns {Promise<{posture: string}>} Token reported by console.
@@ -248,7 +340,7 @@ export const ConsoleClient = {
     const devicePosture = this._collectDevicePosture();
     const url = this.constructURI(this._paths.DEVICE_POSTURE);
 
-    const res = await fetch(url, {
+    const res = await this._xhrFetch(url, {
       method: "POST",
       headers: {
         "Content-Type": "application/json",
@@ -305,12 +397,13 @@ export const ConsoleClient = {
     const headers = new Headers({});
     const accessToken = await this.getAccessToken();
     headers.set("Authorization", `Bearer ${accessToken}`);
+    headers.set("Accept", "application/json");
     if (jsonBody !== null) {
       headers.set("Content-Type", "application/json");
     }
 
     const url = this.constructURI(path);
-    const res = await fetch(url, {
+    const res = await this._xhrFetch(url, {
       method,
       headers,
       body: jsonBody === null ? undefined : JSON.stringify(jsonBody),
@@ -404,7 +497,7 @@ export const ConsoleClient = {
       let res;
       try {
         const url = this.constructURI(this._paths.TOKEN);
-        res = await fetch(url, {
+        res = await this._xhrFetch(url, {
           method: "POST",
           headers: {
             "Content-Type": "application/json",
@@ -417,7 +510,7 @@ export const ConsoleClient = {
         });
       } catch (cause) {
         throw new InvalidAuthError(
-          "Token refresh request failed",
+          `Token refresh request failed: ${cause.message}`,
           "TOKEN_REFRESH_FAILED",
           { cause }
         );
--- a/browser/extensions/felt/content/felt.xhtml
+++ b/browser/extensions/felt/content/felt.xhtml
@@ -82,11 +82,15 @@
           </form>
         </div>
         <div class="felt-browser-error is-hidden">
-          <span
+          <div
             class="felt-browser-error-multiple-crashes is-hidden"
             data-l10n-id="felt-browser-error-multiple-crashes"
           >
-          </span>
+          </div>
+          <div class="felt-browser-error-connection is-hidden">
+            <span data-l10n-id="felt-browser-error-connection"></span>
+            <span class="felt-browser-error-details"></span>
+          </div>
         </div>
         <span
           class="felt-powered-by text-deemphasized"
--- a/browser/extensions/felt/content/styles/felt.css
+++ b/browser/extensions/felt/content/styles/felt.css
@@ -168,7 +168,13 @@ body:has(#notification-popup-box[open]) .felt:not(#notification-popup-box .felt)
   /* stylelint-disable-next-line */
   left: 8px;
   font-size: medium;
-  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
-  color: var(--icon-color-critical);
+  color: var(--text-color-error);
   font-weight: var(--font-weight-bold);
+}
+
+.felt-browser-error-details {
+  display: block;
+  font-weight: normal;
+  font-size: small;
+  margin-top: var(--space-xsmall);
 }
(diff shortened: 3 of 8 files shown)


lissyx (approved):
LGTM

gcp:
Kind of want to add a test before landing.

gcp:
Also let's wait until https://github.com/mozilla/enterprise-firefox/pull/291 lands so we can use the divs that adds to the Felt window.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -239,6 +239,96 @@ export const ConsoleClient = {
...
+   * - json()/text() can be called multiple times (no body consumption)
+   *
+   * @param {string} url - The URL to fetch
+   * @param {object} options - Fetch-like options
+   * @param {string} [options.method="GET"] - HTTP method
+   * @param {object} [options.headers={}] - Request headers
+   * @param {string|null} [options.body=null] - Request body
+   * @returns {Promise<{ok: boolean, status: number, json: Function, text: Function}>}
+   */
+  _xhrFetch(url, { method = "GET", headers = {}, body = null } = {}) {
```
gcp, line 286: note: can't use async here because we already construct the Promise manually, it would get wrapped in another Promise. And we can't get rid of that promise because we need to resolve/reject in the callbacks from XMLHttpRequest.

From a callers POV it behaves the same as fetch(), minus the limitations in the comments.

testing/enterprise/test_felt_browser_console_error.py
```diff
@@ -0,0 +1,62 @@
...
+            """
+            arguments[0].value = arguments[1];
+            arguments[0].dispatchEvent(new Event('input', { bubbles: true }));
+            """,
+            email,
+            "random@mozilla.com",
+        )
+
+        btn = self.get_elem("#felt-form__sign-in-btn")
+        btn.click()
```
1rneh: Could we add a method in `FeltTests` e.g. `submit_email` that we can call in this test (`super.submit_email()`) and in `test_felt_00_chrome_on_email_submit` in `FeltTests` to avoid duplicating code.

testing/enterprise/test_felt_browser_console_error.py
```diff
@@ -0,0 +1,62 @@
...
+        assert message != f"Failed to connect to {console_addr}:", "No error details"
+
+        self._driver.set_context("content")
+        return True
+
+    def test_felt_0_load_sso(self, exp):
+        return True
+
+    def test_felt_1_perform_sso_auth(self, exp):
+        return True
```
1rneh: The do nothing and can be removed, right?

testing/enterprise/test_felt_browser_console_error.py
```diff
@@ -0,0 +1,62 @@
...
+class FeltConsoleError(FeltTests):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def teardown(self):
+        if not hasattr(self, "_child_driver"):
+            self._manually_closed_child = True
+        return super().teardown()
+
+    def test_felt_00_chrome_on_email_submit(self, exp):
```
1rneh: Could we make the method name describe the test here?

browser/extensions/felt/content/styles/felt.css
```diff
@@ -168,7 +168,6 @@ body:has(#notification-popup-box[open]) .felt:not(#notification-popup-box .felt)
   /* stylelint-disable-next-line */
   left: 8px;
   font-size: medium;
-  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
-  color: var(--icon-color-critical);
+  color: var(--text-color-error);
```
1rneh, line 171: Thanks!

browser/extensions/felt/content/window.js
```diff
@@ -20,8 +20,45 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
+  }
+  if (errorContainer) {
+    const crashError = errorContainer.querySelector(
+      ".felt-browser-error-multiple-crashes:not(.is-hidden)"
+    );
+    if (!crashError) {
+      errorContainer.classList.add("is-hidden");
+    }
+  }
+}
```
1rneh: I guess this method generally resets the visible error state. Probably there will be more errors to be added to the list so let's maybe refactor the method to handel that. 
```suggestion
function resetErrorReport() {
  const errorWrapper = document.querySelector(".felt-browser-error");
  if (!errorWrapper.classList.contains("is-hidden")) {
    // No visible error report.
    return;
  }
  errorWrapper.classList.add("is-hidden");
  const errors = errorWrapper.querySelectorAll(".felt-browser-error span:not(.is-hidden)");
  errors.forEach(e => e.classList.add("is-hidden"));
}
```

1rneh: On another thought, I think we could make the error reporting a bit more generic. I think we need the three methods: 
- `ErrorReport.init()` which we call once the document is loaded 
- `ErrorReport.reset()` when we retry the login flow
- `ErrorReport.update(error)` if we encounter an error during the login process and want to add it to the error report.

And then we can have an enum will all errors we're facing to the user.

browser/extensions/felt/content/felt.xhtml
```diff
@@ -87,6 +87,7 @@
             data-l10n-id="felt-browser-error-multiple-crashes"
           >
           </span>
+          <span class="felt-browser-error-connection is-hidden"></span>
```
1rneh: Could we convert them to `div` elements so they are not all displayed in one line?

browser/extensions/felt/content/window.js
```diff
@@ -20,8 +20,45 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
-  const posture = await lazy.ConsoleClient.sendDevicePosture();
+  hideConnectionError();
+
+  let posture;
+  try {
+    posture = await lazy.ConsoleClient.sendDevicePosture();
+  } catch (err) {
+    console.error(`FeltExtension: Failed to connect to console: ${err}`);
+    const consoleUrl = lazy.ConsoleClient.consoleBaseURI.origin;
+    showConnectionError(`Failed to connect to ${consoleUrl}: ${err.message}`);
```
1rneh: We are localizing the crash error, so we should probably localize `Failed to connect to` as well. 

Tbh I'm not so happy about what we display though. We should have something actionable for the user such as `We couldn't connect to the console. Please contact your admin for support.` and then a maybe a details section with that error message. Let's make sure we define the UX for this soon.

1rneh (changes requested):
This is part 1 of my review. I had quite a lot of thoughts on our error report handling. 
I’ll share a follow-up review (part 2) covering the changes to `ConsoleClient` later today.

browser/extensions/felt/content/window.js
gcp: >And then we can have an enum will all errors we're facing to the user.

Not sure about this part. The code is a bit messy because it tries to surface up the actual error instead of a generic "couldn't connect to the console", but this means enumeration would be problematic.

1rneh: We could still display user-friendly error messages:
```
const ERR_MESSAGES = {
    CONSOLE_CONNECT: "We couldn't connect to the console.", 
    MULTIPLE_CRASHES: "We couldn't launch Firefox Enterprise due to multiple crashes on startup."
}
```
And if there is more to show such as the actual error message, we can append that to the error message or in a details section. What do you think?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -248,14 +338,22 @@ export const ConsoleClient = {
...
-      method: "POST",
-      headers: {
-        "Content-Type": "application/json",
-        Accept: "application/json",
-      },
-      body: JSON.stringify(devicePosture),
-    });
+    let res;
+    try {
+      res = await this._xhrFetch(url, {
```
1rneh: Hm, calling fetch directly should have been replaced here when we introduced `_post` in a766fb8. Seems like I missed this one. Could you change that? It also wraps `_xhrFetch` :)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -248,14 +338,22 @@ export const ConsoleClient = {
...
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+          Accept: "application/json",
+        },
+        body: JSON.stringify(devicePosture),
+      });
```
1rneh: Hm, calling fetch directly should have been replaced here when we introduced `_post` in a766fb8. Seems like I missed this one. Could you change that? It also wraps `_xhrFetch` :)
```suggestion
      res = await this._post(url, JSON.stringify(devicePosture));
```

1rneh: And then in `_fetch` we also need to add `headers.set("Accept", "application/json");`.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -239,6 +239,96 @@ export const ConsoleClient = {
...
+          text: () => Promise.resolve(xhr.responseText),
+        };
+        resolve(response);
+      };
+
+      xhr.onerror = () => {
+        let errorName = "NetworkError";
+        if (xhr.channel) {
+          errorName = this._getErrorNameForStatus(xhr.channel.status);
+        }
```
1rneh: nit :) (we already do it so nicely with the headers above)
```suggestion
        const errorName = xhr.channel 
          ? this._getErrorNameForStatus(xhr.channel.status) 
          : "NetworkError"
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -248,14 +338,22 @@ export const ConsoleClient = {
...
+          Accept: "application/json",
+        },
+        body: JSON.stringify(devicePosture),
+      });
+    } catch (err) {
+      console.error(
+        `ConsoleClient: Failed to connect to ${url}: ${err.message}`
+      );
+      throw err;
+    }
```
1rneh: Logging and rethrowing the same error here seems redundant, as the caller already handles the logging. I’d suggest removing this catch unless we want to add semantic context to the error.

1rneh (commented):
Part 2:

Thank you for the detailed documentation :)

Just some small suggestion about the changes in ConsoleClient. The rest of it looks good.

gcp:
Screenshot 2026-01-22 at 12 19 50
Screenshot 2026-01-22 at 12 16 41

browser/extensions/felt/content/window.js
gcp: I provide a translatable string for common errors now, or the raw NSS error if it's more complicated.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -248,7 +340,7 @@ export const ConsoleClient = {
     const devicePosture = this._collectDevicePosture();
     const url = this.constructURI(this._paths.DEVICE_POSTURE);
 
-    const res = await fetch(url, {
+    const res = await this._xhrFetch(url, {
```
1rneh, line 343: I had left a comment about this [here](https://github.com/mozilla/enterprise-firefox/pull/337#discussion_r2698274051). Could we use `this._post` instead of directly calling `this._xhrFetch`?

1rneh (approved):
Just one remark, otherwise this looks good to me! Thank you!

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: No, because `this._post()` wants authentication (it does `this.getAccessToken()` internally) and the device posture is sent _before_ we authenticate. It's step 2 in the diagram below.

```
  ┌─────────────────────────────────────────────────────────────────────────────┐
  │                        FELT LOGIN FLOW                                      │
  └─────────────────────────────────────────────────────────────────────────────┘

  ┌──────────────┐                                          ┌──────────────────┐
  │  FELT UI     │                                          │  ENTERPRISE      │
  │  (window.js) │                                          │  CONSOLE         │
  └──────┬───────┘                                          └────────┬─────────┘
         │                                                           │
         │  1. User enters email, clicks "Sign In"                   │
         │                                                           │
         ▼                                                           │
  ┌──────────────────┐                                               │
  │ connectToConsole │                                               │
  │    (email)       │                                               │
  └──────┬───────────┘                                               │
         │                                                           │
         │  2. sendDevicePosture() [UNAUTHENTICATED]                 │
         │     POST /sso/device_posture                              │
         │     Body: { os, security, build, network, secureBootEnabled }
         │──────────────────────────────────────────────────────────►│
         │                                                           │
         │◄──────────────────────────────────────────────────────────│
         │     Response: { posture: "<devicePostureToken>" }         │
         │                                                           │
         │  3. Construct SSO URL with email + devicePostureToken     │
         │     /sso/login?target=browser&email=...&devicePostureToken=...
         │                                                           │
         ▼                                                           │
  ┌──────────────────┐                                               │
  │ Load SSO in      │                                               │
  │ <browser> elem   │───────────────────────────────────────────────►
  └──────┬───────────┘      4. User performs SSO auth                │
         │                     (IdP login, MFA, etc.)                │
         │                                                           │
         │◄──────────────────────────────────────────────────────────│
         │  5. Redirect to /sso/callback?code=...                    │
         │                                                           │
         ▼                                                           │
  ┌──────────────────┐                                               │
  │ FeltCallbackChild│  6. Extract tokens from callback              │
  │ (JSActor)        │     { access_token, refresh_token, expires_in }
  └──────┬───────────┘                                               │
         │                                                           │
         │  7. Services.felt.setTokens(...)                          │
         │     Store tokens in XPCOM                                 │
         │                                                           │
         ▼                                                           │
  ┌──────────────────────────────────────────────────────────────────┐
  │                    USER IS NOW AUTHENTICATED                     │
  │                    Browser window launches                       │
  └──────────────────────────────────────────────────────────────────┘
```

testing/enterprise/test_felt_browser_console_error.py
gcp: For the record, I didn't do this here because this test already existed and it would mean a ton of renaming throughout the testing code, which is out of scope here.

PR #343, opened 2026-01-14
Bug 2010300 - Enable searchfox job on enterprise-firefox.

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=2010300

If the automation is using the `.cron.yml` file in this repository, then I think modifying the file here will enable the searchfox jobs there.
Otherwise we would need to apply the change on f-m side.

### DIFF
--- a/.cron.yml
+++ b/.cron.yml
@@ -181,6 +181,7 @@ jobs:
           - mozilla-esr140
           - elm
           - cypress
+          - enterprise-firefox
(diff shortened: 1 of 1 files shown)


lissyx:
@ahal is `cron.yml` doable there?

ahal:
Yes, I think it just needs a quick tweak in `fxci-config`.

lissyx (approved):

PR #344, opened 2026-01-14
Enterprise repack default console

lissyx:


taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -818,27 +818,26 @@ def make_job_description(config, jobs):
...
+                        for repack_name in this_repack[enterprise_name]:
+                            for platform_name in this_repack[enterprise_name][
+                                repack_name
+                            ]["platforms"]:
+                                for repack_locale in this_repack[enterprise_name][
+                                    repack_name
+                                ]["locales"]:
+                                    if platform_name == build_platform:
+                                        repack_final_name = f"{enterprise_name}/{repack_name}/{repack_locale}"
+                                        if repack_final_name not in repacks:
```
ahal, line 839: Consider making `repacks` a set so you don't need to test for existence.

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -818,27 +818,26 @@ def make_job_description(config, jobs):
...
+                                            repacks += [repack_final_name]
```
ahal, line 840: Try:
```python
from itertools import product

repacks = set()
for enterprise_name, repack_configs in this_repack.items():
    for repack_name, config in repack_configs.items():
        repacks.update([
            f"{enterprise_name}/{repack_name}/{locale}"
            for platform, locale in product(config["platforms"],  config["locales"]):
            if platform == build_platform
        ])  
```

Haven't tested but should be equivalent.

lissyx: I tried quickly and could not fix it. We'll have to revisit this later anyway

PR #349, opened 2026-01-16
Bug 2010334 - Support cron tasks in enterprise-firefox

ahal:
Cron tasks are currently enabled (e.g: https://firefox-ci-tc.services.mozilla.com/tasks/SleLuE1XR4eya2hRxOKDyw/runs/0/logs/public/logs/live.log), but they would go down the "hg" branch of the `.taskcluster.yml` and then fail to clone due to the commit hashes / repo urls being wrong.

This patch ensures we go down the "git" branch of `.taskcluster.yml` and ensures we're setup to handle cron there.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -56,7 +56,7 @@ tasks:
     # takes place, and only the first task is included in the `in_tree_action` hook.
     - $switch:
           # Decision task for events originating from hg.mozilla.org
-          'tasks_for in ["hg-push", "cron"] || (tasks_for == "action" && parameters["repository_type"] == "hg")':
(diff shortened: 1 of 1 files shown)


lissyx (approved):
LGTM, how can we test that?

ahal:
I can test it against `github.com/mozilla-releng/staging-firefox`, but it'll take a patch to fxci-config first. I don't mind though, probably best to set it up regardless.

Fwiw given the Decision task passed on this PR, I'm not too worried about it breaking on the push events either.

ahal:
I partially tested against staging-firefox, but getting the searchfox cron to run there is a bit annoying and  I don't think worth it.

PR #350, opened 2026-01-16
Bug 2006563, Bug 2006564 - Small improvements to FELT login UI email field

jonathanmendez:
- Since access keys are not available in the UI, remove the 'E' access key from "Work email" to avoid confusing underlining of the 'e'.
- Focus the work email field on launch since the only expected interaction is to fill in the field.


1rneh (approved):
Thank you!

PR #351, opened 2026-01-16
Bug 2046414 - Add Windows Aarch64

lissyx:
Bug: Bug-2046414

### DIFF
--- a/python/mozbuild/mozbuild/mach_commands.py
+++ b/python/mozbuild/mozbuild/mach_commands.py
@@ -3237,7 +3237,17 @@ def repackage_msi(
 @CommandArgument(
     "--channel",
     type=str,
-    choices=["official", "beta", "esr", "aurora", "nightly", "unofficial"],
+    choices=[
+        "official",
+        "beta",
+        "esr",
+        "aurora",
+        "nightly",
+        "unofficial",
+        "nightly-enterprise",
+        "beta-enterprise",
+        "release-enterprise",
+    ],
     help="Release channel.",
 )
 @CommandArgument(
@@ -3342,6 +3352,9 @@ def repackage_msix(
         "beta": "beta",
         "aurora": "aurora",
         "nightly": "nightly",
+        "nightly-enterprise": "nightly-enterprise",
+        "beta-enterprise": "beta-enterprise",
+        "release-enterprise": "release-enterprise",
     }
 
     if not input:
--- a/taskcluster/gecko_taskgraph/transforms/repackage.py
+++ b/taskcluster/gecko_taskgraph/transforms/repackage.py
@@ -449,6 +449,7 @@ def handle_keyed_by(config, jobs):
     fields = [
         "mozharness.config",
         "msi.display-name",
+        "msix.identity-name",
         "package-formats",
         "worker.max-run-time",
         "flatpak.name",
@@ -572,7 +573,10 @@ def make_job_description(config, jobs):
             elif "macosx64" in build_platform:
                 dep_th_platform = "osx-cross-enterprise/opt"
             elif "win64" in build_platform:
-                dep_th_platform = "windows2012-64-enterprise/opt"
+                if "aarch64" in build_platform:
+                    dep_th_platform = "windows2012-aarch64-enterprise/opt"
+                else:
+                    dep_th_platform = "windows2012-64-enterprise/opt"
             else:
                 raise ValueError(f"Unsupported {build_platform}")
 
@@ -580,7 +584,7 @@ def make_job_description(config, jobs):
         treeherder.setdefault("tier", 1)
         treeherder.setdefault("kind", "build")
 
-        # Search dependencies before adding langpack dependencies.
+        # Search dependencies before adding langpack dependencies
         signing_task = None
         repackage_signing_task = None
         for dependency in dependencies.keys():
@@ -609,10 +613,22 @@ def make_job_description(config, jobs):
         elif config.kind == "repackage-msix":
             assert not locale
 
-            # Like "MSIXs(Bs)".
-            treeherder["symbol"] = "MSIX({})".format(
-                dep_job.task.get("extra", {}).get("treeherder", {}).get("symbol", "B")
-            )
+            if identity_name := job.get("msix", {}).get("identity-name"):
+                attributes["msix_identity_name"] = identity_name
+
+            if "enterprise-repack" in dep_job.label:
+                # Similar to DEB, populate later
+                # TODO: I cannot remember when ...
+                treeherder["symbol"] = "MSIX-Ent({repack_id})"
+            else:
+                # Like "MSIXs(Bs)".
+                msix_symbol = (
+                    dep_job.task
+                    .get("extra", {})
+                    .get("treeherder", {})
+                    .get("symbol", "B")
+                )
+                treeherder["symbol"] = f"MSIX({msix_symbol})"
 
         elif config.kind == "repackage-shippable-l10n-msix":
             assert not locale
@@ -925,6 +941,12 @@ def make_job_description(config, jobs):
                 locale=locale,
                 existing_fetch=task["fetches"],
                 enterprise_repack=repack,
+                msi_archive_format="target.installer.exe"
+                if "msi_display_name" in repack_task["attributes"]
+                else None,
+                msix_archive_format="target.zip"
+                if "msix_identity_name" in repack_task["attributes"]
+                else None,
             )
 
             yield repack_task
@@ -939,6 +961,8 @@ def _generate_download_config(
     locale=None,
     existing_fetch=None,
     enterprise_repack=None,
+    msi_archive_format=None,
+    msix_archive_format=None,
 ):
     locale_path = f"{locale}/" if locale else ""
     fetch = {}
@@ -948,11 +972,7 @@ def _generate_download_config(
     if enterprise_repack:
         locale_path = f"{enterprise_repack}/"
 
-    if repackage_signing_task and build_platform.startswith("win"):
-        fetch.update({
-            repackage_signing_task: [f"{locale_path}target.installer.exe"],
-        })
-    elif build_platform.startswith("linux") or build_platform.startswith("macosx"):
+    if build_platform.startswith("linux") or build_platform.startswith("macosx"):
         signing_fetch = [
             {
                 "artifact": f"{locale_path}target{archive_format(build_platform)}",
@@ -966,18 +986,36 @@ def _generate_download_config(
             })
         fetch.update({signing_task: signing_fetch})
     elif build_platform.startswith("win"):
-        fetch.update({
-            signing_task: [
-                {
-                    "artifact": f"{locale_path}target.zip",
-                    "extract": False,
-                },
-                f"{locale_path}setup.exe",
-            ],
-        })
+        if repackage_signing_task:
+            if msi_archive_format:
+                fetch.update({
+                    repackage_signing_task: [f"{locale_path}{msi_archive_format}"],
+                })
+            else:
+                raise NotImplementedError("repackage_signing_task with non -msi- task")
+        if signing_task:
+            if msix_archive_format:
+                fetch.update({
+                    signing_task: [
+                        {
+                            "artifact": f"{locale_path}{msix_archive_format}",
+                            "extract": False,
+                        }
+                    ],
+                })
+            else:
+                fetch.update({
+                    signing_task: [
+                        {
+                            "artifact": f"{locale_path}target.zip",
+                            "extract": False,
+                        },
+                        f"{locale_path}setup.exe",
+                    ],
+                })
 
         use_stub = task.attributes.get("stub-installer")
-        if use_stub:
+        if use_stub and signing_task:
             fetch[signing_task].append(f"{locale_path}setup-stub.exe")
 
     if fetch:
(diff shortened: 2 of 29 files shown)


taskcluster/gecko_taskgraph/transforms/repackage_signing.py
```diff
@@ -81,6 +81,14 @@ def make_repackage_signing_description(config, jobs):
                 treeherder["symbol"] = "MSIs({})".format(locale or "N")
                 attributes["repackage_type"] = f"{config.kind}-{locale or 'N'}"
 
+        if config.kind == "repackage-signing-msix":
```
ahal: should be `elif`

taskcluster/gecko_taskgraph/transforms/repackage_signing.py
```diff
@@ -81,6 +81,14 @@ def make_repackage_signing_description(config, jobs):
...
+            if "enterprise-repack" in dep_job.label:
+                dep_symbol = dep_job.task.get("extra").get("treeherder").get("symbol")
+                treeherder["symbol"] = f"MSIXs-Ent({dep_symbol})"
+            else:
+                treeherder["symbol"] = "MSIXs({})".format(locale or "N")
+                attributes["repackage_type"] = f"{config.kind}-{locale or 'N'}"
+
         elif config.kind in (
             "repackage-signing-msix",
```
ahal, line 85: This is now dead code, should at least remove this label from the condition.. But why make a new arm to this if block rather than edit this existing one? You'll probably want the same fix to apply to `repackage-signing-shippable-l10n-msix` if we ever end up running those tasks on Enterprise right?

taskcluster/gecko_taskgraph/transforms/repackage_signing.py
```diff
@@ -81,6 +81,14 @@ def make_repackage_signing_description(config, jobs):
                 treeherder["symbol"] = "MSIs({})".format(locale or "N")
                 attributes["repackage_type"] = f"{config.kind}-{locale or 'N'}"
 
+        if config.kind == "repackage-signing-msix":
+            if "enterprise-repack" in dep_job.label:
+                dep_symbol = dep_job.task.get("extra").get("treeherder").get("symbol")
+                treeherder["symbol"] = f"MSIXs-Ent({dep_symbol})"
+            else:
+                treeherder["symbol"] = "MSIXs({})".format(locale or "N")
```
ahal: Use format string for consistency

taskcluster/gecko_taskgraph/transforms/repackage_signing.py
```diff
@@ -98,11 +106,14 @@ def make_repackage_signing_description(config, jobs):
...
                 .replace("/", "_")
             )
+
             job["label"] = job["label"].replace(
                 "repackage-signing", f"repackage-signing-enterprise-repack-{repack_id}"
             )
 
         label = job["label"]
+        if "None" in label:
+            assert False
```
ahal: Let's raise `ValueError` here instead. Assertions can be disabled at runtime

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -572,15 +572,18 @@ def make_job_description(config, jobs):
...
+                    dep_th_platform = "windows2012-64-enterprise/opt"
             else:
                 raise ValueError(f"Unsupported {build_platform}")
 
         treeherder.setdefault("platform", dep_th_platform)
         treeherder.setdefault("tier", 1)
         treeherder.setdefault("kind", "build")
 
-        # Search dependencies before adding langpack dependencies.
+        # Search dependencies before adding langpack dependencies CONTINUE.
```
ahal: mistake?

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -609,10 +612,18 @@ def make_job_description(config, jobs):
...
             assert not locale
 
-            # Like "MSIXs(Bs)".
-            treeherder["symbol"] = "MSIX({})".format(
-                dep_job.task.get("extra", {}).get("treeherder", {}).get("symbol", "B")
-            )
+            if "enterprise-repack" in dep_job.label:
+                # Similar to DEB, populate later
+                # TODO: I cannot remember when ...
+                treeherder["symbol"] = "MSIX-Ent({repack_id})"
```
ahal, line 622: Missing f-string

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -609,10 +612,18 @@ def make_job_description(config, jobs):
...
+            else:
+                # Like "MSIXs(Bs)".
+                treeherder["symbol"] = "MSIX({})".format(
+                    dep_job.task
+                    .get("extra", {})
+                    .get("treeherder", {})
+                    .get("symbol", "B")
+                )
```
ahal, line 630: Stick to f-strings

taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -485,6 +485,9 @@ def target_tasks_enterprise_firefox_with_tests(
     )
 
     def filter(task):
+        if "shippable" in task.label and "win64" in task.label:
+            return True
+
```
ahal: This seems like it's going to catch a lot of things we didn't intend. Can it be more specific? Maybe it can use `task.kind` instead of parsing labels.

taskcluster/gecko_taskgraph/util/verify.py
```diff
@@ -192,7 +192,7 @@ def verify_task_graph_no_shippable_enterprise_level_not_1(
 
     level = int(parameters["level"])
     if level < 3 and "enterprise" in task.label and "shippable" in task.label:
-        raise Exception(
+        print(
```
ahal: This should be reverted or removed

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -966,18 +975,36 @@ def _generate_download_config(
...
-        })
+        if repackage_signing_task:
+            if "-msi-" in repack_label:
+                fetch.update({
+                    repackage_signing_task: [f"{locale_path}target.installer.exe"],
+                })
+            else:
+                raise NotImplementedError("repackage_signing_task with non -msi- task")
+        if signing_task:
+            if "-msix-" in repack_label:
```
ahal: Is there an attribute we can use for this (and msi) rather than parsing labels? If not, can one be created?

taskcluster/kinds/build/windows.yml
```diff
@@ -589,6 +589,175 @@ win64-enterprise/opt:
...
+        symbol: Bpgo(Bent)
+        tier: 1
+    run-on-projects: ["enterprise-firefox"]
+    run-on-repo-type: ["git"]
+    worker-type: b-linux-docker-xlarge-amd
+    worker:
+        max-run-time: 10800
+        env:
+            PERFHERDER_EXTRA_OPTIONS: enterprise
+            # MOZ_AUTOMATION_PACKAGE_TESTS: "1"
```
ahal, line 669: How come these are commented out, do we not want to run tests?

taskcluster/kinds/repackage-msix/kind.yml
```diff
@@ -27,19 +25,24 @@ only-for-build-platforms:
...
+    - win64-aarch64-enterprise-shippable/opt
     - win32-devedition/opt
     - win64-devedition/opt
     - win64-aarch-devedition/opt
 
 tasks:
     repackage-msix:
+        # from-deps:
+        #     group-by: platform
+        #     unique-kinds: false
```
ahal, line 38: nit: don't forget to remove

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -939,6 +951,7 @@ def _generate_download_config(
     locale=None,
     existing_fetch=None,
     enterprise_repack=None,
+    repack_label=None,
```
ahal: Instead of passing in a label, only to then parse a package format out of it, let's pass in the package format directly

ahal (changes requested):

taskcluster/gecko_taskgraph/transforms/repackage.py
lissyx: It is not missing, that's the point of the comment: it will be f-formatted later (except I cannot remember when)

taskcluster/gecko_taskgraph/transforms/repackage.py
lissyx: OK I'm introducing two, I think it should be enough

taskcluster/gecko_taskgraph/transforms/repackage.py
lissyx: maybe `repack_task["attributes"]["repackage_type"`

lissyx: better, we have `display_name` on MSI, and we can copy `identity-name` on MSIX

taskcluster/gecko_taskgraph/transforms/repackage_signing.py
lissyx: I think it was just the result of quick copy/paste here

taskcluster/gecko_taskgraph/util/verify.py
lissyx: Yes, I have it on the PR to be able to iterate ; try was not working correctly for reasons ...

taskcluster/gecko_taskgraph/target_tasks.py
lissyx: No that is just a `DO NOT MERGE` to be able to iterate with the PR, since try was not happy

taskcluster/kinds/build/windows.yml
lissyx: Yes I disabled aarch64 win tests since I think they rely on scarce hardware, I did not want to put to much stress.

lissyx:
@bhearsum Can I get your opinion on the changes for the repackaging MSIX part: https://github.com/mozilla/enterprise-firefox/pull/351/changes/9f9b22318c729097b724c1e7de172c46f85bfbab

python/mozbuild/mozbuild/mach_commands.py
```diff
@@ -3237,7 +3237,15 @@ def repackage_msi(
...
     type=str,
-    choices=["official", "beta", "esr", "aurora", "nightly", "unofficial"],
+    choices=[
+        "official",
+        "beta",
+        "esr",
+        "aurora",
+        "nightly",
+        "unofficial",
+        "firefoxenterprise",
```
bhearsum: Is this really the channel for this product? (How are we handling beta/release/nightly distinctions if so?)

taskcluster/gecko_taskgraph/transforms/repackage_signing.py
```diff
@@ -98,11 +107,14 @@ def make_repackage_signing_description(config, jobs):
                 .get("symbol")
                 .replace("/", "_")
             )
+
             job["label"] = job["label"].replace(
                 "repackage-signing", f"repackage-signing-enterprise-repack-{repack_id}"
             )
 
         label = job["label"]
+        if "None" in label:
```
bhearsum: Is this really needed? I can't imagine the taskgraph would succeed overall if this is the case. Is this perhaps something temporary you added when debugging, before adjusting something further up?

bhearsum (approved):
If it works, I guess this is fine. The repackage transforms are a mess...but that's consistent with their existing state.

taskcluster/gecko_taskgraph/transforms/repackage_signing.py
lissyx: It is something that came up at some point and I wanted to make sure it surfaced as early as possible

python/mozbuild/mozbuild/mach_commands.py
lissyx: I added it to get things working, I am not sure how things are supposed to work there wrt how MSIX gets updated. It will not, as much as I know, be pushed to the store (are there local store existing?)

bhearsum: I strongly recommend you think a bit ahead on this - it's not easy (or sometimes possible) to change this later.

You may want, for example, to additional key this by release-type like we do for Firefox: https://searchfox.org/firefox-main/rev/db790003e5df7dcfced5bcf58823a594a1e8e2c2/taskcluster/kinds/repackage-msix/kind.yml#68-72

taskcluster/gecko_taskgraph/transforms/repackage_signing.py
bhearsum: I can't really stop you from merging this here...but I really don't see what value this is adding.

lissyx: I do not have a firm attachment to it, it was there as a precaution, but since I'm not seeing it  anymore ...

python/mozbuild/mozbuild/mach_commands.py
lissyx: I need to check if we have `release-type` properly set ...

lissyx: OK I'll just re-use the same as on `build` kind

ahal (approved):

PR #355, opened 2026-01-20
Enterprise fix marionette

lissyx:


browser/components/customizableui/test/marionette/test_assert_no_toolbar_changes.py
```diff
@@ -34,6 +34,10 @@ def check_toolbar_placements(self):
         )
         navbarPlacements = self.get_area_default_placements("AREA_NAVBAR")
         navbarPlacements.append("unified-extensions-button")
+
+        if self.marionette.session_capabilities.get("browserName") == "firefoxenterprise":
```
whimboo, line 38: For the `browserName` we still wanted to return `Firefox`. But for the enterprise flag there could be a `moz:browserFlavor` capability as returned by the new session command instead? We talked about this yesterday and you may not have applied it yet to this version of the PR. There are a lot of other parts in this PR affected as well.

testing/marionette/harness/marionette_harness/tests/unit/test_capabilities.py
```diff
@@ -74,7 +74,7 @@ def test_mandated_capabilities(self):
         self.assertEqual(self.caps["platformName"], self.os_name)
         self.assertEqual(self.caps["proxy"], {})
 
-        if self.appinfo["name"] == "Firefox":
+        if self.appinfo["name"] in ["Firefox", "FirefoxEnterprise"]:
```
whimboo, line 77: Isn't enterprise offering a kiosk-like mode not allowing the user to switch into other window states? We probably should not allow to use the various window manipulation commands.

whimboo (commented):
My feedback for the Remote Protocol implementation which might be backported to mozilla-central in the future. I miss the items we talked about yesterday.

whimboo:
Oh looks like I've taken a look at the wrong PR, which I still had open in a tab. You pushed a v2 as https://github.com/mozilla/enterprise-firefox/pull/363.

testing/marionette/harness/marionette_harness/tests/unit/test_capabilities.py
lissyx: We are using the same build for this launcher and the actual browser, I'll see if we could limit window manipulation commands on the launcher part without breaking our tests

PR #358, opened 2026-01-21
Enterprise repack new locales

lissyx:


testing/mozharness/configs/single_locale/enterprise.py
```diff
@@ -0,0 +1,10 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+config = {
+    "app_name": "browser",
+    "mozconfig_variant": "l10n-mozconfig-enterprise",
+    "locales_dir": "browser/locales",
+    "git_repository": "https://github.com/lissyx/firefox-l10n",
```
gcp: Put this under mozilla? Why is this in a personal branch?

taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -486,6 +486,9 @@ def target_tasks_enterprise_firefox_with_tests(
     )
 
     def filter(task):
+        if "shippable" in task.label and "enterprise" in task.label:
```
gcp: This blanket enable feels weird, especially given the level checks for shippable builds below.

taskcluster/gecko_taskgraph/transforms/l10n.py
```diff
@@ -161,6 +161,7 @@ def setup_name(config, jobs):
         # Set the name to the same as the dep task, without kind name.
         # Label will get set automatically with this kinds name.
         job["name"] = job.get("name", task_name(dep))
+        print("setup_name", job["name"])
```
gcp: ton of debug spam

gcp (commented):
This needs review from someone else, I don't feel I can meaningfully comment on this.

testing/mozharness/configs/single_locale/enterprise.py
lissyx: Sorry, it's WIP, it's not intended for review :(

browser/locales/en-US/browser/enterprise/felt.ftl
1rneh, line 1: I'm not sure anymore about the `felt-error-...` strings that were added to surface console connection errors.
I found that we already have localised error messages (https://searchfox.org/firefox-main/source/browser/locales/en-US/chrome/overrides/appstrings.properties#5) and a method that maps status to string id: https://searchfox.org/firefox-main/rev/ab8a90487516e59b3ac3b5cc503d024a2be8ea80/docshell/base/nsDocShell.cpp#3435-3753. We should probably re-use those.

So we could keep those as "previewed" fluent strings for now and not merge them to m-c. I filed bug for the follow-up work: [Bug 2014493](https://bugzilla.mozilla.org/show_bug.cgi?id=2014493).

Two more remarks:
- I believe `felt-browser-error-connection` is missing a dot at the end.
- Could we structure and describe the file a bit differently?

```suggestion
# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.

### FELT (Firefox Enterprise Launcher Tool) SSO Login

felt-window-title = { -brand-short-name } — Login
felt-sso-title = Sign in
felt-sso-input-email =
    .description = Use your organizational-issued email
    .label = Work email
felt-sso-continue-btn =
    .label = Continue
felt-powered-by =
    Powered by { -vendor-short-name }
# $version is the version of Felt, not the version of Firefox.
felt-version =
    Version { $version }

## Error details when launching the browser crashes

felt-browser-error-multiple-crashes = { -brand-short-name } crashed multiple times.
felt-browser-error-connection = Unable to connect to the console. Please contact your administrator.
```

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -19,3 +19,10 @@ enterprise-signout-prompt-title = Sign out of { -brand-short-name }?
 enterprise-signout-prompt-message = You’re signing out of your { -brand-short-name } browser. To use it again, you’ll need to re-authenticate through your company’s SSO provider.
 enterprise-signout-prompt-checkbox-label = Show this message when signing out.
 enterprise-signout-prompt-primary-btn-label = Sign out
+
+restart-forced-title = Restart Forced
+restart-forced-heading = Restart to Keep Using { -brand-short-name }
+restart-forced-intro = Policy requires that { -brand-short-name } needs to be restarted.
+window-restoration-info = Your windows and tabs will be quickly restored, but private ones will not.
+
```
1rneh, line 27:
```suggestion
```

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -32,3 +32,7 @@ extension-firefox-enterprise-light-description=A soft pastel theme with a touch
 
 extension-firefox-enterprise-dark-name=Firefox Enterprise Dark
 extension-firefox-enterprise-dark-description=A deep midnight theme with dark petrol blues and subtle, lighter blue gradients.
+
+blocked-by-policy-title = Access to this site is restricted
+neterror-blocked-by-policy-page-title2 = Access to this site is restricted
```
1rneh: Should these be unified to one?

1rneh (changes requested):

browser/locales/en-US/browser/enterprise/enterprise.ftl
1rneh: This refers to the one empty line.

browser/locales/en-US/browser/enterprise/felt.ftl
lissyx: As I said, you linked to `.properties` and JS uses Fluent here, it's not the same string.

browser/locales/en-US/browser/enterprise/enterprise.ftl
lissyx: I dont think so, one is the body of the page while the other is the title:
```
alex@portable-alex:~/codaz/Mozilla/gecko-cinnabar$ git grep -w blocked-by-policy-title
browser/locales/en-US/browser/enterprise/enterprise.ftl:blocked-by-policy-title = Access to this site is restricted
toolkit/content/aboutNetError.mjs:  "blocked-by-policy-title",
toolkit/content/aboutNetError.mjs:      bodyTitleId = "blocked-by-policy-title";
alex@portable-alex:~/codaz/Mozilla/gecko-cinnabar$ git grep -w neterror-blocked-by-policy-page-title2
browser/locales/en-US/browser/enterprise/enterprise.ftl:neterror-blocked-by-policy-page-title2 = Access to this site is restricted
toolkit/content/aboutNetError.mjs:      pageTitleId = "neterror-blocked-by-policy-page-title2";

File: browser/locales/en-US/browser/enterprise/felt.ftl
Comment (lissyx) [reply browser/locales/en-US/browser/enterprise/felt.ftl:1]: > So we could keep those as "previewed" fluent strings for now and not merge them to m-c. I filed bug for the follow-up work: [Bug 2014493](https://bugzilla.mozilla.org/show_bug.cgi?id=2014493).

We cannot do this, we need to land those strings properly for the rest of the l10n process to be finished, source strings extracted and be added to the l10n repo that will be exposed to pontoon, us to localize there, changes commited and finally our l10n changesets json file to be updated to pick those new locales.

Strings under `preview/` will get skipped.

Comment (1rneh) [reply browser/locales/en-US/browser/enterprise/felt.ftl:1]: We'd put them under `preview/` to have a place to put them now to not block merging the `.ftl` files to m-c. For the pilot we will need to migrate to the strings from `appstring.properties` then. Then all strings are localized.

I put a little pro/contra list together to help us out whether we should use the strings from `.properties` instead of duplicating them in fluent:

Pro:
- Better wording.
- No string dupes (semantically)
- No extra translation work needed
- Maybe more error messages coverage instead of only [the limited set we chose](https://github.com/mozilla/enterprise-firefox/blob/daf78f62ce3741d60a2ddc140be8e7f0c1fc4136/browser/components/enterprise/modules/ConsoleClient.sys.mjs#L256-L262)
- There is an open bug for the migration of the strings to fluent: [Bug 1790187](https://bugzilla.mozilla.org/show_bug.cgi?id=1790187). I talked to the triage owner Manuel. They said they might be able to migrate them to fluent for us.
- We'll revamp those strings when we implement Matt's new design. I'd avoid merging strings to m-c that we know will change soon anyway.

Contra (until the strings are migrated to fluent):
- Brand name `Firefox` is hardcoded into the strings.
- We'd use legacy localisation in felt (but I don't think that's too bad, since we're not adding new legacy localization strings.)

Comment (lissyx) [reply browser/locales/en-US/browser/enterprise/felt.ftl:1]: > We'd put them under `preview/` to have a place to put them now to not block merging the `.ftl` files to m-c. For the pilot we will need to migrate to the strings from `appstring.properties` then. Then all strings are localized.

Strings under `preview/` do not get localized at all.

Comment (lissyx) [reply browser/locales/en-US/browser/enterprise/felt.ftl:1]: Ok it's done in `browser/components/enterprise/modules/ConsoleClient.sys.mjs`, so we might use a string bundle there.

Comment (1rneh) [reply browser/locales/en-US/browser/enterprise/felt.ftl:1]: > Strings under preview/ do not get localized at all.

That's the point. We don't need them localized if we migrate to the strings form `appstrings.properties`.

Comment (lissyx) [reply browser/locales/en-US/browser/enterprise/felt.ftl:1]: > > Strings under preview/ do not get localized at all.
> 
> That's the point. We don't need them localized if we migrate to the strings form `appstrings.properties`.

Then I'm sorry but I dont understand your plan here.

Comment (lissyx) [reply browser/locales/en-US/browser/enterprise/felt.ftl:1]: The `.properties` accept parameters for mostly all the strings we want to format: https://searchfox.org/enterprise-main/rev/fc98b717f157e6287da2b87bd86090a94703cdc8/browser/locales/en-US/chrome/overrides/appstrings.properties#11,14-16,22,24

So we need to surface the URL of the console there.

Comment (lissyx) [reply browser/locales/en-US/browser/enterprise/felt.ftl:1]: @1rneh Let me know, I've found a way to perform the localization closer to what you wanted

File: browser/components/enterprise/modules/ConsoleClient.sys.mjs [hunk 6]
```diff
@@ -253,15 +253,22 @@ export const ConsoleClient = {
...
+        [Cr.NS_ERROR_NET_TIMEOUT]: "netTimeout",
+        [Cr.NS_ERROR_NET_RESET]: "netReset",
+        [Cr.NS_ERROR_NET_INTERRUPT]: "netInterrupt",
+        [Cr.NS_ERROR_OFFLINE]: "netOffline",
       };
-      return networkErrors[status] || `error-code-${status}`;
+      // 'network-error' here is a generic catch-call expected to be in Fluent
+      return status
+        ? networkErrors[status] || `error-code-${status}`
+        : `network-error`;
```
1rneh: This is probably out of scope, but I'd really like to move this function out of the `ConsoleClient`. We should just return the error status here. The ConsoleClient shouldn't do any status to string id mapping. That should happen wherever we use those strings hence in `window.js`. If we don't have a status, then probably we don't need to show the `felt-error-network-error ` message since it contains no information really. We already show `Unable to connect to the console. Please contact your administrator.` .

1rneh: What does `error-code-${status}` do?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -313,10 +320,8 @@ export const ConsoleClient = {
       };
 
       xhr.onerror = () => {
-        const errorName = xhr.channel
-          ? this._getErrorNameForStatus(xhr.channel.status)
-          : "NetworkError";
-        reject(new TypeError(errorName));
+        const errorName = this._getErrorNameForStatus(xhr.channel?.status);
+        reject(new TypeError(errorName, { cause: new URL(url).host }));
```
1rneh: Could we pass `{ cause : { host : new URL(url).host } }`?

browser/extensions/felt/content/window.js
```diff
@@ -54,14 +61,33 @@ const ErrorReport = {
         ".felt-browser-error-details"
       );
       if (detailsElement) {
-        const l10nId = `felt-error-${details}`;
-        const translated = await document.l10n.formatValue(l10nId);
+        const translated = await this.localizeString(details, cause);
         detailsElement.textContent = translated || details;
```
1rneh:
```suggestion
        const message = await this.getLocalisedErrorString(details, cause);
        detailsElement.textContent = message;
```

browser/extensions/felt/content/window.js
```diff
@@ -54,14 +61,33 @@ const ErrorReport = {
...
+  formatStringBundle(msgId, url) {
+    try {
+      return this._stringBundles.app.formatStringFromName(msgId, [url]);
+    } catch (ex) {
+      console.error(
+        `FELT error localization failed for '${msgId}'. Expected for NSS errors.`
+      );
+      return null;
+    }
+  },
```
1rneh: `netOffline` and `netReset` don't take a parameter. This approach seems cleaner to me as it doesn't require trying to format strings in the wrong localization system.

```suggestion
  async getLocalisedErrorString(id, host = null) {
    switch (id) {
      case "dnsNotFound2":
      case "connectionFailure":
      case "netInterrupt":
      case "netTimeout":
        return this._stringBundles.app.formatStringFromName(id, [host]);
      case "netOffline":
      case "netReset":
        return this._stringBundles.app.getStringFromName(id);
      case "felt-error-network":
        return document.l10n.formatValue(id);
      default:
        // This is not a string id but a error status. 
        // No localization needed.
        return id;
    }
  },
```

1rneh (changes requested):
Thank you for migrating to the m-c network error strings!

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: > This is probably out of scope, but I'd really like to move this function out of the `ConsoleClient`. We should just return the error status here. The ConsoleClient shouldn't do any status to string id mapping. That should happen wherever we use those strings hence in `window.js`. If we don't have a status, then probably we don't need to show the `felt-error-network-error ` message since it contains no information really. We already show `Unable to connect to the console. Please contact your administrator.` .

I would say this is something we can look at once the refactoring is done.

lissyx: > What does `error-code-${status}` do?

I defer to @gcp for the intention here, I suspect the goal was to have Fluent strings for specific status as we run into the issues?

browser/extensions/felt/content/window.js
lissyx: I think we still want to `|| details` to show raw error string when we dont have localization, like for some NSS error codes.

browser/extensions/felt/content/window.js
lissyx: I'm really not sure this is a good idea, because now to deal with a few duplicated string, we have two places where we duplicate their IDs to do something with them, kinda defeating the purpose.

And the original code has no problem passing identifiers with parameters when they dont need parameters: https://searchfox.org/enterprise-main/rev/fc98b717f157e6287da2b87bd86090a94703cdc8/docshell/base/nsDocShell.cpp#3850, this is perfectly handled and fine.

lissyx:
@1rneh Should be good, I fixed all except the `formatStringName()` as I explained above.

lissyx:
Once this lands we can land:
 - https://github.com/mozilla-partners/enteprise-manifests/pull/2

browser/extensions/felt/content/window.js
1rneh: I think we are smallowing nss error codes right now because we only display fluent or string bundle strings, right?

lissyx: No because if the string identifier is not found then we return null and it should fallback to `|| details`

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: Yeah if needed. But as noted, if we don't hit any of the known errors we're probably in a bad state. (Yet, that's what you got, for some reason I don't understand).

Just simplify this to return the (fixed) `network-error`.

lissyx: Should I just make it `return networkErrors[status] || "network-error";` then @gcp ?

browser/extensions/felt/content/window.js
```diff
@@ -54,14 +61,33 @@ const ErrorReport = {
...
+        const message = await this.getLocalisedErrorString(details, cause);
+        detailsElement.textContent = message || details;
       }
     }
     errorElement.classList.remove("is-hidden");
     this._wrapper.classList.remove("is-hidden");
   },
+
+  async getLocalisedErrorString(details, cause) {
+    const l10nId = `felt-error-${details}`;
```
1rneh: We only have one `felt-error` string. Can we pass that instead of constructing it here?

browser/extensions/felt/content/window.js
```diff
@@ -54,14 +61,33 @@ const ErrorReport = {
...
+    const fluentValue = await document.l10n.formatValue(l10nId);
+    if (fluentValue) {
+      return fluentValue;
+    }
```
1rneh:
```suggestion
    const errorMessage = await document.l10n.formatValue(l10nId);
    if (errorMessage) {
      return errorMessage;
    }
```

browser/locales/en-US/browser/enterprise/felt.ftl
```diff
@@ -2,4 +2,27 @@
...
+    Version { $version }
+
+## Error details when launching the browser crashes
+
+felt-browser-error-multiple-crashes = { -brand-short-name } crashed multiple times.
+felt-browser-error-connection = Unable to connect to the console. Please contact your administrator.
+
+## Network error details.
+
+felt-error-network-error = Unknown network error
```
1rneh: Can we rename this to `felt-error-network`. The second error seems redundant.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: I don't think this is resolved yet.

browser/extensions/felt/content/window.js
1rneh: I'd rather have strings in two places temporarily. The error code to string id mapping ultimately doesn't belong in the `ConsoleClient`. I'd rather have it only return the status and let `window.js` do the error name to string id mapping in one place. I think the current approach is not a good solution. But given the urgency of this, this is not a blocker.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Could you file a follow-up bug to move the status to string id mapping out of the `ConsoleClient`?

1rneh (approved):
Approved once my comments are addressed.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Done https://bugzilla.mozilla.org/show_bug.cgi?id=2015092

lissyx: > I don't think this is resolved yet.

Yep, it slipped through.

browser/extensions/felt/content/window.js
lissyx: I think we still need it, we will receive either `network` as error code for the fluent string that we default to, if the code in `ConsoleClient` could not map a known network error. However if it has been able to, then we get a known error code that does not match the fluent id so we expect the `document.l10n.formatValue` to fail here so we can fallback to the string bundle.

lissyx: Ok I hope what I did is what you expected. Either way, it's not something that changes strings themselves, so we can always fix that later anyway without much pain.

lissyx:
Sync to the main repo is in place: https://github.com/mozilla-l10n/firefox-l10n/pull/149

PR #359, opened 2026-01-21
add support for profile per user

fiji-flo:
Add a suffix based on the first 40bits of the sha256 of the user id. Fall back to the legacy name.

### DIFF
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -54,6 +54,8 @@ export class FeltProcessParent extends JSProcessActorParent {
     this.firefoxReady = false;
     // Track extension ready state (extension must register its observer)
     this.extensionReady = false;
+    // Current loggedInUserInfo
+    this.loggedInUserInfo = null;
 
     this.abnormalExitCounter = 0;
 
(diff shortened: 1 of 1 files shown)


lissyx:
This is the fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1990452 ?

gcp:
>This is the fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1990452 ?

yes

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -347,22 +349,30 @@ export class FeltProcessParent extends JSProcessActorParent {
         "@mozilla.org/toolkit/profile-service;1"
       ].getService(Ci.nsIToolkitProfileService);
 
+      let profileName = await this.profileName();
       let foundProfile = null;
+
       for (let profile of profileService.profiles) {
-        if (profile.name === lazy.FeltCommon.ENTERPRISE_PROFILE) {
-          foundProfile = profile;
+        if (profile.name === profileName) {
```
gcp, line 356: I think this needs `foundProfile = profile` or you will always use the legacy profile.

gcp: ~~Unfortunately this probably can only be reasonably tested with an integration test.~~

This is testable, i.e. you can assert you are creating an encoded profile instead of a legacy one.

gcp (changes requested):

gcp (changes requested):
This looks broken.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
fiji-flo: How did that got lost 🤷 

I was looking if there are any test for the existing profile.

gcp: tl;dr fails because the test framework makes the profile and even https://bugzilla.mozilla.org/show_bug.cgi?id=2011328 won't help

gcp (approved):

PR #367, opened 2026-01-23
Add enterprise-specific addon install event with addon name

jonathanmendez:
The console would like the addon name as part of this telemetry event. Also this was the only re-used normal telemetry event that we included for enterprise, so now we are instead cloning an enterprise-specific event.

### DIFF
--- a/toolkit/mozapps/extensions/AddonManager.sys.mjs
+++ b/toolkit/mozapps/extensions/AddonManager.sys.mjs
@@ -5595,6 +5595,32 @@ AMTelemetry = {
   },
 
   /**
+   * Retrieve the addon name for the given AddonInstall instance.
+   *
+   * @param {AddonInstall} install
+   *        The AddonInstall instance to retrieve the addon name from.
+   *
+   * @returns {string | null}
+   *          The addon name for the given AddonInstall instance (if any).
+   */
+  getAddonNameFromInstall(install) {
+    // Returns the name of the extension that is being installed, as soon as the
+    // addon is available in the AddonInstall instance (after being downloaded
+    // and validated successfully).
+    if (install.addon) {
+      return install.addon.name;
+    }
+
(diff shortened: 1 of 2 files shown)


toolkit/mozapps/extensions/AddonManager.sys.mjs
```diff
@@ -5899,7 +5926,25 @@ AMTelemetry = {
       })
     );
 #ifdef MOZ_ENTERPRISE
-    GleanPings.enterprise.submit();
+    if (eventMethod == "install" && extra.step == "completed") {
```
gcp, line 5929: Do we care about updates?

gcp (approved):

lissyx:
Is it related to https://bugzilla.mozilla.org/show_bug.cgi?id=2012401 and https://bugzilla.mozilla.org/show_bug.cgi?id=2012400 ?

toolkit/mozapps/extensions/AddonManager.sys.mjs
jonathanmendez: Not for now, maybe for MVP, per @rtestard

jonathanmendez:
> Is it related to https://bugzilla.mozilla.org/show_bug.cgi?id=2012401 and https://bugzilla.mozilla.org/show_bug.cgi?id=2012400 ?

I think this may fix those tests, at least bug 2012400. Before this change, the "install" glean event would get immediately sent because of the enterprise ping. The normal glean pings do not immediately get sent, and tests can rely on that to inspect the glean queue (for some enterprise-specific tests we instrument the code to not send the enterprise ping under a test environment so we can inspect the queue).

Since this change reverts the change to the "install" glean event in favor of a new enterprise-specific event, we should no longer be messing up the glean queue that the tests are expecting.

PR #370, opened 2026-01-24
Bug 2012261 - Remove redeclaration of AppConstants

lissyx:


1rneh (approved):

lissyx:
@gcp I assume you just added it out of precaution in 73a8d1abaec37952024665f6a1d888efd2da6faa but it is in fact not useful since it is breaking tests

PR #373, opened 2026-01-26
Add mitm, portal and connectivity endpoints

fiji-flo:
repack is already merged https://github.com/mozilla/enterprise-repack/pull/3

### DIFF
--- a/browser/branding/enterprise/pref/firefox-enterprise.js
+++ b/browser/branding/enterprise/pref/firefox-enterprise.js
@@ -10,3 +10,20 @@ pref("browser.profiles.enabled", false);
 pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
 
 pref("enterprise.loglevel", "Error");
+
+pref(
+  "security.certerrors.mitm.priming.endpoint",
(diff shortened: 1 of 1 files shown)


lissyx:
@fiji-flo I think we also want a PR against https://github.com/mozilla/enterprise-repack/tree/master/enterfox. I'd like us to remove those from in-tree, but we may need to keep them for local builds to work without too much pain.

fiji-flo:
Sure. I wasn't sure what's our strategy there.

fiji-flo:
https://github.com/mozilla/enterprise-repack/pull/4 @lissyx

lissyx:
@fiji-flo After today's meeting, should we do this?

gcp:
After discussion with the release/updater teams, we want to repack as little information as possible because it's problematic to change later. This means we only want to repack the console URL. 

The endpoints that are required for networking, so the ones that we need before console comms are fully established, hence have to be derivable from the console address.

The console then sets the remaining ones via prefs.

gcp:
Replaced by https://github.com/mozilla/enterprise-firefox/pull/445

PR #385, opened 2026-01-29
Fix Crash on AsyncDBus

lissyx:


dom/system/linux/ModemManagerImeiProvider.h
```diff
@@ -27,8 +28,11 @@ class ModemManagerImeiProvider final : public nsIImeiProvider {
   ~ModemManagerImeiProvider() = default;
 
   RefPtr<GCancellable> mCancellable;
```
gcp, line 36: Are you using this? It's just being passed around as `nullptr` the entire time no?

```
  ModemManagerImeiProvider() : mCancellable(dont_AddRef(g_cancellable_new())) {}                             
                                                                                                             
  ~ModemManagerImeiProvider() {                                                                              
      if (mCancellable) {                                                                                    
          g_cancellable_cancel(mCancellable);                                                                
      }                                                                                                      
  }                                                                                                          
```

gcp (approved):

dom/system/linux/ModemManagerImeiProvider.h
lissyx: ouch

lissyx: you're right I missed it when looking at how it was done elsewhere

PR #389, opened 2026-01-30
Bug 2013652 - Enterprise: allow missing MAC address when reporting ne…

lissyx:
…twork interfaces

### DIFF
--- a/netwerk/system/netlink/NetlinkService.cpp
+++ b/netwerk/system/netlink/NetlinkService.cpp
@@ -277,11 +277,7 @@ class NetlinkLink {
     iface = (ifinfomsg*)NLMSG_DATA(aNlh);
     len = aNlh->nlmsg_len - NLMSG_LENGTH(sizeof(*iface));
 
-    bool shouldStop = false;
     bool hasName = false;
-#if defined(MOZ_ENTERPRISE)
-    bool hasMAC = false;
-#endif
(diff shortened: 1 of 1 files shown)


netwerk/system/netlink/NetlinkService.cpp
```diff
@@ -291,26 +287,15 @@ class NetlinkLink {
 #if defined(MOZ_ENTERPRISE)
       if (attr->rta_type == IFLA_ADDRESS) {
         memcpy(mMAC, RTA_DATA(attr), ETH_ALEN);
-        hasMAC = true;
```
jonathanmendez, line 294: `NetlinkNeighbor` just above has a similar concept but adds a `HasMAC()` method. I think it makes sense to add the same for NetlinkLink.

jonathanmendez (approved):
I think that the current implementation is probably safe for how it's being used (we'd end up reporting the MAC as "00:00:00:00:00:00"?). But I have a suggestion for adding a `HasMAC()` method.

netwerk/system/netlink/NetlinkService.cpp
lissyx: Do you mind filing a follow up bug on bugzilla with your idea in details ? I'm not sure of the use we would make of such `HasMAC` here

PR #400, opened 2026-02-04
Bug 2014434 - Enterprise: add enterprise-searchfox for specific build…

lissyx:
… config

### DIFF
--- a/taskcluster/gecko_taskgraph/transforms/build.py
+++ b/taskcluster/gecko_taskgraph/transforms/build.py
@@ -322,3 +322,16 @@ def add_enterprise_secret_scopes(config, jobs):
             ])
 
         yield job
+
+
+@transforms.add
+def add_enterprise_to_searchfox(config, jobs):
+    for job in jobs:
+        if (
+            config.params["project"] == "enterprise-firefox"
+            and config.kind == "searchfox"
+        ):
+            job["run"].setdefault("extra-config", {}).setdefault(
+                "extra_mozconfig_content", []
+            ).extend(["ac_add_options --enable-enterprise"])
+        yield job


taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -1196,6 +1200,60 @@ def target_tasks_searchfox(full_task_graph, parameters, graph_config):
...
+                if expiry > datetime.utcnow() + timedelta(days=7):
+                    print("Skipping index tasks")
+                    return []
+        if not create.testing:
+            insert_index(index_path, os.environ["TASK_ID"])
+
+    return [
+        "enterprise-searchfox-linux64-searchfox/opt",
+        "enterprise-searchfox-linux64-searchfox/debug",
+        "enterprise-searchfox-macosx64-searchfox/debug",
```
arai-a: The corresponding `macosx64-searchfox/debug` section is removed in the kind.yml file.

arai-a (commented):
I don't know the taskgraph code details, so I'll skip those parts.
The searchfox-specific  looks good, except for the reference to possibly missing entry.

taskcluster/kinds/enterprise-searchfox/kind.yml
```diff
@@ -0,0 +1,263 @@
...
+            platform: linux64-enterprise/opt
+        worker:
+            max-run-time: 10800
+            env:
+                MOZSEARCH_PLATFORM: "linux64-opt"
+        run:
+            config:
+                - builds/releng_base_firefox.py
+                - builds/releng_sub_linux_configs/64_searchfox_and_opt.py
+            extra-config:
```
jonathanmendez: As discussed in meeting, consider using a transform to inject the config instead of defining a whole kind. (While also taking into account the `run-on-projects` value)

taskcluster/gecko_taskgraph/transforms/build.py
```diff
@@ -322,3 +322,18 @@ def add_enterprise_secret_scopes(config, jobs):
...
+        if config.params["project"] == "enterprise-firefox" and config.kind == "searchfox":
+            try:
+                job["run"].setdefault("extra-config", {})["extra_mozconfig_content"] += [
+                    "ac_add_options --enable-enterprise"
+                ]
+            except KeyError:
+                job["run"].setdefault("extra-config", {})["extra_mozconfig_content"] = [
+                    "ac_add_options --enable-enterprise"
+                ]
+        yield job
```
ahal, line 337: No need to make this a transform. Just add it to the tasks directly in `taskcluster/kinds/searchfox/kind.yml`.

ahal (changes requested):

taskcluster/gecko_taskgraph/transforms/build.py
lissyx: The goal here was to limit the changes to `searchfox` kind and avoid merge conflicts.

taskcluster/gecko_taskgraph/transforms/build.py
```diff
@@ -322,3 +322,18 @@ def add_enterprise_secret_scopes(config, jobs):
             ])
 
         yield job
+
+
+@transforms.add
+def add_enterprise_to_searchfox(config, jobs):
+    for job in jobs:
+        if config.params["project"] == "enterprise-firefox" and config.kind == "searchfox":
```
jonathanmendez: Will this ever be true, given that the searchfox kind.yml defines `run-on-projects: ['mozilla-central']`?

taskcluster/gecko_taskgraph/transforms/build.py
ahal: Adding lots of transforms tends to make it harder to reason about things, so I tend to prefer keeping changes in the `kind.yml` where possible.

> The goal here was to limit the changes to searchfox kind

Editing the `searchfox/kind.yml` file also limits to the searchfox kind :)

> and avoid merge conflicts.

This is a valid reason to use a transform. Though it's worth pointing out that you could also avoid merge conflicts in the kind.yml:
```
task-defaults:
    run:
        extra-config:
            extra_mozharness_content:
                by-project:
                    enterprise-firefox: ["--ac_add_options --enable-enterprise"]
                    default: []
```

(though you'd need to ensure that `run.extra-config.extra_mozharness_config` gets filtered through `evaluate_keyed_by` in some transform).

This could land in both `firefox-main` and `enterprise-main`.

Anyway, using the transform is fine for now.

taskcluster/gecko_taskgraph/transforms/build.py
ahal: It'll never be triggered via push, but it's configured to run on a cron schedule:
https://github.com/mozilla/enterprise-firefox/blob/enterprise-main/.cron.yml#L184

taskcluster/gecko_taskgraph/transforms/build.py
```diff
@@ -322,3 +322,18 @@ def add_enterprise_secret_scopes(config, jobs):
...
+            try:
+                job["run"].setdefault("extra-config", {})["extra_mozconfig_content"] += [
+                    "ac_add_options --enable-enterprise"
+                ]
+            except KeyError:
+                job["run"].setdefault("extra-config", {})["extra_mozconfig_content"] = [
+                    "ac_add_options --enable-enterprise"
+                ]
```
ahal: Use a second `.setdefault("extra_mozharness_content", [])` rather than catching `KeyError`.

jonathanmendez (approved):
Looks good to me with @ahal's suggestion

PR #403, opened 2026-02-05
Enterprise bug2012083 handle shutdown crash

lissyx:


browser/extensions/felt/rust/src/client.rs
jonathanmendez: "...to send quit-application messages directly"?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -308,7 +315,7 @@ export class FeltProcessParent extends JSProcessActorParent {
             `firefox exit: PID:${this.proc.pid} exitCode:${JSON.stringify(this.proc.exitCode)}`
           );
           if (!this.restartReported && !this.logoutReported) {
-            if (this.proc.exitCode === 0) {
+            if (this.proc.exitCode === 0 || this.exitReported) {
```
jonathanmendez: This probably is still the right logic after https://bugzilla.mozilla.org/show_bug.cgi?id=2012442 gets implemented? Regardless, if there are changes needed they can be done later.

jonathanmendez (approved):

toolkit/components/terminator/nsTerminator.cpp
```diff
@@ -330,7 +335,7 @@ void nsTerminator::StartWatchdog() {
 
   UniquePtr<Options> options(new Options());
   // crashAfterTicks is guaranteed to be > 0 as
-  // crashAfterMS >= ADDITIONAL_WAIT_BEFORE_CRASH_MS >> HEARTBEAT_INTERVAL_MS
+  // crashAfterMS >= additionalWaitBeforeCrashMs >> HEARTBEAT_INTERVAL_MS
```
jonathanmendez: This comment asserts crashAfterTicks will be > 0, but that's not necessarily true anymore depending on the prefs. If the crashAfterTicks is 0, will that cause a problem?
If yes, then maybe this needs to be set to Max(1, crashAfterMS/HEARTBEAT_INTERVAL_MS).
If it doesn't cause a problem, at least the comment needs updating.

testing/enterprise/test_felt_browser_shutdown_crash_exit.py
```diff
@@ -0,0 +1,71 @@
...
+            self._logger.info("Trying to connect to new browser")
+            self.connect_child_browser()
+            new_browser_pid = self._child_driver.session_capabilities["moz:processID"]
+            assert self._browser_pid != new_browser_pid, (
+                f"PID should not be the same: {self._browser_pid} != {new_browser_pid}"
+            )
+            assert new_browser_pid is None, (
+                f"Should not have started a new browser: {new_browser_pid}"
+            )
+            assert False, (
```
jonathanmendez, line 66: The `assert False` suggests something before these asserts will throw. These asserts are confusing, if the expectation is they never execute anyway. Can the asserts before `assert False` be removed?

toolkit/components/terminator/nsTerminator.cpp
lissyx: @gabrielesvelto Is this going to be a problem?

testing/enterprise/test_felt_browser_shutdown_crash_exit.py
lissyx: The idea was to make sure that we always end up reporting an assertion failure, whatever may happen

jonathanmendez (changes requested):
I did testing with the current implementation and it is working.
Just need resolution on the comment in nsTerminator.cpp

testing/enterprise/test_felt_browser_shutdown_crash_exit.py
jonathanmendez: Since there is a fall back to the final `assert False` I think just leaving that one simplifies things, but it's not a big deal.

toolkit/components/terminator/nsTerminator.cpp
lissyx: I'm going to add an extra enforcement similar to `crashAfterMs` to make sure `additionalWaitBeforeCrashMs` cannot be `<=`

lissyx:
> I did testing with the current implementation and it is working. Just need resolution on the comment in nsTerminator.cpp

This is fixed. I've got a pending review on https://phabricator.services.mozilla.com/D282693 so we can pick final version from central when it lands there

toolkit/components/terminator/nsTerminator.cpp
lissyx: gabriele made the same suggestion :)

PR #416, opened 2026-02-09
Bug 2005507 - Add cache to check recently blocked URLs and avoid duplicate event

jonathanmendez:


dom/security/nsContentSecurityManager.cpp
```diff
@@ -112,6 +149,11 @@ static void RecordBlocklistDomainBrowsedTelemetry(nsIChannel* aChannel,
     return;
   }
 
+  static RecentBlockedUrlCache cache;
```
gcp, line 161: `MOZ_ASSERT(NS_IsMainThread())`

dom/security/nsContentSecurityManager.cpp
```diff
@@ -70,6 +70,43 @@ mozilla::LazyLogModule sUELLog("UnexpectedLoad");
...
+  bool RecentlyBlocked(nsIURI* aURI) {
+    if (!aURI) {
+      return false;
+    }
+    auto now = TimeStamp::Now();
+    nsCString url;
+    aURI->GetSpec(url);
+    ClearStaleEntries(now);
+    bool returnVal = ContainsUrl(url);
+    entries.push_back(std::make_pair(url, now));
```
gcp: Note this in theory can block an URL indefinitely, if I keep inserting it at the 0.99 second mark.

dom/security/nsContentSecurityManager.cpp
```diff
@@ -70,6 +70,43 @@ mozilla::LazyLogModule sUELLog("UnexpectedLoad");
...
+    return returnVal;
+  };
+
+ private:
+  std::deque<std::pair<nsCString, TimeStamp>> entries;
```
gcp: Does this compile in non-unified mode? (I don't see the include)

dom/security/nsContentSecurityManager.cpp
```diff
@@ -70,6 +70,43 @@ mozilla::LazyLogModule sUELLog("UnexpectedLoad");
...
+    }
+  };
+  bool ContainsUrl(const nsACString& url) {
+    for ([[maybe_unused]] const auto& [entry_url, _] : entries) {
+      if (url.Equals(entry_url)) {
+        return true;
+      }
+    }
+    return false;
+  }
```
gcp, line 115: Could use `clang-format`. (It's inconsistent with the code right above)

gcp (approved):

dom/security/nsContentSecurityManager.cpp
jonathanmendez: Did some manual testing with MOZ_RELEASE_ASSERT too just to be sure.

dom/security/nsContentSecurityManager.cpp
jonathanmendez: Updated the logic to only record new entries if an existing entry within the window was not found.

dom/security/nsContentSecurityManager.cpp
jonathanmendez: This did actually compile in non-unified mode. But the includes are now added as they should.

PR #421, opened 2026-02-10
Enterprise bug2006637 ensure startup always

lissyx:


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -259,6 +255,24 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      "felt-firefox-logout",
+    ];
+
+    observerTopics.forEach(aTopic => {
+      const num = Array.from(Services.obs.enumerateObservers(aTopic)).length;
+      if (num !== 0) {
+        console.debug(
+          `FeltExtension: ParentProcess: observerTopics[${aTopic}]: ${num} INCORRECT, WTF`
+        );
+        throw new Error(`Too many observers: ${aTopic}:${num}`);
```
jonathanmendez, line 278: Based on wording of the logs and the other code changes, this feels effectively like a debug assert, i.e. something we expect should never happen (as opposed to a runtime error that may sometimes happen). I'm less familiar with js patterns, so forgive if this is obvious, but is throwing an Error the right fit for that?

Separate observation: This is enforcing that the only observer of these particular events is this particular observer. This is probably fine, certainly for now, but it at least seems plausible that we could someday build other functionality that would want to observe the same events. I guess we'd just adjust if we ever come to that.

lissyx: > Based on wording of the logs and the other code changes, this feels effectively like a debug assert, i.e. something we expect should never happen (as opposed to a runtime error that may sometimes happen). I'm less familiar with js patterns, so forgive if this is obvious, but is throwing an Error the right fit for that?

I'm not sure how we can assert that differently

> 
> Separate observation: This is enforcing that the only observer of these particular events is this particular observer. This is probably fine, certainly for now, but it at least seems plausible that we could someday build other functionality that would want to observe the same events. I guess we'd just adjust if we ever come to that.

I let it for now as a debugging mean, to make sure we dont hit path where we would not properly unregister.

lissyx: @1rneh what's your opinion here?

lissyx:
QA confirmed that this is fixing the issue from bug 2006637

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Ok, this is going to block/make conflicts for https://bugzilla.mozilla.org/show_bug.cgi?id=2012080 and https://bugzilla.mozilla.org/show_bug.cgi?id=2012442

jonathanmendez (approved):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: https://bugzilla.mozilla.org/show_bug.cgi?id=2016195

PR #432, opened 2026-02-11
Bug 2012080 - Enterprise: send back tokens on shutdown

lissyx:


testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -0,0 +1,87 @@
...
+        self._child_driver.set_context("chrome")
+        self._child_driver.execute_script(
+            """
+            Services.felt.setTokens(arguments[0], arguments[1], arguments[2]);
+            """,
+            [self.access_token, self.refresh_token, self.expires_in],
+        )
+        self._child_driver.set_context("content")
+
+        self.perform_quit()
```
jonathanmendez: Test readability: Move this to the top level `test_browser_exit_tokens()` instead of it being a side effect of `run_browser_set_fake_tokens()`

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self.access_token = "1bf8d2cb-9f72-4788-a770-3cf9cc60f30c"
+        self.refresh_token = "82b2d9eb-f0e3-44af-8106-790e7a744d1f"
+        self.expires_in = 3600
+
+        assert browser_tokens[0] != self.access_token, (
+            f"Access token differs: {browser_tokens[0]} vs {self.access_token}"
+        )
+        assert browser_tokens[1] != self.refresh_token, (
+            f"Refresg token differs: {browser_tokens[1]} vs {self.refresh_token}"
```
jonathanmendez: nit: spelling
```suggestion
            f"Refresh token differs: {browser_tokens[1]} vs {self.refresh_token}"
```

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self.perform_quit()
+
+    def run_felt_check_fake_tokens(self):
+        felt_tokens_after_signout = self.get_tokens(self._driver)
+        assert felt_tokens_after_signout[0] == self.access_token, (
+            f"Access token matches: {felt_tokens_after_signout[0]} vs {self.access_token}"
+        )
+        assert felt_tokens_after_signout[1] == self.refresh_token, (
+            f"Refresg token matches: {felt_tokens_after_signout[1]} vs {self.refresh_token}"
```
jonathanmendez: nit: spelling
```suggestion
            f"Refresh token matches: {felt_tokens_after_signout[1]} vs {self.refresh_token}"
```

browser/extensions/felt/rust/src/client.rs
```diff
@@ -87,6 +87,31 @@ impl FeltIpcClient {
...
+                ));
+                if let Some(tx) = &self.tx {
+                    match tx.send(msg) {
+                        Ok(()) => trace!("FeltIpcClient::send_back_tokens() SENT"),
+                        Err(err) => trace!("FeltIpcClient::send_back_tokens() TX ERROR: {}", err),
+                    }
+                }
+            }
+            Err(_) => {
+                trace!("FeltXPCOM::send_back_tokens failed: couldn't acquire lock",);
```
jonathanmendez:
```suggestion
                trace!("FeltIpcClient::send_back_tokens failed: couldn't acquire lock",);
```

browser/extensions/felt/rust/src/client.rs
```diff
@@ -87,6 +87,31 @@ impl FeltIpcClient {
         }
     }
 
+    pub fn send_back_tokens(&self) {
+        trace!("FeltIpcClient::send_back_tokens()");
+        match TOKENS.read() {
+            Ok(tokens) => {
+                trace!(
+                    "FeltXPCOM::send_back_tokens()"
```
jonathanmendez:
```suggestion
                    "FeltIpcClient::send_back_tokens()"
```
Is this redundant with the trace on line 91? Maybe to match the one from FeltXPCOM, you'd want to include the tokens data in the trace? (`"FeltIpcClient::send_back_tokens({} {} {})", tokens.access_token, tokens.refresh_token, tokens.expires_at`)

browser/extensions/felt/rust/src/components.rs
```diff
@@ -111,23 +118,33 @@ impl FeltXPCOM {
...
-                    tokens.access_token,
-                    tokens.refresh_token,
-                    tokens.expires_at
-                );
-                self.send(FeltMessage::Tokens((
-                    tokens.access_token.clone(),
-                    tokens.refresh_token.clone(),
-                    tokens.expires_at,
-                )))
+        if self.is_felt_browser {
```
jonathanmendez, line 121: I probably need a deep-dive on the connections involved here, but at least based on my current understanding, I'm not clear on why there needs to be a separate path for "send back" instead of just using the existing logic/message.

I think the main necessary piece is receiving the message in the felt server thread, which you add below. Then I think you can call SendTokens from the console client, which would send a Tokens message from the browser to the server, which would be received in the server thread, no?

jonathanmendez (changes requested):
My main question is about the need for send_back_tokens as a separate path. Otherwise it's mostly nitpicks.

testing/enterprise/test_felt_browser_exit_tokens.py
lissyx: This is fun, I was doing it before and I felt it was cleaner to scope the `perform_quit()` within the `run_browser_set_fake_tokens()`

browser/extensions/felt/rust/src/client.rs
lissyx: Yes it is redundant. I've removed the tokens explicitely to avoid risking leaking data by mistake

browser/extensions/felt/rust/src/components.rs
lissyx: > Then I think you can call SendTokens from the console client, which would send a Tokens message from the browser to the server, which would be received in the server thread, no?

I am not clear what you have in mind here. We need mostly the same logic flow as for signout notification here or even `SendExtensionReady`, so on the XPCOM side we need ot check in what case we fall through to perform the send on the right socket

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: `this.tokenData = null;` can be removed here.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -656,6 +657,22 @@ export const ConsoleClient = {
    */
   init() {
     Services.obs.addObserver(this, "xpcom-shutdown");
+
+    if (Services.felt.isFeltBrowser()) {
+      lazy.AsyncShutdown.appShutdownConfirmed.addBlocker(
+        `Sending back tokens on shutdown`,
```
1rneh: Let's follow the string pattern from the other usages of this API.
```suggestion
        `ConsoleClient: Sending back tokens to felt on shutdown`
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -656,6 +657,22 @@ export const ConsoleClient = {
...
+        () => {
+          try {
+            Services.felt.sendTokens();
+          } catch (ex) {
+            console.error(
+              `FeltExtension: appShutdownConfirmed sent failed: ${ex}`
```
```suggestion
              `ConsoleClient: Failed to send back tokens to felt on shutdown: ${e}`
```

testing/enterprise/test_felt_browser_signout.py
```diff
@@ -30,12 +30,6 @@ def felt_whoami(self):
         self._child_driver.set_context("content")
         return rv
 
-    def force_window(self):
-        self._driver.set_context("chrome")
-        assert len(self._driver.chrome_window_handles) == 1, "One window exists"
-        self._driver.switch_to_window(self._driver.chrome_window_handles[0])
-        self._driver.set_context("content")
-
```
1rneh, line 38: `force_window` can also be removed in `BrowserCrashes`, because it already inherits it from `FeltTestsBase`?

browser/extensions/felt/api.js
```diff
@@ -193,15 +193,28 @@ this.felt = class extends ExtensionAPI {
...
         );
-        Services.startup.quit(
-          Ci.nsIAppStartup.eAttemptQuit | Ci.nsIAppStartup.eConsiderQuit
+
+        // This is only useful for testing purpose when we need to exit the
+        // browser cleanly but need to keep felt alive for some processing after
+        const inhibitExit = Services.prefs.getBoolPref(
+          "enterprise.felt_tests.inhibit_exit",
+          false
         );
```
1rneh: Can we move this to `EnterpriseCommon` as we do for `enterprise.is_testing`?

1rneh: And also make clear that it's a boolean, e.g. by renaming `inhibitExit` to `isBlockingShutdown`

browser/extensions/felt/api.js
```diff
@@ -193,15 +193,28 @@ this.felt = class extends ExtensionAPI {
...
+        if (!inhibitExit) {
+          Services.startup.quit(
+            Ci.nsIAppStartup.eAttemptQuit | Ci.nsIAppStartup.eConsiderQuit
+          );
+        } else {
+          Services.felt.makeBackgroundProcess(false);
+          this.showWindow();
```
1rneh, line 211:
```suggestion
          const success = Services.felt.makeBackgroundProcess(false);
          console.debug(`FeltExtension: makeBackgroundProcess? ${success}`);
          this.showWindow();
```
Do we care about logging `success` here like we do for the other occurrences?

1rneh (approved):
Just some nits. Otherwise r+ for the JS changes.

testing/enterprise/test_felt_browser_signout.py
lissyx: Yes, my mistake, I forgot

browser/extensions/felt/api.js
lissyx: Not really, and I'm not sure we really need to log it elsewhere in fact, I had placed that in early days to try and make sure it was working for the macOS dock icon issue, but I'm far from convinced it's any useful.

browser/extensions/felt/rust/src/components.rs
jonathanmendez: I'm approving now since this at least matches existing code, and I'm following up separately to better understand the structure of communication more generally.

jonathanmendez (approved):

PR #439, opened 2026-02-13
add device id to login

fiji-flo:
This add a device id persisted in the `felt.json`

### DIFF
--- a/browser/extensions/felt/FeltStorage.sys.mjs
+++ b/browser/extensions/felt/FeltStorage.sys.mjs
@@ -39,6 +39,22 @@ export const FeltStorage = {
   },
 
   /**
+   * Gets the device ID (generates a ID one if not available)
+   *
+   * @returns {string} email
+   */
+  getDeviceId() {
+    let id = this._feltStorage.data?.deviceId;
+    if (id) {
+      return id;
+    }
+    id = globalThis.crypto.randomUUID();
+    this._feltStorage.data.deviceId = id;
+    this._feltStorage.saveSoon();
+    return id;
+  },
+
+  /**
    * Updates the email that was used to sign in the last time
    *
    * @param {string} email
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -8,6 +8,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   TelemetryEnvironment: "resource://gre/modules/TelemetryEnvironment.sys.mjs",
   EnterpriseCommon: "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
   AsyncShutdown: "resource://gre/modules/AsyncShutdown.sys.mjs",
+  FeltStorage: "resource:///modules/FeltStorage.sys.mjs",
 });
 
 /**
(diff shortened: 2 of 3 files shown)


browser/extensions/felt/FeltStorage.sys.mjs
```diff
@@ -38,6 +38,22 @@ export const FeltStorage = {
...
+   * Gets the device ID (generates a ID one if not available)
+   *
+   * @returns {string | undefined} email
+   */
+  getDeviceId() {
+    let id = this._feltStorage.data?.getDeviceId;
+    if (id) {
+      return id;
+    }
+    id = globalThis.crypto.randomUUID();
```
lissyx, line 51: Any reason not to re-use existing telemetry Client ID ? Having it may help correlate things.

https://searchfox.org/enterprise-main/rev/fc98b717f157e6287da2b87bd86090a94703cdc8/toolkit/components/telemetry/app/ClientID.sys.mjs#61

lissyx (changes requested):

browser/extensions/felt/FeltStorage.sys.mjs
fiji-flo: This is a device and not client ID. Is the felt profile persistent enough to use that?

lissyx: As much as I can tell `Client ID` at telemetry level really goes to device level/ installation level.
What kind of "persistent enough" do you care about?

gcp: `felt.json` is in the user directory, so it's not really a device ID, and it's not per installation either (because multiple Firefox Enterprise versions would share it). But it's not really a user-id from Firefox Enterprise's view because it can support multiple logins.

Naming is hard.

fiji-flo: I need some ID here. Naming is hard. I think device id comes close. Ideally it identifies a Firefox Enterprise installation. I'm reluctant or repurposing the Glean ClientId.

gcp: The Telemetry Client ID is per profile. The "deviceID" added here is per OS-user. ~~I think that in Firefox Enterprise those **are** the exact same thing?~~

If you login multiple users in the same Firefox Enterprise install, they would have a different Telemetry Client ID? At least as of https://bugzilla.mozilla.org/show_bug.cgi?id=1990452

So I think this is correct in not using the Telemetry identifier. r+ on this part (but the code is buggy)

browser/extensions/felt/FeltStorage.sys.mjs
```diff
@@ -38,6 +38,22 @@ export const FeltStorage = {
     return this._feltStorage.data?.lastSignedInUserEmail;
   },
 
+  /**
+   * Gets the device ID (generates a ID one if not available)
+   *
+   * @returns {string | undefined} email
+   */
+  getDeviceId() {
+    let id = this._feltStorage.data?.getDeviceId;
```
gcp: You write `data.deviceId` so this is always undefined and generates a new number every run?

gcp (changes requested):

browser/extensions/felt/FeltStorage.sys.mjs
```diff
@@ -38,6 +38,22 @@ export const FeltStorage = {
     return this._feltStorage.data?.lastSignedInUserEmail;
   },
 
+  /**
+   * Gets the device ID (generates a ID one if not available)
+   *
+   * @returns {string | undefined} email
```
gcp: I don't think you can return undefined now.

gcp (approved):

gcp (approved):

PR #442, opened 2026-02-16
Bug 2016195 - Handle window observers once on actor creation/destruction

1rneh:


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -67,6 +67,13 @@ function notifyFirefoxReady() {
   Services.obs.notifyObservers(null, "felt-firefox-window-ready");
 }
 
+const topics = [
```
lissyx, line 70: So for clarity, those observers are bound to the lifecycle of the browser process we start, because they relay the IPC messages we receive from it via our XPCOM.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -183,6 +190,12 @@ export class FeltProcessParent extends JSProcessActorParent {
     };
   }
 
+  actorCreated() {
```
lissyx, line 193: This would technically work, but I find it confusing because it mixes the lifecycle of the JSActor and the lifecycle of the browser process

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -657,6 +648,12 @@ export class FeltProcessParent extends JSProcessActorParent {
     console.error(`FeltExtension: loggedInUserInfo not set`);
     return lazy.FeltCommon.ENTERPRISE_PROFILE;
   }
+
+  didDestroy() {
```
lissyx, line 652: This is going no to work, because we loose this actor when the other pair dies. The JS object will remain and continue its life so things will run fine, except if we remove the observers during the actor destroy phase, then we effectively remove then from the **running** browser process so we end up breaking the communication back from IPC

gcp:
Per discussion with @lissyx and checking with nika, this can be closed.

gcp:
There's still the question whether we should ```throw new Error(`Too many observers: ${aTopic}:${num}`);``` or not.

gcp:
Take a look at https://github.com/mozilla/enterprise-firefox/pull/465

>So for clarity, those observers are bound to the lifecycle of the browser process we start, because they relay the IPC messages we receive from it via our XPCOM.

I mean...yes...and no. They obviously don't do anything if the process is gone, but I'm not clear why we need to kill and re-add them.

PR #443, opened 2026-02-16
Bug 2015697 - Enterprise: set updateURL to console for FELT only

lissyx:


lissyx:
Some context in https://bugzilla.mozilla.org/show_bug.cgi?id=2015697#c2

browser/app/profile/firefox.js
```diff
@@ -153,6 +153,15 @@ pref("app.update.notifyDuringDownload", false);
   pref("app.update.auto", true);
 #endif
 
+#ifdef MOZ_ENTERPRISE
+// On Enterprise we want to enforce updates so we force it
+// TODO: Should we force directly when starting FELT ?
+pref("app.update.auto", true);
+pref("app.update.checkOnlyInstance.enabled", false);
+pref("app.update.background.enabled", true);
+pref("app.update.staging.enabled", true);
```
gcp: Does it make sense to force this (staging)? Doesn't the updater code automatically determine whether staged updates are possible?

Edit: Okay, the code assumes we always want to do staged updates. We might need to check if this has no unexpected problems.

browser/app/profile/firefox.js
```diff
@@ -153,6 +153,15 @@ pref("app.update.notifyDuringDownload", false);
   pref("app.update.auto", true);
 #endif
 
+#ifdef MOZ_ENTERPRISE
+// On Enterprise we want to enforce updates so we force it
+// TODO: Should we force directly when starting FELT ?
```
gcp: I don't understand what you mean (and it should have a follow-up bug).

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+    this.setUpdateState();
+
+    // Inform early
+    this.show(".felt-updates-checking");
+
+    if (lazy.isUpdatesTesting()) {
+      // on Windows at least, during testing, make sure we let time for the UI to show up
+      // before we start those actions. For some reason, not doing this ends up
+      // in Marionette not starting the test code while the update applies.
+      lazy.setTimeout(this.forceUpdateCheck.bind(this), 5000);
```
gcp: Hardcoded delay = yikes.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    Services.prefs.setStringPref(
+      "enterprise.felt.previousBuildID",
+      Services.appinfo.appBuildID
+    );
+  },
+
+  forceUpdateCheck() {
+    Services.prefs.setBoolPref("app.update.auto", true);
```
gcp: You're already setting this to true in the prefs file.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+
+    debug("AppUpdater check");
+    this._updatePromise = this._appUpdater.check();
+
+    this._updatePromise.catch(err => {
+      this._appUpdater.removeListener(this.appUpdaterCallback.bind(this));
+      this.unsetUpdateState();
+      debug(`AppUpdater failure: ${err}`);
+      console.error(err);
+      // TODO: How to deal with those errors?
```
gcp: "Please notify your administrator"?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+      case lazy.AppUpdater.STATUS.READY_FOR_RESTART:
+        this.updateFinished();
+        this.automaticRestart();
+        break;
+
+      // Below are status codes that are handling error states or unexpected
+      // states.
+
+      case lazy.AppUpdater.STATUS.UPDATE_DISABLED_BY_POLICY:
+        // TODO: This should not happen?
```
gcp: Why not?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+        this.unsetUpdateState();
+        break;
+
+      case lazy.AppUpdater.STATUS.OTHER_INSTANCE_HANDLING_UPDATES:
+        // TODO: This should not happen?
```
gcp: I think it can happen in server scenarios? But we can't deal with it for now.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+      case lazy.AppUpdater.STATUS.NO_UPDATES_FOUND:
+        this.unsetUpdateState();
+        break;
+
+      case lazy.AppUpdater.STATUS.UNSUPPORTED_SYSTEM:
+        this.unsetUpdateState();
+        break;
+
+      case lazy.AppUpdater.STATUS.MANUAL_UPDATE:
+        // TODO: What should we do?
```
gcp: "Notify admin"?

browser/extensions/felt/content/styles/felt.css
```diff
@@ -96,6 +98,71 @@ body {
...
+.felt-updates-checking::before {
+  content: "";
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
+  height: 20px;
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
+  width: 20px;
+  animation: spinner 1s linear infinite;
+  position: absolute;
+  margin: var(--space-xxsmall) 0 0 calc(var(--space-small) * -4);
+  /* TODO: We probably want one specific for us? Or a copy of this one? */
```
gcp: File a bug.

browser/locales/en-US/browser/enterprise/felt.ftl
```diff
@@ -5,12 +5,19 @@
...
 felt-window-title = { -brand-short-name } — Login
+
 felt-sso-title = Sign in
 felt-sso-input-email =
     .description = Use your organizational-issued email
     .label = Work email
 felt-sso-continue-btn =
     .label = Continue
+
+felt-updates-title = Good morning
```
gcp: 🤔

testing/enterprise/test_felt_browser_updates_basic.py
```diff
@@ -0,0 +1,117 @@
...
+                            value = Services.prefs.getFloatPref(pref);
+                        }
+                        break;
+
+                    default:
+                        throw new Error(`Unhandled type ${prefType} for ${pref}`);
+                        break;
+                }
+
+                if (value) {
```
gcp: This seems very suspicious given that you have a `getBoolPref` branch. What's the expected behavior _here_ if that returns `false`?

testing/enterprise/felt_tests.py
```diff
@@ -271,12 +302,25 @@ def do_GET(self):
                     content = file.read()
 
                 self.send_response(200, "Success")
-                self.send_header("Content-type", "application/x-xpinstall")
                 self.send_header("Content-Length", len(content))
+                if path.endswith(".xpi"):
+                    self.send_header("Content-Type", "application/x-xpinstall")
+                if path.endswith(".mar"):
+                    self.send_header("Content-Type", "application/octet-stream")
                 self.end_headers()
```
gcp, line 276: I think you lost this return so now you will fall through.

testing/enterprise/felt_tests.py
```diff
@@ -310,11 +354,34 @@ def do_POST(self):
             self.check_auth()
             m = json.dumps(None)
 
+        elif path.startswith("/api/browser/updates"):
+            self.server.serve_updates = not self.server.serve_updates
+            print(f"Server Updates: {self.server.serve_updates}")
```
gcp: Missing return? This just 404's due to the fallthrough below.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+
+  forceUpdateCheck() {
+    Services.prefs.setBoolPref("app.update.auto", true);
+
+    this._appUpdater = new lazy.AppUpdater();
+    if (!this._appUpdater) {
+      throw new Error("Updater failure");
+    }
+
+    const updaterCallback = this.appUpdaterCallback.bind(this);
```
gcp: `this._updaterCallback`

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+    this._appUpdater.addListener(updaterCallback);
+
+    debug("AppUpdater check");
+    this._updatePromise = this._appUpdater.check();
+
+    this._updatePromise.catch(err => {
+      this._appUpdater.removeListener(this.appUpdaterCallback.bind(this));
```
gcp: `this._updaterCallback` (what you have now won't work)

testing/enterprise/test_felt_browser_updates_apply.py
```diff
@@ -0,0 +1,152 @@
...
+        updates_dir = os.path.join(self._update_root, "updates")
+
+        if os.path.isdir(updates_dir):
+            shutil.rmtree(updates_dir)
+
+        def rm_if_exists(path):
+            if os.path.isfile(path):
+                os.unlink(path)
+
+        rm_if_exists(os.path.join(updates_dir, "active-update.xml"))
```
gcp: You rmtree above.

browser/extensions/felt/content/styles/felt.css
```diff
@@ -96,6 +98,71 @@ body {
   line-height: normal;
 }
 
+.felt-updates {
+  display: flex;
+  align-items: left;
```
gcp: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/align-items ??

browser/extensions/felt/content/felt.xhtml
```diff
@@ -115,7 +115,29 @@
...
+            <span
+              class="felt-updates-applying"
+              data-l10n-id="felt-updates-applying"
+            >
+            </span>
+            <progress id="felt-updates-progress" max="100" value="0"></progress>
+          </div>
+        </div>
+      </div>
+      <div class="felt-login is-hidden">
```
gcp: Consider making this visible until update setup/init succeeds.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+
+  checkBuildID() {
+    debug(`checkBuildID`);
+    const previousBuildID = parseInt(
+      Services.prefs.getStringPref(
+        "enterprise.felt.previousBuildID",
+        Services.appinfo.appBuildID
+      )
+    );
+    const actualBuidID = parseInt(Services.appinfo.appBuildID);
```
gcp: typo here (which means something in testing must also be off, or you'd have noticed...)

testing/enterprise/test_felt_browser_updates_basic.py
```diff
@@ -0,0 +1,117 @@
...
+                        value = Services.prefs.getBoolPref(pref);
+                        break;
+
+                    case "string":
+                        value = Services.prefs.getStringPref(pref);
+                        break;
+
+                    case "number":
+                        const num = arguments[pref];
+                        if (Number(num) === num && num % num === 0) {
```
gcp: The ```%``` in ```num % num == 0``` works for float numbers too...

gcp (changes requested):

browser/app/profile/firefox.js
lissyx: Yes, that is why I left a TODO, I was not sure of the behavior we want here.

browser/app/profile/firefox.js
lissyx: @gcp Should those prefs be set directly in `firefox.js` file, or should we force them in `api.js` or somewhere else when we start `FELT`, that's what I was unsure about.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Yes, but only for tests. If someone is able to figure out why Marionette stays blocked at some point, I'm happy to fix better.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: @gcp Yes, but it falls back to the `TODO` there, some decision needs to be made.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: @gcp OK, but in term of UX this needs to be addressed, the UX proposed had no error handling so far. I could be willing to re-use the existing error notification system, but I was not sure it was OK in term of UX: one could assume that since the version is reported in device posture, a failure to update will anyway end up in blocking at device posture check and thus notify user and administrator. Hence, is it really important we notify of the error here? Especially since the user may have no mean to act

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: I was under the impression we would not get into the state that a policy would **block** app update, since we want to enforce it?

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: @gcp In part of the PR I'm forcing ignoring the other instance, so we should be fine. It's also some question we may want to think about and answer, but I think we are safe to assume we should not care for other instances?

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: We may, but same as the UX remark I made on https://github.com/mozilla/enterprise-firefox/pull/443#discussion_r2835318243

browser/extensions/felt/content/styles/felt.css
lissyx: I think I should ask Matt for this? I forgot

browser/locales/en-US/browser/enterprise/felt.ftl
lissyx: That was part of the UX shared :)

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: I think testing is fine

lissyx: I carried over the typo :)

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Thanks, that is what I was supposed to do but I think I lost track of it :(

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: Indeed I need to check `undefined`

testing/enterprise/felt_tests.py
lissyx: I am not sure I see the issue here. We properly send headers, then we send the content, either MAR (with slow path) or XPI (with fast path). Then `m = None` so we do nothing else?

browser/components/enterprise/modules/Updates.sys.mjs
gcp: I'm mostly worried this might cause intermittents.

browser/components/enterprise/modules/Updates.sys.mjs
gcp: What if we break something in webcompat and this causes an internal site to no longer work? That could cause the admin to want to block updates at all costs.

testing/enterprise/felt_tests.py
gcp: If `m = none` you are 404-ing `self.not_found(path)`.

(I see you fixed this anyway?)

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: I agree, but I tried a lot of other options already and nothing has been good enough, MOZ_CHAOSMODE repeatedly was still stable so far even on windows debug builds.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: I assume updates would stop being served anyway then ?

lissyx: But let's assume we have to handle it, we just need to show login then, right ? I don't see any special handling to perform

browser/app/profile/firefox.js
Mossop: Staging is enabled by default for normal Firefox so I would assume there are no known issues

browser/components/enterprise/modules/Updates.sys.mjs
Mossop: 5s is unlikely to cause any racing I guess but what makes me more uncomfortable about this is that it changes the order of operations in tests, the `forceUpdateCheck` will happen later than everything in the rest of this init function (and anything after the init function is called). Would it be an issue to make this function async and await the timeout before proceeding?

browser/extensions/felt/content/styles/felt.css
lissyx: OOps, fixed

testing/enterprise/felt_tests.py
lissyx: Yes, if `m = None` the default path is 404, but here we set 200 and we `return` so we are fine?

browser/extensions/felt/content/styles/felt.css
lissyx: Fixed, with light/dark/HCM modes that should be handled via CSS media queries

browser/extensions/felt/content/felt.xhtml
lissyx: This is what `setUpdateState()` does, I'm going to revert the html-level changes

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,282 @@
...
+    debug(`AppUpdater: '${status}'`);
+
+    switch (status) {
+      case lazy.AppUpdater.STATUS.CHECKING:
+        this.show(".felt-updates-checking");
+        this.hide(".felt-updates-application");
+        break;
+
+      // Downloading will move progress bar from 10% to 80%
+      // downloadedBytes / totalBytes being "undefined" will move from 0 to 10%
```
Mossop: Do these numbers come from somewhere? Can we define them as constants in case we need to change them easily later?

browser/app/profile/firefox.js
Mossop: Should these go in the `firefox-enterprise.js` prefs file instead?

Mossop (approved):

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: No, they just come from my mind, as a way to show some "real looking" progress.

browser/app/profile/firefox.js
lissyx: Yes, likely

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: I'll check

lissyx: It may work indeed

lissyx: Fixed

browser/app/profile/firefox.js
```diff
@@ -153,6 +153,15 @@ pref("app.update.notifyDuringDownload", false);
   pref("app.update.auto", true);
 #endif
 
+#ifdef MOZ_ENTERPRISE
```
gcp: We should only put those in `branding/enterprise/pref` BTW.

gcp (changes requested):

browser/app/profile/firefox.js
lissyx: Oops, I forgot to remove them from here

1rneh: You resolved the previous comments but the preferencs are now both in `firefox.js` and `firefox-enterprise.js`, should they be removed here?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,290 @@
...
+  UpdateUtils: "resource://gre/modules/UpdateUtils.sys.mjs",
+  isUpdatesTesting: "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
+  setTimeout: "resource://gre/modules/Timer.sys.mjs",
+});
+
+const FELT_UPDATE_APPLY_PERCENT_INIT = 10;
+const FELT_UPDATE_APPLY_PERCENT_END = 90;
+
+// eslint-disable-next-line no-console
+const debug = msg => console.debug(`FeltUpdates: ${msg}`);
```
1rneh: Can we create a console instance here? I guess then we also don't need to disable the eslint for that line.
```
  return console.createInstance({
    prefix: "EnterpriseUpdates",
    maxLogLevelPref: lazy.EnterpriseCommon.ENTERPRISE_LOGLEVEL_PREF,
  });
});
```

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,290 @@
...
+    debug(`Loaded`);
+
+    // Make sure that we always refer to the correct document, so we can show
+    // back the login UI in any circumstance
+    this._document = doc;
+    this._errorReporter = errorReporter;
+
+    this.checkBuildID();
+
+    if (this._inited) {
```
1rneh: nit
```suggestion
    if (this._initialized) {
```

testing/enterprise/test_felt_browser_updates_basic.py
```diff
@@ -0,0 +1,120 @@
...
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_tests import FeltTests
+
+
+class FeltUpdatesBasicChecks(FeltTests):
+    def test_felt_updates_basic_checks(self):
+        super().run_felt_chrome_on_email_submit()
```
1rneh: You can call `self.run_felt_chrome_on_email_submit()` directly since this class doesn’t override that method. Same for the others usages of `super()` in this and the other tests. 

The only valid usage is `super().setup()` and `super().teardown()` in `test_felt_browser_updates_apply`.

testing/enterprise/test_felt_browser_updates_basic.py
```diff
@@ -0,0 +1,120 @@
...
+            {"app.update.background.enabled": True},
+            {"app.update.staging.enabled": True},
+        ]
+
+        driver.set_context("chrome")
+        rv = driver.execute_script(
+            """
+            console.debug(`FeltTest: checking app update prefs`);
+            console.debug(`FeltTest: arguments`, arguments);
+            for (let prefId of Object.keys(arguments)) {
```
1rneh: `arguments` is an array of the passed argument, in our case we only passed `prefs`. So here `Object.keys(arguments)` is basically `["0"]`.

I don't really unterstand what we are doing here? Could you elaborate?

testing/enterprise/test_felt_browser_updates_basic.py
```diff
@@ -0,0 +1,120 @@
...
+
+        final_rv = {}
+        for pref in rv:
+            pref_name = list(pref.keys())[0]
+            final_rv[pref_name] = pref[pref_name]
+
+        return final_rv
+
+    def run_verify_felt_app_update_url(self):
+        update_url = self.get_app_update_url(self._driver)
```
1rneh: Can we pass the `ENVIRONMENT` here instead of the driver?

browser/extensions/felt/content/felt.xhtml
```diff
@@ -115,6 +115,28 @@
...
+            class="felt-updates-checking is-hidden"
+            data-l10n-id="felt-updates-checking"
+          >
+          </span>
+          <div class="felt-updates-application is-hidden">
+            <span
+              class="felt-updates-applying"
+              data-l10n-id="felt-updates-applying"
+            >
+            </span>
```
1rneh: Can we make this a `<label for="felt-updates-progress" data-l10n-id="felt-updates-applying">` for better accessibility?

browser/extensions/felt/content/styles/felt.css
```diff
@@ -68,7 +70,8 @@ body {
   flex: 1 0 0;
 }
 
-.felt-login__wordmark {
+.felt-login__wordmark,
+.felt-updates__wordmark {
```
1rneh: This is actually not specific to `-login__` nor `-updates__`. Let's maybe rename the class to something like `firefox-enterprise-wordmark`?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,290 @@
...
+      return;
+    }
+
+    debug(`Inited`);
+    this._updatePromise = undefined;
+    this._receivedStaging = false;
+    this.setUpdateState();
+
+    // Inform early
+    this.show(".felt-updates-checking");
```
1rneh: I think this should move to `setUpdateState()` because we will always check for an update first.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,290 @@
...
+    this._appUpdater = new lazy.AppUpdater();
+    if (!this._appUpdater) {
+      throw new Error("Updater failure");
+    }
+
+    this._updaterCallback = this.appUpdaterCallback.bind(this);
+    this._appUpdater.addListener(this._updaterCallback);
+
+    debug("AppUpdater check");
+    this._updatePromise = this._appUpdater.check();
```
1rneh: There is no reason to store `_updatePromise` on `this`, or is it supposed to be used outside of this scope?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,290 @@
...
+      debug(`AppUpdater failure: ${err}`);
+      console.error(err);
+      this.updateError("contact-admin");
+    });
+
+    // Resolves when we are ready to restart?
+    this._updatePromise.then(result => {
+      this._appUpdater.removeListener(this._updaterCallback);
+      debug(`AppUpdater success: ${result}`);
+    });
```
1rneh: There is no reason to store `_updatePromise` on this, or is it supposed to be used outside of this scope?

```suggestion
    this._appUpdater.check().catch(err => {
      lazy.log.error(`Failed to check for updates: ${err}`)
      this.unsetUpdateState();
      this.updateError("contact-admin");
    }).finally(() => {
      this._appUpdater.removeListener(this._updaterCallback);
    })
```

1rneh (commented):
Thanks for putting this together - I’ve left a first round of review comments below.

1rneh (changes requested):

browser/app/profile/firefox.js
lissyx: I clearly remember removing them and making them only in `firefox-enterprise.js` so I dont know what is going on ...

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Sure

browser/app/profile/firefox.js
lissyx: Oh ok? I fixed it in the first commit but then I have another one adding it ...

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: I guess it looks weird to me that we use `self.` on higher level methods :)

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: It's all just so that we can have one list of prefs names and types that we pass to JS to collect its values and get back to python, rather than manually having a bunch of `Services.prefs.getXXXPref` calls

lissyx: On the type of `arguments` I need to verify, but the layer of translation between python and js, either at marionette or webdriver level, makes weird things hence the complication here. I'll make sure I have not regressed something though with all the rebases

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Well maybe we dont need all those debug statement anymore, when I look again at them ...

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: Did your changes land yet? I cannot find `ENVIRONMENT` there

lissyx: Ah it's `Environment`

browser/extensions/felt/content/felt.xhtml
lissyx: I did it at some point and for whatever reason I lost it, thanks

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: No specific reason, I think I was not yet sure how we could / would use it, your suggestion makes sense

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: This is the generated log:
```
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: checking app update prefs"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: arguments" ({0:{'app.update.auto':true}, 1:{'app.update.channel':""}, 2:{'app.update.checkOnlyInstance.enabled':false}, 3:{'app.update.background.enabled':true}, 4:{'app.update.staging.enabled':true}})
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: pref" "app.update.auto"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: prefType" "boolean"
[Parent 4060307: Main Thread]: D/console debug: "FeltTests: get_app_update_prefs: app.update.auto=>boolean"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: pref" "app.update.channel"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: prefType" "string"
[Parent 4060307: Main Thread]: D/console debug: "FeltTests: get_app_update_prefs: app.update.channel=>string"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: pref" "app.update.checkOnlyInstance.enabled"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: prefType" "boolean"
[Parent 4060307: Main Thread]: D/console debug: "FeltTests: get_app_update_prefs: app.update.checkOnlyInstance.enabled=>boolean"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: pref" "app.update.background.enabled"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: prefType" "boolean"
[Parent 4060307: Main Thread]: D/console debug: "FeltTests: get_app_update_prefs: app.update.background.enabled=>boolean"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: pref" "app.update.staging.enabled"
[Parent 4060307: Main Thread]: D/console debug: "FeltTest: prefType" "boolean"
[Parent 4060307: Main Thread]: D/console debug: "FeltTests: get_app_update_prefs: app.update.staging.enabled=>boolean"

Comment (1rneh) [reply testing/enterprise/test_felt_browser_updates_basic.py]: I think the readability really suffers from it and it would be more clear to call `Services.prefs.getXXXPref` directly.

File: testing/enterprise/test_felt_browser_updates_basic.py [hunk 33]
```diff
@@ -0,0 +1,123 @@
...
+                    case "string":
+                        value = Services.prefs.getStringPref(pref);
+                        break;
+
+                    case "number":
+                        const num = arguments[pref];
+                        // Check for "." because we want e.g. "1.0" to report as
+                        // a float, and Number.isInteger() and friends will
+                        // report it as an int
+                        if (String(num).includes(".")) {
```
1rneh: If we keep this logic, let's use `Number.isInteger()`

testing/enterprise/test_felt_browser_updates_basic.py
```diff
@@ -0,0 +1,123 @@
...
+                    case "boolean":
+                        value = Services.prefs.getBoolPref(pref);
+                        break;
+
+                    case "string":
+                        value = Services.prefs.getStringPref(pref);
+                        break;
+
+                    case "number":
+                        const num = arguments[pref];
```
1rneh:
```suggestion
                        const num = arguments[prefId][pref];
```

testing/enterprise/test_felt_browser_updates_basic.py
1rneh: Or are we expecting that dictionary of update preferences to grow significantly?

browser/branding/enterprise/pref/firefox-enterprise.js
```diff
@@ -13,3 +13,10 @@ pref("browser.profiles.enabled", false);
 pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
 
 pref("enterprise.loglevel", "Error");
+
+// On Enterprise we want to enforce updates so we force it
+// TODO: Should we force directly when starting FELT ?
+pref("app.update.auto", true);
+pref("app.update.checkOnlyInstance.enabled", false);
+pref("app.update.background.enabled", true);
+pref("app.update.staging.enabled", true);
```
1rneh, line 25: Have you also checked whether we need to override any `app.update.*` preference from `browser/branding/enterprise/pref/firefox-branding.js`?

1rneh (changes requested):
I built this locally to test and I made a few observations, both shown in the following screenshots:

**Expected design (theme difference irrelevant):**

Image


**Actual implementation:**

- The title and the spinner + "Checking for updates" UI is not in line vertically with the wordmark.
- No gap between spinner and string
- "Powered by Mozilla" is gone

Image

- The "Checking for updates" UI might only displayed for less than half a second before the email form is shown. So in that case (which will be probably be the most prominent on) It currently looks like a glitch.

https://github.com/user-attachments/assets/c7a5f466-3533-4be2-ae2f-c5a9fef7b84c

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: No, `Number.isInteger()` would return true for e.g. `1.0` and we dont want that there

lissyx:
> I built this locally to test and I made a few observations, both shown in the following screenshots:
> 
> **Expected design (theme difference irrelevant):**
> Image
> 
> **Actual implementation:**
> 
>     * The title and the spinner + "Checking for updates" UI is not in line vertically with the wordmark.

Will fix

> 
>     * No gap between spinner and string

There is one, it's too small maybe?

> 
>     * "Powered by Mozilla" is gone
> 

Will fix

> 
> Image
> 
>     * The "Checking for updates" UI might only displayed for less than half a second before the email form is shown. So in that case (which will be probably be the most prominent on) It currently looks like a glitch.

Unfortunately, either we introduce a useless delay here, or we directly remove the `checking`, or we live with "it looks like a glitch", because we show it during the time we do the check indeed. If the network or server is slow, it will be visible more, but if it's fast it's going to behave like that. And if we dont see it and network is not blazing fast, then people would end up on an empty screen?

> 
> 
> Screen.Recording.2026-03-02.at.14.58.20.mov

browser/branding/enterprise/pref/firefox-enterprise.js
lissyx: > Have you also checked whether we need to override any `app.update.*` preference from `browser/branding/enterprise/pref/firefox-branding.js`?

There are so many for specific edge cases, but from what I have checked, this should be enough for us for now, given the scope we have.

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: It may grow significantly yes, when you see how many prefs can control updates. The first implementation I had was a bunch of `Services.prefs.` calls, and it's also quickly easy to shoot ourselves in the foot this way

lissyx:
> > I built this locally to test and I made a few observations, both shown in the following screenshots:
> > **Expected design (theme difference irrelevant):**
> > Image
> > **Actual implementation:**
> > ```
> > * The title and the spinner + "Checking for updates" UI is not in line vertically with the wordmark.
> > ```
> 
> Will fix
> 
> > ```
> > * No gap between spinner and string
> > ```
> 
> There is one, it's too small maybe?
> 
> > ```
> > * "Powered by Mozilla" is gone
> > ```
> 
> Will fix

All three should be fixed

browser/branding/enterprise/pref/firefox-enterprise.js
```diff
@@ -13,3 +13,10 @@ pref("browser.profiles.enabled", false);
 pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
 
 pref("enterprise.loglevel", "Error");
+
+// On Enterprise we want to enforce updates so we force it
+// TODO: Should we force directly when starting FELT ?
```
1rneh: The current implementation enforces the update in felt. So do we still need to answer this question?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
...
+      Services.prefs.getStringPref(
+        "enterprise.felt.previousBuildID",
+        Services.appinfo.appBuildID
+      )
+    );
+    const actualBuildID = parseInt(Services.appinfo.appBuildID);
+
+    if (previousBuildID < actualBuildID) {
+      this.show(".felt-updates-message");
+      this.show(".felt-updates-finished");
```
1rneh: We can remove the class `is-hidden` from the embedded `.felt-updates-finished` element. Then we only need to call `this.show(".felt-updates-message");` once.
Alternatively we could also combine both into one `<div>` element:
`<div class="felt-updates-finished is-hidden" data-l10n-id="felt-updates-uptodate" />`

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
...
+  showDelayedUpdateChecking() {
+    if (this._checkingTimeout) {
+      return;
+    }
+
+    this._checkingTimeout = lazy.setTimeout(() => {
+      this.hide(".felt-updates-application");
+      this.show(".felt-updates-checking");
+      this.show(".felt-updates");
+      lazy.clearTimeout(this._checkingTimeout);
```
1rneh: No need to clear the timeout, we just finished running the callback here.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
...
+
+    if (previousBuildID < actualBuildID) {
+      this.show(".felt-updates-message");
+      this.show(".felt-updates-finished");
+    }
+
+    Services.prefs.setStringPref(
+      "enterprise.felt.previousBuildID",
+      Services.appinfo.appBuildID
+    );
```
1rneh, line 73: I guess we only need to update the preference if the previous id is smaller then the current.

```suggestion
    if (previousBuildID < actualBuildID) {
      this.show(".felt-updates-message");
      this.show(".felt-updates-finished");
      
      Services.prefs.setStringPref(
        "enterprise.felt.previousBuildID",
        Services.appinfo.appBuildID
      );
    }
```

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
...
+
+      case lazy.AppUpdater.STATUS.OTHER_INSTANCE_HANDLING_UPDATES:
+        // This should not happen, sending specific console error?
+        this.unsetUpdateState();
+        break;
+
+      case lazy.AppUpdater.STATUS.CHECKING_FAILED:
+        this.unsetUpdateState();
+        break;
+
```
1rneh, line 151: I feel like this status should be handled more critical than just unsetting the update state.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
...
+    const pane = this._document.querySelector(cls);
+    if (pane.classList.contains("is-hidden")) {
+      pane.classList.remove("is-hidden");
+    }
+  },
+
+  // Delay so that either the check is quick (fast network, no update) and there
+  // is nothing valuable to show except a quick flash, or it takes longer and
+  // there is value in informing the user.
+  showDelayedUpdateChecking() {
```
1rneh:
```suggestion
  scheduleDelayedUpdateCheckUI() {
```

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
...
+        break;
+
+      case lazy.AppUpdater.STATUS.NO_UPDATER:
+      default:
+        this.unsetUpdateState();
+        break;
+    }
+  },
+
+  hide(cls) {
```
1rneh:
```suggestion
  hide(selector) {
```

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
...
+    const pane = this._document.querySelector(cls);
+    if (pane.classList.contains("is-hidden")) {
+      return;
+    }
+    pane.classList.add("is-hidden");
+  },
+
+  show(cls) {
```
1rneh:
```suggestion
  show(selector) {
```

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
1rneh, line 3: Can we generally make sure all method and variable names are descriptive and don't hide any further actions. E.g. `checkBuildId` assumes we are checking a build id but in fact we are also toggling the visibility of the update feedback message.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,280 @@
...
+    }, 500);
+  },
+
+  setUpdateState() {
+    this.hide(".felt-login");
+    this.showDelayedUpdateChecking();
+  },
+
+  unsetUpdateState() {
+    lazy.clearTimeout(this._checkingTimeout);
```
1rneh: Multiple suggestions regarding clearing the timeout callback.

Lets outfactor it to it's own method e.g. `cancelDelayedUpdateCheckUI()` and 
1. Check if `this._checkingTimeout === null`
1. Clear the timeout
2. Set `this._checkingTimeout = null`

browser/extensions/felt/content/felt.xhtml
```diff
@@ -115,9 +115,42 @@
       <div class="felt-logo">
         <div class="felt-logo__container"></div>
       </div>
+      <div class="felt-updates is-hidden">
```
1rneh, line 118: Can we improve the class names? Examples for improvement:
- There is reason why `email` is in `felt-updates__email-pane`
- There is `felt-updates-applying` and `felt-updates-application`

1rneh (changes requested):
Since I'll be out until tomorrow afternoon, I commented on everything I noticed from another quick look.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: I can understand when we see the name, but the fact is this state happens quite easily. In fact, right now, with no update handling server side, we hit this. I agree it's an edge case, and in the end the console would not return an error on the update query.

browser/branding/enterprise/pref/firefox-enterprise.js
lissyx: I think we can live with that, I'll file a follow up to make sure we dont forget to re-analyze this.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: It should be fine yes, since we default to the current buildID when there is no existing pref value.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: `Unexpected failure while checking for an update. Please contact your administrator.` We'll show something like that, and until our stage/prod server implements update on its side, it will show

browser/extensions/felt/content/felt.xhtml
lissyx: Should be better, let me know if others are confusing

lissyx:
Capture d’écran du 2026-03-03 20-50-50
Capture d’écran du 2026-03-03 20-51-01

Capture d’écran du 2026-03-03 21-02-59
Capture d’écran du 2026-03-03 21-03-03

Capture d’écran du 2026-03-03 21-14-08
Capture d’écran du 2026-03-03 21-14-14

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+  maybeShowUpdateSuccess() {
+    const previousBuildID = parseInt(
+      Services.prefs.getStringPref(
+        "enterprise.felt.previousBuildID",
+        Services.appinfo.appBuildID
+      )
+    );
+    const actualBuildID = parseInt(Services.appinfo.appBuildID);
+
+    if (previousBuildID < actualBuildID) {
```
gcp, line 66: Given that you _only_ write `previousBuildID` inside this if, you can't actually ever get here. The comparison is always `false` because both branches will always be `Services.appinfo.appBuildID`.

testing/enterprise/test_felt_browser_updates_apply.py
```diff
@@ -0,0 +1,151 @@
...
+            """
+        )
+        self._driver.set_context("content")
+        return rv
+
+    def set_previous_buildID(self):
+        self._driver.set_context("chrome")
+        self._driver.execute_script(
+            """
+            Services.prefs.setStringPref("enterprise.felt.previousBuildID", "20250701120000");
```
gcp: You're not actually doing this in the default prefs so this ends up hiding the bug marked above where this never gets set.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+      // Downloading will move progress bar from 10% to 80%
+      // downloadedBytes / totalBytes being "undefined" will move from 0 to 10%
+      case lazy.AppUpdater.STATUS.DOWNLOADING: {
+        this.cancelDelayedUpdateCheckUI();
+        this.hide(".felt-updates-checking");
+        this.show(".felt-updates-application");
+
+        let percent = FELT_UPDATE_APPLY_PERCENT_INIT;
+        if (downloadedBytes && totalBytes) {
+          percent = parseInt((downloadedBytes / totalBytes) * 100);
```
gcp: This should be rescaled to max 80% (and +10% added to it) or your progress bar will be broken.

testing/enterprise/test_felt_browser_updates_basic.py
```diff
@@ -0,0 +1,118 @@
...
+        self.run_verify_browser_prefs()
+
+    def get_app_update_url(self, env):
+        driver = self.get_driver(env)
+        driver.set_context("chrome")
+        rv = driver.execute_script("return Services.appinfo.updateURL;")
+        driver.set_context("content")
+        return rv
+
+    def get_app_update_prefs(self, env):
```
gcp, line 32: The argument here is "env" but that's not what your callers are passing (they're passing the driver!).

This begs the question how it works at all.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+      Services.prefs.setStringPref(
+        "enterprise.felt.previousBuildID",
+        Services.appinfo.appBuildID
+      );
+    }
+  },
+
+  forceUpdateCheck() {
+    this._appUpdater = new lazy.AppUpdater();
+    if (!this._appUpdater) {
```
gcp: This can't happen - was the intent to calls some initializer on AppUpdater?

toolkit/xre/nsAppRunner.cpp
```diff
@@ -4718,6 +4721,14 @@ Maybe<ShouldNotProcessUpdatesReason> ShouldNotProcessUpdates(
     return Some(ShouldNotProcessUpdatesReason::FirstStartup);
   }
 
+#  if defined(MOZ_ENTERPRISE)
+  // Don't process updates when launcing a Browser from FELT, only Felt should
```
gcp: nit: typo

toolkit/mozapps/update/tests/marionette/manifest.toml
```diff
@@ -3,6 +3,6 @@ subsuite = "integration"
 
 ["test_no_window_update_restart.py"]
 run-if = [
-  "os == 'mac'",
+  "os == 'mac' && !enterprise", # Bug 20193408 - Test is failing and there is a "profile missing" window visible
```
gcp: nit: typo? "Missing Bug ID"

browser/extensions/felt/content/styles/felt.css
```diff
@@ -178,3 +263,26 @@ body:has(#notification-popup-box[open]) .felt:not(#notification-popup-box .felt)
...
+  font-size: 16px;
+  font-style: normal;
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
+  font-weight: 400;
+  line-height: normal;
+}
+
+.felt-updates-uptodate::before {
+  content: var(--felt-update-checkmark-img-url);
+  margin: 0 calc(var(--space-small) * 1) 0 0;
```
gcp: ```x * 1 = x```

browser/extensions/felt/content/styles/felt.css
```diff
@@ -96,6 +100,87 @@ body {
...
+.felt-updates {
+  display: flex;
+  align-items: stretch;
+  flex-direction: column;
+}
+
+.felt-updates__title {
+  color: var(--text-color, #15141a);
+  /* font-family: "Mozilla Text"; */
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
```
gcp, line 111: Is there any reason why we're not using some design tokens and hardcoding all of these? Do we need to check with Matt if this is intended?

browser/extensions/felt/content/styles/felt.css
```diff
@@ -96,6 +100,87 @@ body {
...
+
+.felt-updates-application progress {
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
+  border-radius: 35px;
+  border: none;
+  display: flex;
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
+  height: 8px;
+  /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */
+  padding-right: 89px;
```
gcp: See remark above, this is a pretty random amount no.

testing/enterprise/test_felt_browser_updates_basic.py
gcp: Shooting ourselves in the foot because we noticed too late some pref has been added/ changed and we don't set it correctly in Felt/Firefox **during** **tests** is probably preferable over finding out in the field.

That said, de-sync between how the prefs are set in reality and how they're set in the tests has already lead to a bug here.

Edit: I'm not arguing for a specific solution here, just ranting that it's worth spending a moment on this puzzle.

testing/enterprise/test_felt_browser_updates_basic.py
```diff
@@ -0,0 +1,118 @@
...
+        driver = self.get_driver(env)
+        driver.set_context("chrome")
+        rv = driver.execute_script(
+            """
+            for (let prefId of Object.keys(arguments)) {
+                const pref = Object.keys(arguments[prefId])[0];
+                const prefType = typeof arguments[prefId][pref];
+
+                let value = undefined;
+                switch (prefType) {
```
gcp: Maybe factor this out into a separate function (which should have tests of its own...)

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+
+    this._updaterCallback = this.appUpdaterCallback.bind(this);
+    this._appUpdater.addListener(this._updaterCallback);
+
+    this._appUpdater
+      .check()
+      .catch(err => {
+        this.unsetUpdateState();
+        console.error(`AppUpdater failure: ${err}`, err);
+        this.updateError("contact-admin");
```
1rneh: `this.updateError` already unsets the update state.

testing/enterprise/test_felt_browser_updates_basic.py
gcp: Can't this be simplified with https://searchfox.org/firefox-main/rev/83d1a08db47b91a4d53341a799745caac9c38bde/testing/marionette/client/marionette_driver/marionette.py#883?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+    }, 500);
+  },
+
+  cancelDelayedUpdateCheckUI() {
+    if (this._checkingTimeout !== null) {
+      lazy.clearTimeout(this._checkingTimeout);
+    }
+    this._checkingTimeout = null;
+  },
+
```
1rneh, line 236: Can we choose more descriptive method names? The current names hide some of the actions that the methods invoke. Here are some suggestions:
- `setUpdateState` to `displayUpdateState`
- `unsetUpdateState` to `displayLoginState`
- `updateFinished` to `hideUpdateState`
- `updateError` to `displayLoginStateWithUpdateError`

gcp (changes requested):

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+        this._document.querySelector("#felt-updates-progress").value = percent;
+        break;
+      }
+
+      // Staging will increment from 80% to 90%
+      case lazy.AppUpdater.STATUS.STAGING:
+        this._receivedStaging = true;
+
+        this.hide(".felt-updates-checking");
+        this.show(".felt-updates-application");
```
1rneh: `STATUS.STAGING` comes after `STATUS.DOWNLOADING` so `.felt-updates-application` should already be displayed.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: no it's just that I'm dumb

lissyx: and i wanted to be defensive

browser/extensions/felt/content/styles/felt.css
lissyx: I'm not sure, that's how the CSS provided by Figma is, I asked @1rneh about whether we should fix that indeed

browser/extensions/felt/content/styles/felt.css
lissyx: Provided by Figma as suggested

toolkit/mozapps/update/tests/marionette/manifest.toml
lissyx: Which missing bug ID ?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+    this.hide(".felt-updates-checking");
+    this.hide(".felt-updates-application");
+  },
+
+  updateError(errorMsg) {
+    this.unsetUpdateState();
+    this.hide(".felt-updates-message");
+    this._errorReporter.update("felt-updates-error-messages", errorMsg);
+  },
+
```
1rneh: `this.unsetUpdateState()` shows the login and on the next to lines we modify the login UI. Let's move `this.unsetUpdateState()` to the end.

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: I did not knew about this marionette pref handling, let's get rid of my poor code

testing/enterprise/test_felt_browser_updates_basic.py
lissyx: or kill it with fire

testing/enterprise/test_felt_browser_updates_apply.py
lissyx: That's worth. It looks like the call to `force_check_buildID` disappeared?

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: I think my CSS was broken and it made me think this code was right ...

1rneh (approved):
Thank you for providing the screenshots. The UI changes look very correct now :)

The remaining comments address descriptive naming of methods and some small issues with the UI state changes.

toolkit/mozapps/update/tests/marionette/manifest.toml
gcp: https://bugzilla.mozilla.org/show_bug.cgi?id=20193408

browser/extensions/felt/content/styles/felt.css
lissyx: And worse, this `padding-right` was in fact breaking the 100% completion of the bar:
- without it and progress.value=100:
image

- with it and progress.value=100:
image

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Yes, I was worried we may have cases where it skips a step.

browser/extensions/felt/content/styles/felt.css
1rneh: I think I know how you ended up with this (this comes from Figma):

Image

lissyx: Yes, and I was too lazy to cross-check and too dumb to realize Figma was showing me a specific state :(

toolkit/mozapps/update/tests/marionette/manifest.toml
lissyx: [2019348](https://bugzilla.mozilla.org/show_bug.cgi?id=2019348)

testing/enterprise/test_felt_browser_updates_apply.py
```diff
@@ -0,0 +1,130 @@
...
+class FeltUpdatesApplyFromFelt(FeltTests):
+    EXTRA_PREFS = {
+        "app.update.log": True,
+        "app.update.disabledForTesting": False,
+        "app.update.BITS.enabled": False,
+        "enterprise.felt_tests.is_updates_testing": True,
+    }
+
+    def setup(self):
+        print("Enabling updates")
```
gcp: FWIW all the other enterprise test code uses ```self._logger```, but this adds a bunch of (random?) ``print``.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+      // before we start those actions. For some reason, not doing this ends up
+      // in Marionette not starting the test code while the update applies.
+      await new Promise((resolve, _) => {
+        lazy.setTimeout(resolve, 5000);
+      });
+    }
+
+    this.forceUpdateCheck();
+
+    Services.obs.addObserver(this, "update-staged");
```
gcp: Do we ever need to remove these?

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -0,0 +1,285 @@
...
+ChromeUtils.defineESModuleGetters(lazy, {
+  AppUpdater: "resource://gre/modules/AppUpdater.sys.mjs",
+  isUpdatesTesting: "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
+  setTimeout: "resource://gre/modules/Timer.sys.mjs",
+  clearTimeout: "resource://gre/modules/Timer.sys.mjs",
+});
+
+const FELT_UPDATE_APPLY_PERCENT_INIT = 10;
+const FELT_UPDATE_APPLY_PERCENT_DOWNLOAD_END = 80;
+const FELT_UPDATE_APPLY_PERCENT_STAGING_END = 90;
```
gcp: We never go above 90% from this code? Maybe that was intended?

browser/branding/enterprise/pref/firefox-enterprise.js
```diff
@@ -13,3 +13,11 @@ pref("browser.profiles.enabled", false);
...
 
 pref("enterprise.loglevel", "Error");
+
+// On Enterprise we want to enforce updates so we force it
+// Bug 2020768: Should those value be set/locked at runtime by FELT only
+//              or is it fine to apply it to any enterprise build?
+pref("app.update.auto", true);
+pref("app.update.checkOnlyInstance.enabled", false);
+pref("app.update.background.enabled", true);
+pref("app.update.staging.enabled", true);
```
gcp, line 25: I think my concern remaining is if we can't do a staged update, i.e. if the Firefox Enterprise was installed system-wide in a directory where the user has no access, and the maintenance service isn't installed. We might need to check that we handle that sensibly - perhaps in a follow-up.

gcp (approved):

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Not from what I've read in other parts of the code dealing with updates, but maybe we can just remove those on xpcom-shutdown for safety?

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Yes maybe we should make the steps 90 and 100%, I thought there would be a case where we want to increment after staging but that may not be true

browser/branding/enterprise/pref/firefox-enterprise.js
lissyx: https://bugzilla.mozilla.org/show_bug.cgi?id=2022450

testing/enterprise/test_felt_browser_updates_apply.py
lissyx: Yes, I think it's because at some point I was doing those steps before marionette was setup so `self._logger` was not available, that's not the case anymore so we can use logger

PR #444, opened 2026-02-16
Bug 2010518 - Get Remote Settings URL from the console

gcp:
Form the remote settings URL based on the console URL, or get it set by the console via prefs.

### DIFF
--- a/services/settings/Utils.sys.mjs
+++ b/services/settings/Utils.sys.mjs
@@ -51,6 +51,11 @@ ChromeUtils.defineLazyGetter(lazy, "isRunningTests", () => {
 // Overriding the server URL is normally disabled on Beta and Release channels,
 // except under some conditions.
 ChromeUtils.defineLazyGetter(lazy, "allowServerURL", () => {
+  if (AppConstants.MOZ_ENTERPRISE) {
+    // Enterprise URL is set by the console.
+    return true;
+  }
+
   if (!AppConstants.RELEASE_OR_BETA) {
     // Always allow to override the server URL on Nightly/DevEdition.
     return true;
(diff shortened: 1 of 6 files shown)


gcp:
Todo: check whether this is correctly reflected in Felt AND Firefox.

fiji-flo (commented):
I'm not getting how this would be only applied in FF and not felt. But the authentication would not work in felt before the user logged in.
In general I did not plan to put this behind auth since it's no secret and that way it. Would also work in felt.

services/settings/Utils.sys.mjs
```diff
@@ -99,6 +111,18 @@ const _cdnURLs = {};
 
 export var Utils = {
   get SERVER_URL() {
+    if (AppConstants.MOZ_ENTERPRISE) {
```
lissyx: Should we enforce checking if we are FELT or Browser as well?

lissyx: Or would there be a good usecase to in fact have it in both? Now that I think of it it may be valuable.

services/settings/Utils.sys.mjs
```diff
@@ -291,6 +321,13 @@ export var Utils = {
         request.setRequestHeader(name, value);
       }
 
+      if (AppConstants.MOZ_ENTERPRISE) {
```
gcp: From discussion, this can be removed entirely.

services/settings/Utils.sys.mjs
gcp: I think we want them in both. But at some point we'll want to filter what goes in there, as Enterprise probably doesn't want to run experiments and stuff.

fiji-flo (approved):
Looks good. Note the console does not yet send the pref.

gcp:
>Looks good. Note the console does not yet send the pref.

Should be OK due to the Firefox bundle. I think :P

PR #445, opened 2026-02-16
Bug 2010518 - Derive networking endpoint prefs

gcp:
Instead of hardcoding MITM detection, captive portal, and connectivity check URLs, derive them from enterprise.console.address at startup. The defaults are cleared in enterprise builds and EnterpriseEndpoints sets them on the default pref branch so user overrides take precedence.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -1583,9 +1583,10 @@ pref("app.feedback.baseURL", "https://ideas.mozilla.org/");
 
 pref("security.certerrors.permanentOverride", true);
 pref("security.certerrors.mitm.priming.enabled", true);
-pref("security.certerrors.mitm.priming.endpoint", "https://mitmdetection.services.mozilla.com/");
 #ifdef MOZ_ENTERPRISE
 pref("security.certerrors.mitm.priming.endpoint", "");
+#else
+pref("security.certerrors.mitm.priming.endpoint", "https://mitmdetection.services.mozilla.com/");
 #endif
 pref("security.certerrors.mitm.auto_enable_enterprise_roots", true);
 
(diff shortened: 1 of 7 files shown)


browser/components/enterprise/EnterpriseEndpoints.sys.mjs
```diff
@@ -0,0 +1,49 @@
...
+    let baseURL;
+    try {
+      baseURL = new URL(consoleAddress);
+    } catch {
+      return;
+    }
+
+    const defaultBranch = Services.prefs.getDefaultBranch("");
+    for (const { pref, path } of ENDPOINT_PREFS) {
+      const url = new URL(path, baseURL).href;
```
fiji-flo, line 45: Probably something broader to decide. I we want keep the option open to move to only path based routing we might want to optionally not override the path of the `baseURL`. For that we would have to change the `path` to not start with `/` but `./`. 

Or we just rename the console address pref to console origin to be explicit. (slight preference for that)

browser/app/profile/firefox.js
```diff
@@ -1574,6 +1574,9 @@ pref("app.feedback.baseURL", "https://ideas.mozilla.org/");
 pref("security.certerrors.permanentOverride", true);
 pref("security.certerrors.mitm.priming.enabled", true);
 pref("security.certerrors.mitm.priming.endpoint", "https://mitmdetection.services.mozilla.com/");
+#ifdef MOZ_ENTERPRISE
+pref("security.certerrors.mitm.priming.endpoint", "");
```
fiji-flo, line 1587: Why not if else here? To not "change" the FF code path? It would make it easier to catch changes that would break this for us.

fiji-flo (approved):
Looks good, just some questions / nits.

browser/components/enterprise/EnterpriseEndpoints.sys.mjs
gcp: OK with renaming the pref but not right now :)

PR #450, opened 2026-02-17
Bug 1998386 - Avoid private browsing AUMID and icon for Felt window (Windows)

jonathanmendez:


widget/windows/nsWindow.cpp
```diff
@@ -1151,6 +1155,9 @@ nsresult nsWindow::Create(nsIWidget* aParent, const LayoutDeviceIntRect& aRect,
     bool usePrivateAumid =
         Preferences::GetBool("browser.privateWindowSeparation.enabled", true) &&
         aInitData.mIsPrivate &&
+#if defined(MOZ_ENTERPRISE)
+        !is_felt_ui() &&
```
lissyx, line 1161: This may be worth a comment / augmenting the one about permanent Private Browsing mode, because it feel a bit confusing at first sight

lissyx (approved):

widget/windows/nsWindow.cpp
```diff
@@ -1151,6 +1155,9 @@ nsresult nsWindow::Create(nsIWidget* aParent, const LayoutDeviceIntRect& aRect,
...
+#endif
         !StaticPrefs::browser_privatebrowsing_autostart();
     RefPtr<IPropertyStore> pPropStore;
     if (!FAILED(SHGetPropertyStoreForWindow(mWnd, IID_IPropertyStore,
```
lissyx, line 1165: This may be something we can test from python code, right? Would be nice if we could cover it in marionette

gcp: You probably need to do the same hack I did in the URL forwarding tests to check whether a window is a private window or not, but not sure there's an easy way to know what icon etc is being used?

jonathanmendez: @gcp I'm not sure which hack you're referring to?

But regardless, I've spent probably far too long today digging through win32 APIs to figure out how I would detect and test this situation. There was a "quite ugly but maybe livable" version that uses pywin32 and looks for the text "Private" in the app user model id; but I'm not sure if the pywin32 dependency is allowed, much less how to declare the dependency in the test framework. The ctypes version of it would be far too ugly I think.
As much as I actually am a fan of automated testing, I don't think this really benefits much from it, as the icon is quite noticeable to any user, and it's not the kind of code that's likely to accidentally get switched off anyway. So I'm going to go ahead and merge the change without a test, but I'm open to ideas if there's something simple enough to add.

gcp: https://github.com/mozilla/enterprise-firefox/pull/310#pullrequestreview-3717539502

https://github.com/mozilla/enterprise-firefox/pull/310/changes#diff-8e8e9cad89fc4da764b6a748ea96220e7950ae83272eb140d05cc7bad5e1994dR16

That said, I guess this doesn't help you to check the icon.

PR #425, opened 2026-02-11
Bug 2016062 - use a specific shipping-product for firefox-enterprise builds

jcristau:
Use it in target_tasks_enterprise_firefox_with_tests to avoid scheduling tons of irrelevant firefox-only tasks.

PR #454, opened 2026-02-18
Run mar-signing tasks on push to enterprise-main, and index them

jcristau:


taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -502,6 +502,9 @@ def filter(task):
         if not build_platform or not build_type:
             return True
 
+        if shippable and "enterprise" not in build_platform:
```
lissyx, line 505: This is for testing the PR @jcristau ? Otherwise I'm not sure I get the idea here

lissyx (approved):

taskcluster/gecko_taskgraph/target_tasks.py
ahal: The idea is to not run regular Firefox shippable builds. I can't think of a reason we'd want to run these on enterprise?

(This could probably be it's own commit though)

taskcluster/kinds/mar-signing-l10n/kind.yml
```diff
@@ -30,13 +27,20 @@ only-for-build-platforms:
...
+    - macosx64-enterprise-shippable/opt
+    - win64-enterprise-shippable/opt
 
 tasks:
     mar-signing-l10n:
         from-deps:
             group-by: single-with-filters
-        shipping-phase: promote
+        run-on-projects: [enterprise-firefox]
+        run-on-git-branches: [enterprise-main]
```
ahal, line 40: I assume we want these on `enterprise-release` as well?

ahal (approved):

taskcluster/gecko_taskgraph/target_tasks.py
lissyx: Ah it makes sense, my brain misparsed `not in` as `in` somehow

jcristau: This is to avoid pulling in tasks like mar-signing-win64-asan-reporter-shippable/opt (and their dependencies) that we're otherwise selecting in this commit, and don't have a shipping-product set, so aren't excluded by #425.

taskcluster/kinds/mar-signing-l10n/kind.yml
jcristau: Eventually yes, though I want to keep that separate and double check signing parameters etc before turning it on.

PR #455, opened 2026-02-18
Convert to expires_at immediately at ingestion.

gcp:
Prevent subtle API bugs mixing `expires_in` vs `expires_at`.

There are regressions with this in https://github.com/mozilla/enterprise-firefox/pull/432

### DIFF
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -184,7 +184,7 @@ export class FeltProcessParent extends JSProcessActorParent {
             Services.felt.setTokens(
               data.access_token,
               data.refresh_token,
-              data.expires_in
+              data.expires_at
             );
             break;
           }
@@ -643,7 +643,8 @@ export class FeltProcessParent extends JSProcessActorParent {
             refresh_token = "",
             expires_in = 0,
           } = message.data;
-          Services.felt.setTokens(access_token, refresh_token, expires_in);
+          const expires_at = Math.floor(Date.now() / 1000) + Number(expires_in);
+          Services.felt.setTokens(access_token, refresh_token, expires_at);
 
           // TODO: Bug 2003001 - Pass user info from Felt to Firefox to avoid network request on startup
           this.loggedInUserInfo =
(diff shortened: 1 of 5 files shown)


browser/extensions/felt/rust/nsIFelt.idl
```diff
@@ -57,7 +57,7 @@ interface nsIFelt : nsISupports {
   ACString getAccessTokenIfValid();
 
   [must_use]
-  void setTokens(in ACString access_token, in ACString refresh_token, in long long expires_in);
+  void setTokens(in ACString access_token, in ACString refresh_token, in long long expires_at);
```
lissyx, line 61: nit: maybe worth improving our documentation on the IDL?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -184,7 +184,7 @@ export class FeltProcessParent extends JSProcessActorParent {
             Services.felt.setTokens(
               data.access_token,
               data.refresh_token,
-              data.expires_in
+              data.expires_at
```
lissyx, line 187: So I assume the bug was here, but I'm not clear as of the root cause, is it because we should have updated the `expires_in` value when extracting from the browser / setting it back in felt ?

lissyx (approved):

browser/extensions/felt/rust/nsIFelt.idl
gcp: I added a comment what the format of the value is here. If you had smth else in mind let me know.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
gcp: There's no actual `expires_in` in that JSON because the other parts of this code send `expires_at`. I think you just get 0 here.

However, setTokens **was** expecting `expires_in`, so you'd have needed to do the back-conversion here even after that fix. (Or the other parts of that code should have done the conversion and then sent `expires_in` instead)

This is a huge footgun all over, and in any case `expires_in` literally stops making sense the _second_ after you receive it, so let's kill it with fire in the codebase 🔥 🔥 🔥

lissyx: Doh, I completely missed it, you're right.

PR #457, opened 2026-02-19
Set expires_at in enterprisepolicies tests

1rneh:
Regressed by #455.


lissyx (approved):
Did you caught this trying to run tests? :)

PR #459, opened 2026-02-19
Bug 2015833 - Fetch policies during engine initialization

1rneh:


Mossop (approved):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -415,14 +420,12 @@ EnterprisePoliciesManager.prototype = {
...
 
     switch (topic) {
-      case "policies-startup":
-        // Before the first set of policy callbacks runs, we must
-        // initialize the service.
-        this._initialize();
-
+      case "policies-startup": {
+        const initializedPromise = this._initialize();
+        this.spinResolve(initializedPromise);
```
jonathanmendez, line 429: I'm assuming/guessing that the reason this uses a spinResolve instead of await is that `observe` is non-async and defined in `niIObserver` so you can't change that, combined with needing `_initialize` to finish before calling `_runPoliciesCallbacks`. Is it worth adding a small comment here explaining the same?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -954,6 +995,22 @@ class RemotePoliciesProvider {
...
     }
   }
+
+  async fetchPoliciesOnStartup() {
+    if (!this._isPollingEnabled) {
+      return;
+    }
+
+    const res = await lazy.ConsoleClient.getRemotePolicies();
+    if (!res.policies) {
```
gcp, line 1016: The previous can throw I think, so you want to catch too, not just nullcheck.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -954,6 +995,22 @@ class RemotePoliciesProvider {
...
+      console.error(
+        `No policies were found in the response: ${JSON.stringify(res)}.`
+      );
+      this._failed = true;
+    } else {
+      this._policies = res.policies;
```
gcp, line 1022: Doesn't this need to do a call/trigger/notify to explicitly process the policies?

gcp (changes requested):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
gcp: ```_ingestPolicies()``` or ```triggerOnPoliciesChanges()``` ?

1rneh: Oh, yes. Good catch! I have to fetch the policies before calling https://github.com/mozilla/enterprise-firefox/blob/8c0694ea871f51bf3f01688d7cc0a47cf010e49e/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#L167

Let me verify, that that is correct and working.

jonathanmendez (approved):
I agree with gcp's feedback but I'll let him handle the followup. Otherwise looks good.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Comment added.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Verified!

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
gcp: Some background is probably in the original PR https://github.com/mozilla/enterprise-firefox/pull/321#discussion_r2720599471

gcp (approved):

PR #462, opened 2026-02-20
Bug 2018152 - Remove fallback for console

1rneh:


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -82,12 +82,13 @@ export const ConsoleClient = {
...
-      Services.prefs.getStringPref(
-        PREFS.CONSOLE_ADDRESS,
-        "https://console.enterfox.eu"
-      )
-    );
+    let consoleURI;
+    try {
+      consoleURI = Services.prefs.getStringPref(PREFS.CONSOLE_ADDRESS);
+    } catch (e) {
+      console.error("Critial misconfiguration: Missing console URI.");
```
lissyx, line 89: Since it's critical, I would assume we want to throw to make sure it's properly broken and we know it asap?

gcp (approved):

lissyx (approved):

PR #463, opened 2026-02-20
Bug 2018157 - Update default console url to Alpha console

1rneh:


browser/branding/enterprise/pref/firefox-enterprise.js
```diff
@@ -4,7 +4,10 @@
 
 /* global pref */
 
-pref("enterprise.console.address", "https://console.enterfox.eu");
+pref(
+  "enterprise.console.address",
+  "https://stage.fx-enterprise.nonprod.webservices.mozgcp.net"
```
lissyx, line 9: We probably want to rename and change it on https://github.com/mozilla-partners/sample-configs/tree/enterprise/enterfox as well

lissyx: https://github.com/mozilla-partners/sample-configs/blob/ad40e903d46dbde9598e3f052008973c4034f97a/enterfox/distribution/distribution.ini#L9

lissyx: Also making sure that the manifest repo defined at https://searchfox.org/enterprise-main/rev/0abeac0773fb124041269eb15b6cfdb21c785cf9/taskcluster/config.yml#535-541 reflects the change:

They need to point to the right commit in their respective repositories:
 - https://github.com/mozilla-partners/enterprise-manifests/blob/main/default.xml#L8
 - https://github.com/mozilla-partners/staging-enterprise-manifests/blob/main/default.xml#L8

As you can see we have two, `sample-configs` https://github.com/mozilla-partners/sample-configs/tree/enterprise as shared earlier that is the staging one shared with other projects and on which you need to PR against the `enterprise` branch, and the `enterprise-repacks` at https://github.com/mozilla-partners/enterprise-repacks that is dedicated to us so you need to PR against `main` branch there. The content is for now the same on both.

The staging repo is for running tests / PRs while the non staging repo is for actual shippable builds

lissyx: (it does not block this change)

lissyx (approved):

browser/branding/enterprise/pref/firefox-enterprise.js
1rneh: Thank you! I'm working on it. In the meantime I'm going to land this.

PR #465, opened 2026-02-20
Bug 2016195 - Register browser observers once at module level

gcp:
Register observers once on first use via a module-level flag instead. The browserObserver dispatches through gFeltProcessParentInstance (module-level singleton), so a single registration remains valid across Firefox restart cycles and JSActor pair re-creations.

### DIFF
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -67,6 +67,26 @@ function notifyFirefoxReady() {
   Services.obs.notifyObservers(null, "felt-firefox-window-ready");
 }
 
+// These observer topics relay IPC events from the Firefox subprocess back
+// through XPCOM. Their lifetime is tied to the Firefox process, not the
+// JSActor pair (which can be destroyed and re-created independently when the
+// content process hosting the SSO page is recycled). We register them once on
+// first use via gObserversRegistered and never remove them: the browserObserver
+// dispatches through gFeltProcessParentInstance (module-level), so a single
+// registration remains valid across actor re-creations and Firefox restart
+// cycles. They are cleaned up implicitly when the Felt UI process exits.
+// See browserObserver.observe() which routes all events via the singleton:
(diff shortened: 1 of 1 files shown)


gcp:
Things don't break when I log out and log in again, so I think the analysis is correct?

lissyx:
> Things don't break when I log out and log in again, so I think the analysis is correct?

Tests will say for sure.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -67,6 +67,26 @@ function notifyFirefoxReady() {
...
+// These observer topics relay IPC events from the Firefox subprocess back
+// through XPCOM. Their lifetime is tied to the Firefox process, not the
+// JSActor pair (which can be destroyed and re-created independently when the
+// content process hosting the SSO page is recycled). We register them once on
+// first use via gObserversRegistered and never remove them: the browserObserver
+// dispatches through gFeltProcessParentInstance (module-level), so a single
+// registration remains valid across actor re-creations and Firefox restart
+// cycles. They are cleaned up implicitly when the Felt UI process exits.
+// See browserObserver.observe() which routes all events via the singleton:
+// https://github.com/mozilla/enterprise-firefox/blob/3caad8cb1f33/browser/extensions/felt/content/FeltProcessParent.sys.mjs#L111-L183
```
lissyx, line 79: Ok, but since they anyway depend on the process we launched what is the added value here of having them always setup?

gcp: ...remember we got here due to a discussion about making sure they don't get added/removed one too many times!

This is simpler no?

lissyx: Ok, from that perspective, yes

lissyx: Note that https://bugzil.la/1991834 should help us as well

Mossop (approved):
This looks no worse than the current code. At some point I think we should move to a true module singleton rather than using the process parent as a singleton

lissyx (approved):

jonathanmendez (approved):

PR #471, opened 2026-02-23
 Enterprise tests refactoring

1rneh:
- Extracts enterprise badge + panel tests from signout tests
- Adds environment enum with members `FELT` and `FIREFOX` to clarify in which environment we are asserting whether the user is signed in or signed out


testing/enterprise/felt_tests.py
```diff
@@ -640,3 +640,14 @@ def run_felt_perform_sso_auth(self):
 
     def await_felt_auth_window(self):
         self._wait.until(lambda mn: len(self._driver.chrome_window_handles) == 1)
+
+    def get_whoami(self):
+        self._child_driver.set_context("chrome")
+        rv = self._child_driver.execute_script(
+            """
+            const { ConsoleClient } = ChromeUtils.importESModule("resource:///modules/enterprise/ConsoleClient.sys.mjs");
+            return ConsoleClient._get(ConsoleClient._paths.WHOAMI);
```
jonathanmendez: 1. It looks like you could use `getLoggedInUserInfo()` and avoid the call to private `_get()`?
2. How does this work out regarding async/promise? I assume this works out because the test passes, but I don't know why/how. Does `execute_script()` handle that?

testing/enterprise/test_felt_browser_signout.py
```diff
@@ -46,54 +35,20 @@ def get_private_cookies(self):
...
 
-        whoami = self.felt_whoami()
-        assert whoami["id"], "Expected user to exist"
-        assert whoami["email"], "Expected user email to exist"
-        assert whoami["picture"], "Expected user picture to exist"
-
-        # Save email for later to test that email is correctly pre-filled
-        # in test_felt_7_prefilled_email_submit
+        # Cache email in case prefilled email input field is expected
+        whoami = self.get_whoami()
```
jonathanmendez: I think that, while mildly redundant, it makes sense to keep a small part of `run_browser_ui_state_when_user_is_logged_in()` how it was before (probably renamed to something like 'assert_user_logged_in()`).

I interpret the flow of `test_browser_signout()` to be something like:
1. Setup - Sign in as usual, *do a precondition assert that we are logged in*, and remember the user credentials
2. Act - Sign out
3. Assert - Check that `whoami` is invalid while signed out. Check that the sign in window shows up and the prefilled email matches the cached credentials from setup. Check various typical sign in actions.

Crucially, the act and assert are not meaningful if we don't meet the precondition that we were logged in already.

So I would leave the original lines 57-64 in a step like `assert_user_logged_in()` instead of caching the whoami email in `run_perform_signout()`. The rest of the code that's currently in `run_browser_ui_state_when_user_is_logged_in()` is properly removed as it becomes redundant with your new test.

testing/enterprise/test_felt_browser_custom_enterprise_toolbar_button.py
```diff
@@ -0,0 +1,51 @@
...
+
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_tests import FeltTests
+
+
+class CustomEnterpriseUITests(FeltTests):
```
jonathanmendez: Maybe
```suggestion
class CustomEnterpriseBadgeTests(FeltTests):
```
? (just because "UI" feels very broad to contain many aspects, including even potentially the FELT UI)

jonathanmendez (changes requested):

testing/enterprise/felt_tests.py
1rneh: Discussed offline. I updated it to use `execute_async_script` which makes an async call to `ConsoleClient.getLoggedInUserInfo()`

testing/enterprise/test_felt_browser_signout.py
1rneh: Also discussed offline.

testing/enterprise/test_felt_browser_signout.py
```diff
@@ -46,51 +33,18 @@ def get_private_cookies(self):
...
-        assert email.get_property("textContent") == whoami["email"], (
-            "User email not correctly set"
-        )
-
-        self._child_driver.set_context("content")
-
-    def run_perform_signout(self):
-        self.open_tab_child("about:newtab")
+        self.assert_user_signed_in(env=Environment.FIREFOX)
+        # Cache email in case prefilled email input field is expected
```
jonathanmendez: nit: reword to be a bit more precise on the 'why'
```suggestion
        # Cache email for later use in prefilled email input field assertion
```

jonathanmendez (approved):

testing/enterprise/test_felt_browser_signout.py
1rneh: Done.

testing/enterprise/base_test.py
```diff
@@ -8,12 +8,18 @@
...
+from enum import Enum
 
 from marionette_driver.marionette import Marionette
 from marionette_driver.wait import Wait
 from marionette_harness import MarionetteTestCase
 
 
+class Environment(Enum):
+    FELT = "felt"
+    FIREFOX = "Firefox"
```
lissyx, line 20: If we introduce that we should update all call sites where we pass either `self._driver` or `self._child_driver`.

testing/enterprise/base_test.py
```diff
@@ -132,3 +138,39 @@ def connect_child_browser(self, capabilities=None):
...
+
+        driver.set_context("chrome")
+        try:
+            user = driver.execute_async_script(
+                """
+                const callback = arguments[arguments.length - 1];
+                const { ConsoleClient } = ChromeUtils.importESModule("resource:///modules/enterprise/ConsoleClient.sys.mjs");
+                ConsoleClient.getLoggedInUserInfo()
+                    .then(callback)
+                    .catch(err => callback({_error: String(err)}))
```
lissyx, line 155: I'm still surprised why it was not being racy before, the `_get(WHOAMI)` should have been async as well. Lucky?

lissyx (approved):

testing/enterprise/base_test.py
1rneh: I filed [Bug 2019218](https://bugzilla.mozilla.org/show_bug.cgi?id=2019218) because the scope of this test was only to extract enterprise badge + panel tests from the signout tests and that scope has already grown a bit now in this patch.

testing/enterprise/base_test.py
1rneh: Tbh I didn't check why it wasn't more racy before, but I verified that this working reliably now.

PR #472, opened 2026-02-23
Bug 2002562 - Add keyboard navigation support to enterpise badge

1rneh:


browser/base/content/navigator-toolbox.js
```diff
@@ -337,7 +337,8 @@ document.addEventListener(
         #unified-extensions-button,
         #library-button,
         #split-view-button,
-        #smartwindow-ask-button
+        #smartwindow-ask-button,
+        #enterprise-badge-toolbar-button
```
jonathanmendez: Put the changes in this file behind `#ifdef MOZ_ENTERPRISE` (which requires marking the file for preprocessing in jar.mn)

testing/enterprise/felt_tests.py
```diff
@@ -640,3 +640,14 @@ def run_felt_perform_sso_auth(self):
 
     def await_felt_auth_window(self):
         self._wait.until(lambda mn: len(self._driver.chrome_window_handles) == 1)
+
+    def get_whoami(self):
+        self._child_driver.set_context("chrome")
+        rv = self._child_driver.execute_script(
+            """
+            const { ConsoleClient } = ChromeUtils.importESModule("resource:///modules/enterprise/ConsoleClient.sys.mjs");
+            return ConsoleClient._get(ConsoleClient._paths.WHOAMI);
```
jonathanmendez: 1. It looks like you could use `getLoggedInUserInfo()` and avoid the call to private `_get()`?
2. How does this work out regarding async/promise? I assume this works out because the test passes, but I don't know why/how. Does `execute_script()` handle that?

testing/enterprise/test_felt_browser_enterprise_toolbar_button.py
```diff
@@ -0,0 +1,91 @@
...
+                return arguments[0].matches(":focus-visible");
+                """, [badge])
+        assert is_focus_visible_on_badge, "Badge is not :focus-visible"
+
+        self._logger.info("Opening enterprise panel by keypress.")
+        chrome_body.send_keys(Keys.ENTER)
+
+        self._logger.info("Checking enterprise panel")
+        self.get_elem_child("#panelUI-enterprise")
+
```
jonathanmendez, line 97: Are there missing asserts here, or is the check just whether `get_elem_child()` succeeds?

This may be a deeper review comment that the name `get_elem_child()` is misleading since it's really `get_displayed_elem_child_or_raise_if_not_displayed_or_found` (which you don't need to address, but I'm sharing to make sure I understand).

testing/enterprise/test_felt_browser_signout.py
```diff
@@ -46,51 +36,18 @@ def get_private_cookies(self):
...
 
-        whoami = self.felt_whoami()
-        assert whoami["id"], "Expected user to exist"
-        assert whoami["email"], "Expected user email to exist"
-        assert whoami["picture"], "Expected user picture to exist"
-
-        # Save email for later to test that email is correctly pre-filled
-        # in test_felt_7_prefilled_email_submit
+        # Cache email in case prefilled email input field is expected
+        whoami = self.get_whoami()
```
jonathanmendez: I think that, while mildly redundant, it makes sense to keep a small part of `run_browser_ui_state_when_user_is_logged_in()` how it was before (probably renamed to something like 'assert_user_logged_in()`).

I interpret the flow of `test_browser_signout()` to be something like:
1. Setup - Sign in as usual, *do a precondition assert that we are logged in*, and remember the user credentials
2. Act - Sign out
3. Assert - Check that `whoami` is invalid while signed out. Check that the sign in window shows up and the prefilled email matches the cached credentials from setup. Check various typical sign in actions.

Crucially, the act and assert are not meaningful if we don't meet the precondition that we were logged in already.

So I would leave the original lines 57-64 in a step like `assert_user_logged_in()` instead of caching the whoami email in `run_perform_signout()`. The rest of the code that's currently in `run_browser_ui_state_when_user_is_logged_in()` is properly removed as it becomes redundant with your new test.

jonathanmendez (changes requested):

testing/enterprise/test_felt_browser_enterprise_toolbar_button.py
1rneh: Yes, its waiting until the element becomes visible: https://searchfox.org/enterprise-main/rev/6d779fbe3a649b27248e41f53e465822295967ac/testing/enterprise/felt_tests.py#517-527,532-538

Let me add a comment.

testing/enterprise/test_felt_browser_signout.py
1rneh: Resolving this here, because it's going to be addressin in #471.

testing/enterprise/felt_tests.py
1rneh: Resolving this here, because it's going to be addressin in #471.

testing/enterprise/test_felt_browser_enterprise_toolbar_button.py
1rneh: I added it in the log information.

jonathanmendez (approved):

PR #473, opened 2026-02-23
Bug 2012458 - Add policy for configuring sidebar AI providers.

mkaply:
This also has a change to allow external icons to work in the sidebar.

I focused on adding/removing providers and removing prompts.

If we need to add prompts, it would be straightforward, but that wasn't part of the original requirements.


browser/locales/en-US/browser/policies/policies-descriptions.ftl
```diff
@@ -232,6 +232,8 @@ policy-SecurityDevices2 = Add or delete PKCS #11 modules.
 
 policy-ShowHomeButton = Show the home button on the toolbar.
 
+policy-SidebarChat = Configure providers and prompts in the chat sidebar.
```
1rneh: This needs to go into `browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl` in order for enterprise l10n to pick it up.

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -1663,6 +1663,69 @@
...
+              }
+            },
+            "RemoveAll": {
+              "type": "boolean"
+            },
+            "Remove": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "enum": ["claude", "chatgpt", "copilot", "gemini", "lechat"]
```
1rneh: For the `SyncPolicy` you feedback-ed that you preferred this to be booleans to avoid multi-select drop downs. Is this different here?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3149,6 +3151,12 @@ export var Policies = {
     },
   },
 
+  SidebarChat: {
+    onBeforeAddons(manager, param) {
+      lazy.SidebarChatPolicies.applySidebarChatPolicy(param);
+    },
```
1rneh, line 3441: Right now this policy can’t be removed live with remote policies. Can we add that capability?

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,152 @@
...
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  GenAI: "resource:///modules/GenAI.sys.mjs",
+});
+
+import { setAndLockPref } from "resource:///modules/policies/Policies.sys.mjs";
+
+export let SidebarChatPolicies = {
```
1rneh:
```suggestion
export const SidebarChatPolicies = {
```

browser/components/enterprisepolicies/schemas/policies-schema.json
1rneh: Is it intentional to not include `huggingchat` and `localhost`?

1rneh: Same for `Prompts.Remove`

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,152 @@
...
+
+      lazy.GenAI.chatProviders.set(url, engineConfig);
+    });
+  },
+
+  /**
+   * Remove prompts based on policy configuration
+   *
+   * @param {object} policy - The policy configuration
+   * @param {boolean} policy.RemoveAllPrompts - Whether to remove all prompts
```
1rneh:
```suggestion
   * @param {boolean} policy.RemoveAll - Whether to remove all prompts
```

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,152 @@
...
+
+      let promptObj;
+      try {
+        promptObj = JSON.parse(value);
+      } catch (e) {
+        continue; // skip invalid JSON
+      }
+
+      if (policy.Remove.includes(promptObj.id)) {
+        setAndLockPref(pref, "");
```
1rneh: I wasn't sure was this was doing so I tried it out. I set all `browser.ml.chat.prompts.*` prefs to `""`. The context menu and the ai shortcut menu display an empty item for each prompt that I cleared (see screenshots). Selecting one of the empty items will still send a message with the context to the respective provider just without the localized prompt message.

Image

Image


Instead of using preferences we can make use of `allowFeature` and `disallowFeature` and prevent these prompt options to be added to the menus, e.g. in `addAskChatItems` ([here](https://searchfox.org/firefox-main/rev/3c52c9c886f7b0423001cef86abfafcf97f2b90a/browser/components/genai/GenAI.sys.mjs#387-394)) or in `getContextualPrompts` ([here](https://searchfox.org/enterprise-main/rev/cd0657b0a96de7b200498c8578fb6dd135b56610/browser/components/genai/GenAI.sys.mjs#919-942)).

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * Remove AI providers based on policy configuration
+   *
+   * @param {object} policy - The policy configuration
+   * @param {boolean} policy.RemoveAll - Whether to remove all built-in providers
+   * @param {Array<string>} policy.Remove - Array of provider IDs to remove
+   */
+  removeProviders(policy) {
+    if (policy.RemoveAll) {
+      lazy.GenAI.chatProviders.clear();
```
1rneh: It would be convenient to be able to enabled/disable the chat entirely.  Maybe add another boolean `Enabled` that controls `browser.ml.chat.enabled`? 

I also noticed that setting `RemoveAll=true` without adding a custom provider will get GenAI into a broken state. In this case we could disable the chat entirely.

Example for a broken state: No providers available, clicking `Continue` throws, but otherwise no feedback to the user (see screenshot). Also the ai shortcut menu items are dead. (not in screenshot)

Image

1rneh (changes requested):
Thanks for the patch! This is my first round of review. I wasn’t previously familiar with the `GenAI` system module or the sidebar integration, so these comments are based on my current understanding.

I’ve included some questions and a few concerns regarding how prompts and providers are currently being cleared.

browser/components/enterprisepolicies/schemas/policies-schema.json
mkaply: Good point :).

My thought process here was that these work better when they match the internal names exactly.

But you're right, the UI would be better with real names and then a mapping internally.

mkaply: Huggingchat was removed from Firefox and localhost isn't available by default. I'm honestly not sure what to do there.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: Good find. I tried to mess with the prompt code a little and honestly it's terrible.

That's why I added the code to just disable the function if all prompts are removed.

I'll see if I can do better.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: Good call out.

mkaply:
I have a couple more things to fix, but this is basically a rewrite that makes things much simpler and will make things much easier to undo.

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: I think what I've done should work.

browser/components/genai/GenAI.sys.mjs
```diff
@@ -846,6 +848,10 @@ export const GenAI = {
 
     // Add remove provider option
     const popup = source === "tool" ? menu : menu.menupopup;
+    if (!popup) {
+      showItem(menu, false);
+      return;
+    }
```
1rneh, line 863: Could you elaborate on what this is doing?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3149,6 +3151,17 @@ export var Policies = {
...
+    onRemove(manager, oldParams) {
+      if (!oldParams) {
+        lazy.SidebarChatPolicies.unapplySidebarChatPolicy();
+      }
```
1rneh, line 3382: From this I take we assume `oldParams` can `undefined`, which would presuppose `param` passed to `onBeforeAddongs` was undefined as well, hence there is nothing to reset because we didn't modify anything in the first place?

From a quick glimpse through the other policies, they usually expect a boolean parameter when doing an early if statement.

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
```diff
@@ -6,4 +6,5 @@ policy-BlocklistDomainBrowsedTelemetry = Enable and configure security logging/t
 policy-DownloadTelemetry = Enable and configure security logging/telemetry when a download is triggered.
 policy-EnterpriseStorageEncryption = Enable enterprise-managed primary password for encrypted storage.
 policy-PrintPageTelemetry = Enable and configure security logging/telemetry when a page is printed.
+policy-SidebarChat = Configure providers and prompts in the chat sidebar.
```
1rneh: Is it called "chat sidebar" or "sidebar chat"?

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * Apply SidebarChat policy configuration
+   *
+   * @param {object} param - The policy parameter
+   */
+  applySidebarChatPolicy(param) {
+    if (param.Providers) {
+      this.configureProviders(param.Providers);
+      this.setDefaultProvider(param.Providers);
```
1rneh: nit:
```suggestion
      this.setDefaultProvider(param.Providers.Default);
```

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+      );
+    }
+
+    // If there are no providers, disable the chat sidebar
+    if (!providerIds.length) {
+      setAndLockPref("browser.ml.chat.enabled", false);
+    }
+  },
+
+  configurePrompts(policy) {
```
1rneh: nit:
```suggestion
  configurePrompts(prompts) {
```

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { setAndLockPref } from "resource:///modules/policies/Policies.sys.mjs";
+import {
+  CHAT_PROVIDERS_DEFAULT,
+  GenAI,
+} from "resource:///modules/GenAI.sys.mjs";
+
+export const SidebarChatPolicies = {
+  configureProviders(policy) {
```
1rneh: nit:
```suggestion
  configureProviders(providers) {
```

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
1rneh: Can we enhance the JSDoc coverage in this file?

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+    }
+  },
+
+  unapplySidebarChatPolicy() {
+    Services.prefs.clearUserPref("browser.ml.chat.providers");
+    Services.prefs.clearUserPref("browser.ml.chat.provider");
+    Services.prefs.clearUserPref("browser.ml.chat.prompts.0");
+    Services.prefs.clearUserPref("browser.ml.chat.prompts.1");
+    Services.prefs.clearUserPref("browser.ml.chat.prompts.2");
+    Services.prefs.clearUserPref("browser.ml.chat.prompts.3");
```
1rneh: I believe we need to reset the preference `browser.ml.chat.enabled` as well.

1rneh (changes requested):
Sorry, it took me a bit longer to take a look at this PR. This is my first round of review. I'm going to continue on the second round tomorrow.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+
+export const SidebarChatPolicies = {
+  configureProviders(policy) {
+    const idToName = new Map(
+      [...GenAI.chatProviders.values()].map(c => [c.id, c.name])
+    );
+
+    let providerIds = CHAT_PROVIDERS_DEFAULT.split(",").filter(
+      id => !policy.BuiltIn || policy.BuiltIn[idToName.get(id)] !== false
+    );
```
1rneh: If `policy.BuiltIn` is false, we don’t need to run `.filter()` at all:
```suggestion
    let providerIds = CHAT_PROVIDERS_DEFAULT.split(",");

    if (policy.BuiltIn) {
      providerIds = providerIds.filter(
        id => policy.BuiltIn[idToName.get(id)] !== false
      );
    }
```

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+
+    policy.Add?.forEach(engine => {
+      const url = engine.url.href || engine.url;
+      const engineConfig = { id: engine.id, name: engine.name };
+      ["iconUrl", "queryParam"].forEach(prop => {
+        if (engine[prop] !== undefined) {
+          engineConfig[prop] = engine[prop];
+        }
+      });
+      GenAI.chatProviders.set(url, engineConfig);
```
1rneh: This can override an existing provider config, e.g. two equal configs with different `queryParam` value. When we remove the policy that config remains active. Might be very unlikely and maybe ok to ignore?

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+    if (policy.BuiltIn || policy.Add?.length) {
+      Services.prefs.setStringPref(
+        "browser.ml.chat.providers",
+        providerIds.join(",")
+      );
+    }
+
+    // If there are no providers, disable the chat sidebar
+    if (!providerIds.length) {
+      setAndLockPref("browser.ml.chat.enabled", false);
```
1rneh: I just learnt about the `GenerativeAI` policy: https://firefox-admin-docs.mozilla.org/reference/policies/generativeai/
I guess instead of using `SidebarChat` and setting all `Providers.Builtin` values to `false` that policy should be set to disable the sidebar chat. Still if `SidebarChat` is used we should likely stick to preventing the broken UI and set all prefs that `GenerativeAI` sets:
https://searchfox.org/firefox-main/rev/7095e662eec49a42856e0d99f2ff96c08130119c/browser/components/enterprisepolicies/Policies.sys.mjs#1671-1673

1rneh: On a second thought, we probably don't want two policies to set the same preferences since that can cause conflict either when both are active or if only one is removed.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+  },
+
+  configurePrompts(policy) {
+    if (policy.Enabled === false) {
+      for (const pref of Services.prefs.getChildList(
+        "browser.ml.chat.prompts."
+      )) {
+        Services.prefs.setStringPref(pref, "");
+      }
+      setAndLockPref("browser.ml.chat.shortcuts", false);
```
1rneh: Why do we lock this preference?
We will need to reset it when removing the policy.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
```diff
@@ -0,0 +1,141 @@
...
+      Quiz: "quiz",
+      Proofread: "proofread",
+    };
+
+    const disabledIds = new Set(
+      Object.entries(policy.BuiltIn)
+        .filter(([, enabled]) => enabled === false)
+        .map(([name]) => PROMPT_NAME_TO_ID[name])
+        .filter(Boolean)
+    );
```
1rneh: This works fine. Since `policy.BuiltIn` only has a few entries, the multiple passes aren’t an issue. If we wanted to make it slightly more explicit, we could also write it as a single loop.
```suggestion
    const disabledIds = new Set();
    for (const [name, enabled] of Object.entries(policy.BuiltIn)) {
      if (enabled === false) {
        const id = PROMPT_NAME_TO_ID[name];
        if (id) {
          disabledIds.add(id);
        }
      }
    }
```
But I don't have a strong preference.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
1rneh: **Edit:** Taking that back since that might cause a conflict with the `GenerativeAI` policy.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
1rneh: We likely want to avoid setting that preference (see review summary).

1rneh (changes requested):
Round No 2.

Thank you for the update! Nice that we have a preference for everything that we want to control.

Two main thoughts:
- Can we use `setDefaultPref` and `unsetDefaultPref` instead of modifying pref values via APIs that write those changes to disk?
- I just learnt about the GenerativeAI policy: https://firefox-admin-docs.mozilla.org/reference/policies/generativeai/. The policy `GenerativeAI` can be used to disable the Sidebar chat entirely. In my previous review I recommended setting `browser.ml.chat.enabled=false` if all providers are removed. But I don't think this is a use case for `SidebarChat`. We probably don't want two policies to set the same preferences since that can cause conflict either when both are active or if only one is removed, hence let's avoid [](https://searchfox.org/firefox-main/rev/7095e662eec49a42856e0d99f2ff96c08130119c/browser/components/enterprisepolicies/Policies.sys.mjs#1671-1673).

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: The sidebar is broken if there are no providers at all, so we do want to turn it off. I know it conflicts with other policy, but not much else to do.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: Claude likes the way it did :)

browser/components/genai/GenAI.sys.mjs
mkaply: I added a comment. Basically If the popup fails to create, it hides the menuitem.

And the popup fails if it has no items at all.

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: I'll spend more time on this. I don' t know enough about how oldParams works.

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
mkaply: chat sidebar.

mkaply: But I named the policy the way I did because it's different sidebars.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: done

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: done

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: Claude's really good for that :)

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: Done

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: Seeing if we can do it better.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: I think it can be ignored.

mkaply:
I feel pretty good about this now. Everything has been addressed and cleaned up.

All your comments should be addressed.

browser/components/genai/GenAI.sys.mjs
1rneh: Have you checked the state if we show no items?

browser/components/enterprisepolicies/Policies.sys.mjs
1rneh: This is still unresolved. Are we expecting this can be `undefined`? If so we should also check for `if (params)` in `onBeforeAddons` for consistency. 
As said above I think it's generally used if `param` and `oldParams` are boolean values, while we are expecting an object

browser/components/enterprisepolicies/helpers/AIChatbotPolicies.sys.mjs
```diff
@@ -0,0 +1,194 @@
...
+   *   remove the prompt.
+   */
+  configurePrompts(prompts) {
+    if (prompts.Enabled === false) {
+      for (const pref of Services.prefs.getChildList(
+        "browser.ml.chat.prompts."
+      )) {
+        PoliciesUtils.setDefaultPref(pref, "");
+      }
+      PoliciesUtils.setDefaultPref("browser.ml.chat.shortcuts", false);
```
1rneh, line 104: We'll need to reset this preference on removal.

1rneh (approved):
Just some nits + the changes still need to be linted.

I'm concerned about policies interfering with each other in general. Might be something to point out in the documentation?

browser/components/genai/GenAI.sys.mjs
mkaply: Yes, the menuitem goes away completely. That's what the code does.

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: I talked to Alex like this and oldParams is the old parameters. It can be object or boolean depending on the policy.

So checking for if (oldParams) is the correct play here (And I verified that this worked).;

browser/components/enterprisepolicies/helpers/AIChatbotPolicies.sys.mjs
mkaply: I do. 

    PoliciesUtils.unsetDefaultPref("browser.ml.chat.shortcuts");

1rneh: Sorry, I missed that.

browser/components/enterprisepolicies/Policies.sys.mjs
gcp: >So checking for if (oldParams) is the correct play here

Either the check is needed, and then it should be in `onBeforeAddons` too, or it's not needed, and then it's dead code (and should be removed).

I think that's what @1rneh is saying.

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
gcp: Claude has been reading too much Haskell.

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: params is always there for onBeforeAddons. oldParams is not always there for onRemove

mkaply: params is always there for onBeforeAddons. oldParams is not always there for onRemove

mkaply: params is always there for onBeforeAddons. oldParams is not always there for onRemove

1rneh: Out of curiosity, how so? And if that's the case and `oldParams` is `undefined`, do we mind since we are not passing `oldParams` to `lazy.SidebarChatPolicies.unapplySidebarChatPolicy()` anyway.

mkaply: oldParams is just how we make the decision as to whether or not we need to unapply. If there is no oldParams, there's nothing to remove. We don't actually use the data for anything in this case.

In theory we could use it to remove the new AI engines from the list, but I don't think it's worth the trouble

browser/components/enterprisepolicies/helpers/SidebarChatPolicies.sys.mjs
mkaply: It aqreed and did a better version :)

browser/components/enterprisepolicies/helpers/AIChatbotPolicies.sys.mjs
```diff
@@ -0,0 +1,197 @@
...
+    const idToName = new Map(
+      [...GenAI.chatProviders.values()].map(c => [c.id, c.name])
+    );
+
+    let providerIds = CHAT_PROVIDERS_DEFAULT.split(",");
+    if (providers.BuiltIn) {
+      providerIds = providerIds.filter(
+        id => providers.BuiltIn[idToName.get(id)] !== false
+      );
+    }
```
1rneh, line 49: If the default provider set in `browser.ml.chat.provider` is being removed, we need to clear the pref.

browser/components/genai/chat.html
```diff
@@ -7,7 +7,7 @@
     <meta charset="utf-8" />
     <meta
       http-equiv="Content-Security-Policy"
-      content="default-src resource: chrome:; object-src 'none'; img-src data: chrome:;"
+      content="default-src resource: chrome:; object-src 'none'; img-src data: chrome: https:;"
```
gcp: `[34672] Hit MOZ_CRASH(Page chrome://browser/content/genai/chat.html must not contain a CSP with the directive img-src that includes https:) at /Users/morbo/git/firefox/dom/security/nsContentSecurityUtils.cpp:1516`

gcp (changes requested):

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -38,6 +38,73 @@
...
+                    "type": "string"
+                  }
+                },
+                "required": ["url", "name", "id"]
+              }
+            },
+            "BuiltIn": {
+              "type": "object",
+              "properties": {
+                "Anthropic Claude": { "type": "boolean" },
```
gcp, line 74: Is this supposed to disable the built-ins? Doesn't work in my testing.

browser/components/genai/chat.js
gcp, line 165: This doesn't work if the provider is selected. (See the code right above, which GitHub hides by default...)

browser/components/genai/GenAI.sys.mjs
gcp, line 1467: Note this just unhides active providers, it doesn't delete inactive ones, this leading to the bug elsewhere if the deleted provider is selected.

gcp, line 1475: Here it recovers the inactive ones but with `hidden = true`.

browser/components/enterprisepolicies/schemas/policies-schema.json
gcp: Edit: Okay, removing providers won't work due to existing bugs in the GenAI panel code.

1rneh (changes requested):

browser/components/enterprisepolicies/Policies.sys.mjs
gcp: I traced through the code (and tested a few things too). I stand behind what I said. I don't see any way that you can end up in `onRemove` with `oldParams` being falsy. Even serving an empty policy `"policies": {"AIChatbot": {}}` will cause this to be truthy (empty object) in `onBeforeAddons`, and this means `onRemove` will also be passed a truthy value. If you pass an invalid value, the validator will return `undefined`, but then `onBeforeAddons` doesn't get called, and `_previousPolicies[policyName]`, which is what feeds `oldParams`, won't be updated.

So the if here is dead code. In order for `onRemove` to be called, there has to have been _something_ valid in the parameters, even an empty object. But that's truthy!

(Note that every single other policy in this file follows the same pattern. Either params can be missing and both guards are there, or none are)

gcp (changes requested):
The assertion failure looks problematic. I assume that guard is there for good reasons, and it will block debug builds.

browser/components/genai/chat.html
gcp: Let's change this to moz-remote-image and use this whitelist, because other AI stuff is already using it? 

https://searchfox.org/firefox-main/rev/34fc93ba6c4d8f57cd154677f0c4cd9907fd1339/dom/security/nsContentSecurityUtils.cpp#1330

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -192,6 +193,17 @@ export var Policies = {
     },
   },
 
+  AIChatbot: {
+    onBeforeAddons(manager, param) {
+      lazy.AIChatbotPolicies.applyAIChatbotPolicy(param);
+    },
+    onRemove(_manager, oldParams) {
+      if (oldParams) {
```
gcp:
```suggestion
```

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -192,6 +193,17 @@ export var Policies = {
...
+        lazy.AIChatbotPolicies.unapplyAIChatbotPolicy();
+      }
```
gcp:
```suggestion
```

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -192,6 +193,17 @@ export var Policies = {
     },
   },
 
+  AIChatbot: {
+    onBeforeAddons(manager, param) {
+      lazy.AIChatbotPolicies.applyAIChatbotPolicy(param);
+    },
+    onRemove(_manager, oldParams) {
+      if (oldParams) {
+        lazy.AIChatbotPolicies.unapplyAIChatbotPolicy();
```
gcp:
```suggestion
      lazy.AIChatbotPolicies.unapplyAIChatbotPolicy();
```

gcp (changes requested):

browser/components/enterprisepolicies/schemas/policies-schema.json
mkaply: Removing providers is working for me. Were just changing the list in prefs (which does work)

It doesn't work if the panel is displayed. You have to close and reopen.,

browser/components/genai/chat.html
mkaply: moz-remote-image isn't a simple switch because it's an API that requires we query for the appropriate size,. I'll see what I can do.,

browser/components/genai/chat.js
mkaply: I don't understand what you mean. What are you seeing in the UI?

browser/components/genai/GenAI.sys.mjs
mkaply: We're not using the hidden property for our work. 

What are you seeing that is causing a problem?

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: Fixed

browser/components/genai/chat.html
mkaply: I'm wrong. It was pretty simple :)

browser/components/enterprisepolicies/schemas/policies-schema.json
gcp: Add a provider. Select it. Remove the policy. You can open and close the tab as much as you want, the provider stays there. I already pointed out why this happens in the code, it quite literally doesn't have any code to delete things!

The code doesn't even try to hide the not-deleted provider if it's selected. I imagine they didn't deal with it because it gets awkward when the provider is deleted from underneath the user (but that's a realistic use case for us though). It's also not an issue if you don't have live policy changes like we do.

browser/components/genai/GenAI.sys.mjs
gcp: >We're not using the hidden property for our work.

Then how do you expect deleting things to work? The code does not delete anything.

mkaply: All we do is modify the list of available providers in a preference. we don't actually change the providers in genai except for when we need to add new ones.

there's a built-in list that gets overridden by a preference.

browser/components/enterprisepolicies/schemas/policies-schema.json
mkaply: Ah, that's a different problem. The selected provider is an index into the array of providers independent of the list.

I'll fix that.

gcp:
r+ but going to land through #623 as I needed to rebase to test this.

PR #476, opened 2026-02-25
Enterprise release merge 20260225

jonathanmendez:


gcp:
Will obsolete with #485

PR #479, opened 2026-02-26
Bug 2019565 - Set enterprise.console.address only once in Firefox

1rneh:


lissyx (approved):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -264,11 +263,6 @@ export class FeltProcessParent extends JSProcessActorParent {
   }
 
   sendPrefsToFirefox() {
-    Services.felt.sendStringPreference(
-      lazy.CONSOLE_ADDRESS_PREF,
-      lazy.ConsoleClient.consoleBaseURI
-    );
```
lissyx, line 270: So as I was sharing on matrix, this is likely making us miss https://searchfox.org/enterprise-main/rev/f1803e33aa50eee6e3b1f4d0c5c460045cdefbd3/browser/extensions/felt/rust/src/client.rs#377-379

PR #481, opened 2026-02-27
Add fr locales to stageMozGCP repack

1rneh:


jonathanmendez (approved):
There are many failures in treeherder that I could guess are related to the DO NOT LAND commit. But maybe worth keeping an eye on.

jonathanmendez:
> There are many failures in treeherder that I could guess are related to the DO NOT LAND commit. But maybe worth keeping an eye on.

Having looked at them more carefully, my diagnosis seems correct. The failing builds are ones that we don't normally build even in CI/non-PR builds.

PR #483, opened 2026-02-27
Bug 2019920 - use release-signing instead of nightly-signing on the enterprise-release branch

jcristau:
Add the ability for `get_signing_type_from_project` to also take `head_ref` in consideration instead of just the project.


ahal (approved):
Thanks, makes sense to me!

gcp (approved):

PR #485, opened 2026-03-01
Enterprise release merge 20260301

gcp:
Merge current Firefox beta (149.0b3) into enterprise-release.

Resolves one conflict in `browser/config/version_display.txt` (kept `149.0b3` from beta).


jonathanmendez (changes requested):
Please do the following:
- [x] Make sure to update browser/locales/enterprise-l10n-changesets.json to the same commit as browser/locales/l10n-changesets.json

jonathanmendez (approved):

browser/config/version_display.txt
```diff
@@ -1 +1 @@
-149.0
```
lissyx, line 1: I dont think we want this, since everything else on `enterprise-release` is moved already to `149.0` ?

gcp: I absolutely want this to make sure the Firefox version is correctly reflected. I'd rather fix the tests.

lissyx: OK, but last time there was a discussion we said we wanted this branch to behave as release the whole time

lissyx: We need to keep in sync (I believe) as well `browser/config/version.txt` and `config/milestone.txt` which are both at `149.0` due to https://github.com/mozilla/enterprise-firefox/commit/cdcaa4016b3a6455c461eb75db2171f1c8b18570. I dont know what is going to happen if we desync all three.

gcp: Behave as release yes, but I'd like the version number to be "true" so we can spot if we make any mistakes there or more easily keep track what Firefox our releases are based off of. I'll try to keep those in sync.

gcp: @lissyx it looks like the beta tree has the same versioning as we have, so it should be fine, see for example: https://github.com/mozilla-firefox/firefox/tree/beta/browser/config

lissyx: Perfect, if beta does it, we can :)

PR #490, opened 2026-03-02
Fix test_firefox_start version comparison to use display version

gcp:


lissyx (approved):
Maybe leave a comment as why we use this one

PR #493, opened 2026-03-03
Bug 2012452 - Users are informed in case of an authentication failure on FELT

jporter-dev:
## Summary

Updates error report elements in FELT to use `moz-message-bar`, which requires updating the Fluent stringbundles to separate the heading into `.heading`. Error details are added using the `message` slot. 

Also adds a new `No network connection` message bar to explicitly show a different, simpler "user advice" message when no network connectivity is detected (`dnsNotFound2` and `netOffline`).

## Links

- [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2012452)
- [Figma](https://www.figma.com/design/xYH6L8QrDz0Q16XFvdo7We/%F0%9F%A6%8A-Client-Design?node-id=1444-9656&t=vdk5gd3mXj7dgHPO-0)

## Screenshots

image

_No network connected (`dnsNotFound2`)_

image

_Regular NSS error (`netReset`)_

## Testing

1. `mach build`
2. `mach run`
3. Disable Wifi or point `enterprise.console.address` to invalid host
4. Attempt login

### DIFF
--- a/browser/extensions/felt/content/felt.xhtml
+++ b/browser/extensions/felt/content/felt.xhtml
@@ -179,23 +179,43 @@
           >
           </span>
         </div>
-        <div class="felt-browser-error is-hidden">
-          <div
+        <div class="felt-browser-error">
+          <moz-message-bar
             class="felt-browser-error-multiple-crashes is-hidden"
-            data-l10n-id="felt-browser-error-multiple-crashes"
-          ></div>
-          <div class="felt-browser-error-connection is-hidden">
-            <span data-l10n-id="felt-browser-error-connection"></span>
-            <span class="felt-browser-error-details"></span>
-          </div>
-          <div
+            data-l10n-id="felt-browser-error-multiple-crashes2"
+            type="error"
+            dismissable="true"
+          ></moz-message-bar>
+          <moz-message-bar
+            class="felt-browser-error-no-network is-hidden"
+            data-l10n-id="felt-browser-error-no-network"
+            type="error"
+            dismissable="true"
+          >
+            <span slot="message" class="felt-browser-error-details"></span>
+          </moz-message-bar>
+          <moz-message-bar
+            class="felt-browser-error-connection is-hidden"
+            data-l10n-id="felt-browser-error-connection2"
+            type="error"
+            dismissable="true"
+          >
+            <span slot="message" class="felt-browser-error-details"></span>
+          </moz-message-bar>
+          <moz-message-bar
             class="felt-browser-error-sso-timeout is-hidden"
             data-l10n-id="felt-browser-error-sso-timeout"
-          ></div>
-          <div class="felt-updates-error-messages is-hidden">
-            <span data-l10n-id="felt-updates-error-messages"></span>
-            <span class="felt-browser-error-details"></span>
-          </div>
+            type="error"
+            dismissable="true"
+          ></moz-message-bar>
+          <moz-message-bar
+            class="felt-updates-error-messages is-hidden"
+            data-l10n-id="felt-error-updates"
+            type="error"
+            dismissable="true"
+          >
+            <span slot="message" class="felt-browser-error-details"></span>
+          </moz-message-bar>
         </div>
         <span
           class="felt-powered-by text-deemphasized"
--- a/browser/extensions/felt/content/felt.xhtml
+++ b/browser/extensions/felt/content/felt.xhtml
@@ -179,7 +179,7 @@
           >
           </span>
         </div>
-        <div class="felt-browser-error">
+        <div class="felt-browser-error is-hidden">
           <moz-message-bar
             class="felt-browser-error-multiple-crashes is-hidden"
             data-l10n-id="felt-browser-error-multiple-crashes2"
@@ -204,7 +204,7 @@
           </moz-message-bar>
           <moz-message-bar
             class="felt-browser-error-sso-timeout is-hidden"
-            data-l10n-id="felt-browser-error-sso-timeout"
+            data-l10n-id="felt-browser-error-sso-timeout2"
             type="error"
             dismissable="true"
           ></moz-message-bar>
--- a/browser/extensions/felt/content/styles/felt.css
+++ b/browser/extensions/felt/content/styles/felt.css
@@ -246,19 +246,14 @@ body:has(#notification-popup-box[open]) .felt:not(#notification-popup-box .felt)
 .felt-browser-error {
   position: absolute;
   /* stylelint-disable-next-line */
-  bottom: 32px;
+  top: 16px;
   /* stylelint-disable-next-line */
-  left: 8px;
-  font-size: medium;
-  color: var(--text-color-error);
-  font-weight: var(--font-weight-bold);
-}
-
-.felt-browser-error-details {
-  display: block;
-  font-weight: normal;
-  font-size: small;
-  margin-top: var(--space-xsmall);
+  right: 16px;
+  /* stylelint-disable-next-line */
+  left: 16px;
+  display: flex;
+  flex-direction: column;
+  gap: var(--space-small);
 }
 
 .felt-updates-message {
--- a/browser/extensions/felt/content/window.js
+++ b/browser/extensions/felt/content/window.js
@@ -33,6 +33,11 @@ const ErrorReport = {
         "chrome://global/locale/appstrings.properties"
       );
     });
+
+    this._wrapper.addEventListener("message-bar:user-dismissed", e => {
+      e.preventDefault();
+      e.target.classList.add("is-hidden");
+    });
   },
 
   reset() {
@@ -43,10 +48,9 @@ const ErrorReport = {
       return;
     }
     this._wrapper.classList.add("is-hidden");
-    const errors = this._wrapper.querySelectorAll(
-      ".felt-browser-error > div:not(.is-hidden)"
-    );
-    errors.forEach(e => e.classList.add("is-hidden"));
+    for (const bar of this._wrapper.querySelectorAll("moz-message-bar")) {
+      bar.classList.add("is-hidden");
+    }
   },
 
   async update(errorType, details = null, cause = null) {
@@ -93,14 +97,27 @@ const ErrorReport = {
 };
 
 async function connectToConsole(email) {
-  ErrorReport.reset();
-
   let posture;
   try {
     posture = await lazy.ConsoleClient.sendDevicePosture();
   } catch (err) {
     console.error(`FeltExtension: Failed to connect to console: ${err}`);
-    ErrorReport.update("felt-browser-error-connection", err.message, err.cause);
+
+    // Show simpler "No Network Connection" only for truly offline scenarios
+    // netOffline for offline mode, dnsNotFound2 for actual network disconnect
+    const NETWORK_ERRORS = new Set(["netOffline", "dnsNotFound2"]);
+    if (NETWORK_ERRORS.has(err.message)) {
+      ErrorReport.update(
+        "felt-browser-error-no-network",
+        "no-network-connection"
+      );
+    } else {
+      ErrorReport.update(
+        "felt-browser-error-connection",
+        err.message,
+        err.cause
+      );
+    }
     return;
   }
 
@@ -272,6 +289,7 @@ async function connectToConsole(email) {
     Ci.nsIWebProgress.NOTIFY_STATE_NETWORK | Ci.nsIWebProgress.NOTIFY_LOCATION
   );
 
+  ErrorReport.reset();
   document.querySelector(".felt-login__email-pane").classList.add("is-hidden");
   document.querySelector(".felt-login__sso").classList.remove("is-hidden");
 
--- a/browser/locales/en-US/browser/enterprise/felt.ftl
+++ b/browser/locales/en-US/browser/enterprise/felt.ftl
@@ -22,13 +22,22 @@ felt-version =
 
 ## Error details when launching the browser crashes
 
-felt-browser-error-multiple-crashes = { -brand-short-name } crashed multiple times.
-felt-browser-error-connection = Unable to connect to the console. Please contact your administrator.
-felt-browser-error-sso-timeout = Sign-in timed out. Please try again, or contact your administrator if the problem persists.
+felt-browser-error-sso-timeout2 =
+    .heading = Sign-in timed out. Please try again, or contact your administrator if the problem persists.
+felt-browser-error-multiple-crashes2 =
+    .heading = { -brand-short-name } crashed multiple times.
+
+## Network error headings
+
+felt-browser-error-connection2 =
+    .heading = Unable to connect. Please contact your administrator.
+felt-browser-error-no-network =
+    .heading = No network connection
 
 ## Network error details.
 
 felt-error-network = Unknown network error
+felt-error-no-network-connection = Please check your internet connection and try again.
 
 ## Updates messages and related errors messages
 
@@ -36,7 +45,8 @@ felt-updates-title = Good morning
 felt-updates-checking = Checking for updates…
 felt-updates-application = Applying updates…
 felt-updates-uptodate = { -brand-short-name } is up to date
-felt-updates-error-messages = An error occurred while applying updates…
+felt-error-updates =
+    .heading = An error occurred while applying updates…
 felt-error-contact-admin = Please contact your administrator.
 felt-error-unsupported-system-contact-admin = Your current system does not support this new version. Please contact your administrator.
 felt-error-checking-failed-contact-admin = Unexpected failure while checking for an update. Please contact your administrator.
--- a/browser/locales/en-US/browser/enterprise/felt.ftl
+++ b/browser/locales/en-US/browser/enterprise/felt.ftl
@@ -23,9 +23,10 @@ felt-version =
 ## Error details when launching the browser crashes
 
 felt-browser-error-sso-timeout2 =
-    .heading = Sign-in timed out. Please try again, or contact your administrator if the problem persists.
+    .heading = Sign-in timed out
+    .message = Please try again, or contact your administrator if the problem persists.
 felt-browser-error-multiple-crashes2 =
-    .heading = { -brand-short-name } crashed multiple times.
+    .heading = { -brand-short-name } crashed multiple times
 
 ## Network error headings
 
--- a/testing/enterprise/test_felt_browser_console_error.py
+++ b/testing/enterprise/test_felt_browser_console_error.py
@@ -17,27 +17,70 @@ class FeltConsoleError(FeltTestsBase):
             self._manually_closed_child = True
         return super().teardown()
 
-    def connection_error_test(self, console_addr, error_msg):
+    def check_error_bar_message(
+        self,
+        console_addr,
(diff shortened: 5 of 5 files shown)


browser/locales/en-US/browser/enterprise/felt.ftl
```diff
@@ -20,8 +20,13 @@ felt-version =
...
 
-felt-browser-error-multiple-crashes = { -brand-short-name } crashed multiple times.
-felt-browser-error-connection = Unable to connect to the console. Please contact your administrator.
+felt-browser-error-multiple-crashes =
+    .heading = { -brand-short-name } crashed multiple times.
+felt-browser-error-network =
+    .heading = No network connection
+    .message = Please check your internet connection and try again.
+felt-browser-error-connection =
+    .heading = Unable to connect. Please contact your administrator.
```
lissyx: We may want to check with l10n people, but we have to update/change the l10n identifier when we perform any string change. I dont know how adding `.heading` here accounts for.

@eerii would you know? is it OK to just add it like that or should we change the l10n id ?

eerii: I think you might have mentioned the wrong person :) I work on some patches in Firefox but mainly on MathML, I don't know much about localization or enterprise

lissyx: Oops.

lissyx: Yes that should have been @eemeli

jporter-dev: > In case of [Fluent](https://firefox-source-docs.mozilla.org/l10n/overview.html#fluent), any changes to the structure of the message such as adding/removing attributes also requires an update of the ID.

Found this [here](https://firefox-source-docs.mozilla.org/l10n/overview.html#string-updates)

lissyx: > > In case of [Fluent](https://firefox-source-docs.mozilla.org/l10n/overview.html#fluent), any changes to the structure of the message such as adding/removing attributes also requires an update of the ID.
> 
> Found this [here](https://firefox-source-docs.mozilla.org/l10n/overview.html#string-updates)

We would fall into the "adding an attribute" of fluent?

eemeli: > We would fall into the "adding an attribute" of fluent?

Yes, moving a pattern from a message's value to an attribute like `.heading` counts as a change to the structure of the message.

jporter-dev:
Sorted out the marionette failures and l10n conversation

- Bumped the two l10n identifiers where the structure was changed
- Updated the tests to use the `heading` attr since the `moz-message-bar` uses shadow dom for its contents
- Added test case for simpler "No Internet Connection" scenario based on `netOffline` and `dnsNotFound2`

browser/extensions/felt/content/window.js
```diff
@@ -99,7 +103,16 @@ async function connectToConsole(email) {
     posture = await lazy.ConsoleClient.sendDevicePosture();
   } catch (err) {
     console.error(`FeltExtension: Failed to connect to console: ${err}`);
-    ErrorReport.update("felt-browser-error-connection", err.message, err.cause);
+    const NETWORK_ERRORS = new Set(["netOffline", "dnsNotFound2"]);
```
gcp, line 108: Lots of context here: https://github.com/mozilla/enterprise-firefox/pull/337/changes#diff-5245eb3c71a276182b4c2046dfa9caac4683fb44514e2ab52dac85bbb5b98002R248

Then we had: https://github.com/mozilla/enterprise-firefox/pull/358#discussion_r2770575953

Which lead to filing: https://bugzilla.mozilla.org/show_bug.cgi?id=2015092

(My main observation here is I'm not sure where you got that those are the 2 errors you want to handle)

browser/locales/en-US/browser/enterprise/felt.ftl
```diff
@@ -20,8 +20,13 @@ felt-version =
 
 ## Error details when launching the browser crashes
 
-felt-browser-error-multiple-crashes = { -brand-short-name } crashed multiple times.
-felt-browser-error-connection = Unable to connect to the console. Please contact your administrator.
+felt-browser-error-multiple-crashes2 =
+    .heading = { -brand-short-name } crashed multiple times.
+felt-browser-error-connection2 =
+    .heading = Unable to connect. Please contact your administrator.
+felt-browser-error-no-network =
```
gcp, line 35: What's the difference between this and the previous one?

gcp (commented):
I have questions :P

Edit: Answered them myself, I'm good with this.

browser/locales/en-US/browser/enterprise/felt.ftl
gcp: Ah, got it, pure network errors vs other errors like cert expired.

browser/extensions/felt/content/window.js
gcp: Actually, given the current state of that function, that does seem reasonable.

browser/extensions/felt/content/styles/felt.css
```diff
@@ -164,17 +164,9 @@ body:has(#notification-popup-box[open]) .felt:not(#notification-popup-box .felt)
...
-}
-
-.felt-browser-error-details {
-  display: block;
-  font-weight: normal;
-  font-size: small;
-  margin-top: var(--space-xsmall);
+  right: 16px;
+  /* stylelint-disable-next-line */
+  left: 16px;
```
lissyx, line 253: nit: since you're changing this, for ease of future work, can you change the comments to
> /* stylelint-disable-next-line stylelint-plugin-mozilla/use-design-tokens */

I feel it will be easier to spot those designs tokens we may have to fix in the future

browser/extensions/felt/content/felt.xhtml
```diff
@@ -136,15 +136,30 @@
...
-            <span data-l10n-id="felt-browser-error-connection"></span>
-            <span class="felt-browser-error-details"></span>
-          </div>
+        <div class="felt-browser-error">
+          <moz-message-bar
+            class="felt-browser-error-multiple-crashes"
+            data-l10n-id="felt-browser-error-multiple-crashes2"
+            type="error"
+            dismissable="true"
+            hidden="true"
```
lissyx: cannot we keep the `is-hidden` class out of consistency with the rest?

browser/extensions/felt/content/window.js
lissyx, line 108: nit: It took me a bit of time as well even with the above comment, so given we already have a list hardcoded here https://searchfox.org/enterprise-main/rev/2a68f0a8c5ba115460a1f261215ece61c7096a41/browser/components/enterprise/modules/ConsoleClient.sys.mjs#267-272,274 , maybe it is worth a comment that you select those two to report `no network` specific error vs `connection failure` ones.

browser/extensions/felt/content/window.js
```diff
@@ -32,6 +32,11 @@ const ErrorReport = {
         "chrome://global/locale/appstrings.properties"
       );
     });
+
+    this._wrapper.addEventListener("message-bar:user-dismissed", e => {
+      e.preventDefault();
+      this.reset();
```
1rneh: It's possible we are showing multiple message bars. This will reset the error reporter and hide all although the user is just dismissing one. Let's hide the event target.

browser/extensions/felt/content/styles/felt.css
jporter-dev: Sure, I can update all 3 comments. Or is this something I can easily fix now by using the proper design token?

lissyx: I defer to @1rneh for that, but I assume it's a non trivial refactoring, so not for now

browser/extensions/felt/content/window.js
jporter-dev: It did seem to cause a good bit of confusion, I can add a comment and rename the variable for clarification since it is not immediately clear why these were chosen from the other NSS errors.

1rneh (changes requested):
Thank you for working on this!

It would be very helpful if you could provide a screenshot showcasing the new UI state and a link to the Figma design?

After going over your changes, we might want to revisit the design decisions again. More explanation in my comment below.

browser/extensions/felt/content/window.js
1rneh: @jporter-dev Could you provide a link to the Figma design?

So my assumption was that the new design is replacing our current rather hacky error reporting. But the html and JS here is keeping the error details (`.felt-browser-error-details`), which we are not populating correctly anymore because `err.cause` and `err.message` get dropped for most network errors. 

With the implementation from this PR, we are loosing the details of the network errors. I'm not sure if that is intentional? @gcp @lissyx 

Currently we re-use six strings from `m-c`
```
const networkErrors = {
        [Cr.NS_ERROR_UNKNOWN_HOST]: "dnsNotFound2",
        [Cr.NS_ERROR_CONNECTION_REFUSED]: "connectionFailure",
        [Cr.NS_ERROR_NET_TIMEOUT]: "netTimeout",
        [Cr.NS_ERROR_NET_RESET]: "netReset", 
        [Cr.NS_ERROR_NET_INTERRUPT]: "netInterrupt",
        [Cr.NS_ERROR_OFFLINE]: "netOffline",
      };
```
and provide a generic fluent one: `felt-error-network = Unknown network error`. 

We will need to pass `err.message` and `err.cause` because in `err.message` will be the hint which string to use (it's a stringbundle id or the string "network" if we are using the generic fluent error message). `err.cause` contains the console address which gets passed to most stringbundle strings.

While I'm writing this out, I'm well aware this is a painful and hacky state right now.

1rneh (commented):
Ups, for a second I thought I lost the comment. Luckily I didn't.

browser/extensions/felt/content/window.js
jporter-dev: Here is the Figma: https://www.figma.com/design/xYH6L8QrDz0Q16XFvdo7We/%F0%9F%A6%8A-Client-Design?node-id=1444-9656

Right now I'm only using `netOffline` and `dnsNotFound2` to display the simpler "No Internet Connection" error bar. I included `dnsNotFound2` as that is the error that came through when I actually toggled my internet offline on my laptop, and from what I understand `netOffline` is for when "offline mode" is enabled. This is basically for the top design in Figma, the simpler "no internet, please connect" message.

The other 4 errors should come through as they were with the `err.cause` on line 110.

browser/extensions/felt/content/felt.xhtml
jporter-dev: Sure I can revert this back to the class-based approach instead of the widget attribute

jporter-dev:
Some screenshots of different error scenarios. All scenarios except `No network connection` (`dnsNotFound2`/`netOffline`) use the `err.message`. 

no-network-connection-netOffline-dnsError2

_No Network Connected_

unknown-network-error

_Unknown Network Error_

network-error-netReset

_Network Connection Reset_

other-network-error

_Non-NSS network error_

network-error-dnsNotFound2

_Reverting the `netOffline`/`dnsNotFound2` exception_

jporter-dev:
Per Matt's comments, we should go with the simpler advice-driven message for scenarios where a user does not need the extended details to act (first image in comment below). I've updated the code to follow the same l10n path as other messages with the override added in `felt.ftl`. Please let me know if there is a different way of handling this, but it felt better than my original implementation.

image

jporter-dev:
Rebasing pulled in the updater errors and uptodate messages. I refactored some things to fit those in with the message handling, but needs re-review just to make sure all got handled properly along with the change to the no network message. @lissyx @1rneh

browser/extensions/felt/content/felt.xhtml
```diff
@@ -172,26 +172,48 @@
...
           >
-          </span>
-        </div>
-        <div class="felt-browser-error is-hidden">
-          <div
+            <span
+              slot="message"
+              data-l10n-id="felt-updates-uptodate"
+              class="felt-updates-uptodate"
+            ></span>
```
1rneh: Is the "Firefox Enterprise is up to date" now part of the message bar? I assumed this is the design for updates: https://www.figma.com/board/B81rTo6MEtbegmUzor6qSE/FELT-updating-and-Restart-Notification-policy?node-id=0-1&p=f&t=8bU7yyOP6NLeYrqk-0

browser/extensions/felt/content/window.js
```diff
@@ -100,7 +104,23 @@ async function connectToConsole(email) {
...
-    ErrorReport.update("felt-browser-error-connection", err.message, err.cause);
+
+    // Show simpler "No Network Connection" only for truly offline scenarios
+    // netOffline for offline mode, dnsNotFound2 for actual network disconnect
+    const NETWORK_ERRORS = new Set(["netOffline", "dnsNotFound2"]);
+    if (NETWORK_ERRORS.has(err.message)) {
+      ErrorReport.update(
+        "felt-browser-error-no-network",
+        "no-network-connection",
+        err.cause
```
1rneh: Do we still need to pass `err.cause`. From my understanding we don't surface the console address in these cases since if we don't use the strings `netOffline` and `dnsNotFound2` anymore?

browser/locales/en-US/browser/enterprise/felt.ftl
```diff
@@ -22,20 +22,29 @@ felt-version =
...
+felt-error-no-network-connection = Please check your internet connection and try again.
 
 ## Updates messages and related errors messages
 
 felt-updates-title = Good morning
 felt-updates-checking = Checking for updates…
 felt-updates-application = Applying updates…
 felt-updates-uptodate = { -brand-short-name } is up to date
-felt-updates-error-messages = An error occurred while applying updates…
+felt-updates-error-messages =
```
1rneh: String changes require changing the fluent id: https://firefox-source-docs.mozilla.org/l10n/fluent/review.html#changes-to-existing-messages
```suggestion
felt-error-updates =
```

testing/enterprise/test_felt_browser_console_error.py
```diff
@@ -17,27 +17,40 @@ def teardown(self):
...
         self._driver.set_context("content")
 
+    def connection_error_test(self, console_addr, error_msg=None):
+        return self.error_bar_test(
+            console_addr,
+            ".felt-browser-error-connection",
+            "Unable to connect",
+            error_msg,
+        )
+
```
1rneh: Could we get rid of `connection_error_test` and have consumers call `error_bar_test` directly for better readability?

testing/enterprise/test_felt_browser_console_error.py
```diff
@@ -17,27 +17,40 @@ def teardown(self):
...
     def test_felt_00_connection_error_fluent(self):
         return self.connection_error_test("http://127.0.0.1:1", "Unknown network error")
 
     def test_felt_01_connection_error_bundle(self):
```
1rneh: I know you are only picking up existing tests, but could you help improve the method names to be more descriptive? Right now from the names of the test methods one can deduce the test's focus is the localization of the strings, but that's an internal details and could be mentioned in a comment.
You'll need to keep the `test_` prefix in order for the tests to be picked up. Pretty sure the numbers can be dropped because the order is irrelevant.

I think `error_bar_test` could also use a better name like `check_error_in_message_bar`

testing/enterprise/test_felt_browser_console_error.py
```diff
@@ -17,27 +17,40 @@ def teardown(self):
...
-        return self.connection_error_test(
+        return self.error_bar_test(
             "http://nonexistent.localdomain:80",
-            "We can’t connect to the server at nonexistent.localdomain.",
+            ".felt-browser-error-no-network",
+            "No network connection",
         )
+
+    def test_felt_02_connection_error_ssl(self):
+        return self.connection_error_test(f"https://localhost:{self.console_port}")
```
1rneh: Can we also add a test that tests that the error details are populated correctly with the console address?

1rneh (changes requested):
Although our error reporting (multiple localization systems, etc.) is still a bit messy, the UI looks really great now. Thank you for the improvements!

browser/extensions/felt/content/felt.xhtml
jporter-dev: Ah I didn't see that design - just assumed it was included since it was near the errors. I'll revert those changes!

browser/locales/en-US/browser/enterprise/felt.ftl
jporter-dev: I knew this and had it in mind to update it but forgot 🤦‍♂️ thanks for pointing it out!

testing/enterprise/test_felt_browser_console_error.py
jporter-dev: Added a refused port test that covers this path: https://github.com/mozilla/enterprise-firefox/pull/493/changes#diff-404759aae67b15c417a0cc590ca258d541454a6bebc73582ce19253164c4a393R77

testing/enterprise/test_felt_browser_console_error.py
jporter-dev: Cleaned up the names to describe the test cases, removed the numbering, and renamed the error bar helper!

browser/extensions/felt/content/window.js
```diff
@@ -43,10 +48,9 @@ const ErrorReport = {
       return;
     }
     this._wrapper.classList.add("is-hidden");
-    const errors = this._wrapper.querySelectorAll(
-      ".felt-browser-error > div:not(.is-hidden)"
-    );
-    errors.forEach(e => e.classList.add("is-hidden"));
+    for (const bar of this._wrapper.querySelectorAll("moz-message-bar")) {
+      bar.classList.add("is-hidden");
+    }
```
1rneh, line 53: `ErrorReport.reset()` is currently called every time the user presses the Continue button and we attempt to connect to the console, which hides all message bars. That's before we know whether the console connection will succeed. This means we may end up clearing existing messages without actually progressing. 

I think we should only call `ErrorReport.reset()` once the connection succeeded and we proceed to the IdP login. This would also enable support for multiple message bars (e.g. showing both an update application error and a console connection error at the same time). Users can still dismiss them individually.

`nit:` Could you also add a bit of spacing between stacked message bars?

Image

1rneh (changes requested):
Just two small requests: one about the timing of calling ErrorReport.reset(), and the other about adding spacing when multiple message bars are shown.

Otherwise, looks good to me 🙂

browser/extensions/felt/content/window.js
jporter-dev: Good catch, that makes sense now that we have different types of message bars added via different paths than non-auth errors now. 

I had the spacing but think I lost it when reverting the uptodate message, will re-add it! Thanks!

jporter-dev: Used `space-small` for the spacing, and moved the `ErrorReport.reset()` after we load the SSO URI.

image

1rneh (approved):
Thank you!

lissyx (approved):

PR #494, opened 2026-03-03
Bug 2019055, 2019267 - Customize CrashReporter UI for policy autosubmits and Enterprise restart behavior

jonathanmendez:
Note: This does not add the behavior of auto-submitting yet, so even with the env variable for policy auto-submit, the crash report will be submitted on pressing the button in the crash reporter app rather than any earlier.

Enterprise crash reporter when policy is not on:
image

Enterprise crash reporter when policy is on:
image

**Note:** The text at the bottom "Your crash report will automatically be submitted." should not end up being visible once the policy autosubmit behavior is implemented; that label is tied to the submit status, and the submit will be kicked off as the UI becomes visible, so users will instead see a progress bar/completion message.

### DIFF
--- a/toolkit/crashreporter/client/app/src/lang/langpack.rs
+++ b/toolkit/crashreporter/client/app/src/lang/langpack.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use super::language_info::LanguageInfo;
-use super::zip::{read_archive_file_as_string, read_zip, Archive};
+use super::zip::{read_archive_file_as_string, read_archive_files_as_string, read_zip, Archive};
 use crate::config::installation_resource_path;
 use crate::std::path::{Path, PathBuf};
 use anyhow::Context;
@@ -163,8 +163,11 @@ fn read_archive_ftl_definitions(
     langpack: &mut Archive,
     language_identifier: &str,
 ) -> anyhow::Result<String> {
-    let path = format!("localization/{language_identifier}/crashreporter/crashreporter.ftl");
-    read_archive_file_as_string(langpack, &path)
+    let mut paths = Vec::new();
+    paths.push(format!("localization/{language_identifier}/crashreporter/crashreporter.ftl"));
+    #[cfg(feature = "enterprise")]
+    paths.push(format!("localization/{language_identifier}/crashreporter/crashreporter-enterprise.ftl"));
+    read_archive_files_as_string(langpack, &paths)
 }
 
 fn read_archive_ftl_branding(
--- a/toolkit/crashreporter/client/app/src/lang/language_info.rs
+++ b/toolkit/crashreporter/client/app/src/lang/language_info.rs
@@ -7,9 +7,22 @@ use anyhow::Context;
 use fluent::{bundle::FluentBundle, FluentResource};
 use unic_langid::LanguageIdentifier;
 
-const FALLBACK_FTL_FILE: &str = include_str!(mozbuild::srcdir_path!(
-    "/toolkit/locales/en-US/crashreporter/crashreporter.ftl"
-));
+cfg_if::cfg_if! {
+    if #[cfg(feature = "enterprise")] {
+        const FALLBACK_FTL_FILE: &str = concat!(
+            include_str!(mozbuild::srcdir_path!(
+                "/toolkit/locales/en-US/crashreporter/crashreporter.ftl"
+            )),
+            include_str!(mozbuild::srcdir_path!(
+                "/toolkit/locales/en-US/crashreporter/crashreporter-enterprise.ftl"
+            ))
+        );
+    } else {
+        const FALLBACK_FTL_FILE: &str = include_str!(mozbuild::srcdir_path!(
+            "/toolkit/locales/en-US/crashreporter/crashreporter.ftl"
+        ));
+    }
+}
 const FALLBACK_BRANDING_FILE: &str = include_str!(mozbuild::srcdir_path!(
     "/browser/branding/official/locales/en-US/brand.ftl"
 ));
--- a/toolkit/crashreporter/client/app/src/lang/omnijar.rs
+++ b/toolkit/crashreporter/client/app/src/lang/omnijar.rs
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use super::language_info::LanguageInfo;
-use super::zip::{read_archive_file_as_string, read_zip};
+use super::zip::{read_archive_file_as_string, read_archive_files_as_string, read_zip};
 use crate::config::installation_resource_path;
 use crate::std::io::{BufRead, BufReader};
 use anyhow::Context;
@@ -30,9 +30,13 @@ pub fn read() -> anyhow::Result<Vec<LanguageInfo>> {
     let loaded: Vec<_> = locales
         .clone()
         .filter_map(|locale| {
-            match read_archive_file_as_string(
+            let mut paths = Vec::new();
+            paths.push(format!("localization/{locale}/crashreporter/crashreporter.ftl"));
+            #[cfg(feature = "enterprise")]
+            paths.push(format!("localization/{locale}/crashreporter/crashreporter-enterprise.ftl"));
+            match read_archive_files_as_string(
                 &mut zip,
-                &format!("localization/{locale}/crashreporter/crashreporter.ftl"),
+                &paths,
             ) {
                 Ok(v) => Some((locale.to_owned(), v)),
                 Err(e) => {
--- a/toolkit/crashreporter/client/app/src/ui/mod.rs
+++ b/toolkit/crashreporter/client/app/src/ui/mod.rs
@@ -205,6 +205,9 @@ impl ReportCrashUI {
             close_window,
         } = state.borrow();
 
+        #[cfg(feature = "enterprise")]
+        let policy_auto_submit = config.policy_auto_submit;
+
         send_report.on_change(cc! { (logic) move |v| {
             let v = *v;
             logic.push(move |s| s.settings.borrow_mut().submit_report = v);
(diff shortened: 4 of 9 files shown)


lissyx:
You will want a review from @afranchuk I guess ?

toolkit/locales/en-US/crashreporter/crashreporter.ftl
```diff
@@ -7,6 +7,7 @@ crashreporter-branded-title = { -brand-short-name } Crash Reporter
 crashreporter-apology = We’re Sorry
 crashreporter-crashed-and-restore = { -brand-short-name } had a problem and crashed. We’ll try to restore your tabs and windows when it restarts.
 crashreporter-plea = To help us diagnose and fix the problem, you can send us a crash report.
+crashreporter-policy-explanation = Per configured policy, a report is automatically sent to your admin.
```
lissyx: Those strings are enterprise specific? We need to introduce a new `crashreporter-enterprise.ftl` for example and isolate all enterprise-specific string there.

lissyx (changes requested):

toolkit/crashreporter/client/app/src/ui/mod.rs
```diff
@@ -218,13 +218,28 @@ impl ReportCrashUI {
             logic.push(move |s| s.settings.borrow_mut().test_hardware = v);
         }});
 
-        let input_enabled = submit_state.mapped(|s| s == &SubmitState::Initial);
+        let input_enabled = if config.policy_auto_submit {
+            data::Synchronized::new(false)
+        } else {
+            submit_state.mapped(|s| s == &SubmitState::Initial)
+        };
```
afranchuk: If `config.policy_auto_submit` is true, maybe it would make more sense to set the submit state to `InProgress` (which would have the side-effect of disabling the inputs). `InProgress` (and the follow-on states) is sort of defined as not allowing any more input (and we could more rigorously document that in comments to be sure that remains true).

toolkit/crashreporter/client/app/src/ui/mod.rs
```diff
@@ -291,16 +306,16 @@ impl ReportCrashUI {
...
                         {
                             Label text(config.string("crashreporter-button-restart"))
                         },
                         Button["quit"] on_click(cc! { (logic) move || logic.push(|s| s.quit()) })
-                            enabled(&input_enabled) hsize(160)
+                            enabled(&quit_enabled) hsize(160)
                         {
-                            Label text(config.string("crashreporter-button-quit"))
+                            Label text(config.string(if cfg!(feature = "enterprise") {"crashreporter-button-ok"} else {"crashreporter-button-quit"}))
                         }
```
afranchuk: Since the behavior of the "Ok" button is different from the "Quit" button (quit potentially triggers submission, whereas ok will just dismiss the window), I'd prefer it be defined as a separate button entirely.

We could add support for attributes in the `ui!` macro, which would allow us to more easily add/remove the buttons with `#![cfg(...)]`. Or just add support for "switching back" to imperative code. I'll take a look at that.

afranchuk (changes requested):

toolkit/crashreporter/client/app/src/ui/mod.rs
afranchuk: Yep, there's a pretty simple change we can make to the `ui!` macro to allow you to insert regular imperative code (the macro is already pretty simple, so it made it easy). I can make a bug to get that into f-m.

jonathanmendez: Sounds good. Yeah I tried the cfg attribute first and ran into the macro issues and guessed it would be hard to deal with that; glad it sounds like it was easy!

afranchuk: I created https://bugzilla.mozilla.org/show_bug.cgi?id=2021010. I'll have a patch in soon.

jonathanmendez: I locally applied your patch and it worked well for me applying the `#[cfg(feature = "enterprise")]`  attribute (or its inverse) to include or not include the buttons as desired! :)

jonathanmendez: Question about the separate button feedback:

In Enterprise, there are two things going on:
1. Firefox automatically restarts after a crash (so "restart firefox" and "quit firefox" don't make sense)
2. The auto-submit policy may or may not be applied

Because of (1), I think it makes sense to have a single button with the text "Ok".
But depending on whether the policy is on, that button will either "potentially trigger submission" or "just dismiss the window". Which makes the implementation more complicated than we'd hope.

For implementation I could see a few paths:
1. Create a single enterprise-only "Ok" button that calls either `quit()` or `close_window()` depending on the policy config, and which is enabled in different states depending on the policy config.
2. Create two separate enterprise-only "Ok" buttons, one of which calls `quit()` and the other `close_window()` with visibility dependent on the policy config. The `quit()` version is enabled if `input_enabled`, while the `close_window()` version is always enabled.
3. Create a single enterprise-only "Ok" button that only calls `close_window()`; change the text of the 'Quit' button to be "Ok" in enterprise-only; and use the policy config to choose whether the 'Quit-that-says-Ok' or the 'Ok' button are visible. The "Quit-that-says-Ok" is enabled if `input_enabled` while the 'Ok' is always enabled.
4. Stick with the implementation currently in the PR, molding the single "Quit" button depending on enterprise and policy.

| Option | Quit Button Text | Quit Button Enabled | Quit Button Logic | Ok Button Text | Ok Button Enabled | Ok Button Logic | Ok Button2 Text | Ok Button2 Enabled | Ok Button2 Logic |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | "Quit Firefox" | n/a in enterprise | `quit()` | "Ok" | if policy then `true` else `input_enabled` | if policy then `close_window()` else `quit()` | n/a | n/a | n/a |
| 2 | "Quit Firefox" | n/a in enterprise | `quit()` | "Ok" | `true` (visible if policy) | `close_window()` | "Ok" | `input_enabled` (visible if not policy) | `quit()` |
| 3 | if enterprise then "Ok" else "Quit Firefox" | `input_enabled` (visible if not policy) | `quit()` | "Ok" | `true` (visible if policy) | `close_window()` | n/a | n/a | n/a |
| 4 | if enterprise then "Ok" else "Quit Firefox" | if policy then `true` else `input_enabled` | if policy then `close_window()` else `quit()` | n/a | n/a | n/a | n/a | n/a | n/a |

Of these, probably option 2 is the most "simple" at the cost of an odd extra button. Maybe option 1 is the next best, where all the enterprise stuff is isolated, even if the extra button has some extra logic associated with it.

Thoughts?

afranchuk: I'm fine with the button having it's own logic wrt checking the policy, but having it as a separate button seems easier than trying to reusing the Quit button. So option 1 feels right to me.

toolkit/locales/en-US/crashreporter/crashreporter.ftl
jonathanmendez: Took some figuring out but I believe I've implemented the isolation into the separate `crashreporter-enterprise.ftl` file.

toolkit/crashreporter/client/app/src/ui/mod.rs
jonathanmendez: While I don't like the verbosity of having the different options to check, I do prefer that there is a definitive "if the policy is on, you can definitely not add input" since the policy is what will drive the state of the settings. But as I build the behavior PR, I will of course see how the logic shapes up there.

toolkit/crashreporter/client/app/src/ui/mod.rs
jonathanmendez: Ok I've implemented option 1. Let me know how it feels in the code.

toolkit/crashreporter/client/app/src/lang/langpack.rs
```diff
@@ -167,6 +172,16 @@ fn read_archive_ftl_definitions(
...
 
+#[cfg(feature = "enterprise")]
+fn read_archive_ftl_definitions_enterprise(
+    langpack: &mut Archive,
+    language_identifier: &str,
+) -> anyhow::Result<String> {
+    let path =
+        format!("localization/{language_identifier}/crashreporter/crashreporter-enterprise.ftl");
+    read_archive_file_as_string(langpack, &path)
+}
```
afranchuk, line 171: I'd prefer changing `read_archive_ftl_definitions` to take a `fluent_file_stem` parameter (which would be either `crashreporter` or `crashreporter-enterprise`, in these cases) rather than duplicating the function.

toolkit/crashreporter/client/app/src/lang/language_info.rs
```diff
@@ -7,9 +7,22 @@ use anyhow::Context;
...
+            include_str!(mozbuild::srcdir_path!(
+                "/toolkit/locales/en-US/crashreporter/crashreporter-enterprise.ftl"
+            ))
+        );
+    } else {
+        const FALLBACK_FTL_FILE: &str = include_str!(mozbuild::srcdir_path!(
+            "/toolkit/locales/en-US/crashreporter/crashreporter.ftl"
+        ));
+    }
+}
```
afranchuk, line 25: If this ever gets more complicated in the future, we'll probably want to do a different approach (like a `LazyLock<String>` to do the concatenation), but it's fine for now.

toolkit/crashreporter/client/app/src/lang/omnijar.rs
```diff
@@ -30,16 +30,32 @@ pub fn read() -> anyhow::Result<Vec<LanguageInfo>> {
...
+                    Ok(v) => Some(s + &v),
+                    Err(e) => {
+                        log::warn!(
+                            "failed to get localized enterprise strings for {locale}: {e:#}"
+                        );
+                        None
+                    }
+                }
+            });
+            ftl.map(|s| (locale.to_owned(), s))
```
afranchuk: I'd prefer this logic to operate over a slice of strings, so that we aren't duplicating the logic and the `#[cfg(feature = "enterprise")]` part can be as simple as adding the other path to the slice.

Something like
```rust
let mut file_stems = Vec::new();
file_stems.push("crashreporter");
#[cfg(feature = "enterprise")]
file_stems.push("crashreporter-enterprise");

file_stems.into_iter().map(|stem| {
    match read_archive_file_as_string(&mut zip, &format!("localization/{locale}/crashreporter/{stem}.ftl")) {
        Ok(v) => Some(v),
        Err(e) => {
            log::warn!("failed to get localized {stem} strings for {locale}: {e:#}");
            None
        }
    }
})
.collect::<Option<Vec<_>>()
.map(|file_contents| (locale.to_owned(), file_contents.join("\n")))
```

toolkit/crashreporter/client/app/src/ui/mod.rs
```diff
@@ -218,13 +221,36 @@ impl ReportCrashUI {
...
+            if #[cfg(feature = "enterprise")] {
+                let input_enabled = if policy_auto_submit {
+                    data::Synchronized::new(false)
+                } else {
+                    submit_state.mapped(|s| s == &SubmitState::Initial)
+                };
+            } else {
+                let input_enabled = submit_state.mapped(|s| s == &SubmitState::Initial);
+            }
+        }
```
afranchuk, line 234: I still think it'd be cleaner if `input_enabled` remained unchanged here, and `policy_auto_submit` changed `submit_state` to be `SubmitState::InProgress` (which semantically makes sense).

Depending on how the `policy_auto_submit` submission is implemented, this will probably end up Just Working™️  because the submission logic will set the `submit_state` on its own. So perhaps nothing needs to change here yet, and it can be simplified when the time comes.

toolkit/crashreporter/client/app/src/lang/langpack.rs
```diff
@@ -25,6 +25,11 @@ pub fn read(
     let mut zip = read_zip(&langpack)?;
 
     let ftl_definitions = read_archive_ftl_definitions(&mut zip, &identifier)?;
+    #[cfg(feature = "enterprise")]
+    let ftl_definitions_enterprise =
+        read_archive_ftl_definitions_enterprise(&mut zip, &identifier)?;
+    #[cfg(feature = "enterprise")]
+    let ftl_definitions = ftl_definitions + &ftl_definitions_enterprise;
```
afranchuk: Since `ftl_definitions_enterprise` isn't used elsewhere, you could condense this to
```rust
#[cfg(feature = "enterprise")]
let ftl_definitions = ftl_definitions + &read_archive_ftl_definitions_enterprise(&mut zip, &identifier)?;
```

afranchuk: In combination with my other suggestions (changing `read_archive_ftl_definitions` to take a file stem, and iterating over the reading of multiple files), it seems like what we could do here is change `read_archive_ftl_definitions` to take a slice of file stems and produce the combined result. I'm just now realizing that this function could have been used for the omnijar case, too, since both store the files in the same place. The omnijar case is calling `read_archive_file_as_string`, but it's basically doing the same thing as this function.

afranchuk (changes requested):
I have a few suggestions to make the logic more consistent, but otherwise the way the UI is put together looks good to me.

afranchuk (approved):
Looks good now!

afranchuk:
The macro patch is now in firefox-main.

lissyx (approved):

PR #504, opened 2026-03-05
Bug 2021094 - Enterprise: fix exit token tests to avoid test-only race conditions

lissyx:
@1rneh Asking your review mostly because of how I have to interact with `ConsoleClient` here.

### DIFF
--- a/testing/enterprise/test_felt_browser_exit_tokens.py
+++ b/testing/enterprise/test_felt_browser_exit_tokens.py
@@ -5,44 +5,51 @@
 
 import os
 import sys
-import time
 
 sys.path.append(os.path.dirname(__file__))
 
+from base_test import Environment
 from felt_tests import FeltTests
 
 
 class BrowserExitTokens(FeltTests):
     def test_browser_exit_tokens(self):
-        self.felt_block_shutdown()
-        super().run_felt_base()
-        self.run_browser_set_fake_tokens()
+        self.get_driver(Environment.FELT).set_prefs(
+            # required to not close felt window when launching browser,
+            # allowing to collect tokens on felt side
+            {
+                "enterprise.felt_tests.should_not_close_window": True,
+                "enterprise.felt_tests.is_blocking_shutdown": True,
+            },
+            default_branch=True,
+        )
+        self.run_felt_base()
+        self.connect_child_browser()
+        # Record process PID we will have to wait for in self.wait_process_exit()
+        self._browser_pid = self._child_driver.session_capabilities["moz:processID"]
+        self.check_felt_and_firefox_tokens_in_sync()
+        self.force_and_refresh_tokens()
+        self.check_firefox_tokens_updated_after_session_refresh()
         self.perform_quit()
+        self.wait_process_exit()
         self.await_felt_auth_window()
         self.force_window()
-        self.run_felt_check_fake_tokens()
-
-    def felt_block_shutdown(self):
-        self._driver.set_context("chrome")
-        self._driver.execute_script(
-            """
-            Services.prefs.setBoolPref("enterprise.felt_tests.is_blocking_shutdown", true);
-            """
-        )
-        self._driver.set_context("content")
+        self.check_felt_received_refreshed_tokens_on_shutdown()
 
     def perform_quit(self):
-        self._child_driver.set_context("chrome")
-        rv = self._child_driver.execute_script(
+        driver = self.get_driver(Environment.FIREFOX)
+        driver.set_context("chrome")
+        rv = driver.execute_script(
             """
             Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
             """,
         )
-        self._child_driver.set_context("content")
+        driver.set_context("content")
         self._manually_closed_child = True
         return rv
 
-    def get_tokens(self, driver):
+    def get_tokens(self, env):
+        driver = self.get_driver(env)
         driver.set_context("chrome")
         rv = driver.execute_script(
             """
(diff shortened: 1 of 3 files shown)


testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -52,36 +48,58 @@ def get_tokens(self, driver):
...
+        # call gets overwritten by execution of a ConsoleClient._refreshPromise
+        # in happening at the same time. This was the root cause of bug 2021094.
+        driver.execute_async_script(
+            """
+            const callback = arguments[arguments.length - 1];
+            const { ConsoleClient } = ChromeUtils.importESModule("resource:///modules/enterprise/ConsoleClient.sys.mjs");
+            ConsoleClient._refreshPromise = new Promise((resolve, reject) => {
+              Services.felt.setTokens(arguments[0], arguments[1], arguments[2]);
+              resolve();
+            }).then(callback);
```
lissyx: From testing and from checking the code, while it may look a bit fishy, this should work well since we have this check at the beginning of `_refreshSession` to re-use existing promise.

1rneh (commented):
Can you confirm for me that this is what this test is testing?

1. Perform authentication in Felt which sends token set `A` from Felt to Firefox.
2. Override token set `A` with `B` in Firefox.
3. Perform app shutdown in Firefox which sends token set `B` from Firefox back to Felt
4. Verify that Felt has token set `B`.

You modified the token set in Firefox from `A` to `B` to verify Felt really gets the latest token set from Firefox on shutdown. The intermittent issue is that Firefox might refresh the token set to `C` and `C` is sent back to felt which breaks the token assertion.

From my understanding, suppressing the session refresh is fixing this test, but it reveals some important problems:
- We don't check for an inflight session refresh when sending back tokens from Firefox to Felt
- We don't clear tokens in Firefox after sending them to Felt. There might be a small chance we still do a session refresh which will invalidate the token set we send to Felt.
- We don't clear tokens in Felt after sending them to Firefox. Felt making any request to the console will invalidate the token set that Firefox has.

lissyx:
> Can you confirm for me that this is what this test is testing?
> 
>     1. Perform authentication in Felt which sends token set `A` from Felt to Firefox.
> 
>     2. Override token set `A` with `B` in Firefox.
> 
>     3. Perform app shutdown in Firefox which sends token set `B` from Firefox back to Felt
> 
>     4. Verify that Felt has token set `B`.
> 

Yes, that is what the test is doing

> 
> You modified the token set in Firefox from `A` to `B` to verify Felt really gets the latest token set from Firefox on shutdown. The intermittent issue is that Firefox might refresh the token set to `C` and `C` is sent back to felt which breaks the token assertion.

Yes, `B` gets overwritten by `C` just before we perform the actual exit.

> 
> From my understanding, suppressing the session refresh is fixing this test, but it reveals some important problems:
> 
>     * We don't check for an inflight session refresh when sending back tokens from Firefox to Felt

On which side? Because on Firefox side, when we hit `Services.felt.setTokens()` call in the test, at that moment, there is no `ConsoleClient._refreshPromise` yet. From what I traced, the promise gets added just after because there's some API request for the account info. At the moment we do the quit, the promise already had resolved so the token was already overwritten from what I checked.

> 
>     * We don't clear tokens in Firefox after sending them to Felt. There might be a small chance we still do a session refresh which will invalidate the token set we send to Felt.

Which seems to be exactly what happens here. We could clear but them I'm wondering about the impact on https://bugzilla.mozilla.org/show_bug.cgi?id=2010529 and related issues?

I can try and test if doing so fixes the issue without my override of the promise

> 
>     * We don't clear tokens in Felt after sending them to Firefox. Felt making any request to the console will invalidate the token set that Firefox has.

Maybe we should fix this in this PR? Or a follow up?

1rneh:
We moved the conversation about this to Bugzilla for better tracebility: https://bugzilla.mozilla.org/show_bug.cgi?id=2021094

1rneh (changes requested):
I’m a bit unsure about this fix and whether it resolves the intermittence with confidence. From your comments in [Bug 2021094](https://bugzilla.mozilla.org/show_bug.cgi?id=2021094), I get the impression that we’re not entirely certain about Felt not triggering a session refresh after Firefox is launched.

I have a suggestion to improve the test structure:

I think we should properly implement the `/sso/token` refresh endpoint in the mock console. Concretely:

- Authenticate requests to /sso/token (it seems to be the only endpoint that we missed to add authentication for).
- Generate and persist a new token set server-side.
- Respond with the newly generated token set.

Note: You've already done most of this with `/sso/token_force`

With that in place, the test flow could be:
1. Authenticate in Felt which launches Firefox
2. Trigger a session refresh from Firefox (Now Felt should no longer be able to refresh the session because it doesn't hold the valid refresh token).
3. Shutdown Firefox
4. Check that the token set Felt holds is the one the server holds.

I’m actually glad we ran into this intermittent failure, since it surfaced the real and foreseeable problem of having two instances each holding their own token set.
Long-term, I think Felt should be the single owner of the refresh token and the only component responsible for refreshing the session. Firefox should request the access token from Felt instead.
This is different from our previous brainstorming on having Firefox talk to the console through Felt - here it is only about centralizing session refresh responsibility and refresh token ownership.

lissyx:
> I’m a bit unsure about this fix and whether it resolves the intermittence with confidence. From your comments in [Bug 2021094](https://bugzilla.mozilla.org/show_bug.cgi?id=2021094), I get the impression that we’re not entirely certain about Felt not triggering a session refresh after Firefox is launched.

It fixes it with confidence assuming there is indeed no refresh on Felt side, which I could never prove. But you mentionned Felt may trigger a refresh with a Glean ping, and so maybe it is hidden when we run test just because we are lucky glean does not trigger a refresh ...

> 
> I have a suggestion to improve the test structure:
> 
> I think we should properly implement the `/sso/token` refresh endpoint in the mock console. Concretely:
> 
>     * Authenticate requests to /sso/token (it seems to be the only endpoint that we missed to add authentication for).
> 
>     * Generate and persist a new token set server-side.
> 
>     * Respond with the newly generated token set.
> 
> 
> Note: You've already done most of this with `/sso/token_force`

OK

> 
> With that in place, the test flow could be:
> 
>     1. Authenticate in Felt which launches Firefox
> 
>     2. Trigger a session refresh from Firefox (Now Felt should no longer be able to refresh the session because it doesn't hold the valid refresh token).

Session refresh rather than forcing the token, right.

> 
>     3. Shutdown Firefox
> 
>     4. Check that the token set Felt holds is the one the server holds.
> 

So the test structure mostly stays the same

> 
> I’m actually glad we ran into this intermittent failure, since it surfaced the real and foreseeable problem of having two instances each holding their own token set. Long-term, I think Felt should be the single owner of the refresh token and the only component responsible for refreshing the session. Firefox should request the access token from Felt instead. This is different from our previous brainstorming on having Firefox talk to the console through Felt - here it is only about centralizing session refresh responsibility and refresh token ownership.

1rneh:
> But you mentioned Felt may trigger a refresh with a Glean ping, and so maybe it is hidden when we run test just because we are lucky glean does not trigger a refresh ...

No, all Glean does is read the access token (https://searchfox.org/enterprise-main/rev/a038f49228d707c6675ef20ce640034a64307d2e/toolkit/components/glean/src/init/viaduct_uploader.rs#151-160). It does never initiate a session refresh. On failure it reschedules the request and hopes the ConsoleClient refreshed teh session in the meantime.

browser/components/enterprise/EnterpriseCommon.sys.mjs
```diff
@@ -15,10 +14,6 @@ export const isUpdatesTesting = () => {
   return Services.prefs.getBoolPref(IS_UPDATES_TESTING, false);
 };
 
-export const isBlockingShutdown = () => {
-  return Services.prefs.getBoolPref(IS_BLOCKING_SHUTDOWN, false);
-};
-
```
lissyx, line 21: we have to keep it in fact :(

lissyx: (for #506 )

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -5,44 +5,50 @@
...
+            # allowing to collect tokens on felt side
+            {
+                "enterprise.felt_tests.should_not_close_window": True,
+                "enterprise.felt_tests.is_blocking_shutdown": True,
+            },
+            default_branch=True,
+        )
+        self.run_felt_base()
+        self.connect_child_browser()
+        self._browser_pid = self._child_driver.session_capabilities["moz:processID"]
```
1rneh, line 29: Is `self._browser_pid` used anywhere in this test or necessary for a later test? Or can we remove it?

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -52,36 +58,57 @@ def get_tokens(self, driver):
...
-            f"Access token differs: {browser_tokens[0]} vs {self.access_token}"
+    def run_browser_refresh_forced_tokens(self):
+        self.new_browser_tokens = self.get_tokens(Environment.FIREFOX)
+        assert len(self.new_browser_tokens[0]) > 0, "Browser access token empty"
+        assert len(self.new_browser_tokens[1]) > 0, "Browser refresh token empty"
+
+        assert self.new_browser_tokens[0] != self.browser_tokens[0] and len(
+            self.new_browser_tokens[0]
+        ) == len(self.browser_tokens[0]), (
+            f"Browser access token differs: {self.new_browser_tokens[0]} vs {self.browser_tokens[0]}"
```
1rneh: Can we split them into two assertions?

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -5,44 +5,50 @@
...
+            {
+                "enterprise.felt_tests.should_not_close_window": True,
+                "enterprise.felt_tests.is_blocking_shutdown": True,
+            },
+            default_branch=True,
+        )
+        self.run_felt_base()
+        self.connect_child_browser()
+        self._browser_pid = self._child_driver.session_capabilities["moz:processID"]
+        self.run_check_all_tokens()
```
1rneh: nit: Can we rename the methods to clarify what they are checking
- `run_check_all_tokens` to `check_felt_and_firefox_tokens_in_sync`
- `run_browser_refresh_forced_tokens` to `check_firefox_tokens_updated_after_session_refresh`
- `run_felt_check_forced_tokens` to `check_felt_received_refreshed_tokens_on_shutdown`

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -52,36 +58,57 @@ def get_tokens(self, driver):
...
+        ) == len(self.new_browser_tokens[0]), (
+            f"FELT access token matches: {felt_tokens_after_exit[0]} vs {self.new_browser_tokens[0]}"
         )
-        assert felt_tokens_after_signout[1] == self.refresh_token, (
-            f"Refresh token matches: {felt_tokens_after_signout[1]} vs {self.refresh_token}"
+        assert felt_tokens_after_exit[1] == self.new_browser_tokens[1] and len(
+            felt_tokens_after_exit[1]
+        ) == len(self.new_browser_tokens[1]), (
+            f"FELT refresh token matches: {felt_tokens_after_exit[1]} vs {self.new_browser_tokens[1]}"
         )
```
1rneh: Same here.

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -52,36 +58,57 @@ def get_tokens(self, driver):
...
+        ) == len(self.browser_tokens[0]), (
+            f"Browser access token differs: {self.new_browser_tokens[0]} vs {self.browser_tokens[0]}"
         )
-        assert browser_tokens[1] != self.refresh_token, (
-            f"Refresh token differs: {browser_tokens[1]} vs {self.refresh_token}"
+        assert self.new_browser_tokens[1] != self.browser_tokens[1] and len(
+            self.new_browser_tokens[1]
+        ) == len(self.browser_tokens[1]), (
+            f"Browser refresh token differs: {self.new_browser_tokens[1]} vs {self.browser_tokens[1]}"
         )
```
1rneh: Same here.

1rneh (approved):
Thanks for testing the tokens in each state - lgtm now!

Just a few nits:

- Let’s have one assertion per statement.
- Could we make the method names more descriptive? It would help if the names clearly indicate not only that tokens are being checked, but also what about the tokens is being validated.

testing/enterprise/test_felt_browser_exit_tokens.py
lissyx: It's used by wait proces

testing/enterprise/test_felt_browser_exit_tokens.py
gcp: :rofl: https://github.com/mozilla/enterprise-firefox/pull/506#discussion_r2940999849

lissyx: I already did it, that's weird it is not reflecting, did I forgot to push after it maybe ? I'll check when I get back to my desk

testing/enterprise/test_felt_browser_exit_tokens.py
lissyx: Sure

testing/enterprise/test_felt_browser_exit_tokens.py
1rneh: I had a similar thought, but I didn't want to `nit` even more. But if it's already pointed out elsewhere, could be have the `AssertionError` message explain what has failed or what was expected, e.g. `Expected old and new tokens to differ after a session refresh: [..]`.

lissyx: :facepalm: I did the change, forgot to `git add` it, got disrupted in something else so I discarded it.

lissyx: So split the assertions and make assertion message clear

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -52,36 +59,57 @@ def get_tokens(self, driver):
...
-        browser_tokens = self.get_tokens(self._child_driver)
+    def check_felt_and_firefox_tokens_in_sync(self):
+        self.felt_tokens = self.get_tokens(Environment.FELT)
+        self.browser_tokens = self.get_tokens(Environment.FIREFOX)
 
-        self.access_token = "1bf8d2cb-9f72-4788-a770-3cf9cc60f30c"
-        self.refresh_token = "82b2d9eb-f0e3-44af-8106-790e7a744d1f"
-        self.expires_at = int(time.time()) + 3600
+        assert self.felt_tokens[0] == self.browser_tokens[0]
+        assert self.felt_tokens[1] == self.browser_tokens[1]
```
1rneh: Can you add an Assertion error message?

lissyx: thansk, forgot this one

PR #505, opened 2026-03-05
Capitalize property descriptions in SyncPolicyParams JSDocs

1rneh:


bsmth (approved):
I think "open tabs" and "payment methods" should be expanded out like normal words in the sentences, but otherwise lgtm.

lissyx (approved):

PR #506, opened 2026-03-05
Bug 2021408 - Enterprise: mirror tokens after setting them

lissyx:


browser/extensions/felt/rust/src/components.rs
```diff
@@ -146,17 +145,25 @@ impl FeltXPCOM {
                     NS_ERROR_FAILURE
                 }
             }
+        };
+
+        match sent {
+            NS_OK => {
+                trace!("FeltXPCOM::SendTokens(): clear existing tokens");
+                self.set_tokens("".to_string(), "".to_string(), 0);
```
gcp: This means we can't ever recover if Firefox crashes.

(Right now we recover it the tokens still turn out to be valid)

I think my conclusion is that Felt should keep a mirror and we should update it when we get new ones?

lissyx: That's something I missed, you're right, and I'm not sure we have tests covering that. I guess mirroring would be the correct solution indeed?

gcp: The mock console doesn't check whether the auth tokens are proper in the tests IIRC.

gcp: Note it's only half of a regression because if Firefox gets new tokens, Felt would still be stuck with the old ones even in the current code.

lissyx: we check in some places: https://searchfox.org/enterprise-main/rev/42649df111db40bdef19fc0aeb6986d180fa9012/testing/enterprise/felt_tests.py#108-121,175,203,310

Maybe we need to check more though?

lissyx: > This means we can't ever recover if Firefox crashes.
> 
> (Right now we recover it the tokens still turn out to be valid)
> 
> I think my conclusion is that Felt should keep a mirror and we should update it when we get new ones?

I think crash tests caught this in fact: https://treeherder.mozilla.org/logviewer?job_id=552124792&repo=enterprise-firefox-pr&task=RwncIt4fSaewPJ1uhPpIDA.1&lineNumber=4246

lissyx: On shutdown on the browser side, we clear tokens: https://searchfox.org/enterprise-main/rev/4c96ba0d0e4207e3e25ddb5634098ae6d1646984/browser/components/enterprise/modules/ConsoleClient.sys.mjs#688, I suspect we want to keep the clear step, but we do not want to mirror the cleared state, that would defeat the whole purpose of the patch

browser/extensions/felt/rust/src/client.rs
```diff
@@ -87,24 +87,41 @@ impl FeltIpcClient {
...
-    pub fn send_back_tokens(&self) {
+    pub fn send_back_tokens(&self) -> nserror::nsresult {
         trace!("FeltIpcClient::send_back_tokens()");
         match TOKENS.read() {
             Ok(tokens) => {
+                let access_token = tokens.access_token.clone();
+                let refresh_token = tokens.refresh_token.clone();
+
+                if access_token.len() == 0 && refresh_token.len() == 0 {
+                    trace!("FeltIpcClient::send_back_tokens() empty tokens, aborting");
```
gcp: ...but now you're not clearing the remote tokens even in cases where you want this?

https://searchfox.org/enterprise-main/rev/42bdad24fb3cd97cd5678896eef3bcde1cd4e4bf/browser/components/enterprise/modules/ConsoleClient.sys.mjs#612

https://searchfox.org/enterprise-main/rev/42bdad24fb3cd97cd5678896eef3bcde1cd4e4bf/browser/components/enterprise/modules/ConsoleClient.sys.mjs#643

https://searchfox.org/enterprise-main/rev/42bdad24fb3cd97cd5678896eef3bcde1cd4e4bf/browser/components/enterprise/modules/ConsoleClient.sys.mjs#688

gcp (changes requested):

browser/extensions/felt/rust/src/client.rs
lissyx: We are cleaning them in those case, but not mirroring the cleanup. I'm not sure if we should mirror the cleanup in all cases except at shutdown? If we do, do we need to cleanup at shutdown then?
If we still want to mirror cleanup and we want to keep the cleanup shutdown but not mirror it, then I think we need to augment the XPCOM API to surface our will ?

lissyx: Re-reading the code I think we want to mirror when we cleanup on browser side. The real question hence is: do we want to cleanup at shutdown? If we do, then we need to notify the XPCOM API we do not want to mirror, we cannot just rely on the tokens being empty strings.

1rneh (commented):
If we start mirroring tokens between Felt and Firefox, we risk introducing a race condition. For example:

1. Felt refreshes the session and receives a new token set A.
2. Firefox refreshes the session and receives a new token set B, which invalidates A on the server.
3. Firefox sends token set B to Felt.
4. Felt sends token set A to Firefox.

As a result Firefox ends up holding the invalid token set A.

The token life cycle is the reason that Felt and Firefox cannot both communicate with the console simultaneously.

lissyx:
> If we start mirroring tokens between Felt and Firefox, we risk introducing a race condition. For example:
> 
>     1. Felt refreshes the session and receives a new token set A.

It was my understanding that FELT would not perform anything, and thus not perform any refresh during the lifetime of the browser?

> 
>     2. Firefox refreshes the session and receives a new token set B, which invalidates A on the server.
> 
>     3. Firefox sends token set A to Felt.

This should be "token set B" I guess?

> 
>     4. Felt sends token set B to Firefox.

Mirroring is only done in the Firefox -> FELT way

1rneh:
> It was my understanding that FELT would not perform anything, and thus not perform any refresh during the lifetime of the browser?

But wasn't that the conclusion of [Bug 2021094](https://bugzilla.mozilla.org/show_bug.cgi?id=2021094), that Felt was still sending out a request at an unexpected timing?

> This should be "token set B" I guess?

Yes, thanks! I have corrected it in the original comment so that it doesn't cause confusion for the next reader. 

> Mirroring is only done in the Firefox -> FELT way

That's still a problem. Felt refreshing the session will invalidate the tokens on Firefox side.

lissyx:
> > It was my understanding that FELT would not perform anything, and thus not perform any refresh during the lifetime of the browser?
> 
> But wasn't that the conclusion of [Bug 2021094](https://bugzilla.mozilla.org/show_bug.cgi?id=2021094), that Felt was still sending out a request at an unexpected timing?

no, I commented (maybe in DM?) but I could never reproduce, so I may have been wrong from the beginning

> 
> > This should be "token set B" I guess?
> 
> Yes, thanks! I have corrected it in the original comment so that it doesn't cause confusion for the next reader.
> 
> > Mirroring is only done in the Firefox -> FELT way
> 
> That's still a problem. Felt refreshing the session will invalidate the tokens on Firefox side.

But Felt has no reason to refresh the session? Maybe this is worth enforcing, blocking session refresh on Felt side once the browser is running?

1rneh:
> Mirroring is only done in the Firefox -> FELT way
> > That's still a problem. Felt refreshing the session will invalidate the tokens on Firefox side.
> > > But Felt has no reason to refresh the session? Maybe this is worth enforcing, blocking session refresh on Felt side once the browser is running?

Yes, to avoid any raciness we should block any session refresh while Firefox is running, although we are ruling out that possibility right now). 

Btw our assumption that Felt doesn't communicate with the console after Firefox is launched is not correct. Our Glean pings in Felt are send to the console **and are authenticated**. Luckily for us Glean is only reading the tokens not initiating any session refresh. Another reason why we need to send back the tokens from Firefox to Felt to be able to send authenticated pings. Currently when Firefox refreshes the session once, we won't get Glean pings from Felt anymore.

lissyx:
> > Mirroring is only done in the Firefox -> FELT way
> > > That's still a problem. Felt refreshing the session will invalidate the tokens on Firefox side.
> > > > But Felt has no reason to refresh the session? Maybe this is worth enforcing, blocking session refresh on Felt side once the browser is running?
> 
> Yes, to avoid any raciness we should block any session refresh while Firefox is running, although we are ruling out that possibility right now).
> 
> Btw our assumption that Felt doesn't communicate with the console after Firefox is launched is not correct. Our Glean pings in Felt are send to the console **and are authenticated**. Luckily for us Glean is only reading the tokens not initiating any session refresh. Another reason why we need to send back the tokens from Firefox to Felt to be able to send authenticated pings. Currently when Firefox refreshes the session once, we won't get Glean pings from Felt anymore.

Sure, I'll update and try to find a way to block refresh while the browser is running. Do you have an opinion as well on our `clearTokenData()` performed on shutdown of the browser? Cf the discussion above. I guess we still want to do this clear call right?

browser/extensions/felt/rust/src/client.rs
1rneh: On an expected shutdown we likely want to clean the tokens in Firefox and in Felt according to the first two lines of this table: https://www.figma.com/board/29dqnGYlMoyVozr9tinfm8/Untitled?node-id=30-158&t=QWZ1ekk9vsG4yq92-0

gcp: On shutdown, it could be policy-controlled whether that forces a logout or not, so I think you want to be ready to handle both cases.

lissyx:
The errors seems to be intermittents from bug [2021094](https://bugzilla.mozilla.org/show_bug.cgi?id=2021094)

lissyx:
In light of https://bugzilla.mozilla.org/show_bug.cgi?id=2021094#c16 I'll rebase this PR on top of https://github.com/mozilla/enterprise-firefox/pull/506

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -489,6 +526,11 @@ export const ConsoleClient = {
    * @returns {Promise<void>}
    */
   async _refreshSession() {
+    if (this.refreshBlockedInFelt) {
+      // Or should a specific object be return? Or throwing?
```
gcp: Probably at least some hard-to-ignore logging so we notice if something accidentally does this?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -489,6 +526,11 @@ export const ConsoleClient = {
...
+      return Promise.resolve();
```
gcp: FYI because the function is async, this is a no-op, it's the same as writing `return`.

browser/extensions/felt/rust/src/components.rs
```diff
@@ -153,27 +153,85 @@ impl FeltXPCOM {
...
+                    expires_at
+                );
+                Ok(())
+            }
+            Err(_) => Err(()),
+        };
+
+        match set {
+            Ok(_) => {
+                trace!(
```
gcp, line 177: nit: You log the same right above.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -616,9 +658,12 @@ export const ConsoleClient = {
 
   /**
    * Clears persisted and in-memory token data.
+   *
+   * @param {boolean} isShutdown - Clearing for a shutdown? XPCOM expects a bool
```
gcp: note that `clearTokenData` and `clearTokens` have their argument the exact inverse (this is confusing if you miss the difference).

browser/extensions/felt/api.js
```diff
@@ -221,7 +221,7 @@ this.felt = class extends ExtensionAPI {
           Services.startup.quit(
             Ci.nsIAppStartup.eAttemptQuit | Ci.nsIAppStartup.eConsiderQuit
           );
-        } else {
+        } else if (!this._win) {
```
gcp, line 224: nit: leftover from another commit?

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -52,36 +74,51 @@ def get_tokens(self, driver):
...
+        assert self.felt_tokens[1] == self.browser_tokens[1]
 
-        assert browser_tokens[0] != self.access_token, (
-            f"Access token differs: {browser_tokens[0]} vs {self.access_token}"
+    def run_browser_refresh_forced_tokens(self):
+        self.new_browser_tokens = self.get_tokens(Environment.FIREFOX)
+        assert self.new_browser_tokens[0] != self.browser_tokens[0] and len(
+            self.new_browser_tokens[0]
+        ) == len(self.browser_tokens[0]), (
+            f"Browser access token differs: {self.new_browser_tokens[0]} vs {self.browser_tokens[0]}"
```
gcp: If they are different the assertion won't trigger?

browser/extensions/felt/rust/src/client.rs
gcp: I don't think this was addressed? If you don't want to do it in this PR, please file a bug with the cases to fix in a follow-up?

gcp (changes requested):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -324,6 +324,7 @@ export class FeltProcessParent extends JSProcessActorParent {
       .then(async () => {
         this.sendPrefsToFirefox();
         Services.felt.sendTokens();
+        lazy.ConsoleClient.refreshBlockedInFelt = true;
```
gcp:
```
      lazy.ConsoleClient.refreshBlockedInFelt = true;

      this.proc.exitPromise.finally(() => {
        lazy.ConsoleClient.refreshBlockedInFelt = false;
      });

      this.sendPrefsToFirefox();
      Services.felt.sendTokens();
...
```

```
...
    .catch(err => {
      lazy.ConsoleClient.refreshBlockedInFelt = false;
      throw err;
    });
```

(the idea is that if Firefox startup fails for any reason, you want to make sure Felt recovers and can keep communicating with the console)

gcp (changes requested):

browser/extensions/felt/api.js
lissyx: No, this is from rebasing on https://github.com/mozilla/enterprise-firefox/pull/504, because it's required now (too many things changing in the same place)

testing/enterprise/test_felt_browser_exit_tokens.py
lissyx: probably we want two assertions anyway

browser/extensions/felt/rust/src/client.rs
lissyx: no, i forgot because i was constantly getting interrupted while working :(

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: I did that on purpose, because on the XPCOM API side we should not care about the shutdown information and only care about whether we want to mirror.

lissyx: I'll change the `ConsoleClient` one to remove the shutdown information and focus on mirroring then

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Yes, but I always find it's more explicit this way

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Yes, I was not sure how to report it so far. `console.warn` ?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Looks like it is also the linter's opinion, this is what I get when I remove it:
```
browser/components/enterprise/modules/ConsoleClient.sys.mjs
  537:7  error  Async method '_refreshSession' expected no return value.  consistent-return (eslint)
  600:5  error  Async method '_refreshSession' expected no return value.  consistent-return (eslint)

File: testing/enterprise/test_felt_browser_exit_tokens.py (hunk 8 above)
Comment (gcp) [reply testing/enterprise/test_felt_browser_exit_tokens.py]: I mean the message you log is the exact opposite of what it should be, because you assert they are different, so if you log, it's because they were not.

**Review by gcp — APPROVED** [review]

File: browser/components/enterprise/modules/ConsoleClient.sys.mjs [hunk 10]
```diff
@@ -77,6 +77,16 @@ class InvalidAuthError extends Error {
...
   _refreshPromise: null,
 
+  /**
+   * This is Felt-specific: once the browser is running, it has ownership of
+   * the access token, so any tentative to perform a refresh on Felt side while
+   * the browser is running may trigger a state where Felt gets an updated token
+   * that invalidates the running browser one. Whenever the browser receives an
+   * updated token it will mirror back to Felt, so blocking any fresh during
+   * browser execution.
+   */
```
```suggestion
  /**
   * This is Felt-specific: once the browser is running, it has ownership of
   * the refresh token, so any tentative to perform a refresh on Felt side while
   * the browser is running may trigger a state where Felt gets an updated token
   * that invalidates the running browser one. Whenever the browser receives an
   * updated token it will mirror back to Felt, so blocking any refresh during
   * browser execution.
   */
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -77,6 +77,16 @@ class InvalidAuthError extends Error {
...
+  _refreshBlockedInFelt: false,
```
1rneh: nit:
```suggestion
  _isSessionRefreshBlocked: false,
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh, line 93: nit: 
```suggestion
  /**
   * This is Felt-specific: While the browser is running, it has ownership of the
   * refresh token. We block Felt from performing a session refresh since
   * it would invalidate the tokens that the browser is holding. Instead whenever the 
   * browser performs a session refresh, it mirrors back the updated tokens to Felt.
   * When Firefox is quit, this block is lifted and Felt regains ownership of the tokens.
   */
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -480,6 +490,33 @@ export const ConsoleClient = {
     return accessToken;
   },
 
+  /**
+   * Returns whether refresh is currently blocked in Felt. Always return false
+   * on browser instances.
+   *
+   * @returns {boolean} reflecting whether refresh is currently blocked
+   */
+  get refreshBlockedInFelt() {
```
1rneh: nit: 
```suggestion
  get isSessionRefreshBlocked() {
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -480,6 +490,33 @@ export const ConsoleClient = {
...
+      return this._refreshBlockedInFelt === true;
+    }
+    return false;
+  },
+
+  /**
+   * Sets whether refresh should be blocked or not in Felt. Always force false
+   * on browser instances.
+   *
+   * @param {boolean} value - Blocking refresh in Felt or not
```
```suggestion
   * @param {boolean} value - whether performing a session refresh is blocked
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -480,6 +490,33 @@ export const ConsoleClient = {
...
+   */
+  set refreshBlockedInFelt(value) {
+    if (Services.felt.isFeltUI()) {
+      this._refreshBlockedInFelt = !!value;
+    } else {
+      this._refreshBlockedInFelt = false;
```
1rneh: I guess we only need to set it in felt. Since in the getter we are always returning `false`.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -489,6 +526,13 @@ export const ConsoleClient = {
    * @returns {Promise<void>}
    */
   async _refreshSession() {
+    if (this.refreshBlockedInFelt) {
+      console.warn(
+        `Felt: ConsoleClient: _refreshSession() skipped because refreshBlockedInFelt`
+      );
+      return Promise.resolve();
```
1rneh: Two things here: 
1. In theory we don't expect the `ConsoleClient` to make any session refresh to the console while the browser is running. But if we run into this we should return a rejected Promise. Otherwise the `ConsoleClient` will just send the request again with an access token that is still invalid. What do you think?
2. Can we create an instance of the console to log in the the `ConsoleClient`? And then an error instead of a warning?
```
  return console.createInstance({
    prefix: "ConsoleClient",
    maxLogLevelPref: lazy.EnterpriseCommon.ENTERPRISE_LOGLEVEL_PREF,
  });
});
```

1rneh (changes requested):
One blocker is my comment inside the `_refreshSession` method regarding rejecting on a blocked session refresh in felt, otherwise it looks good to me!

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: > Two things here:
> 
>     1. In theory we don't expect the `ConsoleClient` to make any session refresh to the console while the browser is running. But if we run into this we should return a rejected Promise. Otherwise the `ConsoleClient` will just send the request again with an access token that is still invalid. What do you think?

if we thinkg it's not something that may happen, then yes, rejecting may be a good idea, my first implem had a TODO where I was asking about that, so let's do it

> 
>     2. Can we create an instance of the console to log in the the `ConsoleClient`? And then an error instead of a warning?
> 

sure

> 
> ```
> ChromeUtils.defineLazyGetter(lazy, "log", () => {
>   return console.createInstance({
>     prefix: "ConsoleClient",
>     maxLogLevelPref: lazy.EnterpriseCommon.ENTERPRISE_LOGLEVEL_PREF,
>   });
> });
> ```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Could you also change it for the setter?

testing/enterprise/test_felt_browser_exit_tokens.py
```diff
@@ -23,19 +23,35 @@ def test_browser_exit_tokens(self):
...
+        self._driver.set_context("chrome")
+        refresh_blocked = self._driver.execute_script(
+            """
+            const { ConsoleClient } = ChromeUtils.importESModule("resource:///modules/enterprise/ConsoleClient.sys.mjs");
+            return ConsoleClient.isSessionRefreshBlocked;
+            """
+        )
+        self._driver.set_context("content")
+        assert refresh_blocked == value, (
+            f"ConsoleClient session refresh is {refresh_blocked} expected {value}"
```
1rneh: In case of an assertion error, e.g. expected session refresh to be blocked, this message says`ConsoleClient session refresh is false expected true.` which sounds like you expect the opposite.

Could you make it more descriptive, something like
`Expected performing session refreshs to be unblocked/blocked in Felt.`

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -480,6 +496,31 @@ export const ConsoleClient = {
...
+   */
+  set isSessionRefreshBlocked(value) {
+    if (Services.felt.isFeltUI()) {
+      this._isSessionRefreshBlocked = !!value;
+    }
+  },
+
   /**
    * Refreshes the session using a refresh token.
    * Uses the provided token if given; otherwise the stored token.
```
1rneh: `nit:` That's an outdated comment.
```suggestion
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -489,6 +530,13 @@ export const ConsoleClient = {
    * @returns {Promise<void>}
    */
   async _refreshSession() {
+    if (this.isSessionRefreshBlocked) {
+      lazy.log.error(
+        `Felt: ConsoleClient: _refreshSession() skipped because isSessionRefreshBlocked`
+      );
+      return Promise.reject();
```
1rneh: Sorry, could we explicitly throw here? 

```suggestion
      throw new InvalidAuthError(
          "Token refresh request blocked in Felt.",
          "TOKEN_REFRESH_BLOCKED"
        );
```

1rneh (changes requested):

1rneh (approved):

PR #511, opened 2026-03-09
Bug 2021822 - Enterprise: ensure window observer and window exists be…

lissyx:
…fore unregister/close call

### DIFF
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -189,7 +189,10 @@ this.felt = class extends ExtensionAPI {
       Services.ppmm.addMessageListener("FeltParent:FirefoxNormalExit", this);
       Services.ppmm.addMessageListener("FeltParent:FirefoxLogoutExit", this);
       Services.ppmm.addMessageListener("FeltParent:FirefoxAbnormalExit", this);
-      Services.ppmm.addMessageListener("FeltParent:FirefoxStarting", this);
+      Services.ppmm.addMessageListener(
+        "FeltParent:TransitionFeltToBackground",
+        this
+      );
     } else if (Services.felt.isFeltBrowser()) {
       // In the real Firefox, register observer to handle URLs
       Services.obs.addObserver(this.urlObserver, "felt-open-url");
@@ -239,7 +242,7 @@ this.felt = class extends ExtensionAPI {
         break;
       }
 
-      case "FeltParent:FirefoxStarting": {
+      case "FeltParent:TransitionFeltToBackground": {
         Services.startup.enterLastWindowClosingSurvivalArea();
         this.closeWindow();
         const success = Services.felt.makeBackgroundProcess(true);
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -15,6 +15,12 @@ ChromeUtils.defineESModuleGetters(lazy, {
 
 console.debug(`FeltExtension: FeltParentProcess.sys.mjs`);
 
+const PROCESS_START_REASON = {
+  INITIAL_START: "initial-start",
+  RESTART: "restart",
+  CRASH: "crash",
+};
+
 // Import the shared pending URLs queue
 import {
   gFeltPendingURLs,
@@ -162,7 +168,9 @@ export class FeltProcessParent extends JSProcessActorParent {
                     console.debug(
                       `FeltExtension: ParentProcess: Starting new Firefox`
                     );
-                    gFeltProcessParentInstance.startFirefox();
+                    gFeltProcessParentInstance.startFirefox(
+                      PROCESS_START_REASON.RESTART
+                    );
                   } else {
                     console.debug(
                       `FeltExtension: ParentProcess: Restart disabled, sending normal exit to restore FELT UI`
@@ -283,7 +291,7 @@ export class FeltProcessParent extends JSProcessActorParent {
     );
   }
 
-  async startFirefox(ssoCollectedCookies = []) {
+  async startFirefox(startReason, ssoCollectedCookies = []) {
     this.restartReported = false;
     this.logoutReported = false;
     this.exitReported = false;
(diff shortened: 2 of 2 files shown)


browser/extensions/felt/api.js
```diff
@@ -274,10 +274,16 @@ this.felt = class extends ExtensionAPI {
...
+
+    if (this._winObserver) {
+      Services.ww.unregisterNotification(this._winObserver);
+      this._winObserver = null;
+    }
+
+    if (this._win) {
+      this._win.close();
+      this._win = null;
+    }
```
1rneh: I'm curious on when we get into that state that we call `closeWindow` but it was already closed.

lissyx: I saw it when running/debugging some cases related to browser crashes

lissyx: Oh and the `this._win` was just security, the real thing I saw was the `Services.wm.unregisterNotification` hitting a null

lissyx: This is it, e.g. running the test `testing/enterprise/test_felt_browser_crash_restart.py`:
```
2026-03-12 09:57:14.952270 UTC - [Parent 935808: Main Thread]: D/console debug: "Trying to restart Firefox again."
console.debug: "Trying to restart Firefox again."
console.debug: WebExtensions:
  FeltExtension: FeltParent:FirefoxStarting handling ...
console.debug: WebExtensions:
  FeltExtension: closeWindow: this._win=null
JavaScript error: file:///home/alex/codaz/Mozilla/gecko-cinnabar/obj-felt-dbg/dist/bin/browser/chrome/browser/builtin-addons/felt/api.js, line 278: NS_ERROR_ILLEGAL_VALUE: Component returned failure code: 0x80070057 (NS_ERROR_ILLEGAL_VALUE) [nsIWindowWatcher.unregisterNotification]
```

So we do the `closeWindow()` call and because we do it during crash handling, we have not shown again the window, so there is no window and thus no window observer

lissyx: Another fix would be to share the reason for the start in the `FeltParent:FirefoxStarting` message, and only try to close the window when it is a real start?

lissyx: Or simpler: we never send an exit message that would bring up back the Felt window during restart phases (either explicit restart or a crash), there is thus likely no good reason to send `FeltParent:FirefoxStarting` during those phases to close a window we have not shown

browser/extensions/felt/content/FeltCommon.sys.mjs
```diff
@@ -6,4 +6,7 @@ export const FeltCommon = {
   PRIVATE_BROWSING_ID: 1,
   ENTERPRISE_PROFILE: "enterprise-profile",
   POLICY_POLLING_FREQUENCY: 60_000,
+  PROCESS_START_REASON_START: "start",
+  PROCESS_START_REASON_RESTART: "restart",
+  PROCESS_START_REASON_CRASH: "crash",
 };
```
1rneh, line 9: Could we make this an enum `FIREFOX_PROCESS_START_REASON` and move it to `FeltProcessParent`, since it’s only consumed there?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -283,15 +286,26 @@ export class FeltProcessParent extends JSProcessActorParent {
     );
   }
 
-  async startFirefox(ssoCollectedCookies = []) {
+  async startFirefox(ssoCollectedCookies = [], startReason = undefined) {
```
1rneh:
```suggestion
  async startFirefox(startReason, ssoCollectedCookies = [])
```

I think `startReason` should always be required, while `ssoCollectedCookies` is optional. Can we flip the parameters so `startReason` comes first? 

That way we wouldn’t need to pass an empty array when calling `startFirefox` on restarts.

1rneh (approved):
Thanks for debugging the root cause of the missing window instead of just guarding with `!this._win`. I have two small requests. Otherwise LGTM! :)

browser/extensions/felt/content/FeltCommon.sys.mjs
lissyx: Sure, one iteration of my patch was making use of it in several places, but now it's not required :)

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Sure

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -283,15 +291,26 @@ export class FeltProcessParent extends JSProcessActorParent {
...
     gFeltFirefoxReadyNotified = false;
-    Services.cpmm.sendAsyncMessage("FeltParent:FirefoxStarting", {});
+
+    // There is no message being sent to the message listener on restart phases
+    // whether it is a requested restart from the browser or from a crash.
+    // However in those cases there would have been a start message being sent
+    // making us trying to close a Felt window that was not re-opened.
+    // Since there is no message sent on browser process exit in both cases,
+    // then make sure to also not send a matching starting message.
+    if (startReason === PROCESS_START_REASON.START) {
```
1rneh: One last thing I forgot in my previous review:

Tbh I was more confused after reading that comment than before, although I actually know what's happening here. So my suggestion would be to rename the ipc message to `FeltParent:TransitionFeltToBackground` and then we can simply add a comment saying that sending this message is only required once on the initial startup.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -15,6 +15,12 @@ ChromeUtils.defineESModuleGetters(lazy, {
 
 console.debug(`FeltExtension: FeltParentProcess.sys.mjs`);
 
+const PROCESS_START_REASON = {
+  START: "start",
```
1rneh: Can we rename this to `INITIAL_START`? As it stands, it reads like the reason for the process start is “start,” which doesn’t convey much.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Ok, makes sense

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: I may have to admit I thought about it and forgot by the time I wrote it 😥

PR #513, opened 2026-03-09
Enterprise bug2021872 partner repacks with public and private artifacts

lissyx:


taskcluster/gecko_taskgraph/transforms/partner_repack.py
```diff
@@ -106,6 +106,7 @@ def add_command_arguments(config, tasks):
     # staging releases - pass reduced set of locales to the repacking script
     all_locales = set()
     partner_config = get_partner_config_by_kind(config, config.kind)
+    print(f"partner_repack: partner_config: {partner_config}")
```
lissyx: those are left for now until we have finished the work

ahal: Can you move them to a separate commit locally so they don't land on enterprise-main?

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -1979,6 +1979,35 @@ def add_github_checks(config, tasks):
...
+    task_pairs = {
+        "repackage-deb": "enterprise-repack",
+        "repackage-msi": "enterprise-repack-repackage",
+        "enterprise-repack-mac-notarization": "enterprise-repack-mac-signing",
+    }
+
+    for task in tasks:
+        if "dependencies" in task and config.kind in task_pairs.keys() and task_pairs[config.kind] in list(task["dependencies"].keys()):
+            symbol = task.get("treeherder", {}).get("symbol")
+            repack_id = symbol.split("(")[1].split(")")[0]
```
ahal: Let's not parse treeherder symbols. I see in other places we use `extra["repack_id"]`:
https://github.com/mozilla/enterprise-firefox/blob/enterprise-main/taskcluster/gecko_taskgraph/transforms/repackage_signing_partner.py#L51

Can we use that here as well? If not, we should use an attribute here. You'd need to create a new one for `repack_id`

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -1979,6 +1979,35 @@ def add_github_checks(config, tasks):
         yield task
 
 
+@transforms.add
+def set_artifact_prefix_enterprise_repacks(config, tasks):
```
ahal: This shouldn't go in `task.py`. This file is meant only for very general purpose things that can apply to every task. Please add it to a transform file that is specific to the tasks that need this, possibly creating a new file if necessary.

ahal (changes requested):

taskcluster/gecko_taskgraph/transforms/task.py
lissyx: It was not exposed, I'm not sure how to add such attribute here

taskcluster/gecko_taskgraph/transforms/task.py
lissyx: I did this out of ease and flexibility, before this solution came up

taskcluster/kinds/repackage-deb/kind.yml
```diff
@@ -11,6 +11,7 @@ transforms:
     - gecko_taskgraph.transforms.repackage_routes
     - gecko_taskgraph.transforms.job
     - gecko_taskgraph.transforms.task
+    - gecko_taskgraph.transforms.enterprise
```
ahal: Can you put this before the `job` and `task` transforms? Generally we try to order the transforms by specificity.. e.g the first ones are highly specific to the kind, and they get less and less specific until `tasks` which is the _last_ transform and applies to _all_ tasks.

taskcluster/gecko_taskgraph/parameters.py
```diff
@@ -139,6 +139,123 @@ def get_release_type(parameters):
...
+                        "linux64-enterprise-shippable",
+                        "macosx64-enterprise-shippable",
+                        "win64-enterprise-shippable",
+                    ],
+                },
+            },
+        },
+    }
+
+
```
ahal: These aren't parameters. I'd suggest putting them in `gecko_taskgraph/util/partners.py`?

taskcluster/gecko_taskgraph/transforms/enterprise.py
```diff
@@ -0,0 +1,84 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from taskgraph.transforms.base import TransformSequence
+
+transforms = TransformSequence()
+
+
+@transforms.add
+def set_artifact_prefix_enterprise_repacks(config, tasks):
```
ahal: I really don't like modifying / deleting things after the fact. Somewhere (maybe in the `repackage.py` transforms?) we're creating these artifacts in the first place. That's where we should be setting the artifact prefix.

ahal (changes requested):

taskcluster/gecko_taskgraph/transforms/enterprise.py
```diff
@@ -0,0 +1,42 @@
...
+            repack_repo = repack_id.split("/")[0]
+            repack_name = repack_id.split("/")[1]
+            repack_config = config.params["release_partner_config"][config.kind][
+                repack_repo
+            ][repack_name]
+
+            # enterprise-repack will produce all repacks so it cannot have a mix of public and private
+            public = (
+                "public" in repack_config.keys()
+                and repack_config["public"]
```
ahal: nit: repack_config.get("public")

ahal (approved):

PR #517, opened 2026-03-09
Bug 2018619 - fix restricted error page

lissyx:


testing/enterprise/felt_tests.py
```diff
@@ -145,7 +145,7 @@ def do_GET(self):
                 self.forbidden()
                 return
 
-            location = f"http://localhost:{self.server.sso_port}/sso_url"
+            location = "https://wrong.host.badssl.com/sso_url"  # f"http://localhost:{self.server.sso_port}/sso_url"
```
lissyx: This is a quick hack, I am unsure if we can really have a test at marionette level at all, and hitting the network during marionette testing is not OK. Problem is that any https over localhost we could hack even with incorrect certs ends up being accepted by Firefox ...

docshell/base/nsDocShell.cpp
```diff
@@ -3743,8 +3743,12 @@ nsDocShell::DisplayLoadError(nsresult aError, nsIURI* aURI,
         break;
       case NS_ERROR_BLOCKED_BY_POLICY:
         // Page blocked by policy
-        error = "blockedByPolicy";
+#if defined(MOZ_ENTERPRISE)
+        error = "blockedByPolicyEnterprise";
         enterpriseErrorDescriptionID = "blockedByPolicyEnterprise";
```
Mossop, line 3748: I don't think this is needed now it is matching the error code.

Mossop (approved):

docshell/base/nsDocShell.cpp
lissyx: we still need to load the string from our string bundle, so I think we still need it?

Mossop: Oh I see it's causing us to use a different string bundle. Could just be a boolean instead but I guess it doesn't really matter much either way.

toolkit/content/errors/net-errors.mjs
1rneh, line 321: `neterror-blocked-by-policy-page-title` was removed. Could you restore it in `toolkit/locales/en-US/toolkit/neterror/netError.ftl`?

Also I noticed `blockedByPolicy-title` replaced with the enterprise id in `KNOWN_ERROR_TITLE_IDS` in `aboutNetError.mjs`. Could you also restore that, so that we don't break the regular m-c build?

toolkit/content/errors/net-errors.mjs
```diff
@@ -328,18 +323,28 @@ export const NET_ERRORS = [
...
+      showTryAgain: false,
+      showGoBack: false,
+    },
+    hasNoUserFix: true,
+  },
+  {
+    id: "blockedByPolicyEnterprise",
+    errorCode: "blockedByPolicyEnterprise",
+    category: "blocked",
+    bodyTitleL10nId: "neterror-blocked-by-policy-page-title",
```
1rneh:
```suggestion
    bodyTitleL10nId: "blocked-by-policy-title-enterprise",
```

1rneh (changes requested):

toolkit/content/errors/net-errors.mjs
```diff
@@ -328,18 +323,28 @@ export const NET_ERRORS = [
...
+    introContent: {
+      dataL10nId: "fp-neterror-offline-intro",
+      dataL10nArgs: { hostname: null },
+    },
+    descriptionParts: [
+      {
+        tag: "p",
+        dataL10nId: "neterror-blocked-by-policy-contact-admin",
+      },
+    ],
```
1rneh, line 347: Let's also add 
```
customNetError: {
      titleL10nId: "blocked-by-policy-title-enterprise",
      whatCanYouDoL10nId: "neterror-blocked-by-policy-contact-admin",
    },
```
Unfortunately we can't use `whyDidThisHappenL10nId: "blockedByPolicyEnterprise"` right away, because `blockedByPolicyEnterprise` is not a fluent id. 
Let's check with Matt whether we need string adjustments for the updated net error pages (can be done in a follow up).

1rneh (approved):
Please add the `customNetError` part. Other than that it looks fine.

toolkit/content/errors/net-errors.mjs
1rneh: We can use https://bugzilla.mozilla.org/show_bug.cgi?id=2022201 as follow-up. I already added some context there and pinged Matt.

PR #519, opened 2026-03-10
Bug 2022201 - Fix neterror pages

1rneh:
@Mossop Which way would you put these changes behind an enterprise flag?

So far I tried these two ways:
- import `AppConstants` with `ChromeUtils` but `ChromeUtils` isn't accessible from the content file `net-error.mjs`.
- import it statically but eslint warns that I can't import system modules from non-system modules see [eslint rule](https://firefox-source-docs.mozilla.org/code-quality/lint/linters/eslint-plugin-mozilla/rules/reject-import-system-module-from-non-system.html).

### DIFF
--- a/toolkit/locales/en-US/toolkit/neterror/netError.ftl
+++ b/toolkit/locales/en-US/toolkit/neterror/netError.ftl
@@ -7,6 +7,7 @@
 neterror-page-title = Problem loading page
 certerror-page-title = Warning: Potential Security Risk Ahead
 certerror-sts-page-title = Did Not Connect: Potential Security Issue
+neterror-blocked-by-policy-page-title = Blocked Page
 neterror-captive-portal-page-title = Log in to network
 neterror-dns-not-found-title = Server Not Found
 neterror-malformed-uri-page-title = Invalid URL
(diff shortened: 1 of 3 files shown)


lissyx:
This may be a better fix than https://github.com/mozilla/enterprise-firefox/pull/517

Mossop:
Oh if the only error is about importing system modules from non-system modules then I think we can ignore that for AppConstants.

toolkit/locales/en-US/toolkit/neterror/netError.ftl
```diff
@@ -7,6 +7,7 @@
 neterror-page-title = Problem loading page
 certerror-page-title = Warning: Potential Security Risk Ahead
 certerror-sts-page-title = Did Not Connect: Potential Security Issue
+neterror-blocked-by-policy-page-title = Blocked Page
```
lissyx, line 10: is it a string we removed? that looks weird to add it and not refer to it in the code

1rneh:
> Oh if the only error is about importing system modules from non-system modules then I think we can ignore that for AppConstants.

I tried it, but it's disallowed: `Security Error: Content at about:neterror?e=blockedByPolicy may not load or link to resource://gre/modules/AppConstants.sys.mjs.`

1rneh:
This fix has already landed in #517.

PR #524, opened 2026-03-10
Bug 2022263 - Enterprise: handle about:certerror/about:neterror in FELT

lissyx:
Bug-2022263

### DIFF
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -232,36 +232,6 @@ export const ConsoleClient = {
   },
 
   /**
-   * Gets the error name for a channel status code.
-   *
-   * @param {number} status - The channel status code
-   * @returns {string} Human-readable error name
-   */
-  _getErrorNameForStatus(status) {
-    try {
-      const nssErrorsService = Cc[
-        "@mozilla.org/nss_errors_service;1"
-      ].getService(Ci.nsINSSErrorsService);
-      return nssErrorsService.getErrorName(status);
-    } catch {
-      // Not an NSS error, check common network error codes
-
-      // Mapping here should follow what nsDocShell::DisplayLoadError uses.
-      // Consumer code will expect those to fail when using Fluent to format
-      // and perform fallback to string bundles where they are defined.
-      const networkErrors = {
-        [Cr.NS_ERROR_UNKNOWN_HOST]: "dnsNotFound2",
-        [Cr.NS_ERROR_CONNECTION_REFUSED]: "connectionFailure",
-        [Cr.NS_ERROR_NET_TIMEOUT]: "netTimeout",
-        [Cr.NS_ERROR_NET_RESET]: "netReset",
-        [Cr.NS_ERROR_NET_INTERRUPT]: "netInterrupt",
-        [Cr.NS_ERROR_OFFLINE]: "netOffline",
-      };
-      return networkErrors[status] || "network";
-    }
-  },
-
-  /**
    * Fetch-like wrapper that exposes detailed network errors.
    * Uses XMLHttpRequest internally to access channel.status on error,
    * which native fetch() does not expose.
@@ -309,9 +279,11 @@ export const ConsoleClient = {
       };
 
       xhr.onerror = () => {
-        const errorName = this._getErrorNameForStatus(xhr.channel?.status);
+        const channelStatus = xhr.channel?.status ?? null;
         reject(
-          new TypeError(errorName, { cause: { hostname: new URL(url).host } })
+          new TypeError("ConsoleClientXHRError", {
+            cause: { hostname: new URL(url).host, channelStatus },
+          })
         );
       };
 
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -311,7 +311,7 @@ export const ConsoleClient = {
       xhr.onerror = () => {
         const errorName = this._getErrorNameForStatus(xhr.channel?.status);
         reject(
-          new TypeError(errorName, { cause: { host: new URL(url).host } })
+          new TypeError(errorName, { cause: { hostname: new URL(url).host } })
         );
       };
 
--- a/browser/extensions/felt/FeltErrorReport.sys.mjs
+++ b/browser/extensions/felt/FeltErrorReport.sys.mjs
@@ -42,6 +42,35 @@ export const FeltErrorReport = {
    */
   _noNetworkErrors: ["netOffline", "dnsNotFound"],
 
+  /**
+   * Gets the error fluent id for a channel status code.
+   *
+   * @param {number} status - The channel status code
+   * @returns {string} Fluent ID
+   */
+  getFluentIdForStatus(status) {
+    lazy.log.debug(`getFluentIdForStatus(${status})`);
+    try {
+      const nssErrorsService = Cc[
+        "@mozilla.org/nss_errors_service;1"
+      ].getService(Ci.nsINSSErrorsService);
+      return nssErrorsService.getErrorName(status);
+    } catch {
+      // Not an NSS error, check common network error codes
+
+      // Mapping here should follow what nsDocShell::DisplayLoadError uses.
+      const networkErrors = {
+        [Cr.NS_ERROR_UNKNOWN_HOST]: "dnsNotFound",
+        [Cr.NS_ERROR_CONNECTION_REFUSED]: "connectionFailure",
+        [Cr.NS_ERROR_NET_TIMEOUT]: "netTimeout",
+        [Cr.NS_ERROR_NET_RESET]: "netReset",
+        [Cr.NS_ERROR_NET_INTERRUPT]: "netInterrupt",
+        [Cr.NS_ERROR_OFFLINE]: "netOffline",
+      };
+      return networkErrors[status] ?? "network";
+    }
+  },
+
   init(doc) {
     this._document = doc;
     lazy.initializeRegistry();
@@ -226,6 +255,23 @@ export const FeltErrorReport = {
       .querySelector(".felt-login__email-pane")
       .classList.remove("is-hidden");
     this._document.querySelector(".felt-login__sso").classList.add("is-hidden");
+  },
+
+  /**
+   * Handle displaying an error produced by XmlHttpRequest.
+   *
+   * @param {object} aError - The TypeError containing the XHR channel status
+   *                          code in aError.cause.channelStatus and the
+   *                          hostname of the URL triggering the error in
+   *                          aError.cause.hostname.
+   */
+  async handleXhrError(aError) {
+    if (aError.cause) {
+      const errorCode = this.getFluentIdForStatus(aError.cause.channelStatus);
+      await this.updateNetworkError(errorCode, aError.cause, ERROR_SOURCE.XHR);
+    } else {
+      await this.updateNetworkError("network", null, ERROR_SOURCE.XHR);
+    }
   },
 
   /**
--- a/browser/extensions/felt/FeltErrorReport.sys.mjs
+++ b/browser/extensions/felt/FeltErrorReport.sys.mjs
@@ -56,6 +56,9 @@ export const FeltErrorReport = {
       ].getService(Ci.nsINSSErrorsService);
       return nssErrorsService.getErrorName(status);
     } catch {
+      lazy.log.debug(
+        `getFluentIdForStatus(${status}) no nssErrorsService.getErrorName`
+      );
       // Not an NSS error, check common network error codes
 
       // Mapping here should follow what nsDocShell::DisplayLoadError uses.
@@ -67,6 +70,9 @@ export const FeltErrorReport = {
         [Cr.NS_ERROR_NET_INTERRUPT]: "netInterrupt",
         [Cr.NS_ERROR_OFFLINE]: "netOffline",
       };
+      lazy.log.debug(
+        `getFluentIdForStatus(${status}) networkErrors[status]:${networkErrors[status]}`
+      );
       return networkErrors[status] ?? "network";
     }
   },
@@ -91,6 +97,7 @@ export const FeltErrorReport = {
   },
 
   async _maybeResetErrorElement(errorElement) {
+    lazy.log.debug(`_maybeResetErrorElement()`);
     errorElement.dismiss();
     errorElement.removeAttribute("source");
 
@@ -128,9 +135,12 @@ export const FeltErrorReport = {
     context,
     detailsElement = null
   ) {
+    lazy.log.debug(`_applyErrorConfig(${errorCode})`);
+
     try {
       const errorConfig = lazy.getResolvedErrorConfig(errorCode, context);
       if (errorConfig.introContent) {
+        lazy.log.debug(`_applyErrorConfig(${errorCode}): resolved errorConfig`);
         this._document.l10n.setAttributes(
           detailsElement,
           errorConfig.introContent.dataL10nId,
@@ -138,6 +148,9 @@ export const FeltErrorReport = {
         );
       }
       if (errorConfig.customNetError && context.canUseCustomNetError) {
+        lazy.log.debug(
+          `_applyErrorConfig(${errorCode}): errorConfig.customNetError`
+        );
         errorElement.setAttribute(
           "heading",
           await this._document.l10n.formatValue(
@@ -177,6 +190,8 @@ export const FeltErrorReport = {
    * @param {string} source - The source of the error, from ERROR_SOURCE
    */
   async update(errorType, errorCode = null, context = null, source = null) {
+    lazy.log.debug(`update(${errorType}, ${errorCode})`);
+
     const errorElement = this._wrapper.querySelector(`.${errorType}`);
     if (!errorElement) {
       lazy.log.error(`Error element ${errorType} not found`);
@@ -201,6 +216,9 @@ export const FeltErrorReport = {
 
         const l10n = this._document.l10n.getAttributes(detailsElement);
         if (!l10n?.id) {
+          lazy.log.debug(
+            `update(${errorType}): this._applyErrorConfig missing, fallback to felt-error-${errorCode}`
+          );
           const message = await this._document.l10n.formatValue(
             `felt-error-${errorCode}`
           );
@@ -226,14 +244,23 @@ export const FeltErrorReport = {
    *                         'u' parameter.
    */
   async handleNetError(aData) {
+    const errorPageURI = aData.errorPageURI;
+    lazy.log.debug(`handleNetError(${errorPageURI})`);
+
     try {
-      const parsedErrorPage = new URL(aData.errorPageURI);
+      const parsedErrorPage = new URL(errorPageURI);
       const errorCode = (() => {
         const parsedErrorCode = parsedErrorPage.searchParams.get("e");
         const mappedErrorCode = this._errorCodesMapping[parsedErrorCode];
         if (mappedErrorCode) {
+          lazy.log.debug(
+            `handleNetError(${errorPageURI}): mapping ${parsedErrorCode} to ${mappedErrorCode}`
+          );
           return mappedErrorCode;
         }
+        lazy.log.debug(
+          `handleNetError(${errorPageURI}): returning ${parsedErrorCode}`
+        );
         return parsedErrorCode;
       })();
 
@@ -248,6 +275,9 @@ export const FeltErrorReport = {
         ERROR_SOURCE.NET
       );
     } catch (ex) {
+      lazy.log.debug(
+        `handleNetError: exception ${ex}, reverting to default network error`
+      );
       await this.updateNetworkError("network", null, ERROR_SOURCE.NET);
     }
 
@@ -266,10 +296,17 @@ export const FeltErrorReport = {
    *                          aError.cause.hostname.
    */
   async handleXhrError(aError) {
+    lazy.log.debug(
+      `handleXhrError(aError.message=${aError.message} aError.cause=${JSON.stringify(aError.cause)})`
+    );
     if (aError.cause) {
       const errorCode = this.getFluentIdForStatus(aError.cause.channelStatus);
+      lazy.log.debug(
+        `handleXhrError() using errorCode:${errorCode} for ${aError.cause.channelStatus}`
+      );
       await this.updateNetworkError(errorCode, aError.cause, ERROR_SOURCE.XHR);
     } else {
+      lazy.log.debug(`handleXhrError: received ${aError} without a cause`);
       await this.updateNetworkError("network", null, ERROR_SOURCE.XHR);
     }
   },
@@ -286,6 +323,8 @@ export const FeltErrorReport = {
    *                          ERROR_SOURCE.NET. Useful for assertions in tests.
    */
   async updateNetworkError(errorCode, context, source) {
+    lazy.log.debug(`updateNetworkError(${errorCode})`);
+
     if (this._noNetworkErrors.includes(errorCode)) {
       await this.update(
         "felt-browser-error-no-network",
--- a/browser/extensions/felt/FeltErrorReport.sys.mjs
+++ b/browser/extensions/felt/FeltErrorReport.sys.mjs
@@ -1,0 +1,259 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  createEnterpriseLogger:
+    "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
+  getResolvedErrorConfig: "chrome://global/content/errors/error-lookup.mjs",
+  initializeRegistry: "chrome://global/content/errors/error-registry.mjs",
+});
+
+ChromeUtils.defineLazyGetter(lazy, "log", () => {
+  return lazy.createEnterpriseLogger("FeltErrorReport");
+});
+
+export const ERROR_SOURCE = {
+  XHR: "xhr",
+  NET: "net",
+  RESET: "reset",
+};
+
+export const FeltErrorReport = {
+  _wrapper: null,
+
+  /* Per UX requirements, there are some error messages where it is required
+   * there is more control over on enterprise builds: it can be because the
+   * way information is displayed is not as efficient as the original
+   * about:neterror page and string needs rewording, incorrect casing,
+   * duplicated error messages that may refers to UI elements that do not
+   * exists, etc.
+   */
+  _errorCodesMapping: {
+    dnsNotFound: "dnsNotFound-enterprise",
+    netOffline: "NS_ERROR_OFFLINE",
+  },
+
+  /* Show simpler "No Network Connection" only for truly offline scenarios:
+   *  - netOffline for offline mode,
+   *  - dnsNotFound for actual network disconnect
+   */
+  _noNetworkErrors: ["netOffline", "dnsNotFound"],
+
+  init(doc) {
+    this._document = doc;
+    lazy.initializeRegistry();
+    this._wrapper = this._document.querySelector(".felt-browser-error");
+    this._wrapper.addEventListener("message-bar:user-dismissed", e => {
+      e.preventDefault();
+      e.target.classList.add("is-hidden");
+    });
+  },
+
+  reset() {
+    if (!this._wrapper) {
+      return;
+    }
+    for (const bar of this._wrapper.querySelectorAll("moz-message-bar")) {
+      bar.classList.add("is-hidden");
+    }
+  },
+
+  async _maybeResetErrorElement(errorElement) {
+    errorElement.dismiss();
+    errorElement.removeAttribute("source");
+
+    errorElement
+      .querySelectorAll(".felt-browser-error-details[added]")
+      .forEach(el => el.remove());
+
+    // Remove so next l10n.setAttributes() will re-translate from original HTML
+    errorElement.removeAttribute("heading");
+    if (errorElement.dataset.l10nId) {
+      this._document.l10n.setAttributes(
+        errorElement,
+        errorElement.dataset.l10nId
+      );
+      await this._document.l10n.translateElements([errorElement]);
+    }
+
+    return errorElement;
+  },
+
+  /**
+   * Leverages existing about:certerror/about:neterror string IDs and error
+   * handling to update the error message displayed to the user
+   *
+   * @param {object} errorElement - The HTML element holding the error message
+   * @param {string} errorCode - The specific error code used for displaying the error message.
+   * @param {object} context - Additional context for the error.
+   * @param {string} context.hostname - The hostname related to the error.
+   * @param {boolean} context.canUseCustomNetError - Indicates if a custom network error can be used.
+   * @param {object} detailsElement - The HTML element holding the error message details, if used.
+   */
+  async _applyErrorConfig(
+    errorElement,
+    errorCode,
+    context,
+    detailsElement = null
+  ) {
+    try {
+      const errorConfig = lazy.getResolvedErrorConfig(errorCode, context);
+      if (errorConfig.introContent) {
+        this._document.l10n.setAttributes(
+          detailsElement,
+          errorConfig.introContent.dataL10nId,
+          errorConfig.introContent.dataL10nArgs
+        );
+      }
+      if (errorConfig.customNetError && context.canUseCustomNetError) {
+        errorElement.setAttribute(
+          "heading",
+          await this._document.l10n.formatValue(
+            errorConfig.customNetError.titleL10nId
+          )
+        );
+        if (errorConfig.customNetError.whatCanYouDoL10nId) {
+          this._document.l10n.setAttributes(
+            detailsElement,
+            errorConfig.customNetError.whatCanYouDoL10nId,
+            errorConfig.customNetError.whatCanYouDoL10nArgs
+          );
+        } else if (errorConfig.customNetError.whatCanYouDoItems) {
+          errorConfig.customNetError.whatCanYouDoItems.forEach(itemL10nId => {
+            const itemElement = this._document.createElement("span");
+            itemElement.slot = "message";
+            itemElement.className = "felt-browser-error-details";
+            itemElement.setAttribute("added", "true");
+            this._document.l10n.setAttributes(itemElement, itemL10nId);
+            errorElement.appendChild(itemElement);
+          });
+        }
+      }
+    } catch (ex) {
+      lazy.log.debug(
+        `_applyErrorConfig(${errorCode}) not resolved: ${ex} -- Falling back`
+      );
+    }
+  },
+
+  /**
+   * Updates the error display based on the error type and details.
+   *
+   * @param {string} errorType - The type of error to display.
+   * @param {string|null} errorCode - The specific error code used for displaying the error message.
+   * @param {object|null} context - Additional context for the error.
+   * @param {string} source - The source of the error, from ERROR_SOURCE
+   */
+  async update(errorType, errorCode = null, context = null, source = null) {
+    const errorElement = this._wrapper.querySelector(`.${errorType}`);
+    if (!errorElement) {
+      lazy.log.error(`Error element ${errorType} not found`);
+      return;
+    }
+
+    await this._maybeResetErrorElement(errorElement);
+
+    if (errorCode) {
+      const detailsElement = errorElement.querySelector(
+        ".felt-browser-error-details"
+      );
+      if (detailsElement) {
+        if (context) {
+          await this._applyErrorConfig(
+            errorElement,
+            errorCode,
+            context,
+            detailsElement
+          );
+        }
+
+        const l10n = this._document.l10n.getAttributes(detailsElement);
+        if (!l10n?.id) {
+          const message = await this._document.l10n.formatValue(
+            `felt-error-${errorCode}`
+          );
+          detailsElement.textContent = message || errorCode;
+        }
+      }
+    }
+
+    if (source) {
+      errorElement.setAttribute("source", source);
+    }
+
+    errorElement.classList.remove("is-hidden");
+  },
+
+  /**
+   * Handle displaying an error produced by about:neterror/about:certerror
+   * sent from FeltErrorWindowChild.
+   *
+   * @param {object} aData - The message containing the error URL in 'errorPageURI'
+   *                         field which is a URL containing error code as 'e'
+   *                         parameter and URL triggering the error as
+   *                         'u' parameter.
+   */
+  async handleNetError(aData) {
+    try {
+      const parsedErrorPage = new URL(aData.errorPageURI);
+      const errorCode = (() => {
+        const parsedErrorCode = parsedErrorPage.searchParams.get("e");
+        const mappedErrorCode = this._errorCodesMapping[parsedErrorCode];
+        if (mappedErrorCode) {
+          return mappedErrorCode;
+        }
+        return parsedErrorCode;
+      })();
+
+      const errorHostname = new URL(parsedErrorPage.searchParams.get("u")).host;
+
+      await this.updateNetworkError(
+        errorCode,
+        {
+          hostname: errorHostname,
+          canUseCustomNetError: true,
+        },
+        ERROR_SOURCE.NET
+      );
+    } catch (ex) {
+      await this.updateNetworkError("network", null, ERROR_SOURCE.NET);
+    }
+
+    this._document
+      .querySelector(".felt-login__email-pane")
+      .classList.remove("is-hidden");
+    this._document.querySelector(".felt-login__sso").classList.add("is-hidden");
+  },
+
+  /**
+   * Handle displaying a network error generated from about: or XHR.
+   * Error codes are filtered to display either the error connection or the
+   * no network display.
+   *
+   * @param {string} errorCode - The error code as a string value
+   * @param {object} context - The error context, with hostname field holding the
+   *                         value of the host triggering the error
+   * @param {string} source - The source of the error, ERROR_SOURCE.XHR or
+   *                          ERROR_SOURCE.NET. Useful for assertions in tests.
+   */
+  async updateNetworkError(errorCode, context, source) {
+    if (this._noNetworkErrors.includes(errorCode)) {
+      await this.update(
+        "felt-browser-error-no-network",
+        "no-network-connection",
+        null,
+        source
+      );
+    } else {
+      await this.update(
+        "felt-browser-error-connection",
+        errorCode,
+        context,
+        source
+      );
+    }
+  },
+};
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -32,6 +32,7 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
 this.felt = class extends ExtensionAPI {
   FELT_PROCESS_ACTOR = "FeltProcess";
   FELT_WINDOW_ACTOR = "FeltWindow";
+  FELT_ERROR_WINDOW_ACTOR = "FeltErrorWindow";
 
   registerChrome() {
     let aomStartup = Cc[
@@ -66,6 +67,23 @@ this.felt = class extends ExtensionAPI {
       },
       allFrames: true,
       matches,
+    });
+
+    // Remove existing error handler installed by default so we can install
+    // our own to provide a dedicated UI.
+    ChromeUtils.unregisterWindowActor("NetError");
+    ChromeUtils.registerWindowActor(this.FELT_ERROR_WINDOW_ACTOR, {
+      child: {
+        esModuleURI: "chrome://felt/content/FeltErrorWindowChild.sys.mjs",
+        events: {
+          DOMContentLoaded: {},
+        },
+      },
+      parent: {
+        esModuleURI: "chrome://felt/content/FeltErrorWindowParent.sys.mjs",
+      },
+      allFrames: true,
+      matches: ["about:certerror?*", "about:neterror?*"],
     });
 
     // We use a much simpler version of the context menu so replace the default actor with our own.
@@ -434,6 +452,7 @@ this.felt = class extends ExtensionAPI {
 
     if (Services.felt.isFeltUI()) {
       ChromeUtils.unregisterWindowActor(this.FELT_WINDOW_ACTOR);
+      ChromeUtils.unregisterWindowActor(this.FELT_ERROR_WINDOW_ACTOR);
       ChromeUtils.unregisterProcessActor(this.FELT_PROCESS_ACTOR);
       lazy.FeltStorage.uninit();
     }
--- a/browser/extensions/felt/content/FeltErrorWindowChild.sys.mjs
+++ b/browser/extensions/felt/content/FeltErrorWindowChild.sys.mjs
@@ -1,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  createEnterpriseLogger:
+    "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
+});
+
+ChromeUtils.defineLazyGetter(lazy, "log", () => {
+  return lazy.createEnterpriseLogger("FeltErrorWindowChild");
+});
+
+/**
+ * Error-capture dedicated window actor. It is expected to be only setup on
+ * matching pages like about:neterror/about:certerror and is there to pass the
+ * error page URI information back to the parent-side of this window actor for
+ * further processing, including extracting the error code to later display it
+ * properly.
+ *
+ * It relies on inspecting the docShell that contains several nsIChannel:
+ *  - failedChannel: this channel is related to the **real** URL that was trying
+ *    to be accessed, i.e. the page that when loading triggered the error.
+ *
+ *  - currentDocumentChannel: this channel is related to the **loaded** content,
+ *    so it is exactly what is needed here, since it holds the about: error
+ *    page's URL.
+ */
+export class FeltErrorWindowChild extends JSWindowActorChild {
+  handleEvent(aEvent) {
+    if (aEvent.type !== "DOMContentLoaded") {
+      lazy.log.error(`Unexpected aEvent.type=${aEvent.type}`);
+      return;
+    }
+
+    lazy.log.debug(
+      `Notify parent with errorPageURI: ${this.docShell.currentDocumentChannel.originalURI.spec}`
+    );
+    this.sendAsyncMessage("ErrorReport", {
+      errorPageURI: this.docShell.currentDocumentChannel.originalURI.spec,
+    });
+  }
+}
--- a/browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
@@ -7,6 +7,7 @@ const lazy = {};
 ChromeUtils.defineESModuleGetters(lazy, {
   createEnterpriseLogger:
     "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
+  FeltErrorReport: "resource:///modules/FeltErrorReport.sys.mjs",
 });
 
 ChromeUtils.defineLazyGetter(lazy, "log", () => {
@@ -25,7 +26,10 @@ export class FeltErrorWindowParent extends JSWindowActorParent {
     );
     switch (message.name) {
       case "ErrorReport": {
-        // Nothing for now.
+        const errorPage = message.data;
+        lazy.FeltErrorReport.handleNetError(errorPage).catch(err =>
+          lazy.log.error(`Error while handling ErrorReport message: ${err}`)
+        );
         break;
       }
     }
--- a/browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
@@ -1,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  createEnterpriseLogger:
+    "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
+});
+
+ChromeUtils.defineLazyGetter(lazy, "log", () => {
+  return lazy.createEnterpriseLogger("FeltErrorWindowParent");
+});
+
+/**
+ * Parent-side of the Felt error window actor to collect the error page URL. It
+ * receives the URL of the loaded error page and forwards that to the error
+ * report component.
+ */
+export class FeltErrorWindowParent extends JSWindowActorParent {
+  receiveMessage(message) {
+    lazy.log.debug(
+      `Received message ${message.name} => ${JSON.stringify(message.data)}`
+    );
+    switch (message.name) {
+      case "ErrorReport": {
+        // Nothing for now.
+        break;
+      }
+    }
+  }
+}
--- a/browser/extensions/felt/content/window.js
+++ b/browser/extensions/felt/content/window.js
@@ -31,7 +31,7 @@ async function connectToConsole(email) {
   try {
     posture = await lazy.ConsoleClient.sendDevicePosture();
   } catch (err) {
-    lazy.log.error(`FeltExtension: Failed to connect to console: ${err}`);
+    lazy.log.error(`Failed to send device posture: ${err}`);
     await lazy.FeltErrorReport.handleXhrError(err);
     return;
   }
@@ -83,7 +83,7 @@ async function connectToConsole(email) {
   lazy.log.debug(
     `FeltExtension: created contentPrincipal with privateBrowsingId=${contentPrincipal.privateBrowsingId}`
   );
-  lazy.log.debug("Load SSO URI: ", ssoLoginURI);
+  lazy.log.debug("Load SSO URI: ", ssoLoginURI.spec);
   browser.fixupAndLoadURIString(ssoLoginURI.spec, {
     triggeringPrincipal: contentPrincipal,
   });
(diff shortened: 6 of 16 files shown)


browser/extensions/felt/content/FeltErrorWindowChild.sys.mjs
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+console.debug(`FeltExtension: FeltErrorWindowChild.sys.mjs`);
+
+/**
+ *
+ */
```
1rneh, line 30: Empty JSDoc.

browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+console.debug(`FeltExtension: FeltErrorWindowParent.sys.mjs`);
+
+/**
+ *
+ */
```
1rneh, line 21: Empty JSDoc.

1rneh (changes requested):
Nice! Thank you!

This might be a bit of a scope creep but I think it would improve the code structure a lot.

Could we convert `ErrorReport` from `window.js` to be a system module (`FeltErrorReport.sys.mjs`). In `window.js` we'd initialize it as we do with `Updates.sys.mjs` and pass it the `document`. This way `FeltErrorWindowParent` (and other consumers) could directly access `FeltErrorReport.update` and wouldn't need to take the detour to send a custom event.

Note: Since #736 has just landed, you'll now need to update your log messages.

lissyx:
> Nice! Thank you!
> 
> This might be a bit of a scope creep but I think it would improve the code structure a lot.
> 
> Could we convert `ErrorReport` from `window.js` to be a system module (`FeltErrorReport.sys.mjs`). In `window.js` we'd initialize it as we do with `Updates.sys.mjs` and pass it the `document`. This way `FeltErrorWindowParent` (and other consumers) could directly access `FeltErrorReport.update` and wouldn't need to take the detour to send a custom event.
> 
> Note: Since #736 has just landed, you'll now need to update your log messages.

I don't see a problem for moving it around but I need to check we may still needs those events for cross process communication

1rneh:
> I don't see a problem for moving it around but I need to check we may still needs those events for cross process communication

The `FeltErrorReport` lives in the parent process and so does the `FeltErrorWindowParent`, so I believe no cross-process communication is needed.

lissyx:
> > I don't see a problem for moving it around but I need to check we may still needs those events for cross process communication
> 
> The `FeltErrorReport` lives in the parent process and so does the `FeltErrorWindowParent`, so I believe no cross-process communication is needed.

I also believe so but I'll feel better once I did the change and I'm sure about it:)

lissyx:
> Nice! Thank you!
> 
> This might be a bit of a scope creep but I think it would improve the code structure a lot.
> 
> Could we convert `ErrorReport` from `window.js` to be a system module (`FeltErrorReport.sys.mjs`). In `window.js` we'd initialize it as we do with `Updates.sys.mjs` and pass it the `document`. This way `FeltErrorWindowParent` (and other consumers) could directly access `FeltErrorReport.update` and wouldn't need to take the detour to send a custom event.

Moved it as suggested

> 
> Note: Since #736 has just landed, you'll now need to update your log messages.

Done

lissyx:
Some UX, from the added test to cover a couple of classic `about:neterror` cases. Heading and filenames patches the `errorCode` from https://searchfox.org/enterprise-main/rev/1341260d60d8c0f3f20e8baf7c6e51de31dba2a0/toolkit/content/errors/net-errors.mjs which is what `about:neterror` uses to perform the display.

I've tried to surface as many as possible valuable information from the error definitions in that file, especially regarding the "what can you do" elements that gives suggestions about unblocking.

testing/enterprise/test_felt_browser_verify_neterror.py
```diff
@@ -0,0 +1,106 @@
...
+        ]
+        for m in object_methods:
+            getattr(self, m)()
+
+    # Generated from errorCode that can be found in toolkit/content/errors/net-errors.mjs:
+    # $ for e in $(grep "errorCode:" toolkit/content/errors/net-errors.mjs | sed -e "s/^.*errorCode: //g" -e "s/,$//g" -e 's/"//g'); do echo -e "def run_netError_$e(self):\n     self.assert_error(\n        \"$e\",\n        \"title\",\n        \"error\",\n    )\n"; done;
+    def run_netError_dnsNotFound(self):
+        self.assert_error(
+            "dnsNotFound",
+            "Server Not Found",
```
lissyx: From discussion with Matthew, this is not super nice so i'll introduce some way to map errors code to something we control in case required

testing/enterprise/test_felt_browser_verify_neterror.py
```diff
@@ -0,0 +1,106 @@
...
+            "MOZILLA_PKIX_ERROR_MITM_DETECTED",
+            "Unable to connect. Please contact your administrator.",
+            "Firefox Enterprise spotted a potentially serious security issue with www.mozilla.org. Someone pretending to be the site could try to steal things like credit card info, passwords, or emails.",
+        )
+
+    def run_netError_netOffline(self):
+        self.assert_error(
+            "netOffline",
+            "Offline mode",
+            "Press “Try Again” to switch to online mode and reload the page.",
```
lissyx: Similar, we will want to map to `NS_ERROR_OFFLINE` because the wording is better

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Given we have to perform some mapping, I'd like to find a way to move that to `FeltErrorReport`

lissyx:
Updated UX:

### NS_ERROR_OFFLINE
screenshot_test_felt_neterror_handling_NS_ERROR_OFFLINE

### netTimeout
screenshot_test_felt_neterror_handling_netTimeout

### netOffline
screenshot_test_felt_neterror_handling_netOffline

### dnsNotFound
screenshot_test_felt_neterror_handling_dnsNotFound

### blockedByPolicyEnterprise
screenshot_test_felt_neterror_handling_blockedByPolicyEnterprise

### MOZILLA_PKIX_ERROR_MITM_DETECTED
screenshot_test_felt_neterror_handling_MOZILLA_PKIX_ERROR_MITM_DETECTED

browser/components/enterprise/moz.build
```diff
@@ -14,6 +14,7 @@ EXTRA_JS_MODULES.enterprise += [
     "modules/ConsoleClient.sys.mjs",
     "modules/EnterpriseAccountsStorage.sys.mjs",
     "modules/EnterpriseOSInfo.sys.mjs",
+    "modules/FeltErrorReport.sys.mjs",
```
1rneh: Is there a reason `FeltErrorReport` is not placed under `browser/extensions/felt`?

browser/extensions/felt/api.js
```diff
@@ -79,6 +80,23 @@ this.felt = class extends ExtensionAPI {
       matches,
     });
 
+    ChromeUtils.unregisterWindowActor("NetError");
```
1rneh, line 74: Can you add a comment on why we unregister `NetError`?

browser/extensions/felt/content/FeltErrorWindowChild.sys.mjs
```diff
@@ -0,0 +1,43 @@
...
+  handleEvent(aEvent) {
+    if (aEvent.type !== "DOMContentLoaded") {
+      lazy.log.error(`Unexpected aEvent.type=${aEvent.type}`);
+      return;
+    }
+
+    lazy.log.debug(
+      `Notify parent with errorPage: ${this.docShell.currentDocumentChannel.originalURI.spec}`
+    );
+    this.sendAsyncMessage("FeltErrorWindow", {
```
1rneh: Can we be more descriptive with the naming here?
```suggestion
    this.sendAsyncMessage("ReportError", {
```

browser/components/enterprise/modules/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,258 @@
...
+   */
+  _noNetworkErrors: ["netOffline", "dnsNotFound"],
+
+  /**
+   * Gets the error name for a channel status code.
+   *
+   * @param {number} status - The channel status code
+   * @returns {string} Human-readable error name
+   */
+  _getErrorNameForStatus(status) {
```
1rneh:
```suggestion
  _getFluentIdForStatus(status) {
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -332,9 +302,11 @@ export const ConsoleClient = {
       };
 
       xhr.onerror = () => {
-        const errorName = this._getErrorNameForStatus(xhr.channel?.status);
+        const channelStatus = xhr.channel?.status ?? null;
         reject(
-          new TypeError(errorName, { cause: { host: new URL(url).host } })
+          new TypeError(channelStatus, {
+            cause: { hostname: new URL(url).host, channelStatus },
+          })
```
1rneh, line 286: We're duplicating the `channelStatus` here.

browser/extensions/felt/content/window.js
```diff
@@ -537,8 +447,8 @@ window.addEventListener(
   "load",
   () => {
     setBuildVersion();
-    ErrorReport.init();
-    lazy.Updates.init(document, ErrorReport);
+    lazy.FeltErrorReport.init(document);
+    lazy.Updates.init(document, lazy.FeltErrorReport);
```
1rneh: `Updates` can lazily import `FeltErrorReport`, no need to pass it.

browser/components/enterprise/modules/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,258 @@
...
+      const nssErrorsService = Cc[
+        "@mozilla.org/nss_errors_service;1"
+      ].getService(Ci.nsINSSErrorsService);
+      return nssErrorsService.getErrorName(status);
+    } catch {
+      // Not an NSS error, check common network error codes
+
+      // Mapping here should follow what nsDocShell::DisplayLoadError uses.
+      // Consumer code will expect those to fail when using Fluent to format
+      // and perform fallback to string bundles where they are defined.
```
1rneh: This is outdated

browser/components/enterprise/modules/FeltErrorReport.sys.mjs
1rneh: Can we also make sure the usage of this method is clear. It returns a fluent id.

browser/components/enterprise/modules/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,258 @@
...
+
+  init(doc) {
+    this._document = doc;
+    lazy.initializeRegistry();
+    this._wrapper = this._document.querySelector(".felt-browser-error");
+    this._wrapper.addEventListener("message-bar:user-dismissed", e => {
+      e.preventDefault();
+      e.target.classList.add("is-hidden");
+    });
+    this._initialized = true;
```
1rneh: Why do we need `this._initialized`?

browser/extensions/felt/content/window.js
1rneh, line 106: Same here (see my comment about the method head in `FeltErrorReport.update`)

browser/components/enterprise/modules/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,258 @@
...
+   * Error codes are filtered to display either the error connection or the
+   * no network display.
+   *
+   * @param {string} errorCode - The error code as a string value
+   * @param {object} cause - The error cause, with hostname field holding the
+   *                         value of the host triggering the error
+   */
+  updateNetworkError(errorCode, cause) {
+    if (this._noNetworkErrors.includes(errorCode)) {
+      this.update("felt-browser-error-no-network", "no-network-connection");
```
1rneh: Missing `await`. Otherwise we'll toggle visibility before `update` completed.

browser/components/enterprise/modules/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,258 @@
...
+    } else {
+      this.update("felt-browser-error-connection", errorCode, cause);
```
1rneh: Same here.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,258 @@
...
+    if (this._wrapper.classList.contains("is-hidden")) {
+      return;
+    }
+    this._wrapper.classList.add("is-hidden");
+    for (const bar of this._wrapper.querySelectorAll("moz-message-bar")) {
+      bar.classList.add("is-hidden");
+    }
+  },
+
+  async update(errorType, details = null, cause = null) {
```
1rneh: I think it's really complicated to follow how we pass down the arguments to `async update(errorType, details = null, cause = null)` in the stack.

One reason is because the current naming of the function arguments kind of obfuscate what information they entail - fluent id is `details` and `cause` is an object containing the hostname and maybe canUseCustomNetError? This issue persists through the stack trace.

Another reason is the different entry points and how they all resolve from different information (channel, about page uri, etc) to the arguments needed to pass to update. I'm wondering if we can simplify it.

1rneh (changes requested):
After my first round of review, I don't feel very confidetn that I have a good overview over the changes as it's quite a complex PR to review. I'd appreciate if it could be broken down into multiple patches or commits such as the following:
1. Extract felt error handling to `FeltErrorReport`.
2. Move `_getErrorNameForStatus` to `FeltErrorReport`
3. Report about:neterror and about:certerror details in Felt.

browser/components/enterprise/moz.build
lissyx: None, I'll fix

browser/components/enterprise/modules/FeltErrorReport.sys.mjs
lissyx: We dont even check for it so I think I was being defensive and lost my track of thought

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: > I think it's really complicated to follow how we pass down the arguments to `async update(errorType, details = null, cause = null)` in the stack.
> 
> One reason is because the current naming of the function arguments kind of obfuscate what information they entail - fluent id is `details` and `cause` is an object containing the hostname and maybe canUseCustomNetError? This issue persists through the stack trace.

We could rename them if it helps?

> 
> Another reason is the different entry points and how they all resolve from different information (channel, about page uri, etc) to the arguments needed to pass to update. I'm wondering if we can simplify it.

I introduced the different entry point to try and make it clearer, so I had hoped it would rather help than confuse :(

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: We are, I'm not sure what other value we can pass, we do not really use it anyway

lissyx:
> After my first round of review, I don't feel very confidetn that I have a good overview over the changes as it's quite a complex PR to review. I'd appreciate if it could be broken down into multiple patches or commits such as the following:
> 
>     1. Extract felt error handling to `FeltErrorReport`.
> 
>     2. Move `_getErrorNameForStatus` to `FeltErrorReport`
> 
>     3. Report about:neterror and about:certerror details in Felt.

Splitting will take me a bit of time, I've addressed the rest and I'll fix this later before requesting review again

lissyx:
> After my first round of review, I don't feel very confidetn that I have a good overview over the changes as it's quite a complex PR to review. I'd appreciate if it could be broken down into multiple patches or commits such as the following:
> 
>     1. Extract felt error handling to `FeltErrorReport`.
> 
>     2. Move `_getErrorNameForStatus` to `FeltErrorReport`
> 
>     3. Report about:neterror and about:certerror details in Felt.

I've split things, maybe not exactly like that, but that should ease reviewing

browser/extensions/felt/content/window.js
```diff
@@ -24,104 +25,13 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
-      );
-    } else {
-      ErrorReport.update(
-        "felt-browser-error-connection",
-        err.message,
-        err.cause
-      );
-    }
-    return;
+    lazy.FeltErrorReport.handleXhrError(err);
```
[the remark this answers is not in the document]
lissyx: Fixed

browser/extensions/felt/content/window.js
lissyx: Sorry, but I think I lost content here, what was the issue?

browser/extensions/felt/content/window.js
```diff
@@ -240,8 +150,8 @@ async function connectToConsole(email) {
         );
         resetToLoginPage(
           "felt-browser-error-connection",
-          lazy.ConsoleClient._getErrorNameForStatus(status),
-          { host: uri.host }
+          lazy.FeltErrorReport._getErrorNameForStatus(status),
```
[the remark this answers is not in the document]
lissyx: Indeed.

browser/extensions/felt/api.js
```diff
@@ -65,6 +66,24 @@ this.felt = class extends ExtensionAPI {
       matches,
     });
 
+    // Remove existing error handler installed by default so we can install
+    // our own to provide a dedicated UI.
+    ChromeUtils.unregisterWindowActor("NetError");
+    ChromeUtils.registerWindowActor(this.FELT_ERROR_WINDOW_ACTOR, {
```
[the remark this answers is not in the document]
lissyx: We probably do not want things to continue to work if we fail at this, because we would be in a broken state

browser/extensions/felt/api.js
```diff
@@ -417,6 +436,7 @@ this.felt = class extends ExtensionAPI {
 
     if (Services.felt.isFeltUI()) {
       ChromeUtils.unregisterWindowActor(this.FELT_WINDOW_ACTOR);
+      ChromeUtils.unregisterWindowActor(this.FELT_ERROR_WINDOW_ACTOR);
       ChromeUtils.unregisterProcessActor(this.FELT_PROCESS_ACTOR);
```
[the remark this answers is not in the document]
lissyx: Similar, broken state is likely something we prefer to surface explicitely with throwing and breaking everything rather than silently handling and error and continuing.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,257 @@
...
+      this.updateNetworkError(errorCode, {
+        hostname: errorHostname,
+        canUseCustomNetError: true,
+      });
+    } catch (ex) {
+      lazy.log.debug(
+        `handleNetError: exception ${ex}, reverting to default network error`
+      );
+      this.updateNetworkError("network", null);
+    }
```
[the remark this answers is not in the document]
lissyx: Fixed with making `handleNetError` being `async`

testing/enterprise/test_felt_browser_console_error.py
```diff
@@ -40,7 +43,7 @@ def check_error_bar_message(
                 details = self.get_elem(f"{selector} .felt-browser-error-details")
                 details_text = details.get_property("textContent").strip()
                 assert details_text == error_msg, (
-                    f"Correct error message: '{details_text}'"
+                    f"Correct error message: '{details_text}' expected: '{error_msg}'"
```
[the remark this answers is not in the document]
lissyx: Yes I lost it in some rewrites it looks.

browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
```diff
@@ -0,0 +1,34 @@
...
+ * report component.
+ */
+export class FeltErrorWindowParent extends JSWindowActorParent {
+  receiveMessage(message) {
+    lazy.log.debug(`Received message ${message.name} => ${message.data}`);
+    switch (message.name) {
+      case "ErrorReport": {
+        (async () => {
+          await lazy.FeltErrorReport.handleNetError(message.data);
+        })();
```
1rneh: Nothing awaits the result nor catches errors.
```suggestion
        const errorPage = message.data;
        lazy.FeltErrorReport.handleNetError(errorPage);
```

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,257 @@
...
+          detailsElement.textContent = message || details;
+        }
+      }
+    }
+    errorElement.classList.remove("is-hidden");
+    this._wrapper.classList.remove("is-hidden");
+  },
+
+  /**
+   * Handle displaying and error produced by about:neterror/about:certerror
```
1rneh: typo
```suggestion
   * Handle displaying an error produced by about:neterror/about:certerror
```

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,257 @@
...
+    } catch (ex) {
+      lazy.log.debug(
+        `handleNetError: exception ${ex}, reverting to default network error`
+      );
+      await this.updateNetworkError("network", null);
+    }
+  },
+
+  /**
+   * Handle displaying and error produced by XmlHttpRequest.
```
1rneh: typo
```suggestion
   * Handle displaying an error produced by XmlHttpRequest.
```

browser/extensions/felt/content/FeltErrorWindowChild.sys.mjs
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Error-capture dedicated window actor. It is expected to be only setup on
+ * matching pages like about:neterror/about:certerror and is there to pass the
+ * error page URI information back to the parent-side of this window actor for
+ * further processing, including extracting the error code to later display it
+ * properly.
+ *
+ * It relies on inspecting the docShell that contains several nsIChannel:
+ *  - failedChannel: this one's originalURI reports the page load that triggered
+ *                   the error
```
1rneh: `failedChannel` isn't used anywhere. Was the intention to pass it along too?

toolkit/content/errors/net-errors.mjs
```diff
@@ -360,6 +360,30 @@ export const NET_ERRORS = [
     checkTrrOnly: true,
     image: NET_ERROR_ILLUSTRATIONS.noConnection,
   },
+  {
```
1rneh, line 363: Please use `AppConstants.MOZ_ENTERPRISE` to only conditionally include this?

testing/enterprise/test_felt_browser_verify_neterror.py
```diff
@@ -0,0 +1,108 @@
...
+            """
+            const { FeltErrorReport } = ChromeUtils.importESModule("resource:///modules/FeltErrorReport.sys.mjs");
+            (async () => {
+              await FeltErrorReport.handleNetError({errorPage: `about:neterror?e=${arguments[0]}&u=https%3A//www.mozilla.org`});
+            })();
+            """,
+            [error],
+        )
+
+    def assert_error(self, error_name, expected_heading, error_msg):
```
1rneh: Can the to-be-tested action be extracted from the assertion method, hence move `trigger_netError` out of `assert_error`?

testing/enterprise/test_felt_browser_console_error.py
```diff
@@ -86,5 +91,18 @@ def test_felt_error_details_include_console_address(self):
...
-            f"Firefox can’t establish a connection to the server at localhost:{refused_port}.",
+            f"Firefox Enterprise can’t connect to the server at localhost:{refused_port}",
+        )
+
+    def test_felt_error_insecure_certs(self):
+        self._manually_closed_child = True
+        requests.post(
+            f"http://localhost:{self.console_port}/sso/login_force_wrong_location",
+            data={"location": "https://wrong.host.badssl.com/sso_url"},
+        )
```
1rneh: I vaguely remember having asked this before, but I don't remember the discussion, so I'll re-address this here. Can we just set `self.server.login_location = "https://wrong.host.badssl.com/sso_url"` here?

testing/enterprise/test_felt_console_error.py
```diff
@@ -86,5 +91,18 @@ def test_felt_error_details_include_console_address(self):
...
+        requests.post(
+            f"http://localhost:{self.console_port}/sso/login_force_wrong_location",
+            data={"location": "https://wrong.host.badssl.com/sso_url"},
+        )
+        return self.check_error_bar_message(
+            f"http://localhost:{self.console_port}",
+            ".felt-browser-error-connection",
+            "Unable to connect",
+            "Firefox Enterprise spotted a potentially serious security issue with wrong.host.badssl.com. Someone pretending to be the site could try to steal things like credit card info, passwords, or emails.",
         )
```
1rneh, line 174: The others tests in this suite instrument the console address while this test instruments the sso redirect uri. I think it makes no difference if we also temper the console address uri here and it would allow for the test suite to follow on approach.

testing/enterprise/test_felt_verify_neterror.py
```diff
@@ -0,0 +1,108 @@
...
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from base_test import Environment
+from felt_tests import FeltTests
+
+
+class FeltVerifyNetError(FeltTests):
+    def trigger_netError(self, error):
```
1rneh, line 15: Calling `FeltErrorReport` directly skips testing the whole FeltErrorWindow actor pair flow. Can we instrument the mock server so that the `about:neterror` pages actually get triggered (at least for the ones where that is possible)? Maybe via setting `self.server.login_location`? This way breaking changes to the FeltErrorWindow actors regress a test failure.

testing/enterprise/test_felt_browser_verify_neterror.py
```diff
@@ -0,0 +1,108 @@
...
+        message = error.get_attribute("heading").strip()
+        assert expected_heading in message, (
+            f"Unexpected error heading: '{message}', expected '{expected_heading}'"
+        )
+
+        details = self.get_elem(f"{selector} .felt-browser-error-details")
+        details_text = details.get_property("textContent").strip()
+        assert details_text == error_msg, (
+            f"Unexpected error message: '{details_text}', expected '{error_msg}'"
+        )
```
1rneh: Let's re-use [`check_error_bar_message`](https://searchfox.org/enterprise-main/rev/6a1f889e0eed9c41cfbc6254963bc8a237d0d39a/testing/enterprise/test_felt_browser_console_error.py#20-53) here.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -321,9 +291,11 @@ export const ConsoleClient = {
       };
 
       xhr.onerror = () => {
-        const errorName = this._getErrorNameForStatus(xhr.channel?.status);
+        const channelStatus = xhr.channel?.status ?? null;
         reject(
-          new TypeError(errorName, { cause: { host: new URL(url).host } })
+          new TypeError("network", {
+            cause: { hostname: new URL(url).host, channelStatus },
```
1rneh, line 285: Can we just pass `{ hostname: new URL(url).host, channelStatus }` instead of wrapping it in another object?

browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
```diff
@@ -0,0 +1,34 @@
...
+  FeltErrorReport: "resource:///modules/FeltErrorReport.sys.mjs",
+});
+
+ChromeUtils.defineLazyGetter(lazy, "log", () => {
+  return lazy.createEnterpriseLogger("FeltErrorWindowParent");
+});
+
+/**
+ * Parent-side of the Felt error window actor to collect the error page URL. It
+ * receives the URL of the loaded error page and forward that to the error
```
1rneh: typo
```suggestion
 * receives the URL of the loaded error page and forwards that to the error
```

browser/extensions/felt/content/FeltErrorWindowChild.sys.mjs
```diff
@@ -0,0 +1,43 @@
...
+    if (aEvent.type !== "DOMContentLoaded") {
+      lazy.log.error(`Unexpected aEvent.type=${aEvent.type}`);
+      return;
+    }
+
+    lazy.log.debug(
+      `Notify parent with errorPage: ${this.docShell.currentDocumentChannel.originalURI.spec}`
+    );
+    this.sendAsyncMessage("ErrorReport", {
+      errorPage: this.docShell.currentDocumentChannel.originalURI.spec,
```
1rneh: `nit:` Can `errorPage` be renamed to `errorPageURI`?

browser/extensions/felt/api.js
```diff
@@ -65,6 +66,24 @@ this.felt = class extends ExtensionAPI {
...
+      child: {
+        esModuleURI: "chrome://felt/content/FeltErrorWindowChild.sys.mjs",
+        events: {
+          DOMContentLoaded: {},
+        },
+      },
+      parent: {
+        esModuleURI: "chrome://felt/content/FeltErrorWindowParent.sys.mjs",
+      },
+      includeChrome: true,
```
1rneh: Do we need `includeChrome:true`? `NetError` [doesn't include it neither](https://searchfox.org/firefox-main/rev/e28b34ab33dbf49364999070168cbb7e11e8e5bd/toolkit/modules/ActorManagerParent.sys.mjs#465-479) and I think the about:* pages in felt will only live in content?

1rneh (changes requested):
Round of review #1.

Mentioning this since this was important to you. The commits currently break bisection, e.g. the first commit bc2780589c05a9ad95c922de00ebe875124a06fe references `FeltErrorReport` which is added in the later commit 16dd62d0441dbdfe686eaecc76654b16ae92c900. I don't mind as long as we squash merge the PR at the end.

1rneh:
@jporter-dev, could you also give it a round of review, as you've also worked on the error reporting on felt?

browser/extensions/felt/content/window.js
```diff
@@ -240,8 +151,8 @@ async function connectToConsole(email) {
         );
         resetToLoginPage(
           "felt-browser-error-connection",
-          lazy.ConsoleClient._getErrorNameForStatus(status),
-          { host: uri.host }
+          lazy.FeltErrorReport._getFluentIdForStatus(status),
```
jporter-dev:
```suggestion
          lazy.FeltErrorReport.getFluentIdForStatus(status),
```

Since `_getFluentIdForStatus` is accessed outside of `FeltErrorReport`, this function should be made "public"

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,257 @@
...
+                );
+              }
+            }
+          } catch {
+            lazy.log.debug(`details=${details} not resolved, falling back`);
+          }
+        }
+
+        const l10n = this._document.l10n.getAttributes(detailsElement);
+        if (!l10n || (l10n && !l10n.id)) {
```
jporter-dev: This can be simplified with optional chaining 

```suggestion
        if (!l10n?.id)) {
```

browser/extensions/felt/FeltErrorReport.sys.mjs
jporter-dev: I think renaming is a good first step to clarifying this. Perhaps `details -> errorCode` and `cause -> context`? We use `errorCode` in the function callers below, and `context` describes that it is additional context for displaying the error. 

Adding a JSDoc for the params would help as well, outlining the supported values for context.

```
  /**
   * Updates the error display based on the error type and details.
   * @param {string} errorType - The type of error to display.
   * @param {string|null} errorCode - The specific error code used for displaying the error message.
   * @param {object|null} context - Additional context for the error.
   * @param {string} context.hostname - The hostname related to the error.
   * @param {boolean} context.canUseCustomNetError - Indicates if a custom network error can be used.
   */
```

jporter-dev: Other than cleaning up and clarifying the params for this, this function has grown a lot. It can orchestrate the update, but can some of the logic within be split up into a private helper method for each concern? 

Previously we had some of the "message building logic" in `this.getLocalisedErrorString` which helped a bit with readability.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,257 @@
...
+    }
+    errorElement
+      .querySelectorAll(".felt-browser-error-details[added]")
+      .forEach(el => el.remove());
+    if (details) {
+      const detailsElement = errorElement.querySelector(
+        ".felt-browser-error-details"
+      );
+      if (detailsElement) {
+        if (cause) {
```
jporter-dev: Everything within here about handling the `errorConfig` could become `_applyErrorConfig`. This is a complex block of code that should be its own function.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,257 @@
...
+
+  async update(errorType, details = null, cause = null) {
+    if (!this._wrapper) {
+      return;
+    }
+    const errorElement = this._wrapper.querySelector(`.${errorType}`);
+    if (!errorElement) {
+      return;
+    }
+    if (errorElement.hasAttribute("heading-changed")) {
```
jporter-dev: Looks like this is resetting an individual error element, we could create `_resetErrorElement` for separation.

jporter-dev (changes requested):
First pass, splitting the FeltErrorReport out to a separate module makes sense and is a good improvement given the added complexity. Nice job sourcing those errors too! Left some comments about cleaning `update` up a bit since it's handling quite a few things now.

lissyx:
> Round of review #1.
> 
> Mentioning this since this was important to you. The commits currently break bisection, e.g. the first commit [bc27805](https://github.com/mozilla/enterprise-firefox/commit/bc2780589c05a9ad95c922de00ebe875124a06fe) references `FeltErrorReport` which is added in the later commit [16dd62d](https://github.com/mozilla/enterprise-firefox/commit/16dd62d0441dbdfe686eaecc76654b16ae92c900). I don't mind as long as we squash merge the PR at the end.

Nice finding, I thought I had fixed it, and running the tests (only the two I'm changing) does not yield the neterror so of course it would not repro in my flow.

browser/extensions/felt/api.js
lissyx: Nope, that is likely a leftover when I was trying, if it's not on NetError there is no reason we have it

browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
lissyx: Catching the error now

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: Fixed

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: Fixed

browser/extensions/felt/content/FeltErrorWindowChild.sys.mjs
lissyx: > `failedChannel` isn't used anywhere. Was the intention to pass it along too?

No, and your comment suggest my comment is not clear. I wanted to make it clear the distinction that exists between `failedChannel` and `currentDocumentChannel`.

e.g. if you try to load https://mozilla.org and it fails due to certificate:
 - `failedChannel` holds `https://mozilla.org` as URI
 - `currentDocumentChannel` holds `about:neterror?...` as URI

toolkit/content/errors/net-errors.mjs
lissyx: Unfortunately, we already broke about:neterror pages in the past exactly this way:
> Security Error: Content at about:neterror?e=blockedByPolicy may not load or link to resource://gre/modules/AppConstants.sys.mjs.

lissyx: We could preprocess, but I remember we wanted to undo those preprocessing we added in the past.
Given it's only data definition, I dont think it's worth preprocessing the file and thus breaking some linting capabilities.

testing/enterprise/test_felt_verify_neterror.py
lissyx: Yes, it's on purpose.

Here the goal of that test was to expose and assert correct rendering and text content of some of the about:neterror, also to provide data for UX validation. it's not here to check the flow of hitting an error page displaying an error message we control, that is the job of the changes in `testing/enterprise/test_felt_browser_console_error.py`.

Triggering different condition by really loading things was too complicated given the error code we want to test.

testing/enterprise/test_felt_browser_verify_neterror.py
lissyx: Sure

testing/enterprise/test_felt_browser_console_error.py
lissyx: Looks like I added it server side and I forgot to use it

testing/enterprise/test_felt_console_error.py
lissyx: Do you mean instead of using `login_location` ?

lissyx: So as mentionned on the other comment, the redirection is required to make sure that we hit the error while loading a page. If we directly set the console address, then what triggers the error is the `sendDevicePosture()` call that triggers XHR error.

We could bypass the device posture in those tests, but I'm not a big fan of  this and I think it's better if we rather detour the redirection.

lissyx: I tried again and many of the conditions trips on things like domain name resolving when it should not (depending on your setup), or insecure https not accepted (this is a second reason why I made the https console mock patch).

So I think it's good enough we keep just this one aside that exercises really the flow of actors, and the rest of the assertions verifies the error messages we receive.

lissyx: I'm updating with a comment to make it clear

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: OK but I clearly remember you insisted for adding `cause` here in the past

testing/enterprise/test_felt_browser_verify_neterror.py
lissyx: Done

browser/extensions/felt/content/FeltErrorWindowParent.sys.mjs
lissyx: Done

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Maybe "ConsoleClientXHRError" instead of "network"?

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: Yep and trying to improve the tests, I think I'm exposing races where explicitely we have that kind of problems

testing/enterprise/test_felt_console_error.py
lissyx: So, good news/bad news, I'm augmenting coverage and thus finding issues, including races/intermittents, in the code and the tests. The code ones are unlikely to really be critical in real usage and are mostly exposed during testing, but still.

lissyx: I should have something in a better shape, and I think one of the behavior of error handling should allow us to also add coverage for the SSO timeout.

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: > Other than cleaning up and clarifying the params for this, this function has grown a lot. It can orchestrate the update, but can some of the logic within be split up into a private helper method for each concern?
> 
> Previously we had some of the "message building logic" in `this.getLocalisedErrorString` which helped a bit with readability.

I believe the current changes I have should improve this.

testing/enterprise/test_felt_console_error.py
lissyx: And now we have SSO timeout coverage :)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Correct me if I'm wrong, but since we are `new TypeError()` here, we have to use `cause`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#cause

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+  async _maybeResetErrorElement(errorElement) {
+    lazy.log.debug(`_maybeResetErrorElement()`);
+    errorElement.dismiss();
+    errorElement.removeAttribute("source");
+
+    errorElement
+      .querySelectorAll(".felt-browser-error-details[added]")
+      .forEach(el => el.remove());
+
+    // Remove so next l10n.setAttribtues() will re-translate from original HTML
```
jporter-dev:
```suggestion
    // Remove so next l10n.setAttributes() will re-translate from original HTML
```

nit: typo in the comment

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+   * no network display.
+   *
+   * @param {string} errorCode - The error code as a string value
+   * @param {object} context - The error context, with hostname field holding the
+   *                         value of the host triggering the error
+   * @param {string} source - The source of the error, "xhr" or "net". Useful for
+   *                          assertions in tests&
+   */
+  async updateNetworkError(errorCode, context, source) {
+    lazy.log.debug(`updateNetworkError(${errorCode}})`);
```
jporter-dev:
```suggestion
    lazy.log.debug(`updateNetworkError(${errorCode})`);
```

nit: extra `}`

jporter-dev (approved):
Couple of minor typos in comments/logs, otherwise LGTM!

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: Linter missed it?

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+      };
+      lazy.log.debug(
+        `getFluentIdForStatus(${status}) networkErrors[status]:${networkErrors[status]}`
+      );
+      return networkErrors[status] ?? "network";
+    }
+  },
+
+  init(doc) {
+    this._document = doc;
```
1rneh, line 81: I think we should be able to access the chrome document in `FelErrorReport` with `this.browsingContext.topChromeWindow.document`, which discards the need to pass the document to `init`.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+    this._wrapper = this._document.querySelector(".felt-browser-error");
+    this._wrapper.addEventListener("message-bar:user-dismissed", e => {
+      e.preventDefault();
+      e.target.classList.add("is-hidden");
+    });
+  },
+
+  uninit() {
+    this._document = undefined;
+  },
```
1rneh: `uninit` is never called.

testing/enterprise/test_felt_browser_verify_neterror.py
```diff
@@ -0,0 +1,104 @@
...
+            """
+            const callback = arguments[arguments.length - 1];
+            const { FeltErrorReport } = ChromeUtils.importESModule("resource:///modules/FeltErrorReport.sys.mjs");
+            FeltErrorReport.handleNetError({errorPageURI: `about:neterror?e=${arguments[0]}&u=https%3A//www.mozilla.org`}).then(_ => callback());
+            """,
+            [error],
+        )
+
+    def assert_error(self, error_name, expected_heading, error_msg):
+        self.assert_error_bar_message(
```
1rneh: As `assert_error` does nothing besides call `assert_error_bar_message`, can the test methods call it directly?

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+  /**
+   * Handle displaying an error produced by XmlHttpRequest.
+   *
+   * @param {object} aError - The TypeError containing the XHR channel status
+   *                          code in aError.cause.channelStatus and the
+   *                          hostname of the URL triggering the error in
+   *                          aError.cause.hostname.
+   */
+  async handleXhrError(aError) {
+    lazy.log.debug(`handleXhrError(${JSON.stringify(aError)})`);
```
1rneh: The TypeError `aError` has the properties `message` and `cause`, which are both non-enumerable, so stringifying `aError` will yield `"{}"`.

browser/extensions/felt/content/window.js
```diff
@@ -24,103 +25,13 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
-        "no-network-connection"
-      );
-    } else {
-      ErrorReport.update(
-        "felt-browser-error-connection",
-        err.message,
-        err.cause
-      );
-    }
+    lazy.log.error(`Failed to send device posture: ${JSON.stringify(err)}`);
```
1rneh: Same here. The error is constructed by the `ConsoleClient` and also doesn't have any additional enumerable properties.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+  /**
+   * Leverages existing about:certerror/about:neterror string IDs and error
+   * handling to update the error message displayed to the user
+   *
+   * @param {object} errorElement - The HTML element holding the error message
+   * @param {string|null} errorCode - The specific error code used for displaying the error message.
+   * @param {object|null} context - Additional context for the error.
+   * @param {string} context.hostname - The hostname related to the error.
+   * @param {boolean} context.canUseCustomNetError - Indicates if a custom network error can be used.
+   * @param {object|null} detailsElement - The HTML element holding the error message details, if used.
```
1rneh: You define `errorCode` as optional since `null` is a potential value. Is that intentional? Because that would mean we'd pass `null` to `getResolvedErrorConfig`?

If a property is optional, can you wrap the property name into squared brackets, e.g. 
`   * @param {object|null} detailsElement - The HTML element holding the error message details, if used.`
and make sure we add guards in the code.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+  },
+
+  uninit() {
+    this._document = undefined;
+  },
+
+  reset() {
+    if (!this._wrapper) {
+      return;
+    }
```
1rneh, line 93: When do we expect `this._wrapper` to be undefined or null?

testing/enterprise/felt_console_error.py
```diff
@@ -0,0 +1,97 @@
...
+            source_attribute = self._wait.until(
+                lambda d: (
+                    h.strip()
+                    if (h := error.get_attribute("source")) and source in h.strip()
+                    else False
+                )
+            )
+            assert source in source_attribute, (
+                f"Error element is missing the {source} attribute. Found: {source_attribute}"
+            )
```
1rneh, line 79: We already assert the content of heading, message and message details. What is the value in asserting the attribute `source` that we only use for tests?

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+    try {
+      const parsedErrorPage = new URL(aData.errorPageURI);
+      const errorCode = (() => {
+        const parsedErrorCode = parsedErrorPage.searchParams.get("e");
+        if (parsedErrorCode in this._errorCodesMapping) {
+          lazy.log.debug(
+            `handleNetError(${aData.errorPageURI}): mapping ${parsedErrorCode} to ${this._errorCodesMapping[parsedErrorCode]}`
+          );
+          return this._errorCodesMapping[parsedErrorCode];
+        }
```
1rneh: We iterating through `this._errorCodesMapping` twice here. Can save the result and then check whether it's undefined.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+        errorCode,
+        context,
+        source
+      );
+    }
+
+    this._document
+      .querySelector(".felt-login__email-pane")
+      .classList.remove("is-hidden");
+    this._document.querySelector(".felt-login__sso").classList.add("is-hidden");
```
1rneh: This will not clear the browser session correctly since it remains in its old state.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+
+    this._wrapper.classList.add("is-hidden");
+
+    const errorElement = this._wrapper.querySelector(`.${errorType}`);
+    if (!errorElement) {
+      lazy.log.error(`Error element ${errorType} not found`);
+      return;
+    }
+
+    await this._maybeResetErrorElement(errorElement);
```
1rneh, line 201: This introduces a behavioral change. Previously we'd have supported multiple message bars at once. But this clears all previous ones.

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+    this._document = undefined;
+  },
+
+  reset() {
+    if (!this._wrapper) {
+      return;
+    }
+    if (this._wrapper.classList.contains("is-hidden")) {
+      return;
+    }
```
1rneh: I think we should discard toggling the visibility on the wrapper element as each message bar element handles its own visibility. If none of the wrapper's children is visible it collapses anyway. This avoids having to maintain visibility in two places.

1rneh (changes requested):
Note: Bisection is still broken. In 95f5c4f your code calls `lazy.FeltErrorReport._getErrorNameForStatus(status)` but `_getErrorNameForStatus` in that commit it's still defined on ConsoleClient. It is moved and renamed to `getFluentIdForStatus` in the next commit 0dbe863.

browser/extensions/felt/FeltErrorReport.sys.mjs
1rneh: This doesn't need to be part of this scope. If I remember correctly, we already have a bug filed for it? I think we assigned it to you?

browser/extensions/felt/FeltErrorReport.sys.mjs
```diff
@@ -0,0 +1,349 @@
...
+   * sent from FeltErrorWindowChild.
+   *
+   * @param {object} aData - The message containing the error URL in 'errorPageURI'
+   *                         field which is a URL containing error code as 'e'
+   *                         parameter and URL triggering the error as
+   *                         'u' parameter.
+   */
+  async handleNetError(aData) {
+    lazy.log.debug(`handleNetError(${aData.errorPageURI})`);
+
```
1rneh, line 249: Can we extract the error page uri only once `const errorPageURI = aData.errorPageURI` and reuse that constant?

testing/enterprise/felt_console_error.py
lissyx: Because the error message can be the same in both code path, but we want to ensure that the right code path was used to generate this one, to avoid asserting to true something that e.g. would have been produced by XHR handling when expecting the error to come from about:neterror handling.

lissyx:
> Note: Bisection is still broken. In [95f5c4f](https://github.com/mozilla/enterprise-firefox/commit/95f5c4fcbaf838dcdb13838bb905d68fac619b87) your code calls `lazy.FeltErrorReport._getErrorNameForStatus(status)` but `_getErrorNameForStatus` in that commit it's still defined on ConsoleClient. It is moved and renamed to `getFluentIdForStatus` in the next commit [0dbe863](https://github.com/mozilla/enterprise-firefox/commit/0dbe863db8aa2397d26a11c1b38cc0ea98202cff).

I will keep bisection correct once it's ready

testing/enterprise/test_felt_browser_verify_neterror.py
lissyx: It was there to make the assertions easier to read in the tests

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: No it was a mistake, the caller checks that it's non `null`

lissyx: Same for `context`

lissyx: And `detailsElement`

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: I'm not sure, it was there before, I just kept it: https://searchfox.org/enterprise-main/rev/6106787d2bc8a58435ae60402458f29ca5defeaf/browser/extensions/felt/content/window.js#60-62

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: I do not see any bug, and especially assigned to me ?

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: I am not sure what you have in mind here, we have no `browsingContext` on this object

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: We still support displaying two different error elements, it's visible during testing.

lissyx: This one only makes sure that one error element we use is properly cleaned of any previous error message we may have exposed, to avoid mixed content between two errors, as some test intermittence were showing

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: Removed it, it's likely a left over from early version and it is indeed not useful.

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: It's part of `reset()` which is supposed to set things back to their original state, so I'm not sure if this really applies here

browser/extensions/felt/FeltErrorReport.sys.mjs
1rneh: I had tested something in the dev toolbox, but my testing `this` was a `ChromeWindow`, which is not the case for `FeltErrorReport`. So please disregard my comment. Sorry for the confusion!

browser/extensions/felt/FeltErrorReport.sys.mjs
1rneh: I couldn't find the bug anymore so I filed a new one: Bug-2045684

browser/extensions/felt/FeltErrorReport.sys.mjs
1rneh: My comment wasn't meant specific to `reset()` but to discard the visibility toggling of the wrapper in all of `FeltErrroReport`.

lissyx: Removing all of those calls then?

lissyx: @1rneh I'm lost to how you want to handle this, if we remove all this code then we are going to keep displaying  things?

1rneh: I'm proposing that we stop toggling the `is-hidden` class on the wrapper, since all `<moz-message-bar>` have the `is-hidden` class. The wrapper will collapse when all children are hidden and become visible when at least one child is shown.

lissyx: So we just remove the code

lissyx: Unfortunately, removing the code (or I removed the wrong part) is breaking everything

lissyx: Ok, I had forgot to update the html :(

browser/extensions/felt/FeltErrorReport.sys.mjs
1rneh: > Removed it, it's likely a left over from early version and it is indeed not useful.

We encounter a about:neterror page only during sso navigation at which point the email login is hidden and the browser is shown. So we need to hide the browser and show the email login, otherwise the user is left on the sso with just the error message bar overlay and no email login to retry.

My comment was only addressing that the browser is currently not correctly cleared, which is an additional task to toggling the visibility of the sso and email pane.

testing/enterprise/felt_console_error.py
1rneh: But do we mind where it comes from unless it affects the error messaging?

If you're certain this is an important piece of information, can we create an enum in `FeltErrorReport` capturing the three possible sources to avoid any magic strings.

1rneh (changes requested):
I also left [this comment](https://github.com/mozilla/enterprise-firefox/pull/524#discussion_r3388172759) about the need to reset to the email login after running into an about:neterror/about:certerror page.

lissyx:
> I also left [this comment](https://github.com/mozilla/enterprise-firefox/pull/524#discussion_r3388172759) about the need to reset to the email login after running into an about:neterror/about:certerror page.

yes and I'm lost at what you want now ...

testing/enterprise/felt_console_error.py
lissyx: Right

browser/extensions/felt/FeltErrorReport.sys.mjs
lissyx: ok then I completely misunderstood your previous comments, I'll add it back

testing/enterprise/felt_console_error.py
lissyx: I've added `FeltErrorReport.ErrorSource`

1rneh (approved):
r+ since this is blocking. It looks proper and we can follow-up on upcoming issues.

lissyx:
> `_getErrorNameForStatus`

OK this should be fixed

PR #525, opened 2026-03-10
Bug 2012442 - Closing Firefox logs out the user from Firefox Enterprise

jporter-dev:
## Summary

- Adds logic in `browser.js` to intercept the browser close and delegate to `EnterpriseHandler.onSignOut` when the last window is closing
  - Last window is closed
- Adds logic in `BrowserGlue.sys.mjs` to intercept the application exit and delegate to `EnterpriseHandler.onSignOut`
  - `Cmd+Q` or right click dock icon and quit on MacOS
  - `Cmd+Shift+Q` or hamburger menu and exit on Windows

## Links

- [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2012442)

- #569 

## Testing

Prefs that are toggled: `enterprise.promptOnSignout` and `browser.tabs.warnOnClose`

1. `mach build`
2. `mach run`
3. Login with FELT
4. Verify Signout Prompt displays on all window close scenarios
  - [ ] `Cmd+Q` on MacOS
  - [ ] Right click dock icon and quit on MacOS
  - [ ] `Cmd+Shift+Q` on Windows
  - [ ] Hamburger menu and exit on Windows
  - [ ] Close last window on any OS

### DIFF
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -3830,23 +3830,6 @@ function WindowIsClosing(event) {
  * @returns true if closing can proceed, false if it got cancelled.
  */
 function warnAboutClosingWindow() {
-  // If this is the last Enterprise Firefox window, warn about signing out.
-  if (AppConstants.MOZ_ENTERPRISE) {
-    let isLastWindow = true;
-    for (let win of browserWindows()) {
-      if (!win.closed && win != window) {
-        isLastWindow = false;
-        break;
-      }
-    }
-    if (isLastWindow) {
-      // Return false to prevent the standard close flow; warnAboutClosing
-      // handles shutdown asynchronously after the user confirms the dialog.
-      EnterpriseHandler.onSignOut(window);
-      return false;
-    }
-  }
-
   // Popups aren't considered full browser windows; we also ignore private windows.
   let isPBWindow =
     PrivateBrowsingUtils.isWindowPrivate(window) &&
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -3830,6 +3830,23 @@ function WindowIsClosing(event) {
  * @returns true if closing can proceed, false if it got cancelled.
  */
 function warnAboutClosingWindow() {
+  // If this is the last Enterprise Firefox window, warn about signing out.
+  if (AppConstants.MOZ_ENTERPRISE) {
+    let isLastWindow = true;
+    for (let win of browserWindows()) {
+      if (!win.closed && win != window) {
+        isLastWindow = false;
+        break;
+      }
+    }
+    if (isLastWindow) {
+      // Return false to prevent the standard close flow; warnAboutClosing
+      // handles shutdown asynchronously after the user confirms the dialog.
+      EnterpriseHandler.onSignOut(window);
+      return false;
+    }
+  }
+
   // Popups aren't considered full browser windows; we also ignore private windows.
   let isPBWindow =
     PrivateBrowsingUtils.isWindowPrivate(window) &&
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -1553,17 +1553,36 @@ BrowserGlue.prototype = {
       AppConstants.MOZ_ENTERPRISE &&
       !lazy.EnterpriseHandler._signoutComplete
     ) {
-      let topWindow = lazy.BrowserWindowTracker.getTopWindow({
+      const topWindow = lazy.BrowserWindowTracker.getTopWindow({
         allowFromInactiveWorkspace: true,
       });
       if (topWindow) {
-        aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
-        lazy.EnterpriseHandler.onSignOut(topWindow, false).then(confirmed => {
-          if (confirmed) {
-            Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
-          }
-        });
-        return;
+        const shouldWarnForShortcut =
+          this._quitSource == "shortcut" &&
+          Services.prefs.getBoolPref("browser.warnOnQuitShortcut");
+        const warnOnSignout = Services.prefs.getBoolPref(
+          "enterprise.promptOnSignout",
+          true
+        );
+        if (
+          !warnOnSignout &&
+          shouldWarnForShortcut &&
+          Services.prefs.getBoolPref("browser.warnOnQuit")
+        ) {
+          // When promptOnSignout is false and warnOnQuitShortcut applies, let
+          // the native dialog run instead of the enterprise one. The signout
+          // will be deferred until the user confirms that dialog.
+          lazy.EnterpriseHandler._signoutComplete = true;
+          lazy.EnterpriseHandler._signoutPending = true;
+        } else {
+          aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
+          lazy.EnterpriseHandler.onSignOut(topWindow, false).then(confirmed => {
+            if (confirmed) {
+              Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
+            }
+          });
+          return;
+        }
       }
     }
 
@@ -1726,6 +1745,19 @@ BrowserGlue.prototype = {
     }
 
     this._quitSource = "unknown";
+
+    if (AppConstants.MOZ_ENTERPRISE && lazy.EnterpriseHandler._signoutPending) {
+      if (buttonPressed == 0) {
+        lazy.EnterpriseHandler._signoutPending = false;
+        aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
+        lazy.EnterpriseHandler.initiateShutdown(false).then(() => {
+          Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
+        });
+        return;
+      }
+      lazy.EnterpriseHandler._signoutPending = false;
+      lazy.EnterpriseHandler._signoutComplete = false;
+    }
 
     aCancelQuit.data = buttonPressed != 0;
   },
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -37,6 +37,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   DistributionManagement: "resource:///modules/distribution.sys.mjs",
   DownloadsViewableInternally:
     "moz-src:///browser/components/downloads/DownloadsViewableInternally.sys.mjs",
+  EnterpriseHandler: "resource:///modules/enterprise/EnterpriseHandler.sys.mjs",
   ExtensionsUI: "resource:///modules/ExtensionsUI.sys.mjs",
   FormAutofillUtils: "resource://gre/modules/shared/FormAutofillUtils.sys.mjs",
   Interactions: "moz-src:///browser/components/places/Interactions.sys.mjs",
@@ -1539,6 +1540,16 @@ BrowserGlue.prototype = {
     // and also "we're quitting by closing the last window".
 
     if (aQuitType == "restart" || aQuitType == "os-restart") {
+      return;
+    }
+
+    // Cancel quitting and instead show the Enterprise signout dialog
+    if (AppConstants.MOZ_ENTERPRISE) {
+      aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
+      let topWindow = lazy.BrowserWindowTracker.getTopWindow({
+        allowFromInactiveWorkspace: true,
+      });
+      lazy.EnterpriseHandler.onSignOut(topWindow);
       return;
     }
 
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -37,7 +37,6 @@ ChromeUtils.defineESModuleGetters(lazy, {
   DistributionManagement: "resource:///modules/distribution.sys.mjs",
   DownloadsViewableInternally:
     "moz-src:///browser/components/downloads/DownloadsViewableInternally.sys.mjs",
-  EnterpriseHandler: "resource:///modules/enterprise/EnterpriseHandler.sys.mjs",
   ExtensionsUI: "resource:///modules/ExtensionsUI.sys.mjs",
   FormAutofillUtils: "resource://gre/modules/shared/FormAutofillUtils.sys.mjs",
   Interactions: "moz-src:///browser/components/places/Interactions.sys.mjs",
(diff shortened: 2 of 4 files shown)


browser/base/content/browser.js
```diff
@@ -3820,6 +3820,23 @@ function WindowIsClosing(event) {
  * @returns true if closing can proceed, false if it got cancelled.
  */
 function warnAboutClosingWindow() {
+  // If this is the last Enterprise Firefox window, warn about closing multiple tabs and signing out.
+  if (AppConstants.MOZ_ENTERPRISE && EnterpriseHandler._isInitialized) {
```
1rneh: Did you get to this point and `EnterpriseHandler._isInitialized` was not initialized? Because `EnterpriseHandler.init()` should run for each browser window that is opened: https://searchfox.org/enterprise-main/rev/a038f49228d707c6675ef20ce640034a64307d2e/browser/components/BrowserComponents.manifest#33

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -146,55 +147,110 @@ export const EnterpriseHandler = {
...
-    const shouldInformOnSignout = Services.prefs.getBoolPref(
+  /**
+   * Shows the enterprise close warning dialog and initiates shutdown if the
+   * user confirms. Returns false if the user cancels.
+   *
+   * @param {Window} window chrome window being closed
+   * @returns {Promise<boolean>}
+   */
+  async warnAboutClosing(window) {
+    const tabCount = window.gBrowser.openTabs.length;
```
1rneh: Multiple browser windows might exist. `window.gBrowser.openTabs.length` only counts the open tabs of the current chrome window.

1rneh (changes requested):
Round of review # 1: 

We also need to handle the case where the app is explicitly quit via the keyboard shortcut, e.g. Cmd+Q on macOS.
The entry point appears to be in `BrowserGlue.sys.mjs` when we receive a [`quit-application-requested` notification.](https://searchfox.org/firefox-main/rev/83a58bd4ba576115c4ecdffa50588200afe6ae47/browser/components/BrowserGlue.sys.mjs#210-212)

For reference, the following dialog is still displayed:
Image

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Thanks for pointing out the MacOS `Cmd+Q` path - Matt gave me a new UX for that. It also made me think to test a bit more in Windows and it looks like the `BrowserGlue` change is needed there as well for the hamburger menu -> exit and `Cmd+Shift+Q`. 

Also FYI, I've split up the UX and signout on close logic to separate PRs. This PR will display the existing signout prompt when closing to make reviewing the exit paths easier without the extra changes for the custom dialog.

UX PR: https://github.com/mozilla/enterprise-firefox/pull/569

I address this comment [here](https://github.com/mozilla/enterprise-firefox/pull/569/changes#diff-59884ab0fc783f2f9bfc6070977b4e3d53e3a21398fabbcddc109ecbb2ceebbbR169)

browser/components/BrowserGlue.sys.mjs
```diff
@@ -4,6 +4,7 @@
 
 import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
 import { XPCOMUtils } from "resource://gre/modules/XPCOMUtils.sys.mjs";
+import { EnterpriseHandler } from "./enterprise/EnterpriseHandler.sys.mjs";
```
1rneh: This import is unused.

browser/components/BrowserGlue.sys.mjs
```diff
@@ -37,6 +38,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   DistributionManagement: "resource:///modules/distribution.sys.mjs",
   DownloadsViewableInternally:
     "moz-src:///browser/components/downloads/DownloadsViewableInternally.sys.mjs",
+  EnterpriseHandler: "resource:///modules/enterprise/EnterpriseHandler.sys.mjs",
```
1rneh: Could you extract this and make it conditional?

```
if (AppConstants.MOZ_ENTERPRISE) {
  ChromeUtils.defineESModuleGetters(lazy, {
    EnterpriseHandler: "resource:///modules/enterprise/EnterpriseHandler.sys.mjs",
  })
}
```

browser/base/content/browser.js
```diff
@@ -3821,6 +3820,23 @@ function WindowIsClosing(event) {
...
+      }
+    }
+    if (isLastWindow) {
+      // Return false to prevent the standard close flow; warnAboutClosing
+      // handles shutdown asynchronously after the user confirms the dialog.
+      EnterpriseHandler.onSignOut(window);
+      return false;
+    }
+  }
+
```
1rneh: I’m not very confident in this solution. 
 
A few issues I noticed:
- on macOS, closing the last window would quit the app, which is unexpected: https://searchfox.org/enterprise-main/rev/a038f49228d707c6675ef20ce640034a64307d2e/browser/base/content/browser.js#3888-3899
- if another PB window is open while closing the last non pb window, we fall back to the non-enterprise dialog.
- we bypass other logic in `warnAboutClosingWindow`, such as notification handling: https://searchfox.org/enterprise-main/rev/a038f49228d707c6675ef20ce640034a64307d2e/browser/base/content/browser.js#3886
- if the user accepts the enterprise sign-out prompt and we quit, we do not actually close the window or run the usual follow-up actions afterward.: https://searchfox.org/firefox-main/rev/fa20ce29b1a82b717b7ff6863af8c12b381b3ec5/toolkit/content/globalOverlay.js#42-47. 

I'm worried about the unknown and unwanted side effects we'll introduce.

A more robust option with fewer side effects might be to call `EnterpriseHandler.onSignOut` from `warnAboutClosingTabs`, return the result of the enterprise sign-out dialog (accepted or canceled), and then let the normal code path continue handling shutdown. For that, we could add an argument like `EnterpriseHandler.onSignOut(window, isTriggerShutdown)` and pass `false` here. Inside `onSignOut`, we could pull `await lazy.ConsoleClient.signoutUser()` out of `initiateShutdown()` and only call `initiateShutdown()` conditionally when shutdown is actually intended.

Let me know what you think of it.

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1541,6 +1542,16 @@ BrowserGlue.prototype = {
...
 
+    // Cancel quitting and instead show the Enterprise signout dialog
+    if (AppConstants.MOZ_ENTERPRISE) {
+      aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
+      let topWindow = lazy.BrowserWindowTracker.getTopWindow({
+        allowFromInactiveWorkspace: true,
+      });
+      lazy.EnterpriseHandler.onSignOut(topWindow)
+      return;
+    }
```
1rneh, line 1561: I have the same concern here.

Could we wrap [this](https://searchfox.org/firefox-main/rev/fa20ce29b1a82b717b7ff6863af8c12b381b3ec5/browser/components/BrowserGlue.sys.mjs#1425-1601) in a large conditional block and await the result of the enterprise sign-out dialog instead? Also here we should not initiate the shutdown but let the normal code path continue. To do that we need to make sure to finish with the following: 

If quit is confirmed:
```
this._quitSource = "unknown";
aCancelQuit.data = false;
```
If quit is canceled:
```
aCancelQuit.data = true;
```

1rneh (changes requested):
Review round # 2:

Sorry, I have quite a few changes to request, but I think they’ll help cover edge cases and avoid unintended side effects.

browser/components/BrowserGlue.sys.mjs
1rneh: I see that's outdated :)

browser/base/content/browser.js
jporter-dev: I suppose I took the criteria of "Close Firefox Window" too literally, that path only makes sense for Windows, so going that route I kept it simple to warn when our last window is exited and relied on `initiateShutdown` to handle the rest. Thinking more on the MacOS quirks and the updates from Matt, it makes sense to follow the regular browser window close and app quit paths with our custom dialog response to control the flow than exiting early with our handler.

I like the idea of passing the dialog response argument and separating signout and shutdown for this purpose. I'll take a look at refactoring this to mesh better with the existing close/exit flows. 

> if another PB window is open while closing the last non pb window, we fall back to the non-enterprise dialog.

As for this scenario, I thought it made sense to leave it alone since this is not a signout path so we don't need the extra signout text. If it makes sense to replace with our custom dialog. I can take another look when refactoring to see if it makes sense to show the custom dialog here.

Just to document some scenarios, and let me know if you spot any others:

- Badge signout: signout only, relaunch FELT
- MacOS
  - Top bar app name + Quit: signout and shutdown
  - Shortcut quit (cmd+q): signout and shutdown
  - Close last window: don't sign out
- Non-MacOS
  - Hamburger menu + quit / File + quit: signout and shutdown
  - Shortcut quit (ctrl+shift+q): signout and shutdown
  - Close last window: signout and shutdown

Lots to think about, thanks for the feedback!

1rneh: > As for this scenario, I thought it made sense to leave it alone since this is not a signout path so we don't need the extra signout text. If it makes sense to replace with our custom dialog. I can take another look when refactoring to see if it makes sense to show the custom dialog here.

You are right! Sorry, my bad - I initially thought this would close both windows, but it doesn't. So leaving it as-is makes sense.

> Just to document some scenarios, and let me know if you spot any others:
> 
>     * Badge signout: signout only, relaunch FELT
> 
>     * MacOS
>       
>       * Top bar app name + Quit: signout and shutdown
>       * Shortcut quit (cmd+q): signout and shutdown
>       * Close last window: don't sign out
> 
>     * Non-MacOS
>       
>       * Hamburger menu + quit / File + quit: signout and shutdown
>       * Shortcut quit (ctrl+shift+q): signout and shutdown
>       * Close last window: signout and shutdown

Tbh, I was simply playing around with your patch and trying different closing scenarios on macos to see when we hit `warnAboutClosingWindow` - that’s how I noticed the inconsistency. But from what I can tell, the different scenarios look correct. 

I think my proposal could simplify things quite a lot. We can rely on the existing logic and focus on solely replacing the dialogs conditionally while returning the expected values.

browser/components/BrowserGlue.sys.mjs
jporter-dev: I think I pushed it as you were reviewing 😄

browser/components/BrowserGlue.sys.mjs
jporter-dev: I tried to implement this the way you mentioned as it made sense, but ran into an issue with the signout process being async. Rather than wrapping it in a conditional, signout flips a flag `_signoutComplete` to act as a guard to essentially block the path until we're confirmed + signed out. I tested on Windows and MacOS and this seems to work. Let me know if you know of a better way to handle this async behavior within a sync function without wrapping everything in the `.then` (unless that's preferred)

I was also able to remove the work in `browser.js` with the `BrowserGlue` implementation since it already handles the non-MacOS final window closed exit path. This way we leave the last window close behavior on MacOS as-is and only prompt on true exits.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -652,7 +652,12 @@ export const ConsoleClient = {
    *
    * This is expected to be executed from the browser side.
    */
-  async signoutUser() {
+  /**
```
1rneh: This can be one comment :)

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -186,31 +198,37 @@ export const EnterpriseHandler = {
...
+    return true;
   },
 
-  async initiateShutdown() {
+  async initiateShutdown(shouldTriggerShutdown = true) {
     // TODO: Bug 2001029 - Assert or force-enable session restore?
 
     try {
-      await lazy.ConsoleClient.signoutUser();
+      await lazy.ConsoleClient.signoutUser(shouldTriggerShutdown);
```
1rneh: The logic in `lazy.ConsoleClient.signoutUser` can and needs to stay as-is. The flag should only determine whether to trigger the shutdown in `EnterpriseHandler`.

browser/components/BrowserGlue.sys.mjs
1rneh: Right, I thought about this a bit.. 

Another approach could be to also support the prompt to be built and triggered synchronously: https://firefox-source-docs.mozilla.org/toolkit/components/prompts/prompts/nsIPromptService.html#javascript-sync

This solution would require
- `onSignoutSync` which builds and prompts synchronously, which is called by `BrowserGlue`
- `onSignoutAsync` which builds and prompts asynchronously, which is called when the user explicitly signs out (enterprise panel).

A lazy solution could be to only support a synchronous prompt.

@Mossop Do you have an opinion on this? We are trying to find out how to best intercept current paths to Firefox shutdown, to inject our close+signout prompt.

1rneh (changes requested):
Another proposal incoming ...

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Will re-add. I removed this because for some reason my exits were leaving me with a FELT window. I misunderstood the logic in this and thought it was causing that issue, but now I'm thinking it was due to a previous implementation of the logic in `BrowserGlue.sys.mjs` not handling the flag properly.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
jporter-dev: Rebase merge conflict got me.. I can remove this altogether when reverting the change in this file

browser/components/BrowserGlue.sys.mjs
jporter-dev: I think I like this idea, it's an existing pattern and will simplify things and let us use our prompt + signout to control the flow without any quirks.

jporter-dev: I came across [this](https://searchfox.org/firefox-main/source/toolkit/components/asyncshutdown/AsyncShutdown.sys.mjs), need to read into it some more but maybe we can leverage this to make things easier by separating signout and prompting?

- Register `signoutOnShutdown` (does not prompt) as an async blocker
- `onSignoutSync` just handles the prompt flow synchronously for use in `BrowserGlue.sys.mjs`
- `onSignout` or `onSignoutAsync` handles the signout flow for the badge signout

I need to read up on how AsyncShutdown is used and what phase our signout fits into, but I _think_ it can help us here. Let me know your thoughts!

jporter-dev: [Here's a rough example](https://github.com/jporter-dev/enterprise-firefox/pull/1/changes#diff-e85d9d1ca59e4bcb90f33de35ee192db1fd813cef1dca910e85410b9c58804a1) using the AsyncShutdown blocker. Blocker gets registered, and only runs when signout is confirmed (to avoid messing with [this existing blocker](https://searchfox.org/enterprise-main/source/browser/components/enterprise/modules/ConsoleClient.sys.mjs#691)). 

`onSignOutSync` (which may be renamed since its focused around prompting now) just handles the prompt logic. This keeps our changes in `BrowserGlue.sys.mjs` to a minimum.

1rneh: I like the idea! By decoupling the prompting and signing out you also solve the problem of having to deal with the async call to `ConsoleClient.signoutUser` when prompting synchronously.

gcp: Note that you will run into `_xhrFetch` throwing `NS_ERROR_ILLEGAL_DURING_SHUTDOWN`  when doing `xhr.open()`, so this is tricky to do if you need to POST the logout to the console.

jporter-dev:
Closing this in favor of #674

PR #526, opened 2026-03-11
 Bug 2040659 - Device posture extensions

gcp:
Note the initial policy fetch won't have this as it's initialized late on startup, so the console won't get this until the live policy polls.

### DIFF
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -218,13 +218,16 @@ export const ConsoleClient = {
   },
 
   /**
-   * Fetches remote enterprise policies.
+   * Fetches remote enterprise policies, optionally including device posture.
    *
+   * @param {object} [devicePosture] - Device posture to include in request.
    * @returns {Promise<{policies: Record<string, any>}>}
    */
-  async getRemotePolicies() {
-    const payload = await this._get(this._paths.REMOTE_POLICIES);
-    return payload;
+  async getRemotePolicies(devicePosture = null) {
+    if (devicePosture) {
+      return this._post(this._paths.REMOTE_POLICIES, devicePosture);
+    }
+    return this._get(this._paths.REMOTE_POLICIES);
   },
 
   /**
@@ -347,7 +350,7 @@ export const ConsoleClient = {
    * @returns {Promise<{posture: string}>} Token reported by console.
    */
   async sendDevicePosture() {
-    const devicePosture = await this._collectDevicePosture();
+    const devicePosture = await this.collectDevicePosture();
     const url = this.constructURI(this._paths.DEVICE_POSTURE);
 
     const res = await this._xhrFetch(url, {
@@ -576,7 +579,7 @@ export const ConsoleClient = {
    *
    * @returns {Promise<DevicePosture>} devicePosture
    */
-  async _collectDevicePosture() {
+  async collectDevicePosture() {
     const getImeiValue = async () => {
       try {
         return await Cc["@mozilla.org/imei/provider;1"]
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -5,6 +5,7 @@
 const lazy = {};
 
 ChromeUtils.defineESModuleGetters(lazy, {
+  AddonManager: "resource://gre/modules/AddonManager.sys.mjs",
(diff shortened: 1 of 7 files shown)


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -587,6 +598,34 @@ export const ConsoleClient = {
       }
     };
 
+    // Best-effort: AddonManager is only available in the full browser
+    // process, not in the FELT login window where this is first called.
+    // Returns null when addon info is unavailable (FELT UI, AddonManager
+    // not ready) to distinguish from an empty list (no addons installed).
+    const getExtensions = async () => {
+      try {
+        if (Services.felt.isFeltUI() || !lazy.AddonManager.isReady) {
```
lissyx: Can we/do we need to have something that waits here? If `AddonManager` is really slow to start, we may send `null` even from browser for some time? Is it a problem?

gcp: There's a chicken and egg problem here as you want policies that apply on startup to be known, but you can't know extensions until you finish startup.

"Send null from browser from time to time", I mean, not after the first poll? ~~This does change the live polling code to first get the extension list ready.~~ Welp this needs an await.

testing/enterprise/test_felt_device_posture_poll.py
```diff
@@ -0,0 +1,68 @@
...
+        # null -> Python None). Verify the extraction logic doesn't crash.
+        null_posture = {"extensions": None}
+        null_exts = null_posture.get("extensions") or []
+        assert [e["id"] for e in null_exts] == [], (
+            "Null extensions must not raise TypeError"
+        )
+
+        # The policy poll interval is 500ms in tests. Wait for a poll cycle
+        # from the browser process to overwrite the posture payload with one
+        # that includes the force-installed extension.
```
lissyx: We try for 20s (40 retries every 500ms if we have not found tree style tab), so it's more than just "wait for a poll cycle" no?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -992,7 +992,13 @@ class RemotePoliciesProvider {
...
 
   _performPolling() {
-    lazy.ConsoleClient.getRemotePolicies()
+    lazy.ConsoleClient.sendDevicePosture()
+      .catch(error => {
+        console.warn(
+          `RemotePoliciesProvider: device posture update failed: ${error}`
+        );
+      })
+      .then(() => lazy.ConsoleClient.getRemotePolicies())
```
lissyx: Would this allow  `sendDevicePosture` to fail and us still call `getRemotePolicies()` ?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -1040,6 +1046,12 @@ class RemotePoliciesProvider {
       return;
     }
 
+    try {
+      await lazy.ConsoleClient.sendDevicePosture();
+    } catch (e) {
+      console.warn(`Device posture update on startup failed: ${e}`);
+    }
```
lissyx: I tought we wanted to make device posture something blocking, is it OK we allow failure here?

testing/enterprise/test_felt_device_posture_poll.py
gcp: I'm changing this to verify that we get a null followed by a non-null (together with the await above).

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
gcp: Discussed with Fiji, the posture has to go in the same call as the policy fetch.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -587,6 +605,42 @@ export const ConsoleClient = {
...
+    // ready so extensions are always reported. When false (startup),
+    // returns null if AddonManager isn't ready yet to avoid blocking.
+    const getExtensions = async () => {
+      try {
+        if (Services.felt.isFeltUI()) {
+          return null;
+        }
+        if (!lazy.AddonManager.isReady) {
+          if (waitForAddons) {
+            await lazy.AddonManager.readyPromise;
```
lissyx, line 620: 💯

lissyx (approved):

PR #527, opened 2026-03-11
Bug 2023530 - add os_long_name os_short_name field to devicePosture

victhorlopez:
Enriches the device posture os object with a new os_long_name field sent to the enterprise console — a human-readable OS string, e.g.:
examples:
long 
macOS 15.3 (Sequoia)
Windows 11 22H2 (build 22621) 
Ubuntu 22.04 (5.15.0)

short
macOS 15 (Sequoia)
Windows 11
Ubuntu 22.04

### DIFF
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -9,6 +9,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   EnterpriseCommon: "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
   AsyncShutdown: "resource://gre/modules/AsyncShutdown.sys.mjs",
   FeltStorage: "resource:///modules/FeltStorage.sys.mjs",
+  composeOSNames: "resource:///modules/enterprise/EnterpriseOSInfo.sys.mjs",
 });
 
 ChromeUtils.defineLazyGetter(lazy, "log", () => {
@@ -614,8 +615,17 @@ export const ConsoleClient = {
       .getService()
       .QueryInterface(Ci.nsINetworkLinkService).networkInterfaces;
 
+    const baseOs = lazy.TelemetryEnvironment.currentEnvironment.system.os;
+    const { long: os_long_name, short: os_short_name } =
+      await lazy.composeOSNames(baseOs);
+    const os = {
+      ...baseOs,
+      ...(os_long_name != null && { os_long_name }),
+      ...(os_short_name != null && { os_short_name }),
+    };
+
     const devicePosturePayload = {
-      os: lazy.TelemetryEnvironment.currentEnvironment.system.os,
+      os,
       security: lazy.TelemetryEnvironment.currentEnvironment.system.sec,
       build: lazy.TelemetryEnvironment.currentEnvironment.build,
       network: {
--- a/browser/components/enterprise/modules/EnterpriseOSInfo.sys.mjs
+++ b/browser/components/enterprise/modules/EnterpriseOSInfo.sys.mjs
@@ -1,0 +1,225 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  WindowsRegistry: "resource://gre/modules/WindowsRegistry.sys.mjs",
+});
+
+const WINDOWS_CURRENT_VERSION_KEY =
+  "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion";
+
+const MACOS_SYSTEM_VERSION_PLIST =
+  "/System/Library/CoreServices/SystemVersion.plist";
+
+let gMacOSVersionPromise = null;
+
+/**
+ * Reads the macOS version from SystemVersion.plist.
+ *
+ * @returns {Promise<string|null>}
+ */
+// TODO: Once bug 2017277 is fixed, os.version can be used instead of this function.
+async function getMacOSVersion() {
+  if (gMacOSVersionPromise) {
+    return gMacOSVersionPromise;
+  }
+  gMacOSVersionPromise = (async () => {
+    try {
+      const content = await IOUtils.readUTF8(MACOS_SYSTEM_VERSION_PLIST);
+      const doc = new DOMParser().parseFromString(content, "application/xml");
+      for (const key of doc.querySelectorAll("dict > key")) {
+        if (key.textContent === "ProductVersion") {
+          return key.nextElementSibling?.textContent ?? null;
+        }
+      }
+    } catch (e) {
+      console.error("getMacOSVersion: failed to read SystemVersion.plist", e);
+    }
+    return null;
+  })();
+  return gMacOSVersionPromise;
+}
+
+/**
+ * Returns the Windows release label (e.g. "22H2", "23H2", "1909") from the
+ * registry. Prefers DisplayVersion (introduced in 20H2), falls back to
+ * ReleaseId for older versions. Returns null if unavailable or not on Windows.
+ *
+ * @returns {string|null}
+ */
+function getWindowsReleaseLabel() {
+  for (const valueName of ["DisplayVersion", "ReleaseId"]) {
+    const value = lazy.WindowsRegistry.readRegKey(
+      Ci.nsIWindowsRegKey.ROOT_KEY_LOCAL_MACHINE,
+      WINDOWS_CURRENT_VERSION_KEY,
+      valueName,
+      Ci.nsIWindowsRegKey.WOW64_64
+    );
+    if (value != null) {
+      return value;
+    }
+  }
+  return null;
+}
+
+/**
+ * Returns true if the current Windows installation is a Server edition.
+ *
+ * @returns {boolean}
+ */
+function isWindowsServer() {
+  const installationType = lazy.WindowsRegistry.readRegKey(
+    Ci.nsIWindowsRegKey.ROOT_KEY_LOCAL_MACHINE,
+    WINDOWS_CURRENT_VERSION_KEY,
+    "InstallationType",
+    Ci.nsIWindowsRegKey.WOW64_64
+  );
+  return installationType === "Server";
+}
+
+const WINDOWS_SERVER_NAMES = [
+  [26100, "2025"],
+  [20348, "2022"],
+  [17763, "2019"],
+  [14393, "2016"],
+];
+
+/**
+ * Returns the Windows Server version string (e.g. "2025", "2022") from a build
+ * number, or null if the build number is not recognized.
+ *
+ * @param {number} buildNumber
+ * @returns {string|null}
+ */
+function getWindowsServerVersion(buildNumber) {
+  const match = WINDOWS_SERVER_NAMES.find(([min]) => buildNumber >= min);
+  return match ? match[1] : null;
+}
+
+const MACOS_NAMES = [
+  ["26", "Tahoe"],
+  ["15", "Sequoia"],
+  ["14", "Sonoma"],
+  ["13", "Ventura"],
+  ["12", "Monterey"],
+  ["11", "Big Sur"],
+  ["10.16", "Big Sur"],
+  ["10.15", "Catalina"],
+];
+
+/**
+ * Returns the Windows major version string ("11", "10", or "Unsupported") from
+ * a build number.
+ *
+ * @param {number} buildNumber
+ * @returns {string}
+ */
+function getWindowsMajorVersion(buildNumber) {
+  const versions = [
+    [22000, "11"],
+    [10240, "10"],
+  ];
+  const match = versions.find(([min]) => buildNumber >= min);
+  return match ? match[1] : "Unsupported";
+}
+
+/**
+ * Composes both the long and short human-readable OS names, reading SystemVersion.plist
+ * at most once on macOS.
+ *
+ * @param {object} os - The system.os object from TelemetryEnvironment.currentEnvironment.
+ * @returns {Promise<{long: string|null, short: string|null}>}
+ */
+export async function composeOSNames(os) {
+  const [long, short] = await Promise.all([
+    composeOSName(os),
+    composeShortOSName(os),
+  ]);
+  return { long, short };
+}
+
+/**
+ * Composes a short human-readable OS name from TelemetryEnvironment's system.os object.
+ *
+ * @param {object} os - The system.os object from TelemetryEnvironment.currentEnvironment.
+ * @returns {Promise<string|null>} A short OS string (e.g. "macOS 15 (Sequoia)", "Windows 11",
+ *   "Windows Server 2025", "Ubuntu 22.04"), or null if the OS cannot be identified.
+ */
+async function composeShortOSName(os) {
+  if (AppConstants.platform === "macosx") {
+    const macosVersion = await getMacOSVersion();
+    if (!macosVersion) {
+      return null;
+    }
+    const match = MACOS_NAMES.find(([prefix]) =>
+      macosVersion.startsWith(prefix)
+    );
+    const major = macosVersion.split(".")[0];
+    return match ? `macOS ${major} (${match[1]})` : `macOS ${major}`;
+  }
+  if (os.windowsBuildNumber) {
+    if (isWindowsServer()) {
+      const serverVersion = getWindowsServerVersion(os.windowsBuildNumber);
+      return serverVersion
+        ? `Windows Server ${serverVersion}`
+        : "Windows Server";
+    }
+    return `Windows ${getWindowsMajorVersion(os.windowsBuildNumber)}`;
+  }
+  if (os.distro && os.distroVersion) {
+    return `${os.distro} ${os.distroVersion}`;
+  }
+  console.error("composeShortOSName: unable to identify OS from", os);
+  return null;
+}
+
+/**
+ * Composes a human-readable OS name from TelemetryEnvironment's system.os object.
+ * On macOS, the version is read directly from SystemVersion.plist.
+ * On Windows, the installation type and release label are read from the registry.
+ *
+ * @param {object} os - The system.os object from TelemetryEnvironment.currentEnvironment.
+ * @returns {Promise<string|null>} A human-readable OS string (e.g. "macOS 15.3 (Sequoia)",
+ *   "Windows 11 22H2 (build 22621)", "Windows Server 2025 (build 26100)",
+ *   "Ubuntu 22.04 (5.15.0)"), or null if the OS cannot be identified.
+ */
+async function composeOSName(os) {
+  if (AppConstants.platform === "macosx") {
+    const macosVersion = await getMacOSVersion();
+    if (!macosVersion) {
+      return null;
+    }
+    const match = MACOS_NAMES.find(([prefix]) =>
+      macosVersion.startsWith(prefix)
+    );
+    return match
+      ? `macOS ${macosVersion} (${match[1]})`
+      : `macOS ${macosVersion}`;
+  }
+  if (os.windowsBuildNumber) {
+    const releaseLabel = getWindowsReleaseLabel();
+    const releaseSuffix = releaseLabel ? ` ${releaseLabel}` : "";
+    if (isWindowsServer()) {
+      const serverVersion = getWindowsServerVersion(os.windowsBuildNumber);
+      const prefix = serverVersion
+        ? `Windows Server ${serverVersion}`
+        : "Windows Server";
+      return `${prefix}${releaseSuffix} (build ${os.windowsBuildNumber})`;
+    }
+    // Maps minimum build numbers to Windows major versions, since
+    // windowsBuildNumber is more reliable than the version string for
+    // distinguishing Windows 10 from 11.
+    const major = getWindowsMajorVersion(os.windowsBuildNumber);
+    return `Windows ${major}${releaseSuffix} (build ${os.windowsBuildNumber})`;
+  }
+  if (os.distro && os.distroVersion && os.version) {
+    return `${os.distro} ${os.distroVersion} (${os.version})`;
+  }
+  console.error("composeOSName: unable to identify OS from", os);
+  return null;
+}
--- a/testing/enterprise/test_felt_device_posture.py
+++ b/testing/enterprise/test_felt_device_posture.py
@@ -4,6 +4,8 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import os
+import platform
+import re
 import sys
 import time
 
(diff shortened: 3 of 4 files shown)


browser/components/enterprise/tests/xpcshell/xpcshell.toml
```diff
@@ -1,4 +1,6 @@
 [DEFAULT]
 firefox-appdir = "browser"
 
+["test_console_client.js"]
```
lissyx: Please note our xpcshell tests are either nor running, or for those who are running, not green, so while it's great to have coverage here, it's not visible right now for our day to day operations

lissyx: The (maybe not uptodate) meta bug tracking making xpcshell tests green is https://bugzilla.mozilla.org/show_bug.cgi?id=2012260

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -11,6 +11,68 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
+    const match = versions.find(([min]) => os.windowsBuildNumber >= min);
+    const major = match ? match[1] : "Unknown";
+    return `Windows ${major} (build ${os.windowsBuildNumber})`;
+  }
+  if (os.distro) {
+    return `${os.distro} ${os.distroVersion} (${os.version})`;
+  }
+  console.error("composeOSName: unable to identify OS from", os, macosVersion);
+  return null;
+}
```
lissyx: I think all of that should be done on `nsSystemInfo`

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -11,6 +11,68 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
+    const macosNames = [
+      ["26", "macOS", "Tahoe"],
+      ["15", "macOS", "Sequoia"],
+      ["14", "macOS", "Sonoma"],
+      ["13", "macOS", "Ventura"],
+      ["12", "macOS", "Monterey"],
+      ["11", "macOS", "Big Sur"],
+      ["10.16", "macOS", "Big Sur"],
+      ["10.15", "macOS", "Catalina"],
+      ["10.14", "macOS", "Mojave"],
```
gcp: Firefox requires macOS 10.15 or later

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -11,6 +11,68 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
+    const match = macosNames.find(([prefix]) =>
+      macosVersion.startsWith(prefix)
+    );
+    return match
+      ? `${match[1]} ${macosVersion} ${match[2]}`
+      : `macOS ${macosVersion}`;
+  }
+  if (os.windowsBuildNumber) {
+    const versions = [
+      [22000, "11"],
```
gcp: Probably want to add `22H2` etc version identifiers.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: If we're the only users, it can go in the JS code too? But probably in a separate file.

gcp (changes requested):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: Can we agree in here on this? 

The reason it was in the js initial it was because it was supposed to be a "composed value" out of telemetry (+ the extra fields).

I would be fine to do it in c++ as long as it does not mean writing again another implementation for several things + running into races on how

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: yeah I am adding this right now, trying that it comes not from a table but programmatically

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: @victhorlopez Well, I dont have a strong no on that, looking at the code it felt more natural for me that everything is done in lower level rather than here, but this can be refactored later. I guess we may want to talk with profiler and other folks who already have similar hacks here and there to unify. So having it on our side in a separate JS file would work as well.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: https://www.firefox.com/en-US/firefox/148.0/system-requirements/

Is this enforced somewhere? (i.e do the installers prevent you from that?)

It is the same for win7 and win8. Just so I know If I need to remove those or it is ok to have them

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: Moved to C++

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: I'm not sure the installers actively prevent it but it won't really "work" on deprecated versions. It's possible you can't even launch it.

victhorlopez: Then I guess we can remove the one not supported (same for windows with the ones with version lower than 10)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -591,8 +591,23 @@ export const ConsoleClient = {
...
+        return Services.sysinfo.getProperty(name);
+      } catch {
+        return null;
+      }
+    };
+
+    const baseOs = lazy.TelemetryEnvironment.currentEnvironment.system.os;
+    const os = {
+      ...baseOs,
+      macos_version: getSysinfoProperty("macosVersion"),
```
victhorlopez: this would be sent to backend as null, not sure if we want this key at all if it is not macos.

Another point here is that we need to have this schema documented somewhere

testing/enterprise/test_felt_device_posture.py
```diff
@@ -49,6 +50,42 @@ def run_device_posture_content(self):
...
         )
         assert "secureBootEnabled" in device_posture
+
+        os_long_name = device_posture["os"].get("os_long_name")
+
+        assert os_long_name and len(os_long_name) > 0, (
+            "Device posture reports os_long_name"
+        )
+
+        if sys.platform == "darwin":
```
victhorlopez, line 60: the checks in here are not super accurate, I had a look at mozinfo(which reports platform info on CI)  but not all info is reported in there. So it is hard to have a "reference value" without doing the same that our code is doing to have a reference value. If you have ideas on how to improve this let me know

xpcom/base/nsSystemInfo.cpp
```diff
@@ -533,9 +533,21 @@ static nsresult GetWindowsSecurityCenterInfo(nsAString& aAVInfo,
 
 #  endif  // __MINGW32__
 
+
```
lissyx: nit: remove

xpcom/base/nsSystemInfo.cpp
```diff
@@ -588,6 +600,113 @@ static nsresult GetWinModelId(nsAutoString& aModelId) {
 }
 #endif
 
+#ifdef XP_MACOSX
+static nsresult GetMacOSOSComplete(nsAutoCString& aResult) {
```
lissyx: I'd like us to talk with other people using similar things (profiler for example) so it can be unified. Can you file a follow up and/or make a m-c patch for it?

I remember of https://searchfox.org/firefox-main/rev/7095e662eec49a42856e0d99f2ff96c08130119c/tools/profiler/core/platform.cpp#3324

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -591,8 +591,23 @@ export const ConsoleClient = {
       .getService()
       .QueryInterface(Ci.nsINetworkLinkService).networkInterfaces;
 
+    const getSysinfoProperty = name => {
+      try {
+        return Services.sysinfo.getProperty(name);
+      } catch {
+        return null;
+      }
```
lissyx: Instead of try catch, can we ensure the property is always here, and just null/empty string if `Get*OSComplete()` call failed?

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1722,6 +1841,16 @@ nsresult nsSystemInfo::Init() {
...
     NS_ENSURE_SUCCESS(rv, rv);
   }
+  nsAutoCString macosVersion;
+  if (NS_SUCCEEDED(GetMacOSVersion(macosVersion))) {
+    rv = SetPropertyAsACString(u"macosVersion"_ns, macosVersion);
+    NS_ENSURE_SUCCESS(rv, rv);
+  }
+  nsAutoCString macOSComplete;
+  if (NS_SUCCEEDED(GetMacOSOSComplete(macOSComplete))) {
+    rv = SetPropertyAsACString(u"osLongName"_ns, macOSComplete);
```
lissyx: I think we should define the properly as empty/null if there's a failure

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1735,6 +1864,12 @@ nsresult nsSystemInfo::Init() {
     rv = SetPropertyAsAString(u"winModelId"_ns, modelId);
     NS_ENSURE_SUCCESS(rv, rv);
   }
+
+  nsAutoCString windowsOSComplete;
+  if (NS_SUCCEEDED(GetWindowsOSComplete(windowsOSComplete))) {
+    rv = SetPropertyAsACString(u"osLongName"_ns, windowsOSComplete);
```
lissyx: same, define empty/null in case of failure

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1819,6 +1954,14 @@ nsresult nsSystemInfo::Init() {
   if (widget::lsb::GetLSBRelease(dist, desc, release, codename)) {
     SetPropertyAsACString(u"distro"_ns, dist);
     SetPropertyAsACString(u"distroVersion"_ns, release);
+    nsAutoCString kernelVersion;
+    GetPropertyAsACString(u"version"_ns, kernelVersion);
+    nsAutoCString linuxOSComplete;
+    if (NS_SUCCEEDED(
+            GetLinuxOSComplete(desc, kernelVersion, linuxOSComplete))) {
+      rv = SetPropertyAsACString(u"osLongName"_ns, linuxOSComplete);
```
lissyx: same, define empty/null in case of failure

xpcom/tests/unit/test_systemInfo.js
```diff
@@ -25,6 +25,98 @@ add_task(function properties() {
   Assert.equal(typeof sysInfo.getProperty("umask"), "number");
 });
 
+add_task(function test_osLongName() {
+  const os = Services.appinfo.OS;
+  const sysInfo = Services.sysinfo;
+  const osLongName = sysInfo.getProperty("osLongName");
```
lissyx: This would throw on any non linux/windows/macOS execution right now

xpcom/base/nsSystemInfo.cpp
victhorlopez: See my comment(on the review) 

```although at some point it would be nice to have a proper "PlatformInfoProvider"?)```

In general how do you handle this? Can we fill an enhancement bug to unify the platform information?  

(As you pointed in the original bug it is also in some gfx code, in rust(glean), in js(in the deprecated telemetry environment and I guess few other places)

Comment (lissyx) [reply xpcom/base/nsSystemInfo.cpp]: File a bug on bugzilla and maybe try with a first iteration of a patch. I hope all those places are parsing the data in a similar way?

Comment (victhorlopez) [reply xpcom/base/nsSystemInfo.cpp]: "File a bug on bugzilla and maybe try with a first iteration of a patch." in this MR or some patch on moz-central?

"I hope all those places are parsing the data in a similar way?" More or less I suppose... I guess as long as the interface gives the data that is expected and it is performant, the rest is just implementation detail

Comment (lissyx) [reply xpcom/base/nsSystemInfo.cpp]: Yes, on `mozilla-central`. I'd suggest extracting what makes sense from your current PR once we merge it here, and take that on m-c, pairing with other users to unify all to your implementation?

File: xpcom/base/nsSystemInfo.cpp (hunk 10 above)
Comment (victhorlopez) [reply xpcom/base/nsSystemInfo.cpp]: done, I set it on line 1838 for simplicity, let me know if you are ok with that

File: xpcom/base/nsSystemInfo.cpp (hunk 11 above)
Comment (victhorlopez) [reply xpcom/base/nsSystemInfo.cpp]: done, I set it on line 1838 for simplicity, let me know if you are ok with that

File: xpcom/base/nsSystemInfo.cpp (hunk 12 above)
Comment (victhorlopez) [reply xpcom/base/nsSystemInfo.cpp]: done, I set it on line 1838 for simplicity, let me know if you are ok with that

File: browser/components/enterprise/modules/ConsoleClient.sys.mjs (hunk 9 above)
Comment (victhorlopez) [reply browser/components/enterprise/modules/ConsoleClient.sys.mjs]: Can fix it for this property.  I am not sure about "Services.sysinfo.getProperty("macosVersion")" 
cause that is really macSpecific

File: xpcom/base/nsSystemInfo.cpp (hunk 10 above)
Comment (lissyx) [reply xpcom/base/nsSystemInfo.cpp]: yes, looks nice

File: browser/components/enterprise/modules/ConsoleClient.sys.mjs (hunk 9 above)
Comment (lissyx) [reply browser/components/enterprise/modules/ConsoleClient.sys.mjs]: You can probably rely on `AppConstants.platform` ?

Comment (victhorlopez) [reply browser/components/enterprise/modules/ConsoleClient.sys.mjs]: this one gives (unless I check the wrong one)
```
AppConstants.platform
"macosx" 
```
we are supposed to have "os version". (i.e 26.3.1) although it is already on the os_long_name (inside the string)

edit: ah got your point.. you want me to check instead of try catch

Comment (lissyx) [reply browser/components/enterprise/modules/ConsoleClient.sys.mjs]: I meant query `macOSVersion` based on `AppConstants.platform`

File: browser/components/enterprise/modules/ConsoleClient.sys.mjs (hunk 5 above)
Comment (gcp) [reply browser/components/enterprise/modules/ConsoleClient.sys.mjs]: >this would be sent to backend as null, not sure if we want this key at all if it is not macos.

Makes more sense to not have the key rather than sending pointless `null`. I think the console probably only really needs the ```os_long_name```?

Comment (victhorlopez) [reply browser/components/enterprise/modules/ConsoleClient.sys.mjs]: yeah.. so in theory I do not like strings on backend. I would prefer just fields, if they are fine getting a name composed of multiple things and parsing.. we can just remove macos_version.

The only additional info we are actually adding in this PR is the mac os marketing name (i.e tahoe)
and the mac os version( as opposed to the kernel one). The rest was already in the system.os

Comment (gcp) [reply browser/components/enterprise/modules/ConsoleClient.sys.mjs]: The console's requirement is to:

- Display the OS version of the client to the admin. It's nice if this is pretty and understandable.
- Be able to verify whether a client is on a "new enough" version to be allowed to connect (or connect to the most sensitive services).

For the latter they need to be able to get the full version somehow, either from parsing the fields from `os` or from parsing `os_long_name`. My point was that for now the console only has implemented 1) but not 2) so they're not actually using the rest. I understand your point that parsing the string is a bit ugly but in reality I don't think it matters much. No issue for us to include the full info of course, but no point in sending `null`s.

File: xpcom/base/nsSystemInfo.cpp (hunk 8 above)
Comment (gcp) [reply xpcom/base/nsSystemInfo.cpp]: This turns out to be complicated because `nsSysInfo.cpp`  is hot on the startup path, we've been moving code from there to TelemetryEnvironment (cached, async) to avoid that, but that's also the place we've been told not to add new stuff to.

**Comment by julienw** [comment]
Just a quick question, was that a conscious decision to go with a string rather than sending individual bits like we have for sessions?

File: browser/components/enterprise/modules/ConsoleClient.sys.mjs (hunk 5 above)
Comment (victhorlopez) [reply browser/components/enterprise/modules/ConsoleClient.sys.mjs]: @julienw See this comment regarding your question

**Comment by gcp** [comment]
> Just a quick question, was that a conscious decision to go with a string rather than sending individual bits like we have for sessions?

See https://github.com/mozilla/enterprise-firefox/pull/527#discussion_r2931525071

File: browser/components/enterprise/modules/EnterpriseOSInfo.sys.mjs [hunk 14]
```diff
@@ -0,0 +1,110 @@
...
+ * On macOS, the version is read directly from SystemVersion.plist.
+ * On Windows, the release label is read from the registry.
+ *
+ * @param {object} os - The system.os object from TelemetryEnvironment.currentEnvironment.
+ * @returns {Promise<string|null>} A human-readable OS string (e.g. "macOS 15.3 Sequoia",
+ *   "Windows 11 22H2 (build 22621)", "Ubuntu 22.04 (5.15.0)"), or null if the OS cannot be identified.
+ */
+export async function composeOSName(os) {
+  if (AppConstants.platform === "macosx") {
+    const macosVersion = await getMacOSVersion();
```
gcp, line 193: nit: this can return null, so you need a null check

testing/enterprise/test_felt_device_posture.py
```diff
@@ -49,6 +51,45 @@ def run_device_posture_content(self):
...
+            assert actual_build in os_long_name, (
+                f"os_long_name '{os_long_name}' should contain build number '{actual_build}'"
+            )
+        elif sys.platform == "linux":
+            os_release = platform.freedesktop_os_release()
+            pretty_name = os_release.get("NAME", "")
+            version_id = os_release.get("VERSION_ID", "")
+            # e.g. "Ubuntu 22.04 (5.15.0-91-generic)"
+            assert re.match(r"^.+ \d+(\.\d+)+ \(\S+\)$", os_long_name), (
+                f"os_long_name '{os_long_name}' does not match expected Linux format"
```
gcp: Fedora and Debian use different versions, but I guess this test assumes it'll always be on our CI with Ubuntu?

gcp (changes requested):

julienw:
> > Just a quick question, was that a conscious decision to go with a string rather than sending individual bits like we have for sessions?
> 
> See [#527 (comment)](https://github.com/mozilla/enterprise-firefox/pull/527#discussion_r2931525071)

I'm already deciding of a string from the individual bits for the sessions, I think it would be better to keep the same way of doing things.

Pros:
* One place to change when OS versions change (new version of Windows or macOS)
* Consistent display (eg: `macOS 26 (Tahoe)` vs `macOS Tahoe 26`).

Cons:
* I don't know, you'll tell me?

victhorlopez:
> > > Just a quick question, was that a conscious decision to go with a string rather than sending individual bits like we have for sessions?
> > 
> > 
> > See [#527 (comment)](https://github.com/mozilla/enterprise-firefox/pull/527#discussion_r2931525071)
> 
> I'm already deciding of a string from the individual bits for the sessions, I think it would be better to keep the same way of doing things.
> 
> Pros:
> 
>     * One place to change when OS versions change (new version of Windows or macOS)
> 
>     * Consistent display (eg: `macOS 26 (Tahoe)` vs `macOS Tahoe 26`).
> 
> 
> Cons:
> 
>     * I don't know, you'll tell me?

I think it is a question for @gcp 

We could just patch os.version for macos on js(and remove the patch once it is fixed on glean/c++) and add whatever we want extra in case backend needs (i.e windows display version)

As far as I understood, we want to build the string on client, hence this PR, but gcp needs to comment in here

gcp:
>I think it is a question for @gcp

No, it's not 😀 You both seem to agree providing the fields and having the console construct the pretty version string is preferable, and I already said above I'm fine with it ("No issue for us to include the full info of course").

I provided a suggestion because @julienw was out on PTO but if he has provided feedback on what the console prefers then definitely go that way unless there's pressing reasons not to (and there aren't - you already commented you prefer that solution yourself!).

victhorlopez:
> > I think it is a question for @gcp
> 
> No, it's not 😀 You both seem to agree providing the fields and having the console construct the pretty version string is preferable, and I already said above I'm fine with it ("No issue for us to include the full info of course").
> 
> I provided a suggestion because @julienw was out on PTO but if he has provided feedback on what the console prefers then definitely go that way unless there's pressing reasons not to (and there aren't - you already commented you prefer that solution yourself!).

@julienw @gcp 

My proposal would be 

## A: Schema untouched, os.version patched, pretty name on backend : 
- os.version patched for macos on our code (it will be eventually fixed in m-c and then we can remove the patch)

Pros: The schema remains t he same, which would create little work on backend
Cons: You will no longer have the kernel version of mac os. Additionally it looks it will be harder to update our backend(rollout) than client ( in case we need to add a new macos version name i.e 27)

## B: schema with new fields, pretty name on backend
Same as A the only difference here is => Do we need extra fields in the schema?
- Do you need the actual kernel version of mac? (which we will lose in this process)
- Do you need any other extra information? ( i.e windows display version? 25H2) or will you build that yourself with a table based on windows build number.

Pros: no tables on some fields ( less maintenance on backend)
Cons: Schema will be different than OS for any extra field

## C: schema with new fields + pretty name on client
Same as B + pretty name on client

Pros:  Easier to change that code in our enterprise setup(i.e rollout backend vs client).
Cons: Schema will be different than OS for any extra field


 see current [schema](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/environment.html)

gcp:
There is no point in (A) and (B), we expect much more frequent updates on the client (Firefox) than on the backend, so this kind of "pretty version" tables need to live in the client. But you should specify explicitly what you mean by (C). What are the fields and schema you are actually proposing.

testing/enterprise/test_felt_device_posture.py
victhorlopez: I fixed your case, this would probably fail on a rolling release distro (do we want to cover that as well?)

gcp: Yeah I think you'll get complaints about this 😁

browser/components/enterprise/modules/EnterpriseOSInfo.sys.mjs
```diff
@@ -0,0 +1,177 @@
...
+    // Maps minimum build numbers to Windows major versions, since
+    // windowsBuildNumber is more reliable than the version string for
+    // distinguishing Windows 10 from 11.
+    const major = getWindowsMajorVersion(os.windowsBuildNumber);
+    const releaseLabel = getWindowsReleaseLabel();
+    const releaseSuffix = releaseLabel ? ` ${releaseLabel}` : "";
+    return `Windows ${major}${releaseSuffix} (build ${os.windowsBuildNumber})`;
+  }
+  if (os.distro) {
+    return `${os.distro} ${os.distroVersion} (${os.version})`;
```
gcp, line 221: This all needs null checks: https://searchfox.org/firefox-main/rev/fa20ce29b1a82b717b7ff6863af8c12b381b3ec5/toolkit/components/telemetry/docs/data/environment.rst#154

testing/enterprise/test_felt_device_posture.py
gcp: Maybe skip this entirely if we're not on Ubuntu to prevent pointless test failures on developer machines?

gcp (changes requested):

testing/enterprise/test_felt_device_posture.py
```diff
@@ -49,6 +51,71 @@ def run_device_posture_content(self):
...
+        elif sys.platform == "win32":
+            # e.g. "Windows 11" or "Windows 10"
+            assert re.match(r"^Windows \d+$", os_short_name), (
+                f"os_short_name '{os_short_name}' does not match expected Windows format"
+            )
+        elif sys.platform == "linux":
+            os_release = platform.freedesktop_os_release()
+            pretty_name = os_release.get("NAME", "")
+            version_id = os_release.get("VERSION_ID", "")
+            assert pretty_name in os_short_name, (
```
gcp: This likely still fails on Arch etc. As already said in the previous review, limit this to Ubuntu for our CI only.

browser/components/enterprise/modules/EnterpriseOSInfo.sys.mjs
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Returns the Windows major version string ("11", "10", or "Unsupported") from
+ * a build number.
+ *
+ * @param {number} buildNumber
+ * @returns {string}
+ */
+function getWindowsMajorVersion(buildNumber) {
+  const versions = [
```
gcp, line 123: https://learn.microsoft.com/en-us/windows/release-health/windows-server-release-info

We may be able to cover Windows Server here too.

gcp (changes requested):

gcp (approved):
LGTM but I would bail the test early (before the first assertion) if we're on a "funny" distro

PR #530, opened 2026-03-11
Bug 2019063 - Tab crash logic to auto send crash report if policy enabled

jporter-dev:
## Summary

This PR adds the logic to automatically submit the tab crash report when the policy is enabled and the user is presented with the Tab Crash page. PR #531 adds the custom UI.

- [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2019063)

- https://github.com/mozilla/enterprise-firefox/pull/531
- https://github.com/mozilla/enterprise-admin-reference/pull/86

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -1066,6 +1066,26 @@ export var Policies = {
     },
   },
 
+  CrashReportsAutoSubmit: {
+    onBeforeAddons(_manager, param) {
+      if (param) {
+        setAndLockPref("browser.crashReports.unsubmittedCheck.enabled", true);
+        setAndLockPref(
+          "browser.crashReports.unsubmittedCheck.autoSubmit2",
+          true
+        );
+        setAndLockPref("browser.tabs.crashReporting.sendReport", true);
+        setAndLockPref("browser.tabs.crashReporting.includeURL", true);
+      }
+    },
+    onRemove(_manager, _oldParam) {
+      unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.enabled");
+      unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.autoSubmit2");
+      unsetAndUnlockPref("browser.tabs.crashReporting.sendReport");
+      unsetAndUnlockPref("browser.tabs.crashReporting.includeURL");
+    },
+  },
+
   DefaultDownloadDirectory: {
     onBeforeAddons(manager, param) {
       PoliciesUtils.setDefaultPref(
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -1066,9 +1066,9 @@ export var Policies = {
     },
   },
 
-  CrashReportsAutoSubmit: {
+  CrashReportsSubmit: {
     onBeforeAddons(_manager, param) {
-      if (param) {
+      if (param.ForceAutoSubmit) {
         setAndLockPref("browser.crashReports.unsubmittedCheck.enabled", true);
         setAndLockPref(
           "browser.crashReports.unsubmittedCheck.autoSubmit2",
--- a/browser/modules/ContentCrashHandlers.sys.mjs
+++ b/browser/modules/ContentCrashHandlers.sys.mjs
@@ -531,6 +531,15 @@ export var TabCrashHandler = {
    *        The browser that has recently crashed.
    */
   sendToTabCrashedPage(browser) {
+    if (Services.policies.getActivePolicies()?.CrashReportsAutoSubmit) {
+      let dumpID = this.getDumpID(browser);
+      if (dumpID) {
+        lazy.CrashSubmit.submit(dumpID, lazy.CrashSubmit.SUBMITTED_FROM_AUTO);
+        let childID = this.browserMap.get(browser);
+        this.childMap.set(childID, null); // Avoid resubmission from about:tabcrashed.
+      }
+    }
+
     let title = browser.contentTitle;
     let uri = browser.currentURI;
     let gBrowser = browser.getTabBrowser();
--- a/browser/modules/ContentCrashHandlers.sys.mjs
+++ b/browser/modules/ContentCrashHandlers.sys.mjs
@@ -531,7 +531,9 @@ export var TabCrashHandler = {
    *        The browser that has recently crashed.
    */
   sendToTabCrashedPage(browser) {
-    if (Services.policies.getActivePolicies()?.CrashReportsAutoSubmit) {
+    if (
+      Services.policies.getActivePolicies()?.CrashReportsSubmit?.ForceAutoSubmit
+    ) {
       let dumpID = this.getDumpID(browser);
       if (dumpID) {
         lazy.CrashSubmit.submit(dumpID, lazy.CrashSubmit.SUBMITTED_FROM_AUTO);
--- a/browser/modules/ContentCrashHandlers.sys.mjs
+++ b/browser/modules/ContentCrashHandlers.sys.mjs
@@ -532,6 +532,7 @@ export var TabCrashHandler = {
    */
   sendToTabCrashedPage(browser) {
     if (
+      AppConstants.MOZ_CRASHREPORTER &&
       Services.policies.getActivePolicies()?.CrashReportsSubmit?.ForceAutoSubmit
     ) {
       let dumpID = this.getDumpID(browser);
--- a/browser/components/enterprisepolicies/schemas/policies-schema.json
+++ b/browser/components/enterprisepolicies/schemas/policies-schema.json
@@ -542,8 +542,13 @@
       }
     },
 
-    "CrashReportsAutoSubmit": {
-      "type": "boolean"
+    "CrashReportsSubmit": {
+      "type": "object",
+      "properties": {
+        "ForceAutoSubmit": {
+          "type": "boolean"
+        }
+      }
     },
 
     "DefaultDownloadDirectory": {
--- a/browser/components/enterprisepolicies/schemas/policies-schema.json
+++ b/browser/components/enterprisepolicies/schemas/policies-schema.json
@@ -542,6 +542,10 @@
       }
     },
 
+    "CrashReportsAutoSubmit": {
+      "type": "boolean"
+    },
+
     "DefaultDownloadDirectory": {
       "type": "string"
     },
--- a/browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
+++ b/browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
@@ -7,4 +7,4 @@ policy-DownloadTelemetry = Enable and configure security logging/telemetry when 
 policy-EnterpriseStorageEncryption = Enable enterprise-managed primary password for encrypted storage.
 policy-PrintPageTelemetry = Enable and configure security logging/telemetry when a page is printed.
 policy-Sync = Enable or disable sync and define which data to include.
-policy-CrashReportsAutoSubmit = Enable or disable automatically sending crash reports.
+policy-CrashReportsSubmit = Configure crash report submission settings.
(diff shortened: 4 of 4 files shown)


browser/modules/ContentCrashHandlers.sys.mjs
```diff
@@ -531,6 +531,18 @@ export var TabCrashHandler = {
    *        The browser that has recently crashed.
    */
   sendToTabCrashedPage(browser) {
+    if (Services.policies.getActivePolicies()?.CrashReportsAutoSubmit) {
+      let dumpID = this.getDumpID(browser);
+      if (dumpID) {
+        UnsubmittedCrashHandler.submitReports(
+          [dumpID],
+          lazy.CrashSubmit.SUBMITTED_FROM_AUTO
```
jonathanmendez: I'm not sure how the SubmittedFrom string is used, but it seems plausible we'd still want this to be `SUBMITTED_FROM_CRASH_TAB` or even add a new string. @afranchuk may be able to weigh in.

browser/modules/ContentCrashHandlers.sys.mjs
```diff
@@ -531,6 +531,18 @@ export var TabCrashHandler = {
    *        The browser that has recently crashed.
    */
   sendToTabCrashedPage(browser) {
+    if (Services.policies.getActivePolicies()?.CrashReportsAutoSubmit) {
+      let dumpID = this.getDumpID(browser);
+      if (dumpID) {
+        UnsubmittedCrashHandler.submitReports(
```
jonathanmendez: It's not clear to me whether this is the "right way" to do the submission (vs `lazy.CrashSubmit.submit()` or `maybeSendCrashReport()`). I think @afranchuk should review the logic (or designate the right person to review).

browser/modules/ContentCrashHandlers.sys.mjs
```diff
@@ -531,6 +531,18 @@ export var TabCrashHandler = {
    *        The browser that has recently crashed.
    */
   sendToTabCrashedPage(browser) {
+    if (Services.policies.getActivePolicies()?.CrashReportsAutoSubmit) {
```
jonathanmendez: Putting this in `sendToTabCrashedPage()` _feels_ right. But just in case, looking at callers, [one of them](https://searchfox.org/firefox-main/source/browser/modules/ContentCrashHandlers.sys.mjs#218) has as its alternative path `sendToRestartRequiredPage()`, and I'm curious whether we need to catch that path and autosubmit from there too?

jonathanmendez (commented):
Reviewing this leaves me with some open questions, but I'm probably not the right person to review it. I'm hoping @afranchuk can review or knows who can.

browser/modules/ContentCrashHandlers.sys.mjs
afranchuk: I'm not that familiar with the content crash handling, but as far as I know, there's no crash report submission flow from the restart required page. I think if we crash and an update recently occurred, we just assume that the update is at fault and don't bother with doing crash reporting things. So this is likely in the right place.

browser/modules/ContentCrashHandlers.sys.mjs
afranchuk: Yeah, this isn't right. You should use `CrashSubmit.submit()` with `SUBMITTED_FROM_AUTO` (or _maybe_ you want a new value for that, if you want to distinguish policy auto submit, but it shouldn't be necessary since, for enterprise, auto submit will always be due to policy). The `UnsubmittedCrashHandler` is a separate thing, though FWIW if you look at `submitReports`, all it does is call `CrashSubmit.submit()` 😄  No need for an additional layer here.

browser/modules/ContentCrashHandlers.sys.mjs
afranchuk: Ah, I didn't see this comment before writing my other comment. As I said there, you could add a new string, but otherwise it can be `SUBMITTED_FROM_AUTO`.

afranchuk (changes requested):

browser/modules/ContentCrashHandlers.sys.mjs
jporter-dev: Thanks for the feedback! I looked at the wrong automatic submission flow for guidance here, I've updated it to use `CrashSubmit`. I'll keep `SUBMITTED_FROM_AUTO` for now unless we have a compelling reason to add our own.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -933,6 +933,29 @@ export var Policies = {
     },
   },
 
+  CrashReportsAutoSubmit: {
+    onBeforeAddons(_manager, param) {
+      if (param) {
+        setAndLockPref(
```
jporter-dev: I added these 4 prefs to be locked with the policy. 

`browser.crashReports.unsubmittedCheck.enabled` and `browser.crashReports.unsubmittedCheck.autoSubmit2` make sure unsubmitted reports get automatically submitted. 

`browser.tabs.crashReporting.includeURL` and `browser.tabs.crashReporting.sendReport` are mostly to indicate to the user that we've submitted with the URL and that the report was automatically sent since they don't go through the manual submission flow where these are used.

afranchuk (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -933,6 +933,26 @@ export var Policies = {
...
+    onBeforeAddons(_manager, param) {
+      if (param) {
+        setAndLockPref("browser.crashReports.unsubmittedCheck.enabled", true);
+        setAndLockPref(
+          "browser.crashReports.unsubmittedCheck.autoSubmit2",
+          true
+        );
+        setAndLockPref("browser.tabs.crashReporting.sendReport", true);
+        setAndLockPref("browser.tabs.crashReporting.includeURL", true);
+      }
```
mkaply, line 1079: If param is false, should we unset everything?

browser/modules/ContentCrashHandlers.sys.mjs
```diff
@@ -531,6 +531,15 @@ export var TabCrashHandler = {
...
```
mkaply: Why do we need code to autosubmit if all the prefs have been changed?

mkaply (changes requested):
Need to handle the false path.

browser/locales/en-US/browser/policies/policies-descriptions.ftl
```diff
@@ -51,6 +51,8 @@ policy-ContentAnalysis = Enable or disable connection to data-loss-prevention ag
 
 policy-Cookies = Allow or deny websites to set cookies.
 
+policy-CrashReportsAutoSubmit = Enable or disable automatically sending crash reports.
```
1rneh: The l10n team asked us to separate enterprise strings from non-enterprise strings. So this needs to go into `enterprise-policies-descriptions.ftl`: 
https://searchfox.org/enterprise-main/source/browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl

browser/modules/ContentCrashHandlers.sys.mjs
jporter-dev: That's explained a bit in [this ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=2019068#c0).

The built-in "autosubmission" does not submit immediately, instead the user is still presented with a form which will submit if they interact (close or restore tab). Otherwise the crash is added to a list to be submitted on next startup.

Keeping the prefs in their locked state along with the autosubmit logic adds immediate autosubmission along with autosubmission on startup for failed submissions (ex: no network connection).

mkaply: Thanks for the explanation!

browser/components/enterprisepolicies/Policies.sys.mjs
jporter-dev: Updated. This makes sense to me, but I did not see any other policies using this pattern so just want to make sure I have this right. From what I can tell, this should handle the following:

- policy is set to true
- policy is set to false
- policy is set and then removed

browser/locales/en-US/browser/policies/policies-descriptions.ftl
jporter-dev: Thanks, didn't catch that! Moved to the proper file

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: Actually, I described what to do wrong, sorry.

The question is what if an enterprise wants to force this to false. Should we set and lock everything to false.

jporter-dev: Ah that makes sense. Updated so now when the policy is explicitly set to false, we lock the prefs to false.

jonathanmendez: I'm not completely sure what we might be aiming for with the `false` case, and I think that overall this gets pretty confusing pretty quickly with the existing auto submit user pref.

We have:
1. Existing auto submit (`browser.crashReports.unsubmittedCheck.autoSubmit2`): Seems to be a convenience feature that a) sometimes sends background crash reports right away and b) automatically sends any pending reports e.g. when you start firefox. It does not prevent you from declining to send any individual report when you get prompted.
2. The new policy which is intended to always and immediately send a report, preventing the user from opting out and not waiting for firefox to notice pending reports.

It seems clear enough what to do when the policy is set to `true`.
It's less clear to me what would make sense when the policy is set to `false` (beyond default/not doing anything).
- We could disable/lock the user pref for autosubmit, but that just seems to be an annoyance rather than something an enterprise would specifically want? It also leans into the mixing of the two separate (but confusingly similar) things, the user pref and the policy.
- We could disable submitting crash reports entirely, but that seems extra weird generally and outside the scope of the policy's name/description.

I've asked @jporter-dev to needinfo @rtestard on the desired behavior, but if there's clarity to add here on handling `false` cases of policies that add behavior, I'd welcome that as well.

mkaply: We definitely have had cases with policy where enterprises never want any crash reports sent and don't want users to send them.

browser/modules/ContentCrashHandlers.sys.mjs
```diff
@@ -531,6 +531,17 @@ export var TabCrashHandler = {
    *        The browser that has recently crashed.
    */
   sendToTabCrashedPage(browser) {
+    if (
+      Services.policies.getActivePolicies()?.CrashReportsSubmit?.ForceAutoSubmit
+    ) {
+      let dumpID = this.getDumpID(browser);
+      if (dumpID) {
+        lazy.CrashSubmit.submit(dumpID, lazy.CrashSubmit.SUBMITTED_FROM_AUTO);
```
jonathanmendez, line 540: The existing code paths for submission go through [maybeSendCrashReport](https://searchfox.org/enterprise-main/source/browser/modules/ContentCrashHandlers.sys.mjs#571), which checks whether MOZ_CRASHREPORTER (build setting/appconstant, not env) is enabled. But `getDumpID` also returns null when it's not enabled, so this should be fine? So more of an FYI or you could choose to put in a check if you prefer.

jonathanmendez (approved):

browser/modules/ContentCrashHandlers.sys.mjs
jporter-dev: Was on the fence since the check happened and was already working, but it made sense to add the check for this just to be more explicit and not rely on `getDumpID` to guard for us.

mkaply (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
jonathanmendez: To capture a conclusion from elsewhere, we decided to restructure the policy and defer handling the `false` case, as the work involved would not fit the pilot timeline. Captured in https://bugzilla.mozilla.org/show_bug.cgi?id=2026983

PR #531, opened 2026-03-11
Bug 2019058 - Enterprise: show policy notice and disable crash report UI when CrashReportsAutoSubmit is enforced

jporter-dev:
## Summary

This PR alongside #530 add the full end to end tab crash report automatic submission. This PR adds a custom UI that replaces some of the default titles and messages with our own, and disables the form submission elements.

- [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2019058)

## Screenshots

Screenshot 2026-03-11 at 2 46 50 PM

_Crashed tab with automatic submission enforced, submission in network tab on the right_

- https://github.com/mozilla/enterprise-firefox/pull/530
- https://github.com/mozilla/enterprise-admin-reference/pull/86


browser/modules/ContentCrashHandlers.sys.mjs
jonathanmendez, line 694: At a high level, we want these to be forced true by the policy. From my own searching, this seems to be the only place these are read, so it's quite possible you can do something like:

```suggestion
let sendReport = policyAutoSubmit || this.prefs.getBoolPref("sendReport");
let includeURL = policyAutoSubmit || this.prefs.getBoolPref("includeURL");
```

browser/locales/en-US/browser/aboutTabCrashed.ftl
```diff
@@ -12,6 +12,8 @@ crashed-single-offer-help-message = Choose { crashed-restore-tab-button } to rel
 crashed-multiple-offer-help-message = Choose { crashed-restore-tab-button } or { crashed-restore-all-button } to reload the page/pages.
 crashed-request-help = Will you help us?
 crashed-request-help-message = Crash reports help us diagnose problems and make { -brand-short-name } better.
+crashed-policy-auto-submit-title = Crash reports help us improve
+crashed-policy-auto-submit-message = Per configured policy, a crash report is automatically sent to your admin.
```
jonathanmendez: @bsmth had provided me the following language that I find better:

```suggestion
crashed-policy-auto-submit-message = Your administrator has configured { -brand-short-name } to send crash reports automatically.
```

browser/base/content/aboutTabCrashed.js
```diff
@@ -164,6 +164,23 @@ var AboutTabCrashed = {
...
+      document.getElementById("requestAutoSubmit").hidden = true;
+      document.getElementById("sendReport").disabled = true;
+      document.getElementById("comments").disabled = true;
+      document.getElementById("includeURL").disabled = true;
+      document.getElementById("autoSubmit").disabled = true;
+
+      // Show the policy-based title and message.
+      document.getElementById("policyAutoSubmitTitle").hidden = false;
+      document.getElementById("policyAutoSubmitMessage").hidden = false;
+      this.showCrashReportUI(true);
```
jonathanmendez: `showCrashReportUI` is being called here a second time after the previous call above. It seems like this could cause flicker or other problems, including if `hasReport` is false. This logic should probably be nested in the `if (data.hasReport) {}` block above

browser/base/content/aboutTabCrashed.js
```diff
@@ -164,6 +164,23 @@ var AboutTabCrashed = {
       document.getElementById("requestAutoSubmit").hidden = false;
     }
 
+    if (data.policyAutoSubmit) {
+      // Report is automatically sent, hide default title and message
+      // and disable reporting options since there's no report to customize.
+      document.documentElement.classList.add("crashDumpAvailable");
+      document.querySelector("#reportBox > h2").hidden = true;
```
jonathanmendez: I think it would be better to give the h2 an id so it's clearer here what is being hidden.

browser/base/content/aboutTabCrashed.js
```diff
@@ -164,6 +164,23 @@ var AboutTabCrashed = {
...
+      document.getElementById("requestAutoSubmit").hidden = true;
+      document.getElementById("sendReport").disabled = true;
+      document.getElementById("comments").disabled = true;
```
jonathanmendez: Consider actually hiding this element. The others, while disabled, communicate "this is the active setting, even though you can't change it". But the empty & disabled comments box doesn't communicate anything, so is noise to the user.

browser/modules/ContentCrashHandlers.sys.mjs
jonathanmendez, line 692: I believe we want this to be true when the policy is true, but since this is surfaced as a user pref maybe we need the policy to set and lock it.

jonathanmendez (changes requested):
I found a few things to look into. Some of it may fit better with the other policy PR.

browser/modules/ContentCrashHandlers.sys.mjs
jporter-dev: I think it makes sense to lock all of the prefs referenced in your comments in the policy PR. Do you think there will ever be a scenario where the policy will need to enable autosubmission but disable `browser.tabs.crashReporting.includeURL`?

Prefs to be locked:

```
browser.crashReports.unsubmittedCheck.autoSubmit2
browser.tabs.crashReporting.sendReport
browser.tabs.crashReporting.includeURL
```

browser/base/content/aboutTabCrashed.js
jporter-dev: In the other PR, I clear out the DumpID since our report was submitted, so I _think_ `hasReport` will not be true in the current setup. I can move it into that block with a `|| data.policyAutoSubmit`, that should avoid any duplicate calls and can avoid duplicating some of the functionality.

browser/base/content/aboutTabCrashed.js
jporter-dev: Good call, I hid it since it doesn't make sense to display here. Now it only shows the two checked and disabled prefs along with the policy language to convey what happened.

Screenshot 2026-03-13 at 2 07 50 PM

browser/modules/ContentCrashHandlers.sys.mjs
jporter-dev: [Locked](https://github.com/mozilla/enterprise-firefox/pull/530/changes#r2933155480) these 3 plus `browser.crashReports.unsubmittedCheck.enabled` which is required for handling failed report submissions

jonathanmendez: > Do you think there will ever be a scenario where the policy will need to enable autosubmission but disable browser.tabs.crashReporting.includeURL?

That's a fair question. Right now the policy isn't nuanced. And if we wanted to add "include URL" as a policy setting it would be either on or off since the user doesn't get a chance to change the setting in the UI before the report is sent. You could ping @rtestard in case he wants it to be a certain way but my guess is we leave it without the nuance for now and wait for feedback?

browser/base/content/aboutTabCrashed.js
```diff
@@ -148,14 +148,29 @@ var AboutTabCrashed = {
   onSetCrashReportAvailable(message) {
     let data = message.data;
 
-    if (data.hasReport) {
+    if (data.hasReport || data.policyAutoSubmit) {
       this.hasReport = true;
```
jonathanmendez: It seems weird to me that in the case of policyAutoSubmit we're still setting `this.hasReport = true;`. I think that it's only used for sending the message for [submitting the report](https://searchfox.org/firefox-main/source/browser/actors/AboutTabCrashedParent.sys.mjs#20) when the user closes or restores the tab, but that's irrelevant for this case because it will have already been sent (it will end up trying to send the report, seeing that the dumpID is missing, and return early). It doesn't seem harmful but it would be nice to have the code not "lie" about having a report.

browser/modules/ContentCrashHandlers.sys.mjs
```diff
@@ -678,7 +678,10 @@ export var TabCrashHandler = {
     }
 
     let dumpID = this.getDumpID(browser);
-    if (!dumpID) {
+    let policyAutoSubmit =
+      !!Services.policies.getActivePolicies()?.CrashReportsSubmit
+        ?.ForceAutoSubmit;
+    if (!dumpID && !policyAutoSubmit) {
```
jonathanmendez, line 686: I think this line and/or below merits a comment explaining that in the case of policy auto submit, we will have submitted any report already (hence why hasReport is false) but that we want to display the state of the locked prefs (hence why we query and return sendReport and includeURL). Otherwise the logic in here seems odd.

jonathanmendez (approved):
A couple of nits that are probably worth fixing, but if schedule needs this to go in without them, that's fine.

browser/modules/ContentCrashHandlers.sys.mjs
jporter-dev: That makes sense, I'll add comments to clarify the reasoning for the special logic here.

browser/base/content/aboutTabCrashed.js
jporter-dev: I updated to set `hasReport` accurately based on the policy state, thanks for pointing that out!

PR #545, opened 2026-03-13
Bug 2023267 - Sync policy should adapt current sync state only when Locked=true

1rneh:
Generally `Locked==false` or `Locked==undefined` respects user choice. The current implementation of the Sync policy doesn't, which can lead to the user and the console fighting over a preference (or in this case the sync state) on each startup when policies get applied.

Once we have a default preference for Sync ([Bug 2017719](https://bugzilla.mozilla.org/show_bug.cgi?id=2017719)), `Enabled=X` and `Locked=false` or `Locked=undefined` will set sync's default state to `X`.

### DIFF
--- a/browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
+++ b/browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
@@ -57,7 +57,7 @@ export const SyncPolicy = {
    * @typedef {object} SyncPolicyParams
    * @property {boolean} [Enabled] Whether the feature sync should be enabled
    * @property {boolean} [Locked] Whether to lock the customized sync settings, hence
-   *                              the user modifications/preferences are not overridden.
+   *                              the user modifications/preferences will be overridden.
    *
    * // Per-engine sync configuration
    * @property {boolean} [Addons] Whether syncing addons should be enabled
--- a/browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
+++ b/browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
@@ -49,13 +49,17 @@ export const SyncPolicy = {
    *
    * @returns {boolean} Whether sync is currently enabled.
    */
-  isSyncEnabled() {
+  isCurrentSyncEnabled() {
     return lazy.Weave.Status.checkSetup() == SYNC_STATUS_OK;
   },
 
   /**
    * @typedef {object} SyncPolicyParams
-   * @property {boolean} [Enabled] Whether Sync should be enabled
+   * @property {boolean} [Enabled] Whether the feature sync should be enabled
+   * @property {boolean} [Locked] Whether to lock the customized sync settings, hence
+   *                              the user modifications/preferences are not overridden.
+   *
+   * // Per-engine sync configuration
    * @property {boolean} [Addons] Whether syncing addons should be enabled
    * @property {boolean} [Addresses] Whether syncing addresses should be enabled
    * @property {boolean} [Bookmarks] Whether syncing bookmarks should be enabled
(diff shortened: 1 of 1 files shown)


browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -49,13 +49,17 @@ export const SyncPolicy = {
    *
    * @returns {boolean} Whether sync is currently enabled.
    */
-  isSyncEnabled() {
+  isCurrentSyncEnabled() {
```
lissyx: nit: `isSyncCurrentlyEnabled`

lissyx (approved):

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -84,27 +87,24 @@ export const SyncPolicy = {
...
+    if (isIgnoringUserPreferences) {
+      const isCurrentSyncEnabled = this.isCurrentSyncEnabled();
+      if (shouldEnableSync && !isCurrentSyncEnabled) {
+        lazy.log.debug("Enable Sync");
         await this.connectSync(manager);
-      }
-    } else if (shouldEnableSync === false) {
-      lazy.log.debug("Disable Sync");
-      if (isSyncEnabled) {
+      } else if (shouldEnableSync === false && isCurrentSyncEnabled) {
```
gcp: nit: there is a mix in this code of strict ```=== false``` checks and truthiness checks.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -49,13 +49,17 @@ export const SyncPolicy = {
...
+  isCurrentSyncEnabled() {
     return lazy.Weave.Status.checkSetup() == SYNC_STATUS_OK;
   },
 
   /**
    * @typedef {object} SyncPolicyParams
-   * @property {boolean} [Enabled] Whether Sync should be enabled
+   * @property {boolean} [Enabled] Whether the feature sync should be enabled
+   * @property {boolean} [Locked] Whether to lock the customized sync settings, hence
+   *                              the user modifications/preferences are not overridden.
```
gcp: Isn't it the other way around?

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -49,13 +49,17 @@ export const SyncPolicy = {
...
    */
-  isSyncEnabled() {
+  isCurrentSyncEnabled() {
     return lazy.Weave.Status.checkSetup() == SYNC_STATUS_OK;
   },
 
   /**
    * @typedef {object} SyncPolicyParams
-   * @property {boolean} [Enabled] Whether Sync should be enabled
+   * @property {boolean} [Enabled] Whether the feature sync should be enabled
```
gcp, line 58: Note that Enabled is now a no-op unless Locked is set together with it.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -84,27 +87,24 @@ export const SyncPolicy = {
...
     } = params;
 
-    const isSyncEnabled = this.isSyncEnabled();
-
-    if (shouldEnableSync === true) {
-      lazy.log.debug("Enable Sync");
-      if (!isSyncEnabled) {
+    if (isIgnoringUserPreferences) {
+      const isCurrentSyncEnabled = this.isCurrentSyncEnabled();
+      if (shouldEnableSync && !isCurrentSyncEnabled) {
```
gcp: Note this exists only within the ```isIgnoringUserPreferences``` branch, so without Locked set the setting has no effect.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
1rneh: `shouldEnableSync` can be true, false, or undefined.
`if (!shouldEnableSync)` treats false and undefined the same, but `shouldEnableSync===false` and `shouldEnableSync===undefined` lead to different actions, so I explicitly distinguish them. The true case doesn’t need that.

PR #547, opened 2026-03-16
Bug 2023486 - Enterprise: ensure no pending-elevate state is left

lissyx:
Michael, can you verify it works for you?

### DIFF
--- a/browser/components/enterprise/modules/Updates.sys.mjs
+++ b/browser/components/enterprise/modules/Updates.sys.mjs
@@ -79,6 +79,7 @@ export const Updates = {
     this._appUpdater.addListener(this._updaterCallback);
     Services.obs.addObserver(this, "update-staged");
     Services.obs.addObserver(this, "update-downloaded");
+    Services.obs.addObserver(this, "update-error");
   },
 
   forceUpdateCheck() {
@@ -95,6 +96,7 @@ export const Updates = {
 
   // Similar to browser/base/content/aboutDialog-appUpdater.js:_onAppUpdateStatus
   appUpdaterCallback(status, downloadedBytes, totalBytes) {
+    console.warn(`FeltUpdates: appUpdaterCallback: status:${status}`);
     switch (status) {
       case lazy.AppUpdater.STATUS.CHECKING:
         this.scheduleDelayedUpdateCheckUI();
@@ -259,9 +261,19 @@ export const Updates = {
   },
 
   automaticRestart() {
-    Services.startup.quit(
-      Ci.nsIAppStartup.eForceQuit | Ci.nsIAppStartup.eRestart
-    );
+    // Ensure the on-disk status is changed from "pending-elevate" to
+    // "pending" before restarting. Without this, ProcessUpdates sees
+    // ePendingElevate on the next startup and skips launching the
+    // updater, while the JS layer sees STATE_PENDING and triggers
+    // another restart, causing an infinite restart loop.
+    Cc["@mozilla.org/updates/update-manager;1"]
+      .getService(Ci.nsIUpdateManager)
+      .elevationOptedIn()
+      .finally(() => {
+        Services.startup.quit(
+          Ci.nsIAppStartup.eForceQuit | Ci.nsIAppStartup.eRestart
+        );
+      });
   },
 
   observe(subject, topic, state) {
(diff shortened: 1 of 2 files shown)


michaelvanstraten (dismissed):
Tested it, works

lissyx:
> Tested it, works

Thanks, but I suspect we need also to make sure we cover the case of background update application

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -273,23 +284,51 @@ export const Updates = {
...
+        Services.obs.removeObserver(this, "update-error");
         Services.obs.removeObserver(this, "xpcom-shutdown");
         break;
       case "update-staged":
-      case "update-downloaded": {
+      case "update-downloaded":
         // states from toolkit/mozapps/update/nsIUpdateService.idl#189-191
         switch (state) {
+          case "pending-elevate":
+            Cc["@mozilla.org/updates/update-manager;1"]
```
michaelvanstraten, line 297: This is wrong, you are now calling the elevationOptedIn twice

lissyx: Have you seen any problem from doing so ?

lissyx: I'm asking because we need to handle it here for the case the update happens after felt startup, and I don't see anything mentioning we cannot do it twice.

lissyx: Like if it's breaking when we do it twice I need to verify we properly trigger the call here when we do the check in felt

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -259,37 +261,78 @@ export const Updates = {
...
+        switch (state) {
+          case "elevation-attempt-failed":
+            Services.felt?.sendUpdateReady();
+            break;
+          case "download-attempt-failed": // this.showUpdateAvailableNotification(update, false);
+          case "check-attempts-exceeded":
+          case "unknown":
+          case "bad-perms":
+          case "download-attempts-exceeded":
+          case "elevation-attempts-exceeded": // this.showManualUpdateNotification(update, false);
```
lissyx, line 323: Those are still under UX discussion, I suggest we address that after

toolkit/mozapps/update/UpdateService.sys.mjs
```diff
@@ -3272,6 +3272,13 @@ export class UpdateService {
               "pending-elevate, but this is a silent startup, so the " +
               "elevation window has been suppressed."
           );
+        } else if (Services.felt?.isFeltUI() || Services.felt?.isFeltBrowser()) {
+          LOG(
+            "UpdateService:#asyncInit - status is " +
+              "pending-elevate, but this is a Felt UI or browser, so the " +
+              "elevation window has been suppressed. Updates.sys.mjs should " +
+              "perform the elevationOptedIn() call."
+          );
```
lissyx: @gcp this should avoid showing the elevation window

toolkit/mozapps/update/UpdateService.sys.mjs
```diff
@@ -3272,6 +3272,16 @@ export class UpdateService {
               "pending-elevate, but this is a silent startup, so the " +
               "elevation window has been suppressed."
           );
+        } else if (
+          Services.felt?.isFeltUI() ||
+          Services.felt?.isFeltBrowser()
```
Mossop, line 3277: Is this the same as AppConstants.MOZ_ENTERPRISE?

Mossop (approved):

toolkit/mozapps/update/UpdateService.sys.mjs
lissyx: Nope, it checks we are a browser and not just an enterprise build

Mossop: If we're not felt UI and not felt Browser then what are we?

Mossop (approved):

toolkit/mozapps/update/UpdateService.sys.mjs
lissyx: That is the `else` below. The new `if` here is that we want to make sure we dont show the `updateElevation.xhtml` dialog in neither FELT or Browser. so it would be for a "normal" firefox handling

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: https://bugzilla.mozilla.org/show_bug.cgi?id=2024713

michaelvanstraten (approved):
Tested it, works

PR #550, opened 2026-03-16
Bug 2017135: close FF and prompt login when token fails refresh (401)

victhorlopez:
This patch implements promptForReauthentication() to force-quit the Firefox browser and trigger a sign-out, so FELT can take over and show the auth window again.

### DIFF
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -404,7 +404,7 @@ export const ConsoleClient = {
     }
 
     if ((res.status === 403 || res.status === 401) && !_didRefresh) {
-      await this._refreshSession();
+      await this.handleSessionRefresh();
       return this._fetch(path, method, { _didRefresh: true, jsonBody });
     }
 
@@ -447,15 +447,8 @@ export const ConsoleClient = {
   async getAccessToken() {
     let accessToken = Services.felt.getAccessTokenIfValid();
     if (!accessToken) {
-      await this._refreshSession();
+      await this.handleSessionRefresh();
       accessToken = Services.felt.getAccessTokenIfValid();
-    }
-    if (!accessToken) {
-      // We're not handling reauthentication just yet.
-      throw new InvalidAuthError(
-        "Unhandled reauthentication",
-        "UNHANDLED_REAUTHENTICATION"
-      );
     }
     return accessToken;
   },
@@ -486,10 +479,28 @@ export const ConsoleClient = {
   },
 
   /**
+   * Wraps _refreshSession(), handling ReauthRequiredError by notifying Felt and re-throwing.
+   *
+   * @throws {ReauthRequiredError|InvalidAuthError}
+   * @returns {Promise<void>}
+   */
+  async handleSessionRefresh() {
+    try {
+      await this._refreshSession();
+    } catch (e) {
+      lazy.log.error("handleSessionRefresh: session refresh failed", e);
+      if (e instanceof ReauthRequiredError) {
+        this.consoleForcedLogout();
+      }
+      throw e;
+    }
+  },
+
+  /**
    * Refreshes the session using a refresh token.
    * Serializes concurrent refreshes via an internal promise.
    *
-   * @throws {InvalidAuthError} If unable to refresh session
+   * @throws {ReauthRequiredError|InvalidAuthError} If unable to refresh session
    * @returns {Promise<void>}
    */
   async _refreshSession() {
@@ -510,13 +521,10 @@ export const ConsoleClient = {
     this._refreshPromise = (async () => {
       let refreshToken = Services.felt.getRefreshToken();
       if (!refreshToken) {
-        const e = new ReauthRequiredError(
+        throw new ReauthRequiredError(
           "No refresh token available",
           "MISSING_REFRESH_TOKEN"
         );
-        console.error(e);
-        this.promptForReauthentication();
-        return;
       }
       let res;
       try {
@@ -541,14 +549,11 @@ export const ConsoleClient = {
       }
 
       if (res.status === 401 || res.status === 403) {
-        const e = new ReauthRequiredError(
+        throw new ReauthRequiredError(
           "Invalid refresh token",
           "INVALID_REFRESH_TOKEN",
           { status: res.status }
         );
-        console.error(e);
-        this.promptForReauthentication();
-        return;
       }
 
       // TODO: Handle network issues, offline support, etc.
@@ -629,12 +634,46 @@ export const ConsoleClient = {
   },
 
   /**
-   * If unable to refresh the session, prompt for user reauthentication
-   * to obtain a valid set of access and refresh token.
+   * Shared logout implementation. Guards against double calls, signals FELT to
+   * take over as a background process, then invokes the provided logout
+   * function to notify FELT of the logout type. Does not quit the browser;
+   * callers are responsible for that via quit().
+   *
+   * @param {Function} performXPCOMLogout - Invokes the appropriate XPCOM logout call.
    */
-  promptForReauthentication() {
-    this.clearTokenData();
-    // TODO: Handle Re-authentication
+  _logout(performXPCOMLogout) {
+    // Only the FELT-managed browser should trigger a quit and logout.
+    // _isLogoutInProgress guards against double calls from concurrent requests.
+    if (!Services.felt.isFeltBrowser() || this._isLogoutInProgress) {
+      return;
+    }
+    this._isLogoutInProgress = true;
+    // Make sure we signal early enough to the system that FELT should take
+    // over. Relevant at least for macOS dock icon. Not having this would
+    // at least intermittently result in missing dock icon for FELT after
+    // signout.
+    Services.felt.makeBackgroundProcess(true);
+    // Notify FELT of the logout type so it can handle shutdown appropriately.
+    performXPCOMLogout();
+  },
+
+  quit() {
+    if (!this._isLogoutInProgress) {
+      return;
+    }
+    Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
+  },
+
+  normalLogout({ withQuit = true } = {}) {
+    this._logout(() => Services.felt.performNormalLogout());
+    if (withQuit) {
+      this.quit();
+    }
+  },
+
+  consoleForcedLogout() {
+    this._logout(() => Services.felt.performConsoleForcedLogout());
+    this.quit();
   },
 
   /**
@@ -664,18 +703,8 @@ export const ConsoleClient = {
     const res = await this._post(this._paths.SIGNOUT);
     // Server should maybe return better JSON?
     if (res == null) {
-      // After successful server-side logout clear local state and notify FELT.
-      this.clearTokenData();
-
-      // Make sure we signal early enough to the system that FELT should take
-      // over. Relevant at least for macOS dock icon. Not having this would
-      // at least intermittently result in missing dock icon for FELT after
-      // signout.
-      Services.felt.makeBackgroundProcess(true);
-
-      // Notify FELT that we are logging out so the shutdown is a normal one
-      // that should not be followed by restarting the process.
-      Services.felt.performSignout();
+      // Quit is called from outside (e.g. by the caller after signout).
+      this.normalLogout({ withQuit: false });
       return;
     }
 
@@ -692,6 +721,11 @@ export const ConsoleClient = {
       lazy.AsyncShutdown.appShutdownConfirmed.addBlocker(
         `ConsoleClient: Sending back tokens to felt on shutdown`,
         () => {
+          // Do not send tokens back to Felt on logout — the session is
+          // being intentionally terminated and should not be restored.
+          if (this._isLogoutInProgress) {
+            return;
+          }
           try {
             Services.felt.sendTokens();
           } catch (ex) {
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -8,6 +8,8 @@
 
 const lazy = {};
 
+// Maps logout types (from nsIFelt) to the CSS class applied to the FELT window
+// to display the appropriate message to the user after logout.
 ChromeUtils.defineLazyGetter(
   lazy,
   "logoutTypeMessageClass",
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -8,6 +8,18 @@
 
 const lazy = {};
 
+ChromeUtils.defineLazyGetter(
+  lazy,
+  "logoutTypeMessageClass",
+  () =>
+    new Map([
+      [
+        Services.felt.logoutTypeConsoleForcedLogout,
+        "felt-browser-info-console-forced-logout",
+      ],
+    ])
+);
+
 ChromeUtils.defineESModuleGetters(lazy, {
   UpdateListener: "resource://gre/modules/UpdateListener.sys.mjs",
   FELT_OPEN_WINDOW_DISPOSITION: "resource:///modules/FeltURLHandler.sys.mjs",
@@ -267,7 +279,9 @@ this.felt = class extends ExtensionAPI {
       case "FeltParent:FirefoxLogoutExit": {
         const success = Services.felt.makeBackgroundProcess(false);
         console.debug(`FeltExtension: makeBackgroundProcess? ${success}`);
-        this.showWindow();
+        this.showWindow(
+          lazy.logoutTypeMessageClass.get(message.data.logoutType) ?? ""
+        );
         break;
       }
 
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -239,7 +239,7 @@ export class FeltProcessParent extends JSProcessActorParent {
           }
 
           case "felt-firefox-logout":
-            gFeltProcessParentInstance.logoutFirefox();
+            gFeltProcessParentInstance.logoutFirefox(aData);
             break;
 
           case "felt-firefox-tokens": {
@@ -592,15 +592,17 @@ export class FeltProcessParent extends JSProcessActorParent {
   }
 
   /**
-   * Perform all the logout operations on FELT side
+   * Perform all the logout operations on FELT side.
+   *
+   * @param {string} logoutType - One of the logout type payload strings from the IPC message.
    */
-  logoutFirefox() {
+  logoutFirefox(logoutType) {
     if (!Services.felt.isFeltUI()) {
       throw new Error("Logout handling should only happen on FELT side.");
     }
 
     console.debug(
-      `FeltExtension: Logout, waiting on ${gFeltProcessParentInstance.proc.pid}`
+      `FeltExtension: Logout (${logoutType}), waiting on ${gFeltProcessParentInstance.proc.pid}`
     );
     gFeltProcessParentInstance.logoutReported = true;
     lazy.ConsoleClient.clearTokenData();
@@ -612,7 +614,9 @@ export class FeltProcessParent extends JSProcessActorParent {
     }
 
     gFeltProcessParentInstance.proc.exitPromise.then(_ => {
-      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLogoutExit", {});
+      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLogoutExit", {
+        logoutType,
+      });
     });
   }
 
--- a/browser/extensions/felt/rust/nsIFelt.idl
+++ b/browser/extensions/felt/rust/nsIFelt.idl
@@ -76,5 +76,12 @@ interface nsIFelt : nsISupports {
   ACString getConsoleUrl();
 
   [must_use]
-  void performSignout();
+  void performNormalLogout();
+
+  [must_use]
+  void performConsoleForcedLogout();
+
+  readonly attribute ACString logoutTypeNormal;
+  readonly attribute ACString logoutTypeConsoleForcedLogout;
+  readonly attribute ACString logoutTypeUnknown;
 };
--- a/browser/extensions/felt/rust/src/components.rs
+++ b/browser/extensions/felt/rust/src/components.rs
@@ -303,34 +303,30 @@ impl FeltXPCOM {
         NS_OK
     }
 
-    fn PerformNormalLogout(&self) -> nserror::nsresult {
-        trace!("FeltXPCOM::PerformNormalLogout");
+    fn perform_logout(&self, logout_type: LogoutType) -> nserror::nsresult {
+        trace!("FeltXPCOM::perform_logout");
         let guard = crate::FELT_CLIENT.lock().expect("Could not get lock");
         match &*guard {
             Some(client) => {
-                trace!("firefox_felt_send_extension_ready(): sending message");
-                client.notify_signout(LogoutType::Normal);
+                trace!("FeltXPCOM::perform_logout(): sending message");
+                client.notify_signout(logout_type);
+                NS_OK
             }
             None => {
-                trace!("firefox_felt_send_extension_ready(): missing client");
+                trace!("FeltXPCOM::perform_logout(): missing client");
+                NS_ERROR_FAILURE
             }
         }
-        NS_OK
+    }
+
+    fn PerformNormalLogout(&self) -> nserror::nsresult {
+        trace!("FeltXPCOM::PerformNormalLogout");
+        self.perform_logout(LogoutType::Normal)
     }
 
     fn PerformConsoleForcedLogout(&self) -> nserror::nsresult {
         trace!("FeltXPCOM::PerformConsoleForcedLogout");
-        let guard = crate::FELT_CLIENT.lock().expect("Could not get lock");
-        match &*guard {
-            Some(client) => {
-                trace!("firefox_felt_send_extension_ready(): sending message");
-                client.notify_signout(LogoutType::ConsoleForcedLogout);
-            }
-            None => {
-                trace!("firefox_felt_send_extension_ready(): missing client");
-            }
-        }
-        NS_OK
+        self.perform_logout(LogoutType::ConsoleForcedLogout)
     }
 
     fn IpcChannel(&self) -> nserror::nsresult {
--- a/browser/extensions/felt/rust/src/components.rs
+++ b/browser/extensions/felt/rust/src/components.rs
@@ -17,7 +17,7 @@ use xpcom::RefPtr;
 
 use log::{error, trace};
 
-use crate::message::{FeltMessage, FELT_IPC_VERSION};
+use crate::message::{FeltMessage, LogoutType, FELT_IPC_VERSION};
 #[cfg(target_os = "linux")]
 use crate::utils;
 use crate::utils::{Tokens, CONSOLE_URL, TOKENS, TOKEN_EXPIRY_SKEW};
@@ -288,14 +288,43 @@ impl FeltXPCOM {
         }
     }
 
-    // Firefox to FELT to notify of logout
-    fn PerformSignout(&self) -> nserror::nsresult {
-        trace!("FeltXPCOM::PerformSignout");
+    fn GetLogoutTypeNormal(&self, value: *mut nsACString) -> nserror::nsresult {
+        unsafe { (*value).assign(LogoutType::Normal.as_str()) };
+        NS_OK
+    }
+
+    fn GetLogoutTypeConsoleForcedLogout(&self, value: *mut nsACString) -> nserror::nsresult {
+        unsafe { (*value).assign(LogoutType::ConsoleForcedLogout.as_str()) };
+        NS_OK
+    }
+
+    fn GetLogoutTypeUnknown(&self, value: *mut nsACString) -> nserror::nsresult {
+        unsafe { (*value).assign(LogoutType::Unknown.as_str()) };
+        NS_OK
+    }
+
+    fn PerformNormalLogout(&self) -> nserror::nsresult {
+        trace!("FeltXPCOM::PerformNormalLogout");
         let guard = crate::FELT_CLIENT.lock().expect("Could not get lock");
         match &*guard {
             Some(client) => {
                 trace!("firefox_felt_send_extension_ready(): sending message");
-                client.notify_signout();
+                client.notify_signout(LogoutType::Normal);
+            }
+            None => {
+                trace!("firefox_felt_send_extension_ready(): missing client");
+            }
+        }
+        NS_OK
+    }
+
+    fn PerformConsoleForcedLogout(&self) -> nserror::nsresult {
+        trace!("FeltXPCOM::PerformConsoleForcedLogout");
+        let guard = crate::FELT_CLIENT.lock().expect("Could not get lock");
+        match &*guard {
+            Some(client) => {
+                trace!("firefox_felt_send_extension_ready(): sending message");
+                client.notify_signout(LogoutType::ConsoleForcedLogout);
             }
             None => {
                 trace!("firefox_felt_send_extension_ready(): missing client");
@@ -387,9 +416,9 @@ impl FeltXPCOM {
                                 trace!("FeltServerThread::felt_server::ipc_loop(): ExtensionReady");
                                 crate::utils::notify_observers("felt-extension-ready".to_string());
                             },
-                            Ok(FeltMessage::LogoutShutdown) => {
-                                trace!("FeltServerThread::felt_server::ipc_loop(): Shutdown for logout");
-                                crate::utils::notify_observers("felt-firefox-logout".to_string());
+                            Ok(FeltMessage::Logout(logout_type)) => {
+                                trace!("FeltServerThread::felt_server::ipc_loop(): Logout {:?}", logout_type);
+                                crate::utils::notify_observers_with_payload("felt-firefox-logout".to_string(), Some(logout_type.as_str().to_string()));
                             },
                             Ok(FeltMessage::Tokens((access_token, refresh_token, expires_at))) => {
                                 trace!("FeltServerThread::felt_server::ipc_loop(): Update tokens from browser");
--- a/browser/extensions/felt/rust/src/message.rs
+++ b/browser/extensions/felt/rust/src/message.rs
@@ -45,6 +45,33 @@ impl nsICookieWrapper {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
+pub enum LogoutType {
+    Normal = 0,
+    ConsoleForcedLogout = 1,
+    Unknown = 255,
+}
+
+impl LogoutType {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            LogoutType::Normal => "normal",
+            LogoutType::ConsoleForcedLogout => "console-forced-logout",
+            LogoutType::Unknown => "unknown",
+        }
+    }
+}
+
+impl From<i32> for LogoutType {
+    fn from(value: i32) -> Self {
+        match value {
+            0 => LogoutType::Normal,
+            1 => LogoutType::ConsoleForcedLogout,
+            _ => LogoutType::Unknown,
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 pub enum FeltMessage {
     VersionProbe(u32),
     VersionValidated(bool),
@@ -59,7 +86,7 @@ pub enum FeltMessage {
     OpenURL((String, i32, Option<FocusHint>)),
     RestartForced,
     Restarting,
-    LogoutShutdown,
+    Logout(LogoutType),
     Exiting,
     UpdateReady,
 }
@@ -70,4 +97,4 @@ pub enum FocusHint {
     Timestamp(u32),
 }
 
-pub const FELT_IPC_VERSION: u32 = 7;
+pub const FELT_IPC_VERSION: u32 = 8;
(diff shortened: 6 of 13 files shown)


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -611,7 +611,13 @@ export const ConsoleClient = {
...
   promptForReauthentication() {
     this.clearTokenData();
-    // TODO: Handle Re-authentication
+    // Only the FELT-managed browser should trigger a forced quit and signout.
+    if (!Services.felt.isFeltBrowser()) {
+      return;
+    }
+    Services.felt.makeBackgroundProcess(true);
+    Services.felt.performSignout();
+    Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
```
lissyx: @gcp is this going to hold if refresh is done on felt side rather than browser side?

gcp: I don't understand what you're asking.

gcp: Ha, I think you refer to the clean tokens thing that's just above what Github is displaying here. I mean yes, if we force signout, the clean tokens signal needs to be transmitted to Felt. So nothing changes here.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -655,13 +662,41 @@ export const ConsoleClient = {
...
+    this.clearTokenData();
+    // Make sure we signal early enough to the system that FELT should take
+    // over. Relevant at least for macOS dock icon. Not having this would
+    // at least intermittently result in missing dock icon for FELT after
+    // signout.
+    Services.felt.makeBackgroundProcess(true);
+    // Notify FELT that we are logging out so the shutdown is a normal one
+    // that should not be followed by restarting the process.
+    Services.felt.performSignout(logoutType);
+    Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
```
gcp, line 664: You need to test this with a website that pops up a modal dialog in the `onClose/onBeforeUnload` (I forgot what it's called) handler. It **must not** block even in that case.

browser/extensions/felt/rust/src/message.rs
```diff
@@ -59,7 +78,7 @@ pub enum FeltMessage {
     OpenURL((String, i32)),
     RestartForced,
     Restarting,
-    LogoutShutdown,
+    Logout(LogoutType),
```
gcp, line 89: Can probably combine these messages.

There is a version number for the IPC that should be incremented.

lissyx: So I hesitated on that and I'm the one who suggested maybe it was cleaner to introduce a dedicated one, maybe it was not a good idea

testing/enterprise/test_felt_browser_token_refresh.py
gcp: I assume this will fail and the failure can be asserted?

gcp (changes requested):

browser/extensions/felt/rust/src/message.rs
gcp: It's just that this introduces so much extra scaffolding that's not really needed. And if we have `Logout(LogoutType)`, then `Logout(Shutdown) `seems logical.

lissyx: I think I suggested also to just rename and add the information, maybe I should have insisted more

testing/enterprise/test_felt_browser_token_refresh.py
lissyx: i hope :)

victhorlopez: can you elaborate? line 37 has this, or you mean something else

victhorlopez: okay... I will see if I can fix it, you want to call
self.assert_user_signed_out(env=Environment.FIREFOX)

Because it looks like the classic egg chicken (i.e calling  self.get_logged_in_user_info(env=Environment.FIREFOX) is meant to close firefox)

lissyx: I think we should already be getting the information out of `get_logged_in_user_info` and can assert there, because all `assert_user_signed_out` does on line 37 is the same ? So you could directly call `self.assert_user_signed_out(Environment.FIREFOX)` right now.

lissyx: Maybe an empty string would be better?

lissyx: You will also likely want to check API end points in the mock console to ensure there is proper auth checking done

lissyx: @victhorlopez yes I think you are right about the chicken and egg problem, not only we are forcing signout but also closing the browser, there is a risk of hitting `NoSuchWindowException`.

Maybe you could wrap the `self.assert_user_signed_out(Environment.FIREFOX)` into a try/except block, explicitely catching `NoSuchWindowException` in which you could assert that the browser was closed, and that the `invalidated-by-server` token was sent back (but we have `test_felt_browser_exit_tokens.py` that should already verify that kind of behavior), and if you dont hit the assertion then you are OK to assert the value in the browser

victhorlopez: This should be already [in here](https://github.com/mozilla/enterprise-firefox/blob/enterprise-main/testing/enterprise/felt_tests.py#L353). or  you mean some extra asserts?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: I will have a look into this, see if I can figure it out

edit: added test for this

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: I had a look, I will check for "console-forced-logout" which is what the felt should receive on the icp message

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -628,13 +662,42 @@ export const ConsoleClient = {
...
    * to obtain a valid set of access and refresh token.
+   *
+   * @param {number} logoutType - One of the Ci.nsIFelt.LOGOUT_TYPE_* constants.
    */
-  promptForReauthentication() {
-    this.clearTokenData();
-    // TODO: Handle Re-authentication
+  promptForReauthentication(logoutType = Ci.nsIFelt.LOGOUT_TYPE_UNKNOWN) {
+    this.logout(logoutType);
+    Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
```
victhorlopez, line 664: @1rneh I move this in here, I know we are also [calling it](https://searchfox.org/enterprise-main/source/browser/components/enterprise/EnterpriseHandler.sys.mjs#204) from outside. 

If the logic makes sense we could move that try catch finally inside signout? Or do we have a case where we want to signout and not close the browser, and have this line in logout
`Services.startup.quit(Ci.nsIAppStartup.eForceQuit);`

testing/enterprise/test_felt_browser_token_refresh.py
lissyx: Just making sure the current state of checking the auth on the mock server is correct, there were cases where we forgot and/or maybe where the way we check for auth may not be correct (e.g. we `return` on auth error but the caller does not have the information so it may proceed with the rest of the execution): https://searchfox.org/enterprise-main/rev/a038f49228d707c6675ef20ce640034a64307d2e/testing/enterprise/felt_tests.py#112,117,121,175

I'm not sure this last line does what we want know that we enforce auth checking

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -524,7 +551,11 @@ export const ConsoleClient = {
           "MISSING_REFRESH_TOKEN"
         );
         console.error(e);
-        this.promptForReauthentication();
+        this.promptForReauthentication(
+          Ci.nsIFelt.LOGOUT_TYPE_NO_REFRESH_TOKEN_AVAILABLE
+        );
+        // Return instead of throwing so getAccessToken() can fall through to
+        // InvalidAuthError, which signals the signed-out state to callers.
```
1rneh: What do you think about altering this a bit so that we don't call `this.promptForReauthentication();` directly in `_refreshSession` multiple times? Maybe we can rename `promptForReauthentication()` to `notifyFeltToReauthenticate()`.

We could wrap `this._refreshSession` in another method such as `handleSessionRefresh`? Inside we call `this._refreshSession`, which throws an `ReauthRequiredError` and adds the logout type as `cause` when ever the user needs to reauthenticate. In `handleSessionRefresh`we'd  catch that, extract the `cause` and call `notifyFeltToReauthenticate`  in which we can move the logic from `this.logout`.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: > If the logic makes sense we could move that try catch finally inside signout?

Actually, I would like to move any logic that is not strictly tied to console communication or token handling out of `ConsoleClient`. That includes things like `Services.startup.quit(Ci.nsIAppStartup.eForceQuit)`, `Services.felt.makeBackgroundProcess`, and `Services.felt.performSignout`, and possibly have `EnterpriseHandler` own that instead. But I do not think this needs to be part of the current change. So I do not think we should move it now.

> Or do we have a case where we want to signout and not close the browser, and have this line in logout
Services.startup.quit(Ci.nsIAppStartup.eForceQuit);

Signing out always goes along with quitting the browser, and vice versa.

1rneh (changes requested):

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -0,0 +1,44 @@
...
+    def setup(self):
+        super().setup()
+        self._logout_checker = _LogoutChecker(self)
+
+    def test_transparent_token_refresh(self):
+        super().run_felt_base()
+        self.connect_child_browser()
+        self.assert_user_signed_in(env=Environment.FIREFOX)
+
+        self.policy_access_token.value = ""
```
1rneh, line 61: Is the expectation that this won't interfere with out tests depending on `self.policy_access_token.value` because the `assert_user_signed_in` will trigger a session refresh which will re-populate  `self.policy_access_token.value`?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: I will give it a try, I did not want to do big changes on the underlying logic to avoid breaking things.

I added [here](https://github.com/mozilla/enterprise-firefox/pull/550/changes#diff-77ab0da61c79c129bc608e37594459cfab3b8b473b919ae19813992a292df501R178) new exceptions

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: Makes sense

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -0,0 +1,44 @@
...
+
+        self.assert_user_signed_in(env=Environment.FIREFOX)
+
+    def test_forced_signout_on_401(self):
+        super().run_felt_base()
+        self.connect_child_browser()
+        self.assert_user_signed_in(env=Environment.FIREFOX)
+
+        self.policy_access_token.value = ""
+        self.policy_refresh_token.value = ""
```
1rneh, line 38: This clears the shared server-side token state, hence other tests can become order-dependent unless you restore or replace them at the end of this test?

1rneh (changes requested):
I have two more remarks about the tests

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: So what I am trying to test here is that triggering an authed endpoint, will succesfully use the refresh token when the current token was invalidated (basically that the normal refresh path is working without closing firefox etc)

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: I just checked and setUp from base tests actually gets called per test method (not class).

lissyx: Yes that's a bit tricky but you can assume `def test_...` means a new console has been spawned. That being said, I'm always in favor of restoring context properly

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -519,13 +537,11 @@ export const ConsoleClient = {
     this._refreshPromise = (async () => {
       let refreshToken = Services.felt.getRefreshToken();
       if (!refreshToken) {
-        const e = new ReauthRequiredError(
+        throw new ReauthRequiredError(
           "No refresh token available",
-          "MISSING_REFRESH_TOKEN"
+          "MISSING_REFRESH_TOKEN",
+          { logoutType: Ci.nsIFelt.LOGOUT_TYPE_NO_REFRESH_TOKEN_AVAILABLE }
```
1rneh: Could you pass `logoutType` inside the `cause` object?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -629,12 +645,43 @@ export const ConsoleClient = {
...
+   * Notifies Felt that reauthentication is required and quits the browser.
+   *
+   * @param {number} logoutType - One of the Ci.nsIFelt.LOGOUT_TYPE_* constants.
+   */
+  notifyFeltToReauthenticate(logoutType = Ci.nsIFelt.LOGOUT_TYPE_UNKNOWN) {
+    this.logout(logoutType);
+    if (!Services.felt.isFeltBrowser() || !this._loggingOut) {
+      return;
+    }
+    Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
```
1rneh, line 664: Sorry, I missed that `notifyFeltToReauthenticate` includes quitting the browser. The name should actually reflect that. Maybe `onQuit` is good enough. Can we combine current code from `logout` and `notifyFeltToReauthenticate` and then add a flag whether to initiate shutdown.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -629,12 +645,43 @@ export const ConsoleClient = {
...
+  logout(logoutType) {
+    // After logout clear local state and notify FELT.
     this.clearTokenData();
-    // TODO: Handle Re-authentication
+    // Only the FELT-managed browser should trigger a quit and logout.
+    // _loggingOut guards against double calls from concurrent requests.
+    if (!Services.felt.isFeltBrowser() || this._loggingOut) {
+      return;
+    }
+    this._loggingOut = true;
```
1rneh: Can we call this something like `_isLogoutInProgress`

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -629,12 +645,43 @@ export const ConsoleClient = {
...
+   *
+   * This does not perform a signout against the console server, nor does it
+   * quit the browser. Callers are responsible for quitting if needed.
+   *
+   * @param {number} logoutType - One of the Ci.nsIFelt.LOGOUT_TYPE_* constants.
    */
-  promptForReauthentication() {
+  logout(logoutType) {
+    // After logout clear local state and notify FELT.
     this.clearTokenData();
```
1rneh: I just thought that clearing the token on signout might introduce race conditions if Firefox still needs the access token to send telemetry. It's also only stored in memory, so we won't need to clear it on Firefox side. We should clear it on Felt side when notified of a logout.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Can we remove this and let consumers of `getAccessToken` run into an `401 Unauthorized` as we already tried to refresh the session a few lines before.

```suggestion
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -456,7 +457,7 @@ export const ConsoleClient = {
   async getAccessToken() {
     let accessToken = Services.felt.getAccessTokenIfValid();
     if (!accessToken) {
-      await this._refreshSession();
+      await this.handleSessionRefresh();
```
1rneh, line 450: Can we wrap this in a try catch and log the error?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -608,15 +608,17 @@ export class FeltProcessParent extends JSProcessActorParent {
...
     if (!Services.felt.isFeltUI()) {
       throw new Error("Logout handling should only happen on FELT side.");
     }
 
     console.debug(
-      `FeltExtension: Logout, waiting on ${gFeltProcessParentInstance.proc.pid}`
+      `FeltExtension: Logout (${logoutType}), waiting on ${gFeltProcessParentInstance.proc.pid}`
     );
     gFeltProcessParentInstance.logoutReported = true;
     lazy.ConsoleClient.clearTokenData();
```
1rneh, line 608: Oh, I see we already clear tokens on Firefox side (see comment about clearing tokens on Firefox side on shutdown).

1rneh (changes requested):
Follow-up points and additional requests:

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -628,7 +630,9 @@ export class FeltProcessParent extends JSProcessActorParent {
     }
 
     gFeltProcessParentInstance.proc.exitPromise.then(_ => {
-      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLogoutExit", {});
+      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLogoutExit", {
+        logoutType,
```
lissyx, line 618: `logoutType` is a string, so you likely miss e.g. `"logoutType": logoutType` to report it in `message.data` ?

browser/extensions/felt/rust/src/components.rs
```diff
@@ -289,13 +289,13 @@ impl FeltXPCOM {
...
     // Firefox to FELT to notify of logout
-    fn PerformSignout(&self) -> nserror::nsresult {
+    fn PerformSignout(&self, logout_type: i32) -> nserror::nsresult {
         trace!("FeltXPCOM::PerformSignout");
         let guard = crate::FELT_CLIENT.lock().expect("Could not get lock");
         match &*guard {
             Some(client) => {
                 trace!("firefox_felt_send_extension_ready(): sending message");
-                client.notify_signout();
+                client.notify_signout(logout_type);
```
lissyx, line 312: I'd prefer we directly move from `i32` to the real message type right now

lissyx, line 312:
```suggestion
                client.notify_signout(LogoutType::from(logout_type));
```

browser/extensions/felt/rust/src/components.rs
```diff
@@ -387,9 +387,15 @@ impl FeltXPCOM {
...
                                 crate::utils::notify_observers("felt-extension-ready".to_string());
                             },
-                            Ok(FeltMessage::LogoutShutdown) => {
-                                trace!("FeltServerThread::felt_server::ipc_loop(): Shutdown for logout");
-                                crate::utils::notify_observers("felt-firefox-logout".to_string());
+                            Ok(FeltMessage::Logout(logout_type)) => {
+                                trace!("FeltServerThread::felt_server::ipc_loop(): Logout {:?}", logout_type);
+                                let payload = match logout_type {
+                                    LogoutType::ConsoleForcedLogout => "console-forced-logout",
+                                    LogoutType::NoRefreshTokenAvailable => "no-refresh-token-available",
```
lissyx: I dont see those strings being used anywhere?

browser/extensions/felt/rust/nsIFelt.idl
```diff
@@ -8,6 +8,11 @@
 
 [scriptable, uuid(373195cb-bbb9-4e01-b276-b799a18ca355)]
 interface nsIFelt : nsISupports {
+  const long LOGOUT_TYPE_NORMAL = 0;
+  const long LOGOUT_TYPE_CONSOLE_FORCED_LOGOUT = 1;
+  const long LOGOUT_TYPE_NO_REFRESH_TOKEN_AVAILABLE = 2;
+  const long LOGOUT_TYPE_UNKNOWN = 255;
```
lissyx: We end up duplicating the information of the logout type.

browser/extensions/felt/rust/src/components.rs
```diff
@@ -289,13 +289,13 @@ impl FeltXPCOM {
     }
 
     // Firefox to FELT to notify of logout
-    fn PerformSignout(&self) -> nserror::nsresult {
+    fn PerformSignout(&self, logout_type: i32) -> nserror::nsresult {
```
lissyx: Since we end up exposing the internal logout type to XPCOM if we do that, I suggest we instead change the interface and introduce e.g. `PerformNormalLogout()`, `PerformConsoleForcedLogout()` etc..

browser/extensions/felt/rust/nsIFelt.idl
lissyx: Rather than that I'd prefer we augment the surface at the methods level, like suggested above

lissyx: And rather maybe we need to expose strings like `"console-forced-logout"` and others that we place in `message.data`

lissyx: (and we could use them in rust from the IDL interface as well?)

testing/enterprise/felt_tests.py
```diff
@@ -464,6 +464,66 @@ def serve(
...
+    action that triggers the logout in a with block.
+    """
+
+    def __init__(self, test):
+        self._test = test
+        self._expected_type = None
+        test._driver.set_context("chrome")
+        test._driver.execute_script(
+            """
+            Services.prefs.clearUserPref("_test.logout_type");
```
lissyx: There is already `enterprise.felt_tests` usage, please stick to it? Since it's browser maybe `enterprise.browser_tests` ?

testing/enterprise/felt_tests.py
```diff
@@ -464,6 +464,66 @@ def serve(
...
+        if exc_type is not None:
+            return False
+
+        handles = self._test._wait.until(
+            lambda mn: self._test._driver.chrome_window_handles
+        )
+        # switch_to_window resets Marionette's internal curBrowser pointer to the new
+        # window; without it, execute_script would throw "browsing context has been
+        # discarded" because it still references the old closed window.
+        self._test._driver.switch_to_window(handles[0])
```
lissyx, line 566: There may be tests where we have strong assumption on what is the current window, please keep track and restore the correct state here

testing/enterprise/felt_tests.py
```diff
@@ -464,6 +464,66 @@ def serve(
...
+    Must be instantiated while the Felt window is open (i.e. in setup()), since it
+    registers a "felt-firefox-logout" observer via execute_script at construction time.
+    Use assert_browser_logouts_with() to set the expected logout type, then wrap the
+    action that triggers the logout in a with block.
+    """
+
+    def __init__(self, test):
+        self._test = test
+        self._expected_type = None
+        test._driver.set_context("chrome")
```
lissyx: `self._driver` is the FELT driver, not the child browser which is `self._child_driver`

testing/enterprise/felt_tests.py
```diff
@@ -464,6 +464,66 @@ def serve(
     )
 
 
+class _LogoutChecker:
+    """Context manager that asserts a Firefox child browser logout of a specific type occurred.
+
+    Must be instantiated while the Felt window is open (i.e. in setup()), since it
+    registers a "felt-firefox-logout" observer via execute_script at construction time.
+    Use assert_browser_logouts_with() to set the expected logout type, then wrap the
+    action that triggers the logout in a with block.
```
lissyx: It's not clear when reading that, that what is asserted is the **Felt** side, it should be made more clear IMHO. Maybe rename the class `FeltLogoutChecker` ?

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -0,0 +1,84 @@
...
+        self._setup_child_browser_with_unload_tab()
+
+        old_access_token = self.policy_access_token.value
+        old_refresh_token = self.policy_refresh_token.value
+        self.policy_access_token.value = ""
+        self.policy_refresh_token.value = ""
+
+        # Trigger an auth request with invalid tokens, expecting a forced logout.
+        with self._logout_checker.assert_browser_logouts_with("console-forced-logout"):
+            self.get_logged_in_user_info(env=Environment.FIREFOX)
```
lissyx, line 44: It's even less obvious from here, having `self._felt_logout_checker.assert...` would improve imho

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assert_child_browser_closed()
+
+        self.await_felt_auth_window()
+        self.force_window()
```
lissyx, line 49: This may subtely break because you use `switch_to` in the context manager without restoring the context

browser/extensions/felt/rust/src/components.rs
victhorlopez: it's for the UI (which is not in this PR) https://github.com/mozilla/enterprise-firefox/pull/576/changes#diff-c47962c16d59bb69cf10f1359cd483c96e3dc8bf5b0862068946358a916af882R11

browser/extensions/felt/rust/nsIFelt.idl
victhorlopez: it was to have it "typed" on the JS side, will have a look at your suggestion

victhorlopez: We could yes, I put ints originally as I thought it was more efficient

testing/enterprise/felt_tests.py
victhorlopez: It's actually  actually on the FELT side, the idea is just an observer on the IPC side (on FELT). That verifies we received the "console-forced-logout" on FELT side after it was shut down

testing/enterprise/felt_tests.py
victhorlopez: See belomw it is supposed to be FELT, my bad to not make it clearer

testing/enterprise/felt_tests.py
victhorlopez: can also leave it outside of the utility, the idea was logout in theory always closes t he child window, but y ou are right putting it here maybe it is too much

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: need to check if I did a mistake here with some refactoring, It was working when I tried some time ago in [here](https://github.com/mozilla/enterprise-firefox/pull/576/changes#diff-c47962c16d59bb69cf10f1359cd483c96e3dc8bf5b0862068946358a916af882R260)

edit: This is just common in js when creating an object, the field then becomes the name of the var i.e {logoutType} expands to {"logoutType": logoutType}

browser/extensions/felt/rust/src/components.rs
victhorlopez: Fine for me, as long as we do not have a lot of reasons for signing out (i.e 100 functions :))

victhorlopez (commented):
I will address, thank you both for the reviews

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -692,6 +721,11 @@ export const ConsoleClient = {
       lazy.AsyncShutdown.appShutdownConfirmed.addBlocker(
         `ConsoleClient: Sending back tokens to felt on shutdown`,
         () => {
+          // Do not send tokens back to Felt on logout — the session is
+          // being intentionally terminated and should not be restored.
+          if (this._isLogoutInProgress) {
```
victhorlopez, line 726: @lissyx can you confirm this is ok? I [removed the clearToken](https://github.com/mozilla/enterprise-firefox/pull/550/changes#diff-5245eb3c71a276182b4c2046dfa9caac4683fb44514e2ab52dac85bbb5b98002L668), and the test_signout_verify test broke, because it was failing in [here](https://searchfox.org/enterprise-main/source/testing/enterprise/test_felt_browser_signout.py#94) (this fixes it)

Is the expectation that we are not calling sendToken() on a normal signout(i.e user signout)? (we were sending empty token in the original code because we were calling clearToken)

browser/extensions/felt/rust/nsIFelt.idl
```diff
@@ -76,5 +76,12 @@ interface nsIFelt : nsISupports {
   ACString getConsoleUrl();
 
   [must_use]
-  void performSignout();
+  void performNormalLogout();
+
+  [must_use]
+  void performConsoleForcedLogout();
+
+  readonly attribute ACString logoutTypeNormal;
```
victhorlopez, line 84: these are the strings coming from rust, used here
https://github.com/mozilla/enterprise-firefox/pull/576/changes#diff-c47962c16d59bb69cf10f1359cd483c96e3dc8bf5b0862068946358a916af882R14

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: the logging was added inside handleSessionRefresh

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: I tried to accommodate for both reviews,  `consoleForcedLogout`, `normalLogout` these are new functions, should I add here in these names somehow also notifyFelt or Shutdown on the name?

`normalLogout({ withQuit = true })`

testing/enterprise/felt_tests.py
victhorlopez: I tried to restore as much context as possible

testing/enterprise/felt_tests.py
```diff
@@ -464,6 +464,84 @@ def serve(
...
+
+    def __init__(self, test):
+        self._test = test
+        self._expected_type = None
+        self._saved_window_handle = None
+
+        test._driver.set_context("chrome")
+        test._driver.execute_script(
+            """
+            Services.prefs.clearUserPref("_test.logout_type");
```
lissyx: can we still keep a consistent namespace?

victhorlopez: done, used enterprise._test.logout_type

testing/enterprise/base_test.py
```diff
@@ -189,6 +190,16 @@ def assert_user_signed_out(self, env):
...
+            "ReauthRequiredError: Invalid refresh token",
+            "InvalidAuthError: Token refresh request blocked in Felt.",
+        ), "Unexpected state after signout"
+
+    def assert_child_browser_closed(self):
+        self._logger.info("Verifying child browser is closed.")
+        try:
+            self._child_driver.get_url()
+            assert False, "Expected child browser to be closed"
+        except (errors.InvalidSessionIdException, OSError):
```
gcp: I think you can get a lot more exceptions here? ` NoSuchWindowException`, random timeouts, etc. Should be a wider catch?

testing/enterprise/felt_tests.py
```diff
@@ -516,6 +516,79 @@ def serve(
...
+
+        try:
+            parent_handles = self._test._driver.chrome_window_handles
+            if self._saved_window_handle in parent_handles:
+                self._test._driver.switch_to_window(self._saved_window_handle)
+        except Exception:
+            # If the parent browser window was closed, there is nothing to restore.
+            pass
+
+        return False
```
gcp, line 585: Restore the prefs you changed here? Can help a lot because all tests after this one would also start failing.

gcp (approved):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: From discussion, the test checks that Felt has no valid tokens after signout. What was likely happening here is that without the `clearToken` here, we send the valid tokens back to Felt, which clears them locally due to signout, but then gets the message with the valid tokens again, which is detected by the test and causes the failure.

If that is is indeed what was happening, this change looks fine.

testing/enterprise/felt_tests.py
victhorlopez: I remove the last "Services.prefs.clearUserPref("enterprise._test.logout_type")",  that will leave the pref in case it needs to be used somehow somewhere else. The other clear is needed as we do not want that previous logouts have an impact on this assertion (and by the end of this context manager, either the assertion is true and we have the right value on the pref, or the assertion fails, and we have the wrong value)

browser/extensions/felt/rust/src/components.rs
```diff
@@ -288,14 +288,43 @@ impl FeltXPCOM {
...
+
+    fn GetLogoutTypeConsoleForcedLogout(&self, value: *mut nsACString) -> nserror::nsresult {
+        unsafe { (*value).assign(LogoutType::ConsoleForcedLogout.as_str()) };
+        NS_OK
+    }
+
+    fn GetLogoutTypeUnknown(&self, value: *mut nsACString) -> nserror::nsresult {
+        unsafe { (*value).assign(LogoutType::Unknown.as_str()) };
+        NS_OK
+    }
```
1rneh, line 304: They are unused.

1rneh (approved):
r+ 

Side note: This will be replaced by #588.

browser/extensions/felt/rust/src/components.rs
victhorlopez: It's for the [readonly attributes ](https://firefox-source-docs.mozilla.org/xpcom/xpidl.html#methods-and-attributes) from [link](https://github.com/mozilla/enterprise-firefox/pull/550/changes#diff-2105140cb7d661ba4c12a79f9965343f40d71ef0c33331f52fd46760599d00d5R84) and then used on the [ui side](https://github.com/mozilla/enterprise-firefox/pull/576/changes#diff-c47962c16d59bb69cf10f1359cd483c96e3dc8bf5b0862068946358a916af882R14)

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -612,7 +614,9 @@ export class FeltProcessParent extends JSProcessActorParent {
...
```
lissyx, line 618: We pass `logoutType` but dont use it?

browser/extensions/felt/rust/src/components.rs
lissyx: @victhorlopez Two PR for the same bug makes it confusing. Can you keep it two commits but in one PR ?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: [here](https://github.com/mozilla/enterprise-firefox/pull/576/changes#diff-c47962c16d59bb69cf10f1359cd483c96e3dc8bf5b0862068946358a916af882R264) in the ui pr

browser/extensions/felt/rust/src/components.rs
victhorlopez: okay

browser/extensions/felt/rust/src/components.rs
```diff
@@ -288,14 +288,43 @@ impl FeltXPCOM {
...
+            Some(client) => {
+                trace!("firefox_felt_send_extension_ready(): sending message");
+                client.notify_signout(LogoutType::Normal);
+            }
+            None => {
+                trace!("firefox_felt_send_extension_ready(): missing client");
+            }
+        }
+        NS_OK
+    }
```
lissyx: Please factorize this and `PerformConsoleForcedLogout`, they do the same except the payload, there is no need to duplicate everything.

Something like:
```suggestion
    fn perform_logout(&self, logout_type: LogoutType) -> nserror::nsresult {
        trace!("FeltXPCOM::perform_logout");
        let guard = crate::FELT_CLIENT.lock().expect("Could not get lock");
        match &*guard {
            Some(client) => {
                trace!("FeltXPCOM::perform_logout(): sending message");
                client.notify_signout(logout_type)
            }
            None => {
                trace!("firefox_felt_send_extension_ready(): missing client");
                NS_ERROR_FAILURE
            }
        }
    }
    
    fn PerformNormalLogout(&self) -> nserror::nsresult {
        trace!("FeltXPCOM::PerformNormalLogout");
        self.perform_logout(LogoutType::Normal)
    }
    
    fn PerformConsoleForcedLogout(&self) -> nserror::nsresult {
        trace!("FeltXPCOM::PerformConsoleForcedLogout");
        self.perform_logout(LogoutType::ConsoleForcedLogout)
    }
```

browser/extensions/felt/api.js
```diff
@@ -8,6 +8,18 @@
...
+ChromeUtils.defineLazyGetter(
+  lazy,
+  "logoutTypeMessageClass",
+  () =>
+    new Map([
+      [
+        Services.felt.logoutTypeConsoleForcedLogout,
+        "felt-browser-info-console-forced-logout",
+      ],
+    ])
```
lissyx, line 22: comment what this is mapping between, especially the second component here

lissyx (approved):
r=me with comments addressed

PR #551, opened 2026-03-16
Bug 2013840  - Attempt to fix SSO callback token extraction failing after cross-process navigation

gcp:
A cross-process navigation during the SSO redirect chain (IdP -> MFA -> callback) can cause the FeltWindowChild JSWindowActor's DOMContentLoaded handler to never fire in the destination content process. When this happens, no tokens are extracted, Firefox is never launched, and the user sees an empty page in Felt.

Add a webProgress listener in the Felt parent process as a fallback. When the callback page finishes loading (STATE_STOP), it explicitly creates the FeltWindow actor pair via getActor() and sends an ExtractTokens query to the child process. This bypasses the flaky event-based actor instantiation. A dedup flag in FeltWindowChild prevents double-processing when both paths succeed.

### DIFF
--- a/browser/extensions/felt/content/FeltWindowChild.sys.mjs
+++ b/browser/extensions/felt/content/FeltWindowChild.sys.mjs
@@ -8,30 +8,51 @@ console.debug(`FeltExtension: FeltWindowChild.sys.mjs`);
  *
  */
 export class FeltWindowChild extends JSWindowActorChild {
+  #tokensSent = false;
+
   actorCreated() {
-    this.actor = ChromeUtils.domProcessChild.getActor("FeltProcess");
+    this.processActor = ChromeUtils.domProcessChild.getActor("FeltProcess");
   }
 
   handleEvent(event) {
     if (event.type !== "DOMContentLoaded") {
-      console.error(`Unexpected event.type=${event.type}`);
       return;
     }
+    this.#extractAndSendTokens(event.target);
+  }
 
-    const tokenData = event.target.querySelector("#token_data");
-    if (tokenData) {
-      console.debug("FeltWindowChild: Extracting token data");
-      const consoleTokenData = JSON.parse(tokenData.textContent);
-      if (
-        consoleTokenData &&
-        "access_token" in consoleTokenData &&
-        consoleTokenData.access_token !== ""
-      ) {
-        console.debug(
-          "FeltWindowChild: Sending token data to ConsoleClient and starting Firefox"
-        );
-        this.actor.sendAsyncMessage("FeltChild:StartFirefox", consoleTokenData);
-      }
+  receiveMessage(message) {
+    if (message.name === "ExtractTokens") {
+      return this.#extractAndSendTokens(this.document);
     }
+    return false;
+  }
+
+  #extractAndSendTokens(doc) {
+    if (this.#tokensSent) {
+      return true;
+    }
+
+    const tokenData = doc.querySelector("#token_data");
+    if (!tokenData) {
+      return false;
+    }
+
+    console.debug("FeltWindowChild: Extracting token data");
+    const consoleTokenData = JSON.parse(tokenData.textContent);
+    if (
+      consoleTokenData &&
+      "access_token" in consoleTokenData &&
+      consoleTokenData.access_token !== ""
+    ) {
+      console.debug("FeltWindowChild: Sending token data to start Firefox");
+      this.#tokensSent = true;
+      this.processActor.sendAsyncMessage(
+        "FeltChild:StartFirefox",
+        consoleTokenData
+      );
+      return true;
+    }
+    return false;
   }
 }
--- a/browser/extensions/felt/content/window.js
+++ b/browser/extensions/felt/content/window.js
@@ -150,8 +150,9 @@ async function connectToConsole(email) {
   // DOMContentLoaded handler to never fire. Monitor the navigation from the
   // parent process and explicitly trigger token extraction when the callback
   // page finishes loading.
-  const callbackPath = lazy.ConsoleClient._paths.SSO_CALLBACK;
-  const callbackHost = lazy.ConsoleClient.consoleBaseURI.host;
+  const callbackPattern = new MatchPattern(
+    lazy.ConsoleClient.ssoCallbackUriMatchPattern
+  );
   const progressListener = {
     QueryInterface: ChromeUtils.generateQI([
       "nsIWebProgressListener",
@@ -167,7 +168,7 @@ async function connectToConsole(email) {
       }
 
       const uri = webProgress.browsingContext?.currentWindowGlobal?.documentURI;
-      if (!uri || uri.host !== callbackHost || uri.filePath !== callbackPath) {
+      if (!uri || !callbackPattern.matches(uri.spec)) {
         return;
       }
 
--- a/browser/extensions/felt/content/window.js
+++ b/browser/extensions/felt/content/window.js
@@ -159,7 +159,7 @@ async function connectToConsole(email) {
       "nsISupportsWeakReference",
     ]),
 
-    onStateChange(webProgress, _request, stateFlags, _status) {
+    onStateChange(webProgress, _request, stateFlags, status) {
       if (
         !(stateFlags & Ci.nsIWebProgressListener.STATE_STOP) ||
         !(stateFlags & Ci.nsIWebProgressListener.STATE_IS_NETWORK)
@@ -174,8 +174,22 @@ async function connectToConsole(email) {
 
       browser.removeProgressListener(progressListener);
 
+      if (!Components.isSuccessCode(status)) {
+        console.error(
+          `FeltExtension: SSO callback page failed to load: 0x${status.toString(16)}`
+        );
+        ErrorReport.update(
+          "felt-browser-error-connection",
+          lazy.ConsoleClient._getErrorNameForStatus(status),
+          { host: uri.host }
+        );
+        return;
+      }
+
       const windowGlobal = browser.browsingContext?.currentWindowGlobal;
       if (!windowGlobal) {
+        console.error("FeltExtension: No WindowGlobal for SSO callback page");
+        ErrorReport.update("felt-browser-error-connection");
         return;
       }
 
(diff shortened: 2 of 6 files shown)


browser/extensions/felt/content/window.js
```diff
@@ -145,6 +145,76 @@ async function connectToConsole(email) {
...
+
+    onStateChange(webProgress, _request, stateFlags, _status) {
+      if (
+        !(stateFlags & Ci.nsIWebProgressListener.STATE_STOP) ||
+        !(stateFlags & Ci.nsIWebProgressListener.STATE_IS_NETWORK)
+      ) {
+        return;
+      }
+
+      const uri = webProgress.browsingContext?.currentWindowGlobal?.documentURI;
```
lissyx, line 197: shouldn't this match against `lazy.ConsoleClient.ssoCallbackUriMatchPattern` instead?

browser/extensions/felt/content/window.js
```diff
@@ -145,6 +145,76 @@ async function connectToConsole(email) {
     triggeringPrincipal: contentPrincipal,
   });
 
+  // Fallback for token extraction: a cross-process navigation during the SSO
+  // redirect chain can cause the FeltWindowChild JSWindowActor's
+  // DOMContentLoaded handler to never fire. Monitor the navigation from the
+  // parent process and explicitly trigger token extraction when the callback
+  // page finishes loading.
+  const callbackPath = lazy.ConsoleClient._paths.SSO_CALLBACK;
+  const callbackHost = lazy.ConsoleClient.consoleBaseURI.host;
```
lissyx: this is bypassing `ConsoleClient.ssoCallbackUriMatchPattern`

browser/extensions/felt/content/window.js
lissyx: e.g. `const isMatch = browser.matchPatterns.matches(lazy.ConsoleClient.ssoCallbackUriMatchPattern, webProgress.browsingContext?.currentWindowGlobal?.documentURI);`

lissyx: this way we keep matching behaving like the js actor does

browser/extensions/felt/content/window.js
```diff
@@ -145,6 +145,76 @@ async function connectToConsole(email) {
...
+          .catch(err => {
+            console.error(
+              `FeltExtension: Fallback token extraction failed: ${err}`
+            );
+          });
+      } catch (err) {
+        console.error(
+          `FeltExtension: Could not reach FeltWindow actor: ${err}`
+        );
+      }
```
lissyx, line 259: I am wondering if all those error handler should not redirect to `about:neterror?e=...&u=...` somehow, so we have a fallback of showing **something** (and we can leverage it with https://bugzilla.mozilla.org/show_bug.cgi?id=2022263 after for better error reporting)

browser/extensions/felt/content/window.js
```diff
@@ -145,6 +145,76 @@ async function connectToConsole(email) {
...
+      const uri = webProgress.browsingContext?.currentWindowGlobal?.documentURI;
+      if (!uri || uri.host !== callbackHost || uri.filePath !== callbackPath) {
+        return;
+      }
+
+      browser.removeProgressListener(progressListener);
+
+      const windowGlobal = browser.browsingContext?.currentWindowGlobal;
+      if (!windowGlobal) {
+        return;
```
lissyx, line 222: maybe we need a way to notify back the handling code that we have not been able to extract tokens from here, and it should coordinate to ensure there is token extraction at some point and fallback to a proper error page if not

browser/extensions/felt/content/FeltWindowChild.sys.mjs
```diff
@@ -8,30 +8,51 @@ console.debug(`FeltExtension: FeltWindowChild.sys.mjs`);
  *
  */
 export class FeltWindowChild extends JSWindowActorChild {
+  #tokensSent = false;
```
lissyx, line 11: I believe we want to leverage this and redirect to proper error page if we never get to it

browser/extensions/felt/content/window.js
```diff
@@ -145,6 +145,76 @@ async function connectToConsole(email) {
...
+  // page finishes loading.
+  const callbackPath = lazy.ConsoleClient._paths.SSO_CALLBACK;
+  const callbackHost = lazy.ConsoleClient.consoleBaseURI.host;
+  const progressListener = {
+    QueryInterface: ChromeUtils.generateQI([
+      "nsIWebProgressListener",
+      "nsISupportsWeakReference",
+    ]),
+
+    onStateChange(webProgress, _request, stateFlags, _status) {
```
lissyx: This is going to fire multiple times (quite a lot from my testing), I am unsure how we can diagnose "we have not been able to reach the SSO callback" without bad false positive (slow network, fastly/frontend extra redirections).

gcp: I don't think we can.

browser/extensions/felt/content/window.js
gcp: Maybe, maybe that's a bit of scope creep.

browser/extensions/felt/content/window.js
lissyx: I think we can detect the case where browsing is finished?

browser/extensions/felt/content/window.js
lissyx: I'm not sure, you already have a boolean holding the information, and if it can avoid us showing an undiagnosable white page, that's not useless?

browser/extensions/felt/content/window.js
gcp: How do you tell the difference between that and "user looked away and the page is waiting for the 2FA to happen".

lissyx: people need to pay attention :p

gcp: I mean my point stands that the 2FA/IdP pages can take action based on user interaction so "browsing is finished" is not a well-defined thing.

lissyx: Yes, hence "I am unsure how we can diagnose", it would be nice though if we had a fallback. Maybe a (big enough) timeout? Like 2 minutes?

browser/extensions/felt/content/window.js
gcp: Showing the error page is one thing "coordinate there is token extraction at some point" is something else entirely (not sure what it would mean here). In this case I agree we can show the error. I guess perhaps that error page can be customized to have a "try again" or we can add the reset/back button from https://bugzilla.mozilla.org/show_bug.cgi?id=2022768

browser/extensions/felt/content/window.js
gcp: Would you prefer a 1 minute timeout over a reset button? Both?

lissyx: > Would you prefer a 1 minute timeout over a reset button? Both?

1 min is maybe too short (slow network, 2FA with mobile phone and slow data network). Maybe go for reset button and sending something to the remote console and console.debug the browser's element state so we can debug why it was blank?

FTR from my experiment, WebProgress was not enough to cover the about:neterror and others URL because of how we inject them, so in cases like that we would only see the requested URL

lissyx (approved):

gcp:
Not understanding the root cause here ain't great, but this version is more likely to recover, can handle timeouts, and avoids any QoS shenanigans. Unless we can reproduce with logging, I'd consider taking this anyway?

browser/extensions/felt/content/window.js
```diff
@@ -145,6 +145,126 @@ async function connectToConsole(email) {
...
+      if (flags & Ci.nsIWebProgressListener.LOCATION_CHANGE_ERROR_PAGE) {
+        clearTimeout(ssoTimeout);
+        resetToLoginPage("felt-browser-error-connection");
+      }
+    },
+
+    onProgressChange() {},
+    onSecurityChange() {},
+    onStatusChange() {},
+    onContentBlockingEvent() {},
```
lissyx: nit: remove them?

browser/extensions/felt/content/window.js
```diff
@@ -145,6 +145,126 @@ async function connectToConsole(email) {
     triggeringPrincipal: contentPrincipal,
   });
 
+  // Fallback for token extraction: a cross-process navigation during the SSO
+  // redirect chain can cause the FeltWindowChild JSWindowActor's
+  // DOMContentLoaded handler to never fire. Monitor the navigation from the
+  // parent process and explicitly trigger token extraction when the callback
+  // page finishes loading.
+  const SSO_TIMEOUT_MS = 60000;
```
lissyx: nit: make this a pref, it would help for writing tests and/or debugging things

browser/extensions/felt/content/window.js
```diff
@@ -145,6 +145,126 @@ async function connectToConsole(email) {
...
+    document
+      .querySelector(".felt-login__email-pane")
+      .classList.remove("is-hidden");
+    ErrorReport.update(errorType, details, cause);
+  }
+
+  const ssoTimeout = setTimeout(() => {
+    console.error("FeltExtension: SSO login timed out after 60s");
+    resetToLoginPage("felt-browser-error-sso-timeout");
+  }, SSO_TIMEOUT_MS);
```
lissyx: Am I reading right that we only make this timeout once? So we expect the whole SSO to be completed in `<= SSO_TIMEOUT_MS` ?

Maybe we could be a bit more conservative and allow ourselves resetting the timeout each time we get a location change, hence we have the notion of "60s on the same page" instead of "60s on the whole flow". This may help avoiding false-positive and properly dealing with slow network / things.

I can clearly see a case where you combine slow network and a bit distracted user and this hit the "sso timeout" too much.

How long does Duo grants for inputting a mobile-sent code?

dom/ipc/ProcessPriorityManager.cpp
```diff
@@ -770,8 +771,9 @@ static bool PriorityUsesLowPowerMainThread(
 }
 // Method reduces redundancy in pref check while addressing the edge case
 // where a pref is flipped to false during active browser use.
-static bool PrefsUseLowPriorityThreads() {
-  return StaticPrefs::threads_use_low_power_enabled() &&
+static bool ShouldUseLowPriorityThreads() {
+  return !is_felt_ui() &&
+         StaticPrefs::threads_use_low_power_enabled() &&
```
lissyx: If it is a pref, why not forcing it during Felt startup instead of changing things here where we risk more merge conflicts? Wouldn't it work if we do it around here: https://searchfox.org/enterprise-main/rev/a038f49228d707c6675ef20ce640034a64307d2e/browser/extensions/felt/api.js#184-192 ?

lissyx (approved):
r=me but maybe we want to fix those comments, especially the way we set the `threads.use_low_power.enabled` pref

PR #554, opened 2026-03-17
Bug 2022759 - Lockdown mode is visible in UI

jporter-dev:
## Summary

Adds a conditional URLBar action based on existence of `SitePolicies` policy for the current page. Displays a custom informational subview when clicked.

Also added mochitests, not 100% sure I added them properly to get picked up by taskcluster.

## Links

- [Figma](https://www.figma.com/design/xYH6L8QrDz0Q16XFvdo7We/%F0%9F%A6%8A-Client-Design?node-id=1793-7549&m=dev&t=ubB0vtnLt2ks0QOY-1)
- [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2022759)

## Screenshots

lockdown-mode-light

_Light mode_

lockdown-mode-dark

_Dark mode_

lockdown-mode-hcm-1

_HCM 1_

lockdown-mode-hcm-2

_HCM 2_

## Testing

1. `mach build`
2. Set the policy

```json
{
  "policies": {
    "SitePolicies": [
      {
        "Match": ["*.mozilla.org"],
        "Policies": {
          "DisableJit": true
        }
      }
    ]
  }
}
```

3. `mach run`
4. Verify policy is applied `about:policies`
5. Navigate to https://mozilla.org
6. Observe icon in URL bar and click for subpanel
7. Navigate to any other page for icon to be removed


browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -71,6 +72,29 @@ export const EnterpriseHandler = {
...
+    });
+
+    window.gBrowser.addTabsProgressListener({
+      onLocationChange(browser, _webProgress, _request, location) {
+        if (browser !== window.gBrowser.selectedBrowser) {
+          return;
+        }
+        let isLockedDown = false;
+        try {
+          isLockedDown = !Services.policies.isAllowedForURI("jit", location);
```
Mossop, line 89: So this will only display the lockdown UI if the JIT is disabled. How should we handle future features of site policies?

jporter-dev: Ah good point, I was going off the current implementation which just has JIT. Just to clarify, the presence of _any_ policy for a given website in `SitePolicies` would mean we are browsing with restrictions?

Mossop: I think that's a safe place to start. We probably want some new API on the enterprise policy service, something like `isAnyPolicyDisallowedForURI`

Mossop: Though I guess maybe it would be better to defer that to another bug and just go with this for now

Mossop (approved):

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Great, thanks! I'll file a new bug to make this more general for any policies.

gcp: @Mossop @jporter-dev is there some way to directly probe the backend whether the JIT is actually disabled for this site/origin/domain? That way the indicator is also a visual debug tool?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -71,6 +72,29 @@ export const EnterpriseHandler = {
...
   },
 
+  _initLockdownModeButton(window) {
+    const button = window.document.getElementById("lockdown-mode-button");
+
+    button.addEventListener("click", event => {
+      window.PanelUI.showSubView("panelUI-lockdown-mode", button, event);
+    });
+
+    window.gBrowser.addTabsProgressListener({
```
Mossop, line 82: Oh I missed this. This should be using addProgressListener instead.

jporter-dev: Thanks, taking a look now!

browser/components/enterprise/EnterpriseHandler.sys.mjs
Mossop: We could check the remote type of the browsing context yes.

PR #555, opened 2026-03-17
[DRAFT] locking v1

fiji-flo:
Allow storing encrypted refresh tokens in a pref. Unlock via OSKeyStore.


lissyx:
This is being worked on by @jporter-dev in multiple PRs

PR #562, opened 2026-03-18
Bug 2012421: intial access connector support

fiji-flo:
This is more hacky than I hoped. But Firefox VPN is evolving fast. We probably need to factor our changes out at some point.

Example policy:

```
{                                                                        
  "Host": "mail.enterfox.eu",                                                               
  "Port": 443,
  "MatchPatterns": ["https://*.mozilla.org", "https://mozilla.org"]                                                                                                             
}
```


browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -16,6 +16,28 @@
...
+          "type": "number"
+        },
+        "MatchPatterns": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "Locked": {
+          "type": "boolean"
```
gcp, line 35: random: I imagine that the admin would allow adding sites to use via the accessconnector, but not REMOVE sites from the list.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -116,6 +116,75 @@ export var Policies = {
...
+        "browser.ipProtection.autoStartEnabled",
+        true,
+        locked
+      );
+      PoliciesUtils.setDefaultPref("browser.ipProtection.mode", 3, locked);
+
+      const serverList = [
+        {
+          code: "US",
+          cities: [
```
gcp, line 141: "yuck"

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -116,6 +116,75 @@ export var Policies = {
...
+        "browser.ipProtection.features.autoStart",
+        true,
+        locked
+      );
+      PoliciesUtils.setDefaultPref(
+        "browser.ipProtection.autoStartEnabled",
+        true,
+        locked
+      );
+      PoliciesUtils.setDefaultPref("browser.ipProtection.mode", 3, locked);
```
gcp: `MODE_INCLUSION` Perhaps doesn't need a comment, but unless you already know the code you can't guess what 3 means :)

toolkit/components/ipprotection/GuardianClient.sys.mjs
```diff
@@ -243,6 +246,14 @@ export class GuardianClient {
    * - 401: The FxA token was rejected, probably guardian and fxa mismatch. (i.e guardian-stage and fxa-prod)
    */
   async fetchUserInfo(abortSignal = null) {
+    if (Services.felt.isFeltBrowser()) {
+      const entitlement = new Entitlement({
+        subscribed: true,
+        uid: 1,
+        maxBytes: "1000000",
```
gcp, line 262: I assume this is ignored elsewhere, just would expect unlimited to be `0` rather than the somewhat random 1MB.

browser/components/ipprotection/IPProtectionPanel.sys.mjs
```diff
@@ -402,7 +402,7 @@ export class IPProtectionPanel {
     );
     if (!headerButton) {
       headerButton = this.#createHeaderButton(ownerDocument);
-      headerArea.appendChild(headerButton);
+      // headerArea.appendChild(headerButton); disabled for enterprise
```
gcp: #ifdef MOZ_ENTERPRISE etc

browser/components/ipprotection/IPProtectionPanel.sys.mjs
```diff
@@ -686,9 +686,12 @@ export class IPProtectionPanel {
     }
 
     const isExclusion = lazy.IPPExceptionsManager.hasExclusion(principal);
+    const isInclusion = lazy.IPPProxyManager.channelFilter().shouldInclude({
```
gcp: ...are we the only ones using `MODE_INCLUSION`? I wasn't expecting us needing to patch this.

browser/components/ipprotection/IPProtectionPanel.sys.mjs
gcp, line 735: note ^^^

browser/components/ipprotection/content/ipprotection-content.mjs
```diff
@@ -451,21 +451,12 @@ export default class IPProtectionContentElement extends MozLitElement {
...
-      this._showMessageBar = true;
-    } else if (!this.state.onboardingMessage && !this.state.bandwidthWarning) {
-      // Remove the message bar if we can no longer render messages before they were dismissed
-      this._showMessageBar = false;
-    }
-
-    const messageBar = this._showMessageBar ? this.messageBarTemplate() : null;
-
-    let content = html`${messageBar}${this.mainContentTemplate()}`;
+    const content = this.state.siteData.isInclusion
```
gcp: `siteData` needs a `null` guard (see below)

gcp: The other users are doing things like `this.state?.siteData?.isExclusion ?? false`

browser/components/ipprotection/IPProtectionToolbarButton.sys.mjs
```diff
@@ -361,6 +369,20 @@ export class IPProtectionToolbarButton {
...
 
+  /**
+   * Checks if the given principal is included in IP Protection.
+   *
+   * @param {nsIPrincipal} principal
+   *  The principal to check.
+   * @returns {boolean}
+   *  True if the site is included, false otherwise.
+   */
+  #isIncludedSite(principal) {
```
gcp, line 380: `isExcludedSite` has a bunch of guards against `null` and `isNullPrincipal`. Probably you should too.

browser/components/ipprotection/IPProtectionPanel.sys.mjs
gcp: The assymetry of not having Includes in the IPPExceptionsManager irks me but that's outside our control I guess.

gcp: I guess forcing `autoStart` above means you can rely on `channelFilter` existing before this code runs? How certain are you of that? 😉

browser/components/ipprotection/GuardianClient.sys.mjs
```diff
@@ -278,6 +289,16 @@ export class GuardianClient {
...
   async fetchProxyUsage(abortSignal) {
+    if (Services.felt.isFeltBrowser()) {
+      const entitlement = new ProxyUsage(
+        "1000000",
+        "1000000",
+        Temporal.Now.plainDateTimeISO()
+          .add(Temporal.Duration.from({ days: 30 }))
+          .toString()
+      );
+      return { status: 200, entitlement };
```
gcp: `return entitlement;` (this looks like a copypasta 🍝 error)

gcp (changes requested):

browser/components/enterprisepolicies/schemas/policies-schema.json
fiji-flo: Yes, but that's probably true for a lot of our block allow lists. Let's do this once we can do this.

browser/components/enterprisepolicies/Policies.sys.mjs
fiji-flo: 👍

toolkit/components/ipprotection/GuardianClient.sys.mjs
fiji-flo: It's somewhat random but 1 seemed to low if it gets converted to MB somewhere and 0 seemed to dangerous.

browser/components/ipprotection/GuardianClient.sys.mjs
fiji-flo: Indeed

browser/components/ipprotection/IPProtectionPanel.sys.mjs
fiji-flo: I do hope. But yes I want to upstream these changes.

gcp: The point was that you could also guard this?

toolkit/components/ipprotection/GuardianClient.sys.mjs
```diff
@@ -65,6 +66,9 @@ export class GuardianClient {
    *  - False: The user is not linked to the Guardian service, they cannot be a proxy user.
    */
   async isLinkedToGuardian(onlyCached = false) {
+    if (AppConstants.I) {
```
gcp: Hmm?

toolkit/components/ipprotection/GuardianClient.sys.mjs
gcp: Looks like this is only really used by the UI, as all accounting is server-side, and remaining quota also checks from what those headers return.

browser/components/ipprotection/IPProtectionPanel.sys.mjs
```diff
@@ -407,7 +407,18 @@ export class IPProtectionPanel {
...
+    if (Services.felt.isFeltBrowser()) {
+      const newHeaderButton =
+        this.#createAccessConnectorStatusLabel(ownerDocument);
+      if (headerButton) {
+        headerArea.replaceChild(newHeaderButton, headerButton);
+      } else {
+        headerArea.appendChild(newHeaderButton);
+      }
+      headerButton = newHeaderButton;
+      headerButton = this.#createAccessConnectorStatusLabel(ownerDocument);
```
gcp: Claude pointed out this is almost certainly a copy paste error too. It's already created above.

gcp (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -116,6 +116,79 @@ export var Policies = {
...
   },
 
+  AccessConnector: {
+    onBeforeAddons(manager, param) {
+      const locked = param.Locked ?? true;
+      PoliciesUtils.setDefaultPref(
+        "browser.ipProtection.enabled",
+        true,
+        locked
+      );
```
fiji-flo: We should move that down, that's probably causing some of the issues. So setting everything first and then enabling it.

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -412,7 +412,19 @@ export function IPProtectionServerlistFactory() {
...
+  {},
+  {
+    get(_target, prop) {
+      if (!_serverlist) {
+        _serverlist = IPProtectionServerlistFactory();
+      }
+      return _serverlist[prop];
+    },
+  }
+);
```
fiji-flo: I we change the order in the policy we can skip this?

toolkit/components/ipprotection/IPProtectionService.sys.mjs
```diff
@@ -230,4 +230,38 @@ XPCOMUtils.defineLazyPreferenceGetter(
...
+        let server = IPProtectionServerlist.selectServer(location?.city);
+        let filter = IPPProxyManager.channelFilter();
+        if (server && filter && !filter.proxyInfo) {
+          let token = "";
+          try {
+            token = Services.felt.getAccessTokenIfValid();
+          } catch (_) {}
+          filter.initialize(token, server);
+        }
+      }
```
fiji-flo: That's not needed if we put the prefs in order.

fiji-flo (commented):
If we move setting `"browser.ipProtection.enabled"` down all gets easier.

PR #563, opened 2026-03-18
Remove /api/browser/hacks/startup endpoint

1rneh:
- Undo the `GET` request to `/api/browser/hacks/startup`.
- Undo the sanitization and writing of preferences into prefs.js.
- Undo sending pref `browser.policies.live_polling.frequency=lazy.FeltCommon.POLICY_POLLING_FREQUENCY` from Felt to Firefox. This preference is already provided via the `/api/browser/hacks/default` `GET` response and send through felt.


lissyx (approved):
Though I can admit I'm a bit worried doing this now and the consequences it may have in term of race conditions :)

PR #564, opened 2026-03-19
Bug 2023485 - Enterprise: check for consecutive updates failures

lissyx:
If for some reason some updates are failing to apply, our force checking, download and restart at the start of Felt may end up in infinite loop. Check the update history to verify for previous app updates failures and report and error in case it makes sense.

### DIFF
--- a/browser/components/enterprise/modules/Updates.sys.mjs
+++ b/browser/components/enterprise/modules/Updates.sys.mjs
@@ -23,6 +23,8 @@ export const Updates = {
     this._errorReporter = errorReporter;
 
     this.maybeShowUpdateSuccess();
+    // Check this early to avoid re-downloading updates when it would fail
+    await this.updateCheckingAllowed();
 
     if (this._initialized) {
       this.displayLoginState();
@@ -38,7 +40,9 @@ export const Updates = {
 
     this._checkingTimeout = null;
     this._receivedStaging = false;
-    this.displayUpdateState();
+    if (this._canDoUpdateChecking) {
+      this.displayUpdateState();
+    }
 
     if (lazy.isUpdatesTesting()) {
       // on Windows at least, during testing, make sure we let time for the UI to show up
@@ -52,6 +56,14 @@ export const Updates = {
     this.forceUpdateCheck();
 
     this._initialized = true;
+  },
+
+  uninit() {
+    this.unobserve();
+    this.cancelDelayedUpdateCheckUI();
+    this._document = undefined;
+    this._errorReporter = undefined;
+    this._initialized = false;
   },
 
   maybeShowUpdateSuccess() {
@@ -73,6 +85,44 @@ export const Updates = {
     );
   },
 
+  async updateCheckingAllowed() {
+    const UM = Cc["@mozilla.org/updates/update-manager;1"].getService(
+      Ci.nsIUpdateManager
+    );
+    // History is limited to 10 items, each install attempt should report a failure
+    const history = await UM.getHistory().catch(ex => {
+      console.error(`FeltUpdates: updateCheckingAllowed failed`, ex);
+      return null;
+    });
+    if (history) {
+      // The UpdaterManager history's capped at 10 max.
+      const maxConsecutiveUpdateFailures = Math.min(
+        Services.prefs.getIntPref(
+          "enterprise.felt.max_consecutive_update_failure",
+          3
+        ),
+        10
+      );
+      const firstNonFailed = history.findIndex(
+        update => update.state !== "failed"
+      );
+      const consecutiveUpdateFailures =
+        firstNonFailed === -1 ? history.length : firstNonFailed;
+      console.warn(
+        `FeltUpdates: updateCheckingAllowed: ${consecutiveUpdateFailures}, max ${maxConsecutiveUpdateFailures}`
+      );
+      if (consecutiveUpdateFailures > maxConsecutiveUpdateFailures) {
+        console.warn(
+          `FeltUpdates: updateCheckingAllowed: skip startup update check because consecutive update failures: ${consecutiveUpdateFailures}, max ${maxConsecutiveUpdateFailures}`
+        );
+        this._canDoUpdateChecking = false;
+        return;
+      }
+    }
+
+    this._canDoUpdateChecking = true;
+  },
+
   prepareUpdateCheck() {
     this._appUpdater = new lazy.AppUpdater();
     this._updaterCallback = this.appUpdaterCallback.bind(this);
@@ -83,10 +133,21 @@ export const Updates = {
   },
 
   forceUpdateCheck() {
+    if (this._canDoUpdateChecking !== true) {
+      console.warn(
+        `FeltUpdates: forceUpdateCheck(): skip because previous updates failures`
+      );
+      this.displayLoginStateWithUpdateError("contact-admin");
+      return;
+    }
+
     this._appUpdater
       .check()
       .catch(err => {
-        console.error(`AppUpdater failure: ${err}`, err);
+        console.error(
+          `Felt: forceUpdateCheck(): AppUpdater failure: ${err}`,
+          err
+        );
         this.displayLoginStateWithUpdateError("contact-admin");
       })
       .finally(() => {
@@ -276,6 +337,13 @@ export const Updates = {
       });
   },
 
+  unobserve() {
+    Services.obs.removeObserver(this, "update-staged");
+    Services.obs.removeObserver(this, "update-downloaded");
+    Services.obs.removeObserver(this, "update-error");
+    Services.obs.removeObserver(this, "xpcom-shutdown");
+  },
+
   observe(subject, topic, state) {
     // We would coerce subject
     //   update = subject && subject.QueryInterface(Ci.nsIUpdate);
@@ -284,10 +352,7 @@ export const Updates = {
     console.warn(`FeltUpdates: observer: topic:${topic} state:${state}`);
     switch (topic) {
       case "xpcom-shutdown":
-        Services.obs.removeObserver(this, "update-staged");
-        Services.obs.removeObserver(this, "update-downloaded");
-        Services.obs.removeObserver(this, "update-error");
-        Services.obs.removeObserver(this, "xpcom-shutdown");
+        this.unobserve();
         break;
       case "update-staged":
       case "update-downloaded":
--- a/testing/enterprise/felt_updater_errors.py
+++ b/testing/enterprise/felt_updater_errors.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+
+import requests
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_tests import FeltTests
+
+
+class FeltUpdaterErrorsBase(FeltTests):
+    def get_update_config_file_path(self):
+        self._driver.set_context("chrome")
+        rv = self._driver.execute_script(
+            """
+            const { UpdateUtils } = ChromeUtils.importESModule("resource://gre/modules/UpdateUtils.sys.mjs");
+            return UpdateUtils.getConfigFilePath();
+            """
+        )
+        self._driver.set_context("content")
+        return rv
+
+    def reload_chrome_window(self):
+        self._driver.set_context("chrome")
+        self._driver.execute_async_script(
+            """
+            const callback = arguments[arguments.length - 1];
+            const UM = Cc["@mozilla.org/updates/update-manager;1"].getService(Ci.nsIUpdateManager);
+            UM.internal.reload(false).then(() => {
+              const { Updates } = ChromeUtils.importESModule("resource:///modules/enterprise/Updates.sys.mjs");
+              Updates.uninit();
+              window.location.reload();
+            }).finally(() => callback());
+            """
+        )
+        self._driver.set_context("content")
+
+    def get_updates_served(self):
+        return requests.get(
+            f"http://localhost:{self.console_port}/api/browser/forced_updates_count"
+        ).json()["serve_forced_updates_count"]
+
+    def assert_updates_served(self, amount):
+        served = self.get_updates_served()
+        self._logger.info(f"Checking updates served: {amount} vs {served}")
+        self._wait.until(lambda mn: self.get_updates_served() == amount)
+        served = self.get_updates_served()
+        assert served == amount, (
+            f"There should have been {amount} update served but {served}"
+        )
+
+    def assert_updates_check_allowed(self, allowed):
+        self._driver.set_context("chrome")
+        rv = self._driver.execute_async_script(
+            """
+            const callback = arguments[arguments.length - 1];
+            const UM = Cc["@mozilla.org/updates/update-manager;1"].getService(Ci.nsIUpdateManager);
+            UM.internal.reload(false).then(() => {
+              const { Updates } = ChromeUtils.importESModule("resource:///modules/enterprise/Updates.sys.mjs");
+              Updates.updateCheckingAllowed().then(() => callback(Updates._canDoUpdateChecking));
+            });
+            """
+        )
+        self._driver.set_context("content")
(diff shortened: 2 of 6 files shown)


browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -73,6 +82,46 @@ export const Updates = {
...
+      );
+      const consecutiveUpdateFailures = previousConsecutiveUpdateFailures + 1;
+      if (consecutiveUpdateFailures > maxConsecutiveUpdateFailures) {
+        console.warn(
+          `FeltUpdates: updateCheckingAllowed: skip startup update check because consecutive update failures: ${consecutiveUpdateFailures}, max ${maxConsecutiveUpdateFailures}`
+        );
+        this.displayLoginStateWithUpdateError("contact-admin");
+        this._canDoUpdateChecking = false;
+        return;
+      }
```
lissyx, line 120: This may trigger if you have the last one update in failed state and that you start consecutively three times, I need to verify this and cover it in tests, and then we may want to reset the pref at a different time

lissyx: Should be covered by `FeltUpdaterErrorsRecovery`

Mossop: Does the update history clear after a restart? Otherwise I'm not seeing how this gets reset.

lissyx: It should not, the current behavior being that we would show there was an error applying updates and that admin should be contacted, not perform a force check, but we can still launch the browser

lissyx: if we successfully apply an update after (in background), it will reset. The idea is really to avoid a case where:
 - start with outdated build
 - something (privileges?) blocks installation
 - felt restarts, try again to download/stage update
 - loop out of control

Mossop: I may be missing something abut the flow here, but if I quit Felt before any new update check happens and then start doesn't that increment the counter again even though there has been no new update attempt? The history would still show the most recent update as failed.

lissyx: Yes, and I'm not sure how we should take care of that right now, but avoiding crash loop is more urgent. The blockage here is only about initial's check for update of Felt, not the background ones

Mossop: Ok, make sure there is a bug on file to solve that

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -73,6 +82,46 @@ export const Updates = {
...
+      Services.prefs.setIntPref(
+        "enterprise.felt.previous_consecutive_update_failure",
+        consecutiveUpdateFailures
+      );
+    } else if (history && history[0] && history[0].state === "succeeded") {
```
Mossop: Should we also reset if there is no history?

Mossop (approved):

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: @Mossop I think I'll try to fix it right away depending on how I progress here, instead of making it a follow up, but yes, that wasthe plan

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Maybe, good idea

browser/components/enterprise/modules/Updates.sys.mjs
gcp: If you have the last 10 items, isn't it less error prone to make this stateless (just scanning the history) instead of maintaining the counter?
I assume nobody's going to set the fail limit to 11 obviously.

browser/components/enterprise/modules/Updates.sys.mjs
gcp, line 39: Shouldn't this be skipped if the above falses `this._canDoUpdateChecking`?

testing/enterprise/manifest.toml
```diff
@@ -46,6 +46,8 @@ run-if = [
 
 ["test_felt_browser_starts.py"]
 
+["test_felt_browser_updater_errors.py"]
```
gcp, line 49: What about `test_felt_browser_updater_errors_recovery.py`

gcp (changes requested):

testing/enterprise/manifest.toml
lissyx: forgot it

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Well from my testing (macOS elevated pkg), it was not 100% certain that **any** update application failure would be reflected into `updates.xml`, I'll check that tomorrow yes

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: no because there we register for the observers and listeners for appUpdater, we want to block only the force check/update at the startup phase to avoid the risk of looping if it fails, but we do not want to block updates that can happen in the background. Imagine you have got a broken one, the verification here avoids looping over and over, we block check/application at startup, you login, and then a new update gets pushed, we apply it in background and fix the mess.

gcp: But it also resets the UI, even though we won't actually check here?

Edit: Now I'm confused, I thought it did because of a `displayUpdateState()` here.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: OK, after testing a lot, I cannot prove my theory, so it is possible this was just a byproduct of the messy state of how I was testing things. I'll remove the pref-based state and rely only on the last values from update history

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: We have a `this.displayUpdateState();` at https://searchfox.org/enterprise-main/rev/a038f49228d707c6675ef20ce640034a64307d2e/browser/components/enterprise/modules/Updates.sys.mjs#41, maybe there's a risk of race between this one and my added `this.displayLoginStateWithUpdateError("contact-admin");`

gcp: Yes, I do think they race, as you call `this.displayUpdateState();` which hides the `.felt-login` that is supposed to show the error.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -73,6 +82,54 @@ export const Updates = {
...
+      const countConsecutiveUpdateFailures = items => {
+        let previous = 0;
+        let current = 0;
+        if (items && items[0] && items[0].state === "failed") {
+          for (const item of items) {
+            if (item.state === "failed") {
+              current++;
+              previous = Math.max(previous, current);
+            } else {
+              current = 0;
```
gcp: You need to break out of the loop here. We don't care there was a large streak of failures a while ago if we successfully updated now.

gcp: This also means the `Math.max` part makes no sense. We only care about the last streak.

gcp: Functional version:
```
  const firstNonFailed = history.findIndex(update => update.state !== "failed");
  const consecutiveUpdateFailures =
    firstNonFailed === -1 ? history.length : firstNonFailed;
```

testing/enterprise/felt_updater_errors.py
```diff
@@ -0,0 +1,105 @@
...
+    def get_updates_served(self):
+        return requests.get(
+            f"http://localhost:{self.console_port}/api/browser/updates_count"
+        ).json()["serve_updates_count"]
+
+    def assert_updates_served(self, amount):
+        self._logger.info("Checking updates served")
+        self._wait.until(lambda mn: self.get_updates_served() == amount)
+        served = self.get_updates_served()
+        assert served, f"There should have been {amount} update served but {served}"
```
gcp: ``served == amount``?

gcp (changes requested):

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Looks like my mind was somewhere else, but now at least it's covered by tests

testing/enterprise/felt_updater_errors.py
lissyx: Since we need to do an extra call to `get_updates_served()` and considering my first implementation was much more checking using this server-side counting and proved to be super racy, I was not sure it was a good idea to add a risk of race here.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: Ok, if we move `this.displayLoginStateWithUpdateError("contact-admin");` into `forceUpdateCheck()` checks, we should be fine and have no race

testing/enterprise/felt_updater_errors.py
gcp: needs `await`

testing/enterprise/test_felt_browser_updater_errors.py
gcp: I fail to understand what repeating this test achieves.

gcp, line 66: Maybe even better to (also) check `self.assert_updates_served(0)`?

gcp, line 83: `self.assert_updates_served(1)`

testing/enterprise/felt_updater_errors.py
gcp: There's no guarantee this finishes loading before the function returns, so depending on why you need this to reload, the test could be racy.

Unfortunately you can't await this so it's a bit more messy to synchronize.

gcp (commented):
I have questions.

Edit: Why Github isn't showing the source next to them is a mistery.

testing/enterprise/test_felt_browser_updater_errors.py
lissyx: Making sure repeated access to not trigger any blockage, since this is how it was decided before

lissyx: I'll remove iit

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -73,6 +82,41 @@ export const Updates = {
...
   },
 
+  async updateCheckingAllowed() {
+    const UM = Cc["@mozilla.org/updates/update-manager;1"].getService(
+      Ci.nsIUpdateManager
+    );
+    // History is limited to 10 items, each install attempt should report a failure
+    const history = await UM.getHistory();
+    if (history) {
+      // The UpdaterManager history's capped at 10 max.
```
gcp, line 98: Note this means setting the pref to 10 means infinite.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -73,6 +82,41 @@ export const Updates = {
...
+      const maxConsecutiveUpdateFailures = Math.min(
+        Services.prefs.getIntPref(
+          "enterprise.felt.max_consecutive_update_failure",
+          3
+        ),
+        10
+      );
+      const firstNonFailed = history.findIndex(
+        update => update.state !== "failed"
```
gcp, line 107: `download-failed` could be considered here too, but I'm not sure.

testing/enterprise/test_felt_browser_updater_errors.py
```diff
@@ -0,0 +1,104 @@
...
+
+class FeltUpdaterErrors(FeltUpdaterErrorsBase):
+    EXTRA_PREFS = {
+        "app.update.log": True,
+        "app.update.disabledForTesting": False,
+        "app.update.BITS.enabled": False,
+        "enterprise.felt_tests.is_updates_testing": False,
+    }
+
+    # This should get executed before test_felt_updater_errors_recover()
```
gcp: Is that in another patch?

testing/enterprise/test_felt_browser_updater_errors.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_felt_updater_errors(self):
+        self._update_root = os.path.dirname(self.get_update_config_file_path())
+        self._updates_history = os.path.join(self._update_root, "updates.xml")
+        self._logger.info(f"Updates hitory {self._updates_history}")
```
gcp: typo

testing/enterprise/test_felt_browser_updater_errors.py
```diff
@@ -0,0 +1,104 @@
...
+
+        if os.path.isfile(self._updates_history):
+            os.unlink(self._updates_history)
+            self._logger.info(f"Updates hitory {self._updates_history} removed.")
```
gcp: same

testing/enterprise/test_felt_browser_updater_errors.py
```diff
@@ -0,0 +1,104 @@
...
+            self.one_xml("failed"),
+            self.one_xml("succeeded"),
+        ])
+        self.reset_updates_served()
+        self.reload_chrome_window()
+        self.assert_updates_check_allowed(False)
+        self.assert_updates_served(0)
+        self.assert_error_displayed()
+
+        # This should cound as two consecutive failures, so we still allow
```
gcp: nit: count

gcp (approved):

testing/enterprise/test_felt_browser_updater_errors.py
lissyx: leftover from previous version of the patch

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: https://bugzilla.mozilla.org/show_bug.cgi?id=2026493

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -73,6 +82,41 @@ export const Updates = {
     );
   },
 
+  async updateCheckingAllowed() {
+    const UM = Cc["@mozilla.org/updates/update-manager;1"].getService(
+      Ci.nsIUpdateManager
+    );
+    // History is limited to 10 items, each install attempt should report a failure
+    const history = await UM.getHistory();
```
gcp:
```suggestion
    const history = await UM.getHistory().catch(ex => {
      console.error(`FeltUpdates: updateCheckingAllowed failed`, ex);
      return null;
    });
```

gcp: This is to ensure we fall through to enabling updates.

browser/components/enterprise/modules/Updates.sys.mjs
```diff
@@ -54,6 +56,13 @@ export const Updates = {
     this._initialized = true;
   },
 
+  uninit() {
+    this.unobserve();
```
gcp, line 62:
```suggestion
    this.unobserve();
    this.cancelDelayedUpdateCheckUI();
```

browser/components/enterprise/modules/Updates.sys.mjs
gcp:
```suggestion
  if (this._canDoUpdateChecking) {
    this.displayUpdateState();
  }
```

gcp: You fixed the race (technically, you're still racing a 500ms timeout, but I'll bet on you winning that race), but this feels cleaner?

gcp (approved):

gcp:
Mostly some robustness nits.

browser/components/enterprise/modules/Updates.sys.mjs
lissyx: nice

lissyx:
There's something hanging between a pair of crash tests, only on windows/opt, `testing/enterprise/test_felt_browser_crash_abort_restart.py` and `testing/enterprise/test_felt_browser_crash_restart.py`, however it's not reproducible locally for me, but making a PR including this one and #568 works better, so likely the `wait_process_exit` fixes we have there are fixing the issue we are now hitting here

PR #566, opened 2026-03-19
Make nulled endpoints point to base console uri

1rneh:
- Retrieves the console base URI from `ConsoleClient` instead of reading it directly from preferences.
- Sets `browser.ipProtection.guardian.endpoint` and `identity.fxaccounts.remote.root` to console base URI
- Restores URI validation checks in `FxAccountsConfig` that were previously bypassed since `identity.fxaccounts.remote.root` was an empty string.


browser/components/enterprise/EnterpriseEndpoints.sys.mjs
```diff
@@ -23,27 +35,26 @@ const ENDPOINT_PREFS = [
...
-    for (const { pref, path } of ENDPOINT_PREFS) {
-      const url = new URL(path, baseURL).href;
+
+    for (const { pref, path } of RELATIVE_CONSOLE_ENDPOINT_PREFS) {
+      const url = new URL(path, consoleBaseURI).href;
       defaultBranch.setStringPref(pref, url);
     }
+
+    for (const pref of BASE_CONSOLE_URI_PREFS) {
+      defaultBranch.setStringPref(pref, consoleBaseURI);
```
gcp: `consoleBaseURI.href` to match the code above (to get a string instead of relying on the implicit conversion).

gcp (approved):

browser/components/enterprise/tests/browser/browser_enterprise_endpoints.js
```diff
@@ -0,0 +1,48 @@
...
+   http://creativecommons.org/publicdomain/zero/1.0/ */
+
+"use strict";
+
+const { ConsoleClient, PREFS } = ChromeUtils.importESModule(
+  "resource:///modules/enterprise/ConsoleClient.sys.mjs"
+);
+const {
+  RELATIVE_CONSOLE_ENDPOINT_PREFS,
+  BASE_CONSOLE_URI_PREFS,
```
gcp, line 11: The test is sort of duplicating the code it's testing, which makes me think it's probably not great for catching mistakes.

Would it make sense to make a list of prefs with `mozilla.org` and error if it shows up somewhere else?

That also prevents upstream merges from accidentally introducing dependencies on our servers.

gcp (approved):

browser/components/enterprise/tests/browser/browser_enterprise_endpoints.js
1rneh: Follow-up bug: [2024702](https://bugzilla.mozilla.org/show_bug.cgi?id=2024702)

PR #568, opened 2026-03-19
Bug 2024587 - Enterprise: restart felt when browser requests restart …

lissyx:
…and there is a ready update

### DIFF
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -147,48 +147,86 @@ export class FeltProcessParent extends JSProcessActorParent {
               "enterprise.disable_restart",
               false
             );
-            console.debug(
-              `FeltExtension: ParentProcess: restart notification, restartDisabled=${restartDisabled}`
+
+            const UM = Cc["@mozilla.org/updates/update-manager;1"].getService(
+              Ci.nsIUpdateManager
             );
-            // Kill Firefox directly instead of broadcasting to receiveMessage()
-            // since gFeltProcessParentInstance is accessible here
-            if (gFeltProcessParentInstance?.proc) {
-              gFeltProcessParentInstance.restartReported = true;
-              gFeltProcessParentInstance.firefox = null;
-              console.debug(
-                `FeltExtension: ParentProcess: Killing Firefox PID=${gFeltProcessParentInstance.proc.pid}`
-              );
-              gFeltProcessParentInstance.proc
-                .kill()
-                .then(() => {
+            UM.getReadyUpdate()
+              .then(readyUpdate => {
+                let pendingUpdate = false;
+                if (readyUpdate) {
+                  // Updates states when restarting will finish the update
+                  const readyStates = [
+                    "pending",
+                    "pending-service",
+                    "pending-elevate",
+                    "applied",
+                    "applied-service",
+                  ];
+                  pendingUpdate = readyStates.includes(readyUpdate.state);
+                }
+                return pendingUpdate;
+              })
+              .catch(err => {
+                console.debug(
+                  `FeltExtension: ParentProcess: getReadyUpdate failed: ${err}`
+                );
+              })
+              .then(pendingUpdate => {
+                console.debug(
+                  `FeltExtension: ParentProcess: restart notification, restartDisabled=${restartDisabled}`
+                );
+                // Kill Firefox directly instead of broadcasting to receiveMessage()
+                // since gFeltProcessParentInstance is accessible here
+                if (gFeltProcessParentInstance?.proc) {
+                  gFeltProcessParentInstance.restartReported = true;
+                  gFeltProcessParentInstance.firefox = null;
                   console.debug(
-                    `FeltExtension: ParentProcess: Killed Firefox, restartDisabled=${restartDisabled}`
+                    `FeltExtension: ParentProcess: Killing Firefox PID=${gFeltProcessParentInstance.proc.pid}`
                   );
-                  if (!restartDisabled) {
-                    console.debug(
-                      `FeltExtension: ParentProcess: Starting new Firefox`
-                    );
-                    gFeltProcessParentInstance.startFirefox(
-                      PROCESS_START_REASON.RESTART
-                    );
-                  } else {
-                    console.debug(
-                      `FeltExtension: ParentProcess: Restart disabled, sending normal exit to restore FELT UI`
-                    );
-                    Services.cpmm.sendAsyncMessage(
-                      "FeltParent:FirefoxNormalExit",
-                      {}
-                    );
-                  }
-                })
-                .catch(err => {
+                  gFeltProcessParentInstance.proc
+                    .kill()
+                    .then(() => {
+                      console.debug(
+                        `FeltExtension: ParentProcess: Killed Firefox, restartDisabled=${restartDisabled}`
+                      );
+
+                      if (!restartDisabled && !pendingUpdate) {
+                        console.debug(
+                          `FeltExtension: ParentProcess: Starting new Firefox`
+                        );
+                        gFeltProcessParentInstance.startFirefox(
+                          PROCESS_START_REASON.RESTART
+                        );
+                      } else if (pendingUpdate) {
+                        console.debug(
+                          `FeltExtension: ParentProcess: Restart requested and pending update, restarting FELT UI`
+                        );
+                        Services.cpmm.sendAsyncMessage(
+                          "FeltParent:FirefoxRestartUpdateExit",
+                          {}
+                        );
+                      } else {
+                        console.debug(
+                          `FeltExtension: ParentProcess: Restart disabled, sending normal exit to restore FELT UI`
+                        );
+                        Services.cpmm.sendAsyncMessage(
+                          "FeltParent:FirefoxNormalExit",
+                          {}
+                        );
+                      }
+                    })
+                    .catch(err => {
+                      console.debug(
+                        `FeltExtension: ParentProcess: Kill failed: ${err}`
+                      );
+                    });
+                } else {
                   console.debug(
-                    `FeltExtension: ParentProcess: Kill failed: ${err}`
+                    `FeltExtension: ParentProcess: No proc to kill!`
                   );
-                });
-            } else {
-              console.debug(`FeltExtension: ParentProcess: No proc to kill!`);
-            }
+                }
+              });
             break;
           }
           case "felt-extension-ready": {
--- a/testing/enterprise/base_test.py
+++ b/testing/enterprise/base_test.py
@@ -135,9 +135,24 @@ class EnterpriseTestsBase(MarionetteTestCase):
         self._logger.info(f"Marionette PORT NEW: {new_marionette_port}")
         assert marionette_port == new_marionette_port, "STILL Valid marionette port"
         assert marionette_port != 2828, "Marionette port should not be default value"
+        self._logger.info(f"Marionette PORT NEW: {new_marionette_port} OK")
 
+        self._logger.info(f"New Marionette port={new_marionette_port}")
         self._child_driver = Marionette(host="127.0.0.1", port=new_marionette_port)
+        self._logger.info(f"New Marionette port={new_marionette_port} OK")
         self._child_driver.start_session(capabilities, timeout=60)
+        self._logger.info(f"New Marionette port={new_marionette_port} OK SESSION")
+
+        port_file_copy = f"{marionette_port_file}.bak"
+        self._logger.info(
+            f"New Marionette port={new_marionette_port} MOVE OUT {marionette_port_file}"
+        )
+        if os.path.isfile(port_file_copy):
+            os.unlink(port_file_copy)
+        os.rename(marionette_port_file, port_file_copy)
+        self._logger.info(
+            f"New Marionette MOVED OUT {marionette_port_file} TO {port_file_copy}"
+        )
 
     def get_driver(self, env):
         return self._driver if env == Environment.FELT else self._child_driver
--- a/testing/enterprise/felt_tests.py
+++ b/testing/enterprise/felt_tests.py
@@ -238,7 +238,7 @@ class ConsoleHttpHandler(LocalHttpRequestHandler):
             )
             if self.server.serve_updates and os.path.isfile(complete_mar):
                 # Versions are important, they need to be equal or higher than the
-                # curernt binary otherwise no update will be downloaded
+                # current binary otherwise no update will be downloaded
                 display_version = self.server.serve_updates_version[
                     "application_version"
                 ][0]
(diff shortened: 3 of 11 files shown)


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -164,13 +182,22 @@ export class FeltProcessParent extends JSProcessActorParent {
...
                       PROCESS_START_REASON.RESTART
                     );
+                  } else if (pendingUpdate) {
+                    console.debug(
+                      `FeltExtension: ParentProcess: Restart requested and pending update, restarting FELT UI`
+                    );
+                    Services.cpmm.sendAsyncMessage(
+                      "FeltParent:FirefoxRestartUpdateExit",
+                      {}
+                    );
```
Mossop: I'm a little confused by this, doesn't cpmm send messages from content to parent, except we're in the parent process already here?

lissyx: That maybe something we want to fix indeed; but that is how we do communication for now.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -147,6 +147,24 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+            if (readyUpdate) {
+              // Updates states when restarting will finish the update
+              const readyStates = [
+                "pending",
+                "pending-service",
+                "pending-elevate",
+                "applied",
+                "applied-service",
+              ];
+              pendingUpdate = readyStates.include(readyUpdate.state);
```
gcp: Puzzled how this code can work given https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

gcp (changes requested):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: If you are suggesting I did a typo, yes :)

lissyx: I was planning to write tests anyway for that, it's progressing correctly right now.

testing/enterprise/felt_tests.py
```diff
@@ -239,6 +227,22 @@ def do_GET(self):
                 os.path.dirname(__file__), os.path.basename("complete.mar")
             )
             if self.server.serve_updates and os.path.isfile(complete_mar):
+                # Versions are important, they need to be equal or higher than the
+                # curernt binary otherwise no update will be downloaded
```
Mossop:
```suggestion
                # current binary otherwise no update will be downloaded
```

Mossop (approved):

testing/enterprise/felt_tests.py
gcp: This should just be `time.sleep(1)` with a loop that repeats up to 30 times or so.

gcp: `process_basename.startswith("firefox")` or this will fail for `firefox.exe`?

gcp (approved):
LGTM but the existing test looks wonky

testing/enterprise/felt_tests.py
lissyx: Right, I thought I had fixed that in the past but I never merged it somehow.

lissyx:
> LGTM but the existing test looks wonky

I'll address that in a follow up to limit risks

testing/enterprise/felt_tests.py
```diff
@@ -653,31 +657,55 @@ def find_elem_by_id(self, e):
...
+                assert not process_basename.startswith("firefox"), (
+                    f"Process PID {pid_to_check} should not be Firefox"
                 )
-                assert process_basename != "firefox", "Process is not Firefox"
+            except psutil.NoSuchProcess:
+                self._logger.info(f"PID disappeared {pid_to_check}")
             except psutil.ZombieProcess:
-                self._logger.info(f"Zombie found as {self._browser_pid}")
+                # If it is a zombie, it is fine as well
+                self._logger.info(f"Zombie found as {pid_to_check}")
```
lissyx, line 734: @gcp this was indeed required to avoid issues on windows testing, that were failing on CI and locally, it's hopefully more robust

toolkit/xre/nsAppRunner.cpp
```diff
@@ -3109,7 +3108,21 @@ static nsresult SelectProfile(nsToolkitProfileService* aProfileSvc,
...
+    const char* xreProfilePath = PR_GetEnv("XRE_PROFILE_PATH");
+    auto savedProfile = xreProfilePath && *xreProfilePath;
+    const char* xreProfileLocalPath = PR_GetEnv("XRE_PROFILE_LOCAL_PATH");
+    auto savedLocalProfile = xreProfileLocalPath && *xreProfileLocalPath;
+
+    // Check if we are FELT UI, and nothing has given us a profile path, either
+    //  - -profile path/to/profile
+    //  - XRE_PROFILE_PATH=path/to/profile /
+    //  XRE_PROFILE_LOCAL_PATH=path/to/local
+    if (is_felt_ui() && !forcedProfile && !savedProfile && !savedLocalProfile) {
```
lissyx, line 3131: @gcp this was also required to make sure the restart happens with the correct profile and not the default felt profile

testing/enterprise/base_test.py
```diff
@@ -135,9 +135,24 @@ def connect_child_browser(self, capabilities=None):
...
+        port_file_copy = f"{marionette_port_file}.bak"
+        self._logger.info(
+            f"New Marionette port={new_marionette_port} MOVE OUT {marionette_port_file}"
+        )
+        if os.path.isfile(port_file_copy):
+            os.unlink(port_file_copy)
+        os.rename(marionette_port_file, port_file_copy)
+        self._logger.info(
+            f"New Marionette MOVED OUT {marionette_port_file} TO {port_file_copy}"
+        )
```
lissyx, line 155: @gcp this was also required to fix what was due to a race during some of the crash behavior tests where we were picking up the port file too fast, at least on macOS

testing/enterprise/test_felt_browser_updates_forward.py
gcp: ```run_felt_click_browser_notification```

toolkit/xre/nsAppRunner.cpp
```diff
@@ -3109,7 +3108,21 @@ static nsresult SelectProfile(nsToolkitProfileService* aProfileSvc,
...
+    // sure the values are there and non empty to consider them as forced.
+    //
+    // AppShutdown::OnShutdownConfirmed() save them both so it is fine assuming
+    // they should be present both.
+    const char* xreProfilePath = PR_GetEnv("XRE_PROFILE_PATH");
+    auto savedProfile = xreProfilePath && *xreProfilePath;
+    const char* xreProfileLocalPath = PR_GetEnv("XRE_PROFILE_LOCAL_PATH");
+    auto savedLocalProfile = xreProfileLocalPath && *xreProfileLocalPath;
+
+    // Check if we are FELT UI, and nothing has given us a profile path, either
```
gcp: Would be clearer if it said what this is actually checking for, rather than what case it doesn't cover.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -134,7 +134,7 @@ export class FeltProcessParent extends JSProcessActorParent {
     this.abnormalExitFirstTime = 0;
 
     this.browserObserver = {
-      observe(aSubject, aTopic, aData) {
+      async observe(aSubject, aTopic, aData) {
```
gcp: I think we had a big discussion about this with @1rneh in another PR, that this is actually pretty dangerous because you have more chances to introduce races.

You can probably keep this sync.

testing/enterprise/test_felt_browser_updates_forward.py
```diff
@@ -9,18 +9,98 @@
...
+        self.install_update_manager_mock("applied")
+        self.run_felt_trigger_update()
+        self.run_felt_check_browser_notification()
+        # This is waiting on FELT to restart
+        self._logger.info(f"Waiting on FELT {felt_pid} restart")
+        self.wait_process_exit(felt_pid)
+        self._logger.info(f"Waited on FELT {felt_pid} restart: done")
+
+        # Verify the restarted process
+        self._driver.quit(in_app=False, clean=True)
```
lissyx: likely to remove

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -147,6 +147,24 @@ export class FeltProcessParent extends JSProcessActorParent {
               "enterprise.disable_restart",
               false
             );
+
+            const UM = Cc["@mozilla.org/updates/update-manager;1"].getService(
+              Ci.nsIUpdateManager
+            );
+            const readyUpdate = await UM.getReadyUpdate();
```
gcp: To keep it sync:
```
      UM.getReadyUpdate().then(readyUpdate => { ...
(existing code goes here)
      }).catch(err => {
        console.debug(
          `FeltExtension: ParentProcess: getReadyUpdate failed: ${err}`
        );
      });
      break;
```

testing/enterprise/test_felt_browser_updates_forward.py
```diff
@@ -9,18 +9,97 @@
...
+        self.run_felt_trigger_update()
+        self.run_felt_check_browser_notification()
+        # This is waiting on FELT to restart
+        self._logger.info(f"Waiting on FELT {felt_pid} restart")
+        self.wait_process_exit(felt_pid)
+        self._logger.info(f"Waited on FELT {felt_pid} restart: done")
+
+        # Verify the restarted process
+        self._driver.start_session()
+        self._logger.info("Reconnected to FELT")
```
gcp, line 78: This part isn't true unless you duplicate the new logic you also added to ```testing/enterprise/base_test.py```. Without it you're not connecting to the old, restarted Felt?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: i was not a big fan either, but i did not thought about the consequences

testing/enterprise/test_felt_browser_updates_forward.py
lissyx: this is the produced log:
```
 0:11.72 INFO Waiting on FELT 1776729 restart
 0:11.72 INFO Checking PID 1776729
 0:11.72 INFO PID 1776729 still exists
 0:11.72 INFO Found PID 1776729: STATUS:running
 0:12.72 INFO PID 1776729 still exists
 0:12.72 INFO Found PID 1776729: STATUS:zombie
 0:13.72 INFO Active waiting for PID 1776729 DONE => continue_checking:False iterations:2 psutil.pid_exists(pid_to_check):True
 0:13.72 INFO PID disappeared 1776729
 0:13.72 INFO All done for PID 1776729
 0:13.72 INFO Waited on FELT 1776729 restart: done
 0:13.74 INFO Reconnected to FELT
 0:13.74 INFO Reconnected to FELT with 1777828
```

So we reconnected. The logic in place in `base_test.py` is for the child browser that needs a different marionette port so we force the pref to 0 https://searchfox.org/enterprise-main/rev/73669e504b64c8ebe02576c85458600499166b03/testing/enterprise/felt_tests.py#170 to pick one and we have to read the marionette port file from the child browser's profile to know where to connect

With FELT we rely on the default 2828 port so we do not need to do that, the `start_session()` call will properly reconnect after seeing that the process it was tracking died.

This part is though indeed dependant on the changes in `toolkit/` because we rely on those to be able to recover the marionette's test run felt profile.

gcp: I'm not sure why this proves we reconnect rather than just connect (to a new one, which can be launched in the code by the `start_session` call). Can we assert there aren't two active Felts? I guess that would also require startup to log profile locks as one might be blocked by that? Messy. Have Firefox log the PID of what it launched (i.e. of the new restarted process)? Maybe easier, would confirm the behavior?

lissyx: the verification is rather manual unfortunately for now, I verified the newly launched process was:
 - using the profile we had marionette setup for us at startup of the test
 - no `-marionette` or other of the CLI args we got
 - using the correct console

We dont have a good way except checking that the process we connected to uses the profile created by marionette at first (so we know it's related to the felt we had), and that it has profile being set via the env variable we added support for, since when we restart we loose all CLI args and it is the only way for us to get back to the profile

I guess it's easy to poke for both over marionette and assert those values

toolkit/xre/nsAppRunner.cpp
lissyx:
```suggestion
+    // When running as FELT UI use a hardcoded profile named "felt" and living
+    // in the OS' temporary directory. There are a few special cases where it is
+    // required to use a profile that is being given from different means:
+    //  - CLI argument "-profile ..."
+    //  - environment variables "XRE_PROFILE_PATH" / "XRE_PROFILE_LOCAL_PATH"
+    //
+    // If either of those are present, the following code is skipped and profile
+    // selection continues. This accounts for manually setting the profile,
+    // which is required including when using "mach run" as well as when FELT
+    // restarts for updates.
```

I hope this is better?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -147,48 +147,84 @@ export class FeltProcessParent extends JSProcessActorParent {
...
-            if (gFeltProcessParentInstance?.proc) {
-              gFeltProcessParentInstance.restartReported = true;
-              gFeltProcessParentInstance.firefox = null;
-              console.debug(
-                `FeltExtension: ParentProcess: Killing Firefox PID=${gFeltProcessParentInstance.proc.pid}`
-              );
-              gFeltProcessParentInstance.proc
-                .kill()
-                .then(() => {
+            UM?.getReadyUpdate()
```
gcp:
```suggestion
            UM.getReadyUpdate()
```

gcp: It either works or throws.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -147,48 +147,84 @@ export class FeltProcessParent extends JSProcessActorParent {
...
-                });
-            } else {
-              console.debug(`FeltExtension: ParentProcess: No proc to kill!`);
-            }
+                }
+              })
+              .catch(err => {
+                console.debug(
+                  `FeltExtension: ParentProcess: getReadyUpdate failed: ${err}`
+                );
```
gcp: If we can't get the state, don't we want to make this equivalent to `!pendingUpdate` and get the same outcomes as https://github.com/mozilla/enterprise-firefox/pull/568/changes#diff-67e3a6ed03942edf9eaf6b59f4c7bb6eefd2cc7bf78503e8e5b0e3174dbf7c3fR187

So either `startFirefox(PROCESS_START_REASON.RESTART);` or `Services.cpmm.sendAsyncMessage("FeltParent:FirefoxNormalExit",{});`

gcp: Maybe 

```
                UM.getReadyUpdate())      
                .then(readyUpdate => {                                                                      
                  let pendingUpdate = false;                                                                
                  if (readyUpdate) {                  
                    const readyStates = [                                                                   
                      "pending",                      
                      "pending-service",                                                                  
                      "pending-elevate",                                                                    
                      "applied",                      
                      "applied-service",
                    ];                                                                                      
                    pendingUpdate = readyStates.includes(readyUpdate.state);                             
                  }                      
                  return pendingUpdate;
                })       
                .catch(err => {                       
                  console.debug(                      
                    `FeltExtension: ParentProcess: getReadyUpdate failed: ${err}, assuming no pending       
  update`                                                                                                   
                  );                                                                                        
                  return false;                       
                })                                                                                        
                .then(pendingUpdate => {                                      
```

So we just use the existing code.

gcp (approved):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Yes I think it would be wise

PR #569, opened 2026-03-19
Bug 2024613 - Closing Firefox logs out the user from Firefox Enterprise UX

jporter-dev:
## Summary

Builds upon #673 by replacing the simple single-checkbox prompt with a custom multi-checkbox prompt base on tab and pref logic.

- Adds `enterprise-close-dialog.xhtml` as custom prompt with multiple checkbox support. Modeled after `commonDialog`
- Adds custom localized signout text to the default prompt when quitting using shortcut (default true on MacOS)

## Links

- [Figma](https://www.figma.com/design/xYH6L8QrDz0Q16XFvdo7We/%F0%9F%A6%8A-Client-Design?node-id=1614-2058&p=f&m=dev)
- [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2024613)

- #674 
- #673 

## Screenshots

close-single-tab-both-checked

_Warn me about **both** - single tab_

close-multi-tab-both-checked

_Warn me about **both** - multiple tabs_

close-multi-tab-signOut-unchecked

_Warn me when **closing multiple tabs ONLY**_

close-multi-tab-tabs-unchecked

_Warn me about **signing out ONLY**_

ask-before-quitting

_Ask before quitting with shortcut (Cmd+Q)_

ask-before-quitting-multi-tab

_Ask before quitting with shortcut (Cmd+Q) with multiple tabs_

## Testing

Prefs that are toggled: 

- Signout warning: `enterprise.promptOnSignout`
- Multi tab warning: `browser.tabs.warnOnClose`
- Ask before quitting with shortcut: `browser.warnOnQuitShortcut`

1. `mach build`
2. `mach run`
3. Login with FELT
4. Verify all window close, signout paths, and pref combos display proper message and sign user out when closing
  - [ ] Close window with single tab open (both prefs checked)
  - [ ] Close window with single tab open (signout pref unchecked)
  - [ ] Close window with multiple tabs open (both prefs checked)
  - [ ] Close window with multiple tabs open (signout pref unchecked)
  - [ ] Close window with multiple tabs open (tab pref unchecked)
  - [ ] Close window with multiple tabs open (no prefs checked)
  - [ ] Signout using Enterprise Badge

### DIFF
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -137,6 +137,13 @@ ChromeUtils.defineLazyGetter(lazy, "gBrowserBundle", function () {
   );
 });
 
+ChromeUtils.defineLazyGetter(lazy, "enterpriseLocalization", () => {
+  return new Localization(
+    ["browser/enterprise/enterprise.ftl", "branding/brand.ftl"],
+    true
+  );
+});
+
 // Seconds of idle time before the late idle tasks will be scheduled.
 const LATE_TASKS_IDLE_TIME_SEC = 20;
 // Time after we stop tracking startup crashes.
@@ -1661,7 +1668,7 @@ BrowserGlue.prototype = {
       checkboxLabelId = "tabbrowser-ask-close-tabs-checkbox";
     }
 
-    const [title, quitButtonLabel, checkboxLabel] =
+    let [title, quitButtonLabel, checkboxLabel] =
       win.gBrowser.tabLocalization.formatMessagesSync([
         titleId,
         quitButtonLabelId,
@@ -1674,6 +1681,23 @@ BrowserGlue.prototype = {
       [closeTabButtonLabel] = win.gBrowser.tabLocalization.formatMessagesSync([
         closeTabButtonLabelId,
       ]);
+    }
+
+    let message = null;
+
+    if (AppConstants.MOZ_ENTERPRISE && shouldWarnForShortcut) {
+      const titleStringId = showCloseCurrentTabOption
+        ? "enterprise-quit-shortcut-prompt-title-with-tabs"
+        : "enterprise-quit-shortcut-prompt-title";
+      const [entTitle, entMessage, entQuitButtonLabel] =
+        lazy.enterpriseLocalization.formatValuesSync([
+          titleStringId,
+          "enterprise-quit-shortcut-prompt-message",
+          "enterprise-quit-shortcut-prompt-primary-btn-label",
+        ]);
+      title = { value: entTitle };
+      message = entMessage;
+      quitButtonLabel = { value: entQuitButtonLabel };
     }
 
     let warnOnClose = { value: true };
@@ -1700,7 +1724,7 @@ BrowserGlue.prototype = {
     let buttonPressed = Services.prompt.confirmEx(
       win,
       title.value,
-      null,
+      message,
       flags,
       quitButtonLabel.value,
       null,
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -1564,7 +1564,6 @@ BrowserGlue.prototype = {
         aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
         this._quitSource = "unknown";
       }
-      return;
     }
 
     // browser.warnOnQuit is a hidden global boolean to override all quit prompts.
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -137,13 +137,6 @@ ChromeUtils.defineLazyGetter(lazy, "gBrowserBundle", function () {
   );
 });
 
-ChromeUtils.defineLazyGetter(lazy, "enterpriseLocalization", () => {
-  return new Localization(
-    ["browser/enterprise/enterprise.ftl", "branding/brand.ftl"],
-    true
-  );
-});
-
 // Seconds of idle time before the late idle tasks will be scheduled.
 const LATE_TASKS_IDLE_TIME_SEC = 20;
 // Time after we stop tracking startup crashes.
@@ -1566,6 +1559,7 @@ BrowserGlue.prototype = {
       }
       if (lazy.EnterpriseHandler.shouldShowClosePrompt()) {
         aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
+        this._quitSource = "unknown";
         const promptWindow = lazy.BrowserWindowTracker.getTopWindow({
           allowFromInactiveWorkspace: true,
         });
@@ -1692,15 +1686,19 @@ BrowserGlue.prototype = {
     let message = null;
 
     if (AppConstants.MOZ_ENTERPRISE && shouldWarnForShortcut) {
-      const [entTitle, entMessage, entQuitButtonLabel] =
-        lazy.enterpriseLocalization.formatValuesSync([
-          {
-            id: "enterprise-quit-shortcut-prompt-title",
-            args: { withTabs: showCloseCurrentTabOption ? "true" : "false" },
-          },
-          "enterprise-quit-shortcut-prompt-message",
-          "enterprise-quit-shortcut-prompt-primary-btn-label",
-        ]);
+      const l10n = new Localization(
+        ["browser/enterprise/enterprise.ftl", "branding/brand.ftl"],
+        true
+      );
+
+      const [entTitle, entMessage, entQuitButtonLabel] = l10n.formatValuesSync([
+        {
+          id: "enterprise-quit-shortcut-prompt-title",
+          args: { withTabs: showCloseCurrentTabOption ? "true" : "false" },
+        },
+        "enterprise-quit-shortcut-prompt-message",
+        "enterprise-quit-shortcut-prompt-primary-btn-label",
+      ]);
       title = { value: entTitle };
       message = entMessage;
       quitButtonLabel = { value: entQuitButtonLabel };
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -1559,17 +1559,26 @@ BrowserGlue.prototype = {
 
     // When Firefox was launched by FELT, show a signout confirmation prompt
     // instead of the standard quit dialog.
-    if (
-      AppConstants.MOZ_ENTERPRISE &&
-      Services.felt?.isFeltBrowser() &&
-      lazy.EnterpriseHandler.isSignoutPromptEnabled()
-    ) {
-      const promptWindow = lazy.BrowserWindowTracker.getTopWindow({
-        allowFromInactiveWorkspace: true,
-      });
-      if (!lazy.EnterpriseHandler.showSignoutPrompt(promptWindow)) {
+    if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltBrowser()) {
+      if (lazy.EnterpriseHandler._skipSignoutPrompt) {
+        lazy.EnterpriseHandler._skipSignoutPrompt = false;
+        return;
+      }
+      if (
+        Services.prefs.getBoolPref("enterprise.prompt_on_signout", true) ||
+        Services.prefs.getBoolPref("browser.tabs.warnOnClose", false)
+      ) {
         aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
-        this._quitSource = "unknown";
+        const promptWindow = lazy.BrowserWindowTracker.getTopWindow({
+          allowFromInactiveWorkspace: true,
+        });
+        lazy.EnterpriseHandler.showSignoutPrompt(promptWindow).then(proceed => {
+          if (proceed) {
+            lazy.EnterpriseHandler._skipSignoutPrompt = true;
+            Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
+          }
+        });
+        return;
       }
     }
 
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -1564,21 +1564,7 @@ BrowserGlue.prototype = {
         lazy.EnterpriseHandler._skipSignoutPrompt = false;
         return;
       }
-      const warnOnSignout = Services.prefs.getBoolPref(
-        "enterprise.prompt_on_signout",
-        true
-      );
-      const warnOnCloseWithTabs = Services.prefs.getBoolPref(
-        "browser.tabs.warnOnClose",
-        false
-      );
-      let tabCount = 0;
-      for (let win of lazy.BrowserWindowTracker.orderedWindows) {
-        if (!win.closed && win.gBrowser) {
-          tabCount += win.gBrowser.openTabs.length;
-        }
-      }
-      if (warnOnSignout || (tabCount > 1 && warnOnCloseWithTabs)) {
+      if (lazy.EnterpriseHandler.shouldShowClosePrompt()) {
         aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
         const promptWindow = lazy.BrowserWindowTracker.getTopWindow({
           allowFromInactiveWorkspace: true,
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -125,6 +125,15 @@ if (AppConstants.MOZ_CRASHREPORTER) {
   });
 }
 
+if (AppConstants.MOZ_ENTERPRISE) {
+  ChromeUtils.defineLazyGetter(lazy, "localization", () => {
+    return new Localization(
+      ["browser/enterprise/enterprise.ftl", "branding/brand.ftl"],
+      true
+    );
+  });
+}
+
 ChromeUtils.defineLazyGetter(lazy, "gBrandBundle", function () {
   return Services.strings.createBundle(
     "chrome://branding/locale/brand.properties"
@@ -1559,11 +1568,16 @@ BrowserGlue.prototype = {
         const promptWindow = lazy.BrowserWindowTracker.getTopWindow({
           allowFromInactiveWorkspace: true,
         });
-        lazy.EnterpriseHandler.showSignoutPrompt(promptWindow).then(proceed => {
-          if (proceed) {
-            Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
-          }
-        });
+        lazy.EnterpriseHandler.showSignoutPrompt(promptWindow)
+          .then(proceed => {
+            if (proceed) {
+              Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
+            }
+          })
+          .catch(e => {
+            console.error("Enterprise signout prompt failed, quitting:", e);
+            Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
+          });
         return;
       }
     }
@@ -1681,19 +1695,15 @@ BrowserGlue.prototype = {
     let message = null;
 
     if (AppConstants.MOZ_ENTERPRISE && shouldWarnForShortcut) {
-      const l10n = new Localization(
-        ["browser/enterprise/enterprise.ftl", "branding/brand.ftl"],
-        true
-      );
-
-      const [entTitle, entMessage, entQuitButtonLabel] = l10n.formatValuesSync([
-        {
-          id: "enterprise-quit-shortcut-prompt-title",
-          args: { withTabs: showCloseCurrentTabOption ? "true" : "false" },
-        },
-        "enterprise-quit-shortcut-prompt-message",
-        "enterprise-quit-shortcut-prompt-primary-btn-label",
-      ]);
+      const [entTitle, entMessage, entQuitButtonLabel] =
+        lazy.localization.formatValuesSync([
+          {
+            id: "enterprise-quit-shortcut-prompt-title",
+            args: { withTabs: showCloseCurrentTabOption ? "true" : "false" },
+          },
+          "enterprise-quit-shortcut-prompt-message",
+          "enterprise-quit-shortcut-prompt-primary-btn-label",
+        ]);
       title = { value: entTitle };
       message = entMessage;
       quitButtonLabel = { value: entQuitButtonLabel };
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -1553,10 +1553,6 @@ BrowserGlue.prototype = {
     // When Firefox was launched by FELT, show a signout confirmation prompt
     // instead of the standard quit dialog.
     if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltBrowser()) {
-      if (lazy.EnterpriseHandler._skipSignoutPrompt) {
-        lazy.EnterpriseHandler._skipSignoutPrompt = false;
-        return;
-      }
       if (lazy.EnterpriseHandler.shouldShowClosePrompt()) {
         aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
         this._quitSource = "unknown";
@@ -1565,7 +1561,6 @@ BrowserGlue.prototype = {
         });
         lazy.EnterpriseHandler.showSignoutPrompt(promptWindow).then(proceed => {
           if (proceed) {
-            lazy.EnterpriseHandler._skipSignoutPrompt = true;
             Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
           }
         });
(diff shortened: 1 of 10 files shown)


testing/enterprise/test_felt_browser_close_warning.py
```diff
@@ -0,0 +1,247 @@
...
+from base_test import Environment
+from felt_tests import FeltTests
+
+PREF_PROMPT_ON_SIGNOUT = "enterprise.prompt_on_signout"
+PREF_WARN_ON_CLOSE = "browser.tabs.warnOnClose"
+
+
+class BrowserCloseWarning(FeltTests):
+    def _set_child_bool_pref(self, pref_name, value):
+        self._child_driver.set_context("chrome")
```
victhorlopez, line 21: Should we make somehow https://searchfox.org/enterprise-main/rev/74d9e8ce44e87b187286e3bae4ac60962218a497/testing/enterprise/felt_tests.py#753 reusable also for child? (similar get get_child_element) or overkill here?

testing/enterprise/test_felt_browser_close_warning.py
```diff
@@ -0,0 +1,247 @@
...
+                Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
+            }
+            """
+        )
+        self._manually_closed_child = True
+
+    def _assert_close_dialog_content(self, expected_title, expected_message):
+        self._logger.info("Waiting for the custom signout dialog to assert its content")
+        self._child_wait.until(
+            lambda _: self._child_driver.execute_script(
```
victhorlopez, line 73: Would some refactor like this work? or overkill?

```
      def _wait_for_close_dialog(self):
        """Wait for the enterprise close dialog to be ready and store a
        reference to it on the chrome window for follow-up scripts."""
        self._child_wait.until(lambda _: self._child_driver.execute_script("""
            try {
                const dialog = document.getElementById('window-modal-dialog')
                    ?.querySelector('.dialogFrame')
                    ?.contentDocument
                    ?.getElementById('enterpriseCloseDialog');
                if (!dialog?.getButton('accept')) return null;
                window.__enterpriseCloseDialog = dialog;
                return true;
            } catch (e) { return null; }
        """))

    def _get_close_dialog_content(self):
        self._wait_for_close_dialog()
        return self._child_driver.execute_script("""
            const doc = window.__enterpriseCloseDialog.ownerDocument;
            return {
                title: doc.getElementById('infoTitle').textContent,
                message: doc.getElementById('infoBody').textContent,
            };
        """)

    def _accept_close_dialog(self):
        self._logger.info("Accepting the enterprise close dialog")
        self._wait_for_close_dialog()
        self._child_driver.execute_script("""
            window.__enterpriseCloseDialog.getButton('accept').click();
            window.__enterpriseCloseDialog = null;
        """)
```

browser/components/enterprise/content/enterprise-close-dialog.css
```diff
@@ -0,0 +1,43 @@
...
+}
+
+#checkboxesList {
+  display: flex;
+  flex-direction: column;
+  gap: var(--space-large);
+}
+
+#checkboxesList .checkbox-label {
+  white-space: nowrap;
```
victhorlopez: Claude has some concerns in here, not sure if it is legit.
"localized strings can be significantly longer than English and would overflow horizontally with no max-width constraint"

browser/components/BrowserGlue.sys.mjs
```diff
@@ -137,6 +137,13 @@ ChromeUtils.defineLazyGetter(lazy, "gBrowserBundle", function () {
   );
 });
 
+ChromeUtils.defineLazyGetter(lazy, "enterpriseLocalization", () => {
```
victhorlopez: should this be surrounded by AppConstants.MOZ_ENTERPRISE ? (i know it is only instantiated when you call it, but not sure what is the pattern to follow in these cases)

testing/enterprise/test_felt_browser_close_warning.py
```diff
@@ -0,0 +1,247 @@
...
+
+        self.assert_child_browser_closed()
+
+    def test_browser_window_close_no_warnings(self):
+        """Sign-out warn off - no dialog, quit proceeds directly."""
+        super().run_felt_base()
+        self.connect_child_browser()
+        self.assert_user_signed_in(env=Environment.FIREFOX)
+
+        self._set_child_bool_pref(PREF_PROMPT_ON_SIGNOUT, False)
```
victhorlopez, line 246: Make sure you don't rely on defaults cause I have a change to use these prefs ( same as all marionette tests do on nonenterprise), i.e not sure you need to set here PREF_WARN_ON_CLOSE
https://searchfox.org/firefox-main/rev/8332a06d47ce7d66623d807068b3410061cd29d3/testing/marionette/client/marionette_driver/geckoinstance.py#666

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1552,19 +1559,24 @@ BrowserGlue.prototype = {
...
-      });
-      if (!lazy.EnterpriseHandler.showSignoutPrompt(promptWindow)) {
+    if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltBrowser()) {
+      if (lazy.EnterpriseHandler._skipSignoutPrompt) {
+        lazy.EnterpriseHandler._skipSignoutPrompt = false;
+        return;
+      }
+      if (lazy.EnterpriseHandler.shouldShowClosePrompt()) {
         aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
-        this._quitSource = "unknown";
```
victhorlopez, line 1565: was this removal on purpose? this._quitSource = "unknown";

jporter-dev: good catch thanks, that was an accident. added this in the other PR and missed this when fixing a conflict

browser/components/BrowserGlue.sys.mjs
jporter-dev: I think I can load this inline down below where it's used rather than this lazyGetter, then it'll be within the `AppConstants.MOZ_ENTERPRISE` check

```javascript
// on line #1695
const l10n = new Localization(
  ["browser/enterprise/enterprise.ftl", "branding/brand.ftl"],
  true
);
```

testing/enterprise/test_felt_browser_close_warning.py
jporter-dev: I can update to set the necessary prefs explicitly avoid relying on defaults. I'll also look into your other comments in the tests to DRY them up. Thanks for pointing that out!

browser/components/enterprise/content/enterprise-close-dialog.css
jporter-dev: I'll just go ahead and remove this rule, it doesn't cause any overflow but it does cause a longer string to widen the parent element. Removing it will let this label wrap naturally for longer localized strings and keep our dialog width consistent.

testing/enterprise/test_felt_browser_close_warning.py
jporter-dev: I don't think I'd want to overload that one, but perhaps `felt_tests.py` is a better place for this. Since it already has a `get_pref_child` function it makes sense for the `set_bool_pref_child` to be defined there.

testing/enterprise/test_felt_browser_close_warning.py
jporter-dev: I went ahead and refactored to explicitly set the prefs and to remove some repetition within the tests, but not sure I want to implement this one. I'd rather avoid storing test state in the browser chrome, but to remove some more repetition here I could move the dialog selector into a constant?

victhorlopez (approved):

victhorlopez:
I approved but I lack a lot of knowledge, I think it makes sense that @1rneh has a look

testing/enterprise/test_felt_browser_close_warning.py
```diff
@@ -0,0 +1,250 @@
...
+    def test_browser_window_close_signout_warning_only(self):
+        """Sign-out warn on, tabs warn off, single tab - enterprise signout dialog shows."""
+        super().run_felt_base()
+        self.connect_child_browser()
+        self.assert_user_signed_in(env=Environment.FIREFOX)
+
+        self._set_child_bool_pref(PREF_PROMPT_ON_SIGNOUT, True)
+        self._set_child_bool_pref(PREF_WARN_ON_CLOSE, False)
+
+        self._close_browser()
```
1rneh: `nit:` Can we rename this to method to `_trigger_browser_closure` as this might not complete the closure but only trigger the prompt.

testing/enterprise/test_felt_browser_close_warning.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+        self._logger.info("Clicking accept on standalone enterprise close dialog")
+        self._child_driver.execute_script(
+            "document.getElementById('enterpriseCloseDialog').getButton('accept').click();"
+        )
+
+        self._child_wait.until(
+            lambda _: set(self._child_driver.chrome_window_handles) == initial_handles
+        )
+        self._child_driver.switch_to_window(list(initial_handles)[0])
```
1rneh: Could you elaborate what we are doing here? 

As far as I understand we are waiting to have the same set of chrome windows again as before (so effectively that the standalone dialog has closed). Then we switch back to the first chrome window? But at the same time we are also initiating a shutdown, so can this race? And should we even have a chrome window at this point as I have assumed we are window-less?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -109,6 +110,12 @@ export const EnterpriseHandler = {
    */
   _isInitialized: false,
 
+  /**
+   * Set to true after the user confirms the enterprise close dialog, so that the
+   * resulting re-quit skips showing it again.
+   */
+  _skipSignoutPrompt: false,
```
1rneh, line 117: Should we remove the `_` prefix since this variable is read and set by `BrowserGlue`, so it isn’t really private?

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1552,19 +1552,25 @@ BrowserGlue.prototype = {
...
-    ) {
-      const promptWindow = lazy.BrowserWindowTracker.getTopWindow({
-        allowFromInactiveWorkspace: true,
-      });
-      if (!lazy.EnterpriseHandler.showSignoutPrompt(promptWindow)) {
+    if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltBrowser()) {
+      if (lazy.EnterpriseHandler._skipSignoutPrompt) {
+        lazy.EnterpriseHandler._skipSignoutPrompt = false;
+        return;
+      }
```
1rneh: Can you elaborate on why you have added this variable? Did we get multiple notifications?

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -15,16 +15,44 @@ enterprise-panel-learn-more = Learn more
...
-        [0] Sign out of { -brand-short-name }?
-        [one] Sign out of { -brand-short-name }?
-       *[other] Sign out and close { $tabCount } tabs?
+        [0] Close { -brand-short-name }?
+       *[other]
+            { $warnSignout ->
+                [true] Close { -brand-short-name } and { $tabCount } tabs?
+               *[false] Close { $tabCount } tabs?
+            }
     }
```
1rneh: Could we split this up into two fluent strings, e.g. `enterprise-close-prompt-title` and `enterprise-close-prompt-title-with-tabcount` as this would make the JS side a bit clearer.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -312,125 +319,166 @@ export const EnterpriseHandler = {
...
+  _getSignoutPromptParams({
+    tabCount,
+    warnOnSignout,
+    warnOnCloseWithTabs,
+  } = {}) {
+    const hasMultipleTabs = tabCount > 1;
+    const hasTabsWarning = hasMultipleTabs && warnOnCloseWithTabs;
 
+    const effectiveTabCount = hasTabsWarning ? tabCount : 0;
     const titleId = {
```
1rneh: After splitting the fluent string into two, we can distinguish between the two depending on the value of `hasTabsWarning` (see my comment in `enterprise.ftl`)

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -15,16 +15,44 @@ enterprise-panel-learn-more = Learn more
...
+#   $warnSignout (String): "true" when the sign-out warning is shown.
+enterprise-close-prompt-message =
+    { $tabCount ->
+        [0] You’re about to sign out of { -brand-short-name } and end your session.
+       *[other]
+            { $warnSignout ->
+                [true] You’re about to sign out of { -brand-short-name } and close { $tabCount } tabs.
+               *[false] Closing { -brand-short-name } will also sign you out.
+            }
+    }
```
1rneh: Same goes for this one. Can we split this into two separate fluent strings, e.g. `enterprise-close-prompt-message` and `enterprise-close-prompt-message-with-tabcount`

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -312,125 +319,166 @@ export const EnterpriseHandler = {
...
+      messageId,
+      { id: "enterprise-close-prompt-primary-btn-label" },
+      { id: "enterprise-close-prompt-message-reauth" },
+      { id: "enterprise-close-prompt-checkbox-label" },
+      { id: "enterprise-close-prompt-tabs-checkbox-label" },
+    ]);
+
+    const message = warnOnSignout
+      ? `${messageBody}\n\n${reauthNotice}`
+      : messageBody;
```
1rneh: I'm not feeling so confident about including line breaks in a string as formatting. The separation between `messageBody` and `reauthNotice` is semantic.

Can we split them into two separate <xul:description> elements, `reauthNotice` can be conditionally included in the DOM rather than conditionally concatenated in JS and then the spacing can be controlled in css?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -312,125 +319,166 @@ export const EnterpriseHandler = {
...
 
-    return {
+    const [
+      title,
+      messageBody,
+      acceptLabel,
+      reauthNotice,
+      checkLabel,
+      tabsCheckLabel,
+    ] = lazy.localization.formatValuesSync([
```
1rneh: If we discard the requirement to have a synchronous prompt call stack, then we can also revert to getting the translations asynchronously with `formatValues`.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -312,125 +319,166 @@ export const EnterpriseHandler = {
...
+        "_blank",
+        "chrome,centerscreen,modal,dialog",
+        params
+      );
+    } else if (!window.gDialogBox.isOpen) {
+      await window.gDialogBox.open(
+        "chrome://browser/content/enterprise/enterprise-close-dialog.xhtml",
+        params
+      );
+    }
```
1rneh: If there is another dialog open, we will want to replace it:
```suggestion
    } else {
      if (window.gDialogBox.isOpen) {
        window.gDialogBox.replaceDialogIfOpen();
      }
      await window.gDialogBox.open(
        "chrome://browser/content/enterprise/enterprise-close-dialog.xhtml",
        params
      );
    }
```

1rneh (changes requested):
First of all, thank you for having such a fixed-up PR. 🚀  
Despite the complexity in this PR, it really helped me while going through the code changes! 

I have one main question: We’re converting the synchronous sign-out prompt back to asynchronous. I assume this is because the synchronous path is somewhat limited for our needs - could you share a bit about that?

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Yeah I can do this. Thanks for spotting this, accidentally left this in after swapping from the confirmEx prompt to the custom dialog.

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Will add the lines to replace the dialog! But I also want to point out that this line here is the reason we have to be async unfortunately 😔. I could not see any way to make this synchronous, and the few ways I tried resulted in the wrong style dialog popup. If you know of a way to make this sync then I would love to make this synchronous again!

browser/locales/en-US/browser/enterprise/enterprise.ftl
jporter-dev: I've split these up, I combined them previously but it makes more sense to handle it in JS than fluent

browser/components/BrowserGlue.sys.mjs
jporter-dev: This is to skip the second prompt when re-triggering shutdown due to making the signout prompt async since we can't block normally

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: I refactored this to make all use of this var happen in `EnterpriseHandler` so it is truly private

testing/enterprise/test_felt_browser_close_warning.py
jporter-dev: This was a bad/wrong implementation of this test scenario, I've updated it to pass `close_all_windows` to `_trigger_browser_closure` to properly test this case by closing all browser windows before triggering closure.

jporter-dev:
@1rneh re-requesting review for everything but the async changes. Explained a bit [here](https://github.com/mozilla/enterprise-firefox/pull/569#discussion_r3102471951) that I could not find a way to show the dialog in the way that we want without using the async call. I could have overlooked it though, or not used the right feature configuration to get the in-browser dialog with the card styling. The few days I could find ended up with the system prompt style browserless popup.

testing/enterprise/test_felt_browser_close_warning.py
```diff
@@ -0,0 +1,255 @@
...
+                    new Promise(r => win.addEventListener('unload', r, { once: true }))
+                );
+                wins.forEach(win => win.close());
+                Promise.all(closed).then(triggerQuit);
+            } else {
+                triggerQuit();
+            }
+            """,
+            [close_all_windows],
+        )
```
1rneh: This is script is currently not awaiting all window closure promises. It needs to be executed asynchronously with `execute_async_script`:
https://firefox-source-docs.mozilla.org/python/marionette_driver.html#marionette_driver.marionette.Marionette.execute_async_script

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1552,19 +1552,20 @@ BrowserGlue.prototype = {
...
         aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
         this._quitSource = "unknown";
+        const promptWindow = lazy.BrowserWindowTracker.getTopWindow({
+          allowFromInactiveWorkspace: true,
+        });
+        lazy.EnterpriseHandler.showSignoutPrompt(promptWindow).then(proceed => {
+          if (proceed) {
+            Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
+          }
+        });
```
1rneh: Can we add a `.catch` block so that this doesn't fail silently? Also do we want to quit on failure to avoid keeping the user in a browser that they can't exit?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -312,125 +319,171 @@ export const EnterpriseHandler = {
...
+        Services.prefs.setBoolPref(WARN_ON_CLOSE_PREF, checked);
+      }
     }
+
     return true;
   },
 
-  isSignoutPromptEnabled() {
-    return Services.prefs.getBoolPref(PROMPT_ON_SIGNOUT_PREF, true);
+  _countOpenTabs() {
```
1rneh, line 422: We're currently counting the tabs twice, once `shouldShowClosePrompt` and again in `showSignoutPrompt`. Can we capture the result and re-use it in `shouldShowClosePrompt`?

browser/components/enterprise/content/enterprise-close-dialog.js
```diff
@@ -0,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const XUL_NS = "http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul";
+
+let gArgs;
```
1rneh:
```suggestion
const gArgs;
```

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1677,6 +1678,27 @@ BrowserGlue.prototype = {
       ]);
     }
 
+    let message = null;
+
+    if (AppConstants.MOZ_ENTERPRISE && shouldWarnForShortcut) {
+      const l10n = new Localization(
+        ["browser/enterprise/enterprise.ftl", "branding/brand.ftl"],
+        true
+      );
```
1rneh: This creates a new instance on each quit shortcut. Can we load this lazily?

1rneh (changes requested):
If we fall short on a synchronous prompting path, then converting to asynchronous works for me.

On a side note: Could we have JSDoc coverage for any newly added?

browser/components/BrowserGlue.sys.mjs
jporter-dev: This makes sense to add these defensive measures. For quitting on failure, I guess that should be an `eForceQuit`?

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Done, cached the response and clears it when done

testing/enterprise/test_felt_browser_close_warning.py
jporter-dev: Updated this to use `execute_async_script`, thank you for showing me that

browser/components/enterprise/content/enterprise-close-dialog.js
```diff
@@ -0,0 +1,80 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const XUL_NS = "http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul";
+
+/**
+ * Initializes the enterprise close dialog, called on DOMContentLoaded.
+ */
+function onLoad() {
+  const gArgs = window.arguments[0].wrappedJSObject ?? window.arguments[0];
```
1rneh: `nit:` Can we move `const gArgs = {}` into the global scope where it was before?

browser/components/BrowserGlue.sys.mjs
1rneh: Do we have an understanding on when to pass which flag? I can't recall a conversation about this.

This is my assumption:
- Explicitly signing out via button passes flag `eForceQuit`
- Quitting with no open windows passes flag `eForceQuit`
- Quitting with open windows passes flag `eAttemptQuit`

What do you think?

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: This still needs to be addressed :)

browser/components/enterprise/content/enterprise-close-dialog.xhtml
```diff
@@ -0,0 +1,50 @@
...
+<?csp default-src chrome:; ?>
+
+<!DOCTYPE window>
+
+<window
+  id="enterpriseCloseDialogWindow"
+  xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+  xmlns:html="http://www.w3.org/1999/xhtml"
+  xmlns:xul="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+  aria-describedby="infoBody"
```
1rneh: Since we split infoBody into two `<description>` elements.
```suggestion
  aria-describedby="infoBody infoReauth"
```

1rneh (commented):
I just have a general question about which flags to pass to `Services.startup.quit` and a few smaller requests :)

browser/components/enterprise/content/enterprise-close-dialog.js
jporter-dev: I did this to address the previous request to make it a `const`. If we move it back into global, then we have to change it back to `let gArgs` since we are setting it here in `onLoad` since we can't reassign a `const`.

I can revert this back if that's the preferred pattern 🙂

jporter-dev: looks like i can just initialize it at the global scope. updated to do that and revert the other changes

1rneh (approved):
Looking good now, thank you!

PR #570, opened 2026-03-19
Bug 2023486 - Wait for elevationOptedIn before notifying Felt

gcp:
`elevationOptedIn` is async, the earlier changes using it in #547 did get that correct.


lissyx (approved):
My mistake , there was an await at first and forgot to chain the promise when changing it

PR #574, opened 2026-03-19
Bug 2002463 - Focus Firefox window when opening external links

gcp:
Of course Linux has to be very special and guard this with tokens, because...I don't know. Can't we just fix the software that steals focus there???

### DIFF
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -136,6 +136,7 @@ this.felt = class extends ExtensionAPI {
 
       try {
         win.openTrustedLinkIn(url, "tab");
+        win.focus();
       } catch (err) {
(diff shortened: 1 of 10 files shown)


gcp:
Bug wasn't reported against Windows, perhaps because of [allowsetforegroundwindow docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-allowsetforegroundwindow
) illustrating Windows is a sane desktop OS that understands process trees:

_Additionally, at least one of the following conditions is true:_ 
_...The calling process was started by the foreground process._

lissyx:
> Of course Linux has to be very special and guard this with tokens, because...I don't know. Can't we just fix the software that steals focus there???

wayland?

lissyx (approved):

PR #575, opened 2026-03-20
Bug 2012082 - Enterprise: Add custom enterprise logo on toolbar

jporter-dev:
## Summary

Adds support for loading the Enterprise Badge logo from `enterprise.logoUrl` using the following URL pattern: `^(https:\/\/|data:image\/(?:png|jpeg|gif|webp|svg\+xml);base64,)`

Also fixes spacing issue with Enterprise Badge on smaller displays. 

Open to trimming the icon format types down as well, for now it's just web raster formats + svg.

## Links

- [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2012082)

## Screenshots

badge-logo-resource

_Custom Enterprise Badge logo_

badge-logo-overlap

_Enterprise Badge overlapping neighboring elements (before fix)_

## Testing

1. `mach build`
2. `mach run`
3. Set `enterprise.logoUrl` to a valid `https` or `data:base64` url
4. Close and reopen or open new window

### DIFF
--- a/browser/components/enterprise/EnterpriseHandler.sys.mjs
+++ b/browser/components/enterprise/EnterpriseHandler.sys.mjs
@@ -245,28 +245,33 @@ export const EnterpriseHandler = {
     const logoUrl = Services.prefs.getStringPref(LOGO_URL, "");
 
     if (!logoUrl) {
-      console.error(`Skipping logo update, ${LOGO_URL} pref is not set.`);
+      console.warn(`${LOGO_URL} pref is not set, skipping logo update`);
       return;
     }
 
-    // allow https: URLs and data: URIs for common image formats with base64 encoding only.
-    const logoUrlPattern =
-      /^(https:\/\/|data:image\/(?:png|jpeg|gif|webp|svg\+xml);base64,)/;
-    if (!logoUrlPattern.test(logoUrl)) {
-      console.error(`Invalid logo URL in pref: ${logoUrl}`);
-      return;
-    }
-
+    let validLogoUrl;
     try {
-      const validLogoUrl = new URL(logoUrl).href;
+      validLogoUrl = new URL(logoUrl);
+    } catch {
+      throw new Error(`Invalid logo URL in pref: ${logoUrl}`);
+    }
 
-      const toolbarLogo = window.document.querySelector(
-        "#enterprise-company-logo__wrapper > image"
-      );
-      toolbarLogo.style.setProperty(
-        "list-style-image",
-        `url("${validLogoUrl}")`
-      );
-    } catch {}
+    if (validLogoUrl.protocol === "https:") {
+      if (validLogoUrl.origin !== lazy.ConsoleClient.consoleBaseURI.origin) {
+        throw new Error(`Logo URL must be hosted from the console: ${logoUrl}`);
+      }
+    } else if (
+      !/^data:image\/(?:png|jpeg|gif|webp|svg\+xml);base64,/.test(logoUrl)
+    ) {
+      throw new Error(`Invalid logo URL in pref: ${logoUrl}`);
+    }
+
+    const toolbarLogo = window.document.querySelector(
+      "#enterprise-company-logo__wrapper > image"
+    );
+    toolbarLogo.style.setProperty(
+      "list-style-image",
+      `url("${validLogoUrl.href}")`
+    );
   },
 };
--- a/browser/components/enterprise/EnterpriseHandler.sys.mjs
+++ b/browser/components/enterprise/EnterpriseHandler.sys.mjs
@@ -27,26 +27,6 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
 const PROMPT_ON_SIGNOUT_PREF = "enterprise.promptOnSignout";
 const ENTERPRISE_TOOLBAR_LOGO_URL_PREF = "enterprise.toolbar.logoUrl";
 
-/**
- * Validate and normalize logo URL
- *
- * @param {*} logoUrl
- */
-function _resolveLogoUrl(logoUrl) {
-  // allow https: URLs and data: URIs for common image formats with base64 encoding only.
-  const logoUrlPattern =
-    /^(https:\/\/|data:image\/(?:png|jpeg|gif|webp|svg\+xml);base64,)/;
-  if (!logoUrlPattern.test(logoUrl)) {
-    return null;
-  }
-
-  try {
-    return new URL(logoUrl).href;
-  } catch {
-    return null;
-  }
-}
-
 export const EnterpriseHandler = {
   /**
    * @type {{name:string, email:string, pictureUrl:string} | null}
@@ -126,17 +106,10 @@ export const EnterpriseHandler = {
     const toolbarLogo = window.document.querySelector(
       "#enterprise-company-logo__wrapper > image"
     );
-    const logoUrl = Services.prefs.getStringPref(
-      ENTERPRISE_TOOLBAR_LOGO_URL_PREF,
-      ""
-    );
-    const validLogoUrl = logoUrl && _resolveLogoUrl(logoUrl);
+    const logoUrl = this._getLogoUrl();
 
-    if (validLogoUrl) {
-      toolbarLogo.style.setProperty(
-        "list-style-image",
-        `url("${validLogoUrl}")`
-      );
+    if (logoUrl) {
+      toolbarLogo.style.setProperty("list-style-image", `url("${logoUrl}")`);
     }
 
     if (!this._signedInUser) {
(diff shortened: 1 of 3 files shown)


browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -25,6 +25,27 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
 });
 
 const PROMPT_ON_SIGNOUT_PREF = "enterprise.promptOnSignout";
+const ENTERPRISE_TOOLBAR_LOGO_URL_PREF = "enterprise.toolbar.logoUrl";
```
lissyx: So @mkaply tells me that anything under `distribution/` on our repacks should gets pushed into the final binary. Can we leverage this ?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -211,6 +227,64 @@ export const EnterpriseHandler = {
 
   uninit() {
     this._signedInUser = {};
+    this._logoUrl = null;
     this._isInitialized = false;
   },
+
+  _setLogoUrlFromPref() {
```
lissyx: Why do we keep it if it works from distribution dir ?

jporter-dev: I only kept it for now since there was talk of the base64 or URL version. I'll remove if the distribution dir is the solution we're going with.

lissyx: My understanding is that we suggested url via pref at first because we did not knew it may work with the repack adding logo files, but if it works and others considers it is viable there is no reason it would not be the only solution we need.

browser/components/enterprise/content/enterprise-badge.css
```diff
@@ -19,6 +19,8 @@
 }
 
 #enterprise-badge-toolbar-button {
+  flex-shrink: 0;
+
```
1rneh, line 23: Nice catch! Thank you!

1rneh (commented):
This approach is new to me. Adding a file to the local distribution directory does not verify that it works correctly with the repack scripts in CI. Additionally, requiring a repack for each company logo update introduces unnecessary overhead, especially since we planned to make this configurable via the console.

lissyx:
> This approach is new to me. Adding a file to the local distribution directory does not verify that it works correctly with the repack scripts in CI. Additionally, requiring a repack for each company logo update introduces unnecessary overhead, especially since we planned to make this configurable via the console.

I suggested it, hacking a local file felt much better than just a random pref value, hitting the network and/or handling base64 data URL. I was not aware we wanted to make the URL configurable via the console.

1rneh:
> I suggested it, hacking a local file felt much better than just a random pref value, hitting the network and/or handling base64 data URL. I was not aware we wanted to make the URL configurable via the console.

We are adding a `/config` endpoint in the console that will provide all configurations for Firefox, which will include the asset url.

jporter-dev:
@1rneh I've reverted it back to the code to read the URL from a pref. I also updated it to pull the pref from `ConsoleClient.sys.mjs` so it will share the pref that is loaded when `/config` is read. For now I used `enterprise.logoUrl` but can update that if there is a pref in mind for this or if we wanted to namespace it specific to the toolbar/badge (ex: `enterprise.toolbar.logoUrl`).

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -23,6 +23,7 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
  */
 export const PREFS = {
   CONSOLE_ADDRESS: "enterprise.console.address",
+  LOGO_URL: "enterprise.logoUrl",
```
1rneh: Let's move this to the `EnterpriseHandler`. `ConsoleClient` really only needs to know of the console address pref.

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: The JSDocs needs updating.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -227,64 +214,31 @@ export const EnterpriseHandler = {
...
     this._isInitialized = false;
   },
 
-  _setLogoUrlFromPref() {
-    const prefLogoUrl = Services.prefs.getStringPref(
-      ENTERPRISE_TOOLBAR_LOGO_URL_PREF,
-      ""
-    );
+  _setLogoUrlFromPref(window) {
+    const prefLogoUrl = Services.prefs.getStringPref(lazy.PREFS.LOGO_URL, "");
```
1rneh: nit
```suggestion
    const logoUrl = Services.prefs.getStringPref(lazy.PREFS.LOGO_URL, "");
```

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -227,64 +214,31 @@ export const EnterpriseHandler = {
...
 
-    // skip if pref is not set
     if (!prefLogoUrl) {
-      return false;
+      return;
```
1rneh, line 249: Let's log an error, if we have no logo uri.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -77,6 +78,8 @@ export const EnterpriseHandler = {
    * @param {Window} window chrome window
    */
   updateBadge(window) {
+    this._setLogoUrlFromPref(window);
```
1rneh: nit
```suggestion
    this._updateLogo(window);
```

1rneh (approved):
Thank you! Only some nits.

This will need hooking up once `api/browser/config` is deployed :)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
jporter-dev: Forget why I moved it there... moved it back :)

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -237,4 +240,33 @@ export const EnterpriseHandler = {
     this._signedInUser = {};
     this._isInitialized = false;
   },
+
+  _updateLogo(window) {
+    const logoUrl = Services.prefs.getStringPref(LOGO_URL, "");
+
+    if (!logoUrl) {
+      console.error(`Skipping logo update, ${LOGO_URL} pref is not set.`);
```
lissyx: `console.warn` ? Or do we consider this should not happen at all? In which case we may need to throw?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -237,4 +240,33 @@ export const EnterpriseHandler = {
...
+      return;
+    }
+
+    // allow https: URLs and data: URIs for common image formats with base64 encoding only.
+    const logoUrlPattern =
+      /^(https:\/\/|data:image\/(?:png|jpeg|gif|webp|svg\+xml);base64,)/;
```
lissyx: Is the logo when hosted over `https` allowed to be sourced from anywhere? Or do we want to serve it from the console ?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -237,4 +240,33 @@ export const EnterpriseHandler = {
...
+    if (!logoUrlPattern.test(logoUrl)) {
+      console.error(`Invalid logo URL in pref: ${logoUrl}`);
```
lissyx: Same question regarding the severity

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -25,6 +25,7 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
 });
 
 const PROMPT_ON_SIGNOUT_PREF = "enterprise.promptOnSignout";
+const LOGO_URL = "enterprise.logoUrl";
```
lissyx: I missed it on `PROMPT_ON_SIGNOUT_PREF` but we should rather have `enterprise.logo_url` instead

jporter-dev: got it, will use snake_case for enterprise prefs from now on 👍

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: `console.warn` makes sense for this missing pref, and we can throw an error when the pref is set and fails validation

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Updated the validation to check against the console base uri

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -237,4 +240,38 @@ export const EnterpriseHandler = {
     this._signedInUser = {};
     this._isInitialized = false;
   },
+
+  _updateLogo(window) {
+    const logoUrl = Services.prefs.getStringPref(LOGO_URL, "");
```
lissyx, line 245: Do we really want to allow having no logo URL ?

lissyx (approved):

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: This acts as a logo override, It would fall back to the default Mozilla logo if unset. Whether this param is required or not should probably be enforced by the console or policy when this gets wired up to the new endpoint.

PR #576, opened 2026-03-20
Bug 2017135: UI changes to display logout reason

victhorlopez:


browser/extensions/felt/api.js
```diff
@@ -8,6 +8,10 @@
 
 const lazy = {};
 
+const LOGOUT_TYPE_MESSAGE_CLASS = new Map([
```
lissyx: Maybe this should just live on the `nsIFelt` interface directly?

victhorlopez: does it make sense? In theory it is a mapping from a logout type to a css class used by one particular UI component.

browser/extensions/felt/content/felt.xhtml
```diff
@@ -216,6 +216,12 @@
           >
             <span slot="message" class="felt-browser-error-details"></span>
           </moz-message-bar>
+          <moz-message-bar
+            class="felt-browser-info-console-forced-logout is-hidden"
+            data-l10n-id="felt-browser-error-console-forced-logout"
```
gcp: Why is one `info` and the other `error`?

browser/extensions/felt/api.js
```diff
@@ -267,7 +261,9 @@ this.felt = class extends ExtensionAPI {
       case "FeltParent:FirefoxLogoutExit": {
         const success = Services.felt.makeBackgroundProcess(false);
         console.debug(`FeltExtension: makeBackgroundProcess? ${success}`);
-        this.showWindow();
+        this.showWindow(
+          lazy.logoutTypeMessageClass.get(message.data.logoutType) ?? ""
```
lissyx, line 265: OK so this is what we send from PR #550, it's confusing that things are split accross

browser/extensions/felt/api.js
```diff
@@ -8,6 +8,15 @@
 
 const lazy = {};
 
+ChromeUtils.defineLazyGetter(lazy, "logoutTypeMessageClass", () =>
+  new Map([
+    [
+      Services.felt.logoutTypeConsoleForcedLogout,
+      "felt-browser-info-console-forced-logout",
+    ],
+  ])
```
lissyx, line 17: Maybe worth explaining what this is mapping from/to

victhorlopez:
comments addressed in https://github.com/mozilla/enterprise-firefox/pull/550

PR #577, opened 2026-03-20
Enterprise main merge 20260320

1rneh:


taskcluster/kinds/repackage/kind.yml
```diff
@@ -111,6 +111,7 @@ tasks:
                 win64\b.*: [mar, installer]
         fetches:
             toolchain:
+                - linux64-7zz
```
lissyx, line 114: we are likely going to need it on `enterprise-repack` kind as well?

taskcluster/kinds/build/windows.yml
```diff
@@ -41,6 +41,7 @@ win32/debug:
     use-sccache: true
     fetches:
         toolchain:
+            - linux64-7zz
```
lissyx, line 44: and you're missing it on the enterprise tasks it seems https://github.com/1rneh/enterprise-firefox/blob/fffad47c3cd3cdb7af2bd47eda524d120039afd4/taskcluster/kinds/build/windows.yml#L255-L265

1rneh: Thanks! I've added them for the following tasks
- win64-enterprise/debug
- win64-enterprise-shippable/opt
- win64-enterprise/opt

taskcluster/kinds/repackage/kind.yml
1rneh: For the `enterprise-repack` kind, we don't fetch any specific toolchain.

lissyx (approved):

PR #578, opened 2026-03-20
 Bug 2024958: fix OnIntrospectFinish double free r=lissyx

victhorlopez:
The pointer used is a refPtr with an init count of 1(RefPtr<GDBusNodeInfo> nodeInfos = dont_AddRef) and will automatically go to 0 when out of scope, no manual reference counting is needed in this case

### DIFF
--- a/dom/system/linux/ModemManagerImeiProvider.cpp
+++ b/dom/system/linux/ModemManagerImeiProvider.cpp
@@ -126,9 +126,8 @@ void ModemManagerImeiProvider::OnIntrospectFinish(RefPtr<Promise> aPromise,
   }
 
   if (!nodeInfos->nodes) {
-    MMI_LOG(Error, "nodeInfos->nodes failed! %s", error->message);
+    MMI_LOG(Error, "nodeInfos->nodes failed! %s", error ? error->message : "(noerror)");
     aPromise->MaybeReject(NS_ERROR_FAILURE);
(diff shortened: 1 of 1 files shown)


dom/system/linux/ModemManagerImeiProvider.cpp
```diff
@@ -126,9 +126,8 @@ void ModemManagerImeiProvider::OnIntrospectFinish(RefPtr<Promise> aPromise,
   }
 
   if (!nodeInfos->nodes) {
-    MMI_LOG(Error, "nodeInfos->nodes failed! %s", error->message);
+    MMI_LOG(Error, "nodeInfos->nodes failed!");
```
victhorlopez: Can you have a look at this one? (we did not hit it), it looks like if the function g_dbus_node_info_new_for_xml works we can get a nullptr by error but I am not 100% sure https://docs.gtk.org/gio/ctor.DBusNodeInfo.new_for_xml.html

lissyx: `error ? error->message : "(noerror)"` ?

lissyx (approved):
r=me but please comment somewhere that we can do this because `RefPtr` and `GDBusNodeInfo`'s destructor already calls the same function

victhorlopez:
I added the description, and did the change regarding the other pointer just in case

PR #581, opened 2026-03-20
Setup `mozilla/enterprise-firefox-try` as a target for `mach try`

ahal:


ahal:
The first two commits are also landing against `firefox-main`, but it's fine if they land here as well.

taskcluster/gecko_taskgraph/transforms/repackage.py
```diff
@@ -837,16 +837,14 @@ def make_job_description(config, jobs):
...
-                                    repack_name
-                                ]["locales"]:
+                release_partner_config = (
+                    config.params.get("release_partner_config") or {}
+                )
+                if this_repack := release_partner_config.get(config.kind):
+                    for enterprise_name, entries in this_repack.items():
+                        for repack_name, entry in entries.items():
+                            for platform_name in entry["platforms"]:
+                                for repack_locale in entry["locales"]:
```
lissyx, line 847: This looks good but I'm always getting confused on changes like that because i never remember what is the payload behind

taskcluster/config.yml
```diff
@@ -7,8 +7,12 @@ project-repo-param-prefix: ''
 product-dir: 'browser'
 treeherder:
     branch-map:
-        enterprise-main: enterprise-main
-        enterprise-release: enterprise-release
+        by-project:
+            enterprise-firefox-try:
+                default: enterprise-firefox-try
```
lissyx, line 12: Is this ready on treeherder side?

ahal: It's not, but it won't hurt anything to land it ahead of time.

tools/tryselect/push.py
```diff
@@ -53,7 +53,7 @@
 MACH_TRY_PUSH_TO_VCS = os.getenv("MACH_TRY_PUSH_TO_VCS") == "1"
 
 HG_TRY_URL = "ssh://hg.mozilla.org/try"
-MACH_TRY_REMOTE = HG_TRY_URL
+MACH_TRY_REMOTE = "git@github.com:mozilla/enterprise-firefox-try.git"
```
lissyx, line 56: This means people need write access to the repo, is it fine? Could we /should we make this work from a fork of that repo instead ?

What if my local configuration needs a different URL to pass the proper SSH keys for example?

.taskcluster.yml
```diff
@@ -412,6 +412,7 @@ tasks:
                       $if: >
                         eventType in ["action", "cron"]
                         || (eventType == "push" && shortRef == "enterprise-main")
+                        || (eventType == "push" && project == "enterprise-firefox-try")
```
lissyx, line 415: if we have `project=enterprise-firefox-try` no risk of index pollution right?

tools/tryselect/push.py
ahal: Yeah, we'll need to grant write access. Long term my plan is to support forks, but that is a bigger change. Some details in:
https://docs.google.com/document/d/1PvT0PCDsoBKxvN9XUeP7b2O9YJAMZOHYVXqJC9vU_Mo/edit?tab=t.ntqqur3dd3nh#heading=h.8rqgi74w363m

This is the "hacky something is better than nothing" phase of try support.

taskcluster/gecko_taskgraph/transforms/repackage.py
ahal: Tbh, I'm also not sure what the format of this object is. I was simplifying based purely on knowing this is equivalent from a Python data structure point of view.

lissyx: We'll need to run it on `try` to verify :-)

taskcluster/config.yml
lissyx: How much is it, just adding the new config?

tools/tryselect/push.py
lissyx: OK, we need a solution to that change before I can approve

taskcluster/config.yml
lissyx: @Archaeopteryx I'll prepare a PR for that

tools/tryselect/push.py
lissyx: @ahal I'm going to take over your PR and add a fix that works for me

lissyx: #631 contains this

taskcluster/config.yml
lissyx: https://github.com/mozilla/treeherder/pull/9358

lissyx (approved):

PR #583, opened 2026-03-23
Fetch linux64-7zz toolchain in enterprise-repack-repackage task

1rneh:


taskcluster/kinds/enterprise-repack-repackage/kind.yml
```diff
@@ -12,6 +12,7 @@ transforms:
     - gecko_taskgraph.transforms.repackage_partner
     - gecko_taskgraph.transforms.job
     - gecko_taskgraph.transforms.task
+    - taskgraph.transforms.run:transforms
```
lissyx: I'm surprised we need this?

lissyx (approved):

taskcluster/kinds/enterprise-repack-repackage/kind.yml
1rneh: I thought so: https://taskcluster-taskgraph.readthedocs.io/en/latest/howto/use-fetches.html
But the decision task fails, so I'm still trying to figure out how to fetch `linux64-7zz` toolchain..

PR #585, opened 2026-03-23
Bug 1998394: Add shutdown on failing to retrieve policies first time on felt browser

victhorlopez:


browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -38,6 +38,20 @@ export const EnterpriseHandler = {
...
+   */
+  beforeUIStartup() {
+    if (
+      Services.felt?.isFeltBrowser() &&
+      Services.policies.status === Ci.nsIEnterprisePolicies.FAILED
+    ) {
+      this.initiateShutdown();
+    }
+  },
+
```
1rneh: I think the `ConsoleClient` is a better place for that. more precisely `ConsoleClient.getRemotePolicies()`.

On another thought... #550 handles shutdown+reauth on 401 and 403. I think any non-5xx error should probably result in sign-out and shutdown? We'll need to go over this again once we defined if and how we support offline mode yet.

1rneh (changes requested):

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -38,6 +38,20 @@ export const EnterpriseHandler = {
...
 
+  /**
+   * Called once via the browser-before-ui-startup category, before any browser
+   * window exists. If the initial remote policy fetch failed, the browser is
+   * shut down immediately so the user is returned to the FELT login screen.
+   */
+  beforeUIStartup() {
+    if (
+      Services.felt?.isFeltBrowser() &&
+      Services.policies.status === Ci.nsIEnterprisePolicies.FAILED
```
lissyx: I'm not sure we can have any guarantee that there has been at least one policy fetch at that moment?

browser/components/enterprise/EnterpriseHandler.sys.mjs
lissyx: 👍  and doing it there ensures us that we know there has been one policy fetch at least

victhorlopez: [EnterprisePoliciesParent remote policies](https://searchfox.org/enterprise-main/source/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#171)
[EnterprisePoliciesParent failed flag](https://searchfox.org/enterprise-main/source/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#143)

I am using this flag in here, regarding the place where this code should be, wasn't ConsoleClient supposed to be just a file where we have functions to query the Console, but other application logic, i.e shutdowns should be done somewhere else? (that is what I understood from a previous comment)

1rneh: Note: We are fetching remote policies on policy engine initialization which is during startup, see https://searchfox.org/enterprise-main/rev/1254e9f1a9161f8d7a26070c2d37d5e10898f051/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#169-171

browser/components/enterprise/EnterpriseHandler.sys.mjs
victhorlopez: About the ordering, as far as I understood ["policies-startup" ](https://searchfox.org/enterprise-main/source/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#437)happens before "browser-before-ui-startup", and the policies has a spinloop to make sure it is done sync. But I guess in case that changes that would break this part of the code

browser/components/enterprise/EnterpriseHandler.sys.mjs
lissyx: > Note: We are fetching remote policies on policy engine initialization which is during startup, see https://searchfox.org/enterprise-main/rev/1254e9f1a9161f8d7a26070c2d37d5e10898f051/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#169-171

It is indeed in the same step of startup, but I'm far from certain that we could assert that the fetch of policies would have happened before we it `beforeUIStartup`

lissyx: https://searchfox.org/enterprise-main/rev/1254e9f1a9161f8d7a26070c2d37d5e10898f051/toolkit/xre/nsXREDirProvider.cpp#650,657-659

testing/enterprise/felt_tests.py
```diff
@@ -430,6 +433,7 @@ def serve(
     policy_extensions=None,
     policy_access_token=None,
     policy_refresh_token=None,
+    policies_fail_request=None,
```
lissyx, line 436: I dont remember why it made sense for access/refresh tokens, but can we work this without introducing such new value and rely on a `POST` to a test-specific endpoint of the mock console to set the state, like we do in other places?
https://searchfox.org/enterprise-main/rev/1254e9f1a9161f8d7a26070c2d37d5e10898f051/testing/enterprise/test_felt_browser_updates_apply.py#30-33,77

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: > I am using this flag in here, regarding the place where this code should be, wasn't ConsoleClient supposed to be just a file where we have functions to query the Console, but other application logic, i.e shutdowns should be done somewhere else? (that is what I understood from a previous comment)

Yes, exactly. What I meant is that `ConsoleClient` may need to trigger e.g. the shutdown flow, but the shutdown logic itself should not live there. Instead, that should be factored out into something like `EnterpriseHandler.initiateShutdown`.

1rneh: > It is indeed in the same step of startup, but I'm far from certain that we could assert that the fetch of policies would have happened before we it beforeUIStartup

I looked into this and we can assume the policy engine has been initialized before `browser-before-ui-startup` is dispatched. 

In `XREMain::XRE_mainRun` we first initialize the policy engine in [`mDirProvider.DoStartup()`](https://searchfox.org/enterprise-main/rev/1254e9f1a9161f8d7a26070c2d37d5e10898f051/toolkit/xre/nsAppRunner.cpp#5833) in which we wait until the remote policies fetch has resolved and then second call [`XRE_mainRun()`](https://searchfox.org/firefox-main/rev/1427c88632d1474d2653928745d78feca1a64ee0/toolkit/xre/nsAppRunner.cpp#6195) which a little in dispatches [finali-ui-startup](https://searchfox.org/firefox-main/rev/1427c88632d1474d2653928745d78feca1a64ee0/toolkit/xre/nsAppRunner.cpp#5853) which [calls the modules from category browser-before-ui-startup](https://searchfox.org/firefox-main/rev/1427c88632d1474d2653928745d78feca1a64ee0/browser/components/BrowserGlue.sys.mjs#414-416).

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -169,6 +171,9 @@ EnterprisePoliciesManager.prototype = {
     // Fetch first set of remote policies during the
     // initialization of the policy engine
     await remoteProvider.fetchPoliciesOnStartup();
+    if (remoteProvider.failed && Services.felt?.isFeltBrowser()) {
+      await lazy.EnterpriseHandler.initiateShutdown();
```
1rneh, line 177: Let's log an error explaining that we couldn't fetch policies on startup.

testing/enterprise/test_felt_browser_init_failures.py
```diff
@@ -0,0 +1,47 @@
...
+from base_test import Environment
+from felt_tests import FeltTests
+
+
+class BrowserInitFailures(FeltTests):
+    def _set_policies_fail(self, fail):
+        requests.post(
+            f"http://localhost:{self.console_port}/api/browser/policies",
+            data=json.dumps({"fail": fail}),
+        )
```
1rneh: Can we set `self.server.policies_fail_request` directly in the tests instead of introducing an endpoint for this?

testing/enterprise/test_felt_browser_init_failures.py
```diff
@@ -0,0 +1,47 @@
...
+        self._set_policies_fail(True)
+        super().run_felt_base()
+        self._manually_closed_child = True
+
+        self.await_felt_auth_window()
+        self.force_window()
+        self.assert_user_signed_out(env=Environment.FELT)
+        self._set_policies_fail(False)
+
+    def test_browser_login_after_policy_fetch_fail(self):
```
1rneh: I’m not sure about the value of `test_browser_login_after_policy_fetch_fail`. It seems to re-test the failure scenario already covered in `test_browser_init_policy_fetch_fail`, and then adds a fairly standard check that the user is signed in after Felt authentication. I think this test could be removed.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -169,6 +171,9 @@ EnterprisePoliciesManager.prototype = {
     // Fetch first set of remote policies during the
     // initialization of the policy engine
     await remoteProvider.fetchPoliciesOnStartup();
+    if (remoteProvider.failed && Services.felt?.isFeltBrowser()) {
```
1rneh: nit:
```suggestion
    if (Services.felt?.isFeltBrowser() && remoteProvider.failed) {
```

1rneh (changes requested):

testing/enterprise/test_felt_browser_init_failures.py
victhorlopez: as per https://github.com/mozilla/enterprise-firefox/pull/585#discussion_r2976661914 I changed it (not sure I implemented it the way I was supposed to)

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
victhorlopez: it's already in the function, let me know If you want that we add another log in here

[log](https://searchfox.org/enterprise-main/source/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#1047)

testing/enterprise/test_felt_browser_init_failures.py
victhorlopez: I will remove it, the idea was not to test always happy path and make sure no state is left behind somewhere and that we can reauth again

testing/enterprise/test_felt_browser_init_failures.py
1rneh: Setting `self.server.policies_fail_request` directly in the test seems functionally equivalent to setting it via the mock console. @lissyx, or am I missing something? If not, I’d prefer to avoid adding the extra endpoint.

lissyx: > Setting `self.server.policies_fail_request` directly in the test seems functionally equivalent to setting it via the mock console. @lissyx, or am I missing something? If not, I’d prefer to avoid adding the extra endpoint.

yes it should be, but for some reason i always felt like this way of sharing states at the process level was more a hack than a correct way to control what we want to control during testing, and extra end point offers a simpler flow (less declaration in multiple places, but yes extra endpoint)

1rneh: Another proposal:

What about having a contextmanager? 
```
@contextmanager
    def scenario(self, name):
        previous = self._scenario
        self._scenario = name
        try:
            yield
        finally:
            self._scenario = previous
```
and then we run the test with `with self.server.scenario("policies_500")` and implement a conditional response for the policies endpoint.

lissyx: I dont see why it would not be a good solution, but it sounds more like a useful refactor we should do at once on all tests and not block this PR on that.

lissyx: But if we decide to refactor then it would make https://github.com/mozilla/enterprise-firefox/pull/612 not relevant anymore.

I'd say we refactor in a follow up and for this PR we live with the `self.server.policies_fail_request` as @1rneh mentionned, and sorry @victhorlopez for doing/undoing :(

victhorlopez: done

testing/enterprise/test_felt_browser_init_failures.py
```diff
@@ -0,0 +1,24 @@
...
+class BrowserInitFailures(FeltTests):
+    def test_browser_init_policy_fetch_fail(self):
+        self.policies_fail_request.value = 1
+        super().run_felt_base()
+        self._manually_closed_child = True
+
+        self.await_felt_auth_window()
+        self.force_window()
+        self.assert_user_signed_out(env=Environment.FELT)
+        self.policies_fail_request.value = 0
```
lissyx: nit: probably we can do that when `self.await_felt_auth_window()` is done

testing/enterprise/felt_tests.py
```diff
@@ -173,6 +173,9 @@ def do_GET(self):
 
         elif path == "/api/browser/policies":
             self.check_auth()
+            if self.server.policies_fail_request.value:
+                self.reply("", 500, "Internal Server Error")
```
lissyx: I've seen cases where some of our code could be picky about payload and content type, let's make sure we pass a correct one:


```suggestion
            if self.server.policies_fail_request.value:
                self.reply("", 500, "Internal Server Error", "text/html")
```

lissyx (approved):

testing/enterprise/felt_tests.py
```diff
@@ -173,6 +173,9 @@ def do_GET(self):
 
         elif path == "/api/browser/policies":
             self.check_auth()
+            if self.server.policies_fail_request.value:
+                self.reply("", 500, "Internal Server Error", "text/html")
```
1rneh: We expect `application/json`: https://searchfox.org/enterprise-main/rev/73669e504b64c8ebe02576c85458600499166b03/browser/components/enterprise/modules/ConsoleClient.sys.mjs#400

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Can we clarify if we want to signal to the user that the access to Firefox Enterprise is blocked because we couldn't apply the required policies? Without some feedback this causes a very inconvenient user experience...

testing/enterprise/felt_tests.py
```diff
@@ -501,6 +482,7 @@ def setUp(self):
         self.sso_port = random.randrange(15000, 20000)
         self.policy_block_about_config = Value("B", 1)
         self.policy_extensions = Value("B", 0)
+        self.policies_fail_request = Value("B", 0)
```
1rneh, line 485: What this `Value("B", 0)` and `Value("B", 1)`?

1rneh (changes requested):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
victhorlopez: The [figma](https://www.figma.com/board/lEW9qt34CKjfrb8FoclKir/FELT-Notification-and-Error-states?node-id=0-1&p=f&t=0c4iRTd3xYYjf5GF-0) has several errors (network, 401 etc). Which one are we supposed to cover in here? Cause we have overlap, network error?

i.e are they actually covered in the other 2 prs?

testing/enterprise/felt_tests.py
victhorlopez: These flags are actually no built ins (like bool int etc) we use them in here and pass to the "[fake console".](https://github.com/mozilla/enterprise-firefox/blob/enterprise-main/testing/enterprise/felt_tests.py#L521) as the "fake console" lives in another process you need this special multiprocessing classes

So that is just value type and default,

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: As by our offline conversation, let's file a follow-up bug that we should fetch the policies in felt and block starting Firefox if we fail to retrieve them. Can you also add the bug id as a comment in the conditional block?

testing/enterprise/felt_tests.py
1rneh: Thank you!

1rneh (approved):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: I filed a follow-up bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2027006

PR #597, opened 2026-03-25
Bug 2022741 - Implement support for policy to force restart

gcp:
Transplanted from https://hg.mozilla.org/mozilla-central/rev/3d336d4d418d

### DIFF
--- a/browser/components/BrowserComponents.manifest
+++ b/browser/components/BrowserComponents.manifest
@@ -82,6 +82,7 @@ category browser-idle-startup resource://gre/modules/ColorwayThemeMigration.sys.
 category browser-idle-startup resource://gre/modules/AppServicesTracing.sys.mjs RustLogForwarder.init
 #ifdef MOZ_UPDATER
 category browser-idle-startup resource://gre/modules/UpdateListener.sys.mjs UpdateListener.maybeShowUnsupportedNotification
+category browser-idle-startup resource:///modules/UpdatePolicyEnforcer.sys.mjs UpdatePolicyEnforcer.registerObservers
 #endif
 #ifdef XP_WIN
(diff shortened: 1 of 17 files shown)


lissyx (approved):

1rneh (commented):
l10n team asked us to not mix enterprise and non-enterprise strings in the same files. Can you check if there is an equivalent file under `browser/locales/en-US/browser/enterprise` or create one for the strings you've are adding? 
https://searchfox.org/enterprise-main/source/browser/locales/en-US/browser/enterprise

1rneh:
> l10n team asked us to not mix enterprise and non-enterprise strings in the same files. Can you check if there is an equivalent file under `browser/locales/en-US/browser/enterprise` or create one for the strings you've are adding? https://searchfox.org/enterprise-main/source/browser/locales/en-US/browser/enterprise

**Edit:** I just saw that this is a commit from upstream/main? Aren't we merging those anyway?

gcp:
>Edit: I just saw that this is a commit from upstream/main? Aren't we merging those anyway?

Yes, but we want to get QA on this one ASAP I think.

lissyx:
> l10n team asked us to not mix enterprise and non-enterprise strings in the same files. Can you check if there is an equivalent file under `browser/locales/en-US/browser/enterprise` or create one for the strings you've are adding? https://searchfox.org/enterprise-main/source/browser/locales/en-US/browser/enterprise

Since it was landed on mozilla-central, we should be fine wrt not mixing l10n identifiers?

gcp:
>Yes, but we want to get QA on this one ASAP I think.

Worst case we close this but they can use the artifacts this will have generated on CI.

1rneh:
This commit was merged as part of #599.

PR #604, opened 2026-03-26
Bug 2026537: Set dom.push.serverURL on test setup

victhorlopez:
The test expects  [](https://searchfox.org/enterprise-main/rev/73669e504b64c8ebe02576c85458600499166b03/dom/push/PushService.sys.mjs#389) a url in there. 

I resetted the pref on the mochi setup.

### DIFF
--- a/testing/profiles/mochitest/user.js
+++ b/testing/profiles/mochitest/user.js
@@ -36,3 +36,7 @@ user_pref("places.history.floodingPrevention.enabled", false);
 // permission, and we can open it and wait for the user to give permission, then
 // don't do that.
 user_pref("geo.prompt.open_system_prefs", false);
+
+// Set a dummy push server URL for mochitest runs. In enterprise builds the
+// default is an empty string, but push tests need a non-empty value.
+user_pref("dom.push.serverURL", "wss://push.services.mozilla.com/");


dom/push/test/test_utils.js
```diff
@@ -216,9 +217,9 @@ async function setupPrefsAndReplaceService(mockService) {
   await setupPrefs();
 }
 
-function setupPrefsAndMockSocket(mockSocket) {
-  setupMockPushSocket(mockSocket);
-  return setupPrefs();
+async function setupPrefsAndMockSocket(mockSocket) {
```
gcp: Why this change? Is the current test racy? (Then this part of the fix could be upstreamed)

victhorlopez: was testing, still in draft

dom/push/test/test_utils.js
```diff
@@ -203,6 +203,7 @@ function setupPrefs() {
     set: [
       ["dom.push.enabled", true],
       ["dom.push.connection.enabled", true],
+      ["dom.push.serverURL", "wss://push.services.mozilla.com/"],
```
lissyx: i know it's a draft, but if we need that then maybe it should be part of the test profiles `testing/profiles/` and we should make sure to have the correct variation there (either a speific files? set of files? or preprocessing?)

victhorlopez: I would look into testing/profiles/. As I said it is a draft, but if the test might assume the url is there and [we changed that](https://searchfox.org/enterprise-main/source/modules/libpref/init/all.js#3228), it would be a setup bug IMO

lissyx: Maybe the assumption holds from firefox point of view and thus from the test and we should just use a different way of disabling push?

lissyx: Or reading the code at https://searchfox.org/enterprise-main/rev/73669e504b64c8ebe02576c85458600499166b03/dom/push/PushService.sys.mjs things looks to be gracefully expecting the pref to be unset but the test really only focuses on when the service is enabled. I guess the question then becomes: do we expect to have usage of push in the future? if we do we should make sure we keep the test running as it was supposed to.

Likely worth a discussion with those owning this code, I guess your fix would make sense, but I'm surprised we in fact have tests that relies on the real URL?

victhorlopez: I some places they use `wss://push.example.org/`

victhorlopez: Basically even if not used in some tests, it fails cause [this if ](https://searchfox.org/enterprise-main/rev/73669e504b64c8ebe02576c85458600499166b03/dom/push/PushService.sys.mjs#389) is expected to have some string in the pref

victhorlopez: @lissyx There is a pref called `dom.push.connection.enabled` to disable push, but in the [original bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2012667) we [disabled the url ](https://github.com/mozilla/enterprise-firefox/commit/937f94b78c4c759f3b2a06c8ac52c5accb688c31) 

`do we expect to have usage of push in the future? if we do we should make sure we keep the test running as it was supposed to`

I think in general we should try to diverge as little as possible from m-c(meaning make everything work, unless it is different from enterprise, i.e FxA and similars)

testing/profiles/mochitest/user.js
```diff
@@ -36,3 +36,7 @@ user_pref("places.history.floodingPrevention.enabled", false);
 // permission, and we can open it and wait for the user to give permission, then
 // don't do that.
 user_pref("geo.prompt.open_system_prefs", false);
+
+// Set a dummy push server URL for mochitest runs. In enterprise builds the
+// default is an empty string, but push tests need a non-empty value.
+user_pref("dom.push.serverURL", "wss://push.example.org/");
```
victhorlopez: as far as I understood this file cannot have preprocessor ifdefs. I am hesitant here, "wss://push.services.mozilla.com/" . Would be the current behaviour the code has (although the url is not used) 
[wss://push.example.org/](https://searchfox.org/enterprise-main/search?q=wss%3A%2F%2Fpush.example.org%2F&path=&case=false&regexp=false) the actual one used in tests (where they override). For now I left it "wss://push.services.mozilla.com/" as it is same behaviour

testing/profiles/mochitest/user.js
```diff
@@ -36,3 +36,7 @@ user_pref("places.history.floodingPrevention.enabled", false);
 // permission, and we can open it and wait for the user to give permission, then
 // don't do that.
 user_pref("geo.prompt.open_system_prefs", false);
+
+// Set a dummy push server URL for mochitest runs. In enterprise builds the
+// default is an empty string, but push tests need a non-empty value.
+user_pref("dom.push.serverURL", "wss://push.services.mozilla.com/");
```
lissyx, line 42: LGTM, this is the default value used on non enterprise builds when running tests?

lissyx (approved):

testing/profiles/mochitest/user.js
victhorlopez: yes

lissyx:
Removing gcp who is on PTO, if we can improve test suites execution and given the patch we have now I dont think it's worth blocking on his review

PR #607, opened 2026-03-26
Bug 2026578 - Hide uptodate message on sso load

1rneh:
**Bug:** The "Firefox Enterprise is up to date" message overlays the loaded `browser` element (see screenshot)
Screenshot 2026-03-26 at 12 05 21

**Fix:** Hide the message once we load the browser element.

### DIFF
--- a/browser/extensions/felt/content/window.js
+++ b/browser/extensions/felt/content/window.js
@@ -290,6 +290,7 @@ async function connectToConsole(email) {
   );
 
   ErrorReport.reset();
+  document.querySelector(".felt-updates-message").classList.add("is-hidden");
   document.querySelector(".felt-login__email-pane").classList.add("is-hidden");
   document.querySelector(".felt-login__sso").classList.remove("is-hidden");
 


browser/extensions/felt/content/window.js
```diff
@@ -290,6 +290,7 @@ async function connectToConsole(email) {
   );
 
   ErrorReport.reset();
+  document.querySelector(".felt-updates-message").classList.add("is-hidden");
```
lissyx, line 293: LGTM but I'm now wondering if we may not have a similar behavior with the other message, like errors, that we display?

lissyx (approved):

browser/extensions/felt/content/window.js
1rneh: These errors are reset by `ErrorReport.reset()`.

lissyx: Right so we are safe, thanks!

PR #611, opened 2026-03-26
Bug 2006564: Focus email input when login is visible

victhorlopez:


browser/extensions/felt/content/window.js
```diff
@@ -311,8 +311,6 @@ async function listenFormEmailSubmission() {
     signInBtn.disabled = emailInput.value.trim() === "";
   });
 
-  emailInput.focus();
```
victhorlopez, line 314: basically now the felt is hidden and this fails(focus does not focus hidden elements), added something to react to [this](https://searchfox.org/enterprise-main/source/browser/components/enterprise/modules/Updates.sys.mjs#248)

lissyx: Why was this not caught by tests, because we manually focus from marionette?

browser/extensions/felt/content/window.js
```diff
@@ -472,6 +470,25 @@ function setupPopupNotifications() {
...
+  new MutationObserver(() => {
+    if (!loginPane.classList.contains("is-hidden")) {
+      emailInput?.focus();
+    }
+  }).observe(loginPane, { attributeFilter: ["class"] });
+
+  if (!loginPane.classList.contains("is-hidden")) {
+    emailInput?.focus();
+  }
+}
```
lissyx, line 490: What about `visibilitychange` event handler? I guess it would be simpler, no need to play around with the classes

lissyx (approved):

browser/extensions/felt/content/window.js
victhorlopez: it's [done](https://searchfox.org/enterprise-main/source/testing/enterprise/felt_tests.py#696) with js

browser/extensions/felt/content/window.js
victhorlopez: the [event](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event) you suggested it is at document level only. The other option was to send a custom event or similar.

1rneh (approved):
Thanks!

PR #614, opened 2026-03-27
Bug 2019061 - CrashReporter: Add logic to automatically submit crash reports upon launch of crash reporter

jonathanmendez:
This is a simplified approach at adding what should be asynchronous behavior (compare to https://github.com/mozilla/enterprise-firefox/pull/602), in favor of accomplishing the goal in a timely manner.

Observed behavior:
When crash reporter launches, a progress indicator shows that a crash report is being prepared and submitted. Once complete, the progress updates to a success or failure message, and the "Ok" button becomes enabled to dismiss the dialog.

This can be tested in local builds with `MOZ_CRASHREPORTER=1 MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT=1 mach run` for now. It will require a follow-up PR to set `MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT` when the policy is enabled.

### DIFF
--- a/toolkit/crashreporter/client/app/src/logic.rs
+++ b/toolkit/crashreporter/client/app/src/logic.rs
@@ -429,7 +429,7 @@ impl ReportCrash {
         let crash_ui = ReportCrashUI::new(
             &*self.settings.borrow(),
             self.config.clone(),
-            logic_remote_queue,
+            logic_remote_queue.clone(),
         );
 
         // Set the UI remote queue.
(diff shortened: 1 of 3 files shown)


toolkit/crashreporter/client/app/src/logic.rs
```diff
@@ -427,6 +445,14 @@ impl ReportCrash {
         let logic_panic_handler = PanicHandler(Arc::downgrade(&crash_ui_async_task));
         self.ui = Some(crash_ui_async_task);
 
+        #[cfg(feature = "enterprise")]
+        // Set up logic thread to immediately send the report
+        // Normally we might do this asynchronously because it blocks,
+        // but with policy_auto_submit, the UI will not be interactive until after the send completes.
```
lissyx: isn't there a risk here, if the upload is slow?

lissyx (approved):

toolkit/crashreporter/client/app/src/logic.rs
jonathanmendez: First, in re-reading this comment, it's confusing and I should at least rephrase the comment to indicate the UI is intentionally not interactive until after the send completes (the Ok button is disabled until the progress completes, then becomes enabled). As the comment is currently written it sounds more like "because this blocks, therefore the UI will not be interactive".

But then further, in the case of a slow upload the behavior is that there is an animated progress indicator with text indicating current upload state. While the state is not final, the button is disabled. Then once the state becomes final the state is enabled.
The ideal case, which the alternate async approach is/was aiming for, lets the Ok button be enabled at all times (and the progress indicator still behaves the same). If the user would press Ok in that case, the UI would disappear but the thread executing the send would continue to completion silently.

So this is a "hopefully good enough" version that I think fits our timeline better since the async version has taken time to come together. We will replace the behavior with the async version once it's ready.

PR #617, opened 2026-03-27
Bug 2011841 - Prevent unauthenticated requests in felt on startup

1rneh:
- Prevents running browser window UI logic in Felt (which initiates guaranteed unauthenticated requests)
- Simulate that no user is signed in into "EnterpriseAccounts" instead of disabling fxa via pref

Note: The expected unauthorized requests on felt-side throw errors that create a lot of noise in the logs.


browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
```diff
@@ -43,7 +43,9 @@ export class EnterpriseStorageManager {
     // unhandled exception when it is GCd - so add an empty .catch handler here
     // to prevent this.
     this.#getAccountDataPromise.catch(() => {});
-    this.#getAccountDataPromise = lazy.ConsoleClient.getFxAccountData();
+    this.#getAccountDataPromise = Services.felt.isFeltUI()
+      ? Promise.resolve(null) // Felt has no signed-in user on startup; avoid guaranteed 401 from FxA
```
lissyx, line 47: This looks OK, but I thought we had disabled FxAccounts via prefs on Felt side ? Are you being extra cautious or did you spot real problems?

lissyx (approved):

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
1rneh: I'm removing the line of code that is setting `identity.fxaccounts.enabled=false` (see changes in `api.js`) since it's enough to simulate that the user is signed out. 

I spot this issue in any arbitrary logs from felt:
```
INFO -  console.error: (new ReauthRequiredError("No refresh token available", "resource:///modules/enterprise/ConsoleClient.sys.mjs", 46))
INFO -  console.warn: "EnterpriseHandler: Unable to initialize enterprise user: " (new InvalidAuthError("Unhandled reauthentication", "resource:///modules/enterprise/ConsoleClient.sys.mjs", 74))
```

victhorlopez: I know this might be a bit of scope creep... does it make sense to also disable this for MOZ_BYPASS_FELT=1? We have a lot of test with this error

1rneh: The scope of this PR is to avoid unauthenticated requests in felt, since the load of failure logs in Felt was also very high.

What you are asking for seems to require more of a test strategy on how to test Firefox with that console dependency. I'd like to keep this out of this scope.

PR #618, opened 2026-03-27
Bug 2026930 - Make CONSOLE_ADDRESS_PREF the only pref in ConsoleClient

1rneh:
I've seen non-console preference being added to `PREFS` in the `ConsoleClient`, which I'd like to avoid since the console address preference is the only preference that the `ConsoleClient` needs to know about.


lissyx (approved):
We should likely document where it would make sense to place those others you saw :)

PR #620, opened 2026-03-27
Bug 2026954: Fix duplicated html id on separator

victhorlopez:


browser/components/enterprise/content/enterprise-panel.inc.xhtml
```diff
@@ -19,7 +19,7 @@
       <html:a id="enterprise-learn-more-link" class="learnMore" data-l10n-id="enterprise-panel-learn-more" />
     </vbox>
 
-    <toolbarseparator id="PanelUI-enterprise-separator" />
+    <toolbarseparator id="PanelUI-enterprise-separator-bottom" />
```
victhorlopez: see line 11, in theory [this](https://searchfox.org/enterprise-main/source/browser/components/enterprise/EnterpriseHandler.sys.mjs#148) code was hidding first separator only?

1rneh: Yes, it's supposed to only hide the first separator when we don't have an email yet to display.

1rneh: Can we rename them more descriptively, e.g.
- the first one: `<toolbarseparator id="PanelUI-enterprise-email-separator-" />`
- the second one: `<toolbarseparator id="PanelUI-enterprise-info-separator-" />`

1rneh (approved):
Nice, thank you!

PR #622, opened 2026-03-27
Bug 2026214 - Fix BlockAboutConfig and DisableDeveloperTools

lissyx:


1rneh:
Could you rename the PR to reflect that it's only providing a fix for `BlockAboutConfig`?

lissyx:
> Could you rename the PR to reflect that it's only providing a fix for `BlockAboutConfig`?

I will include more fix later today when I can look over :)

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -530,13 +530,19 @@ export var Policies = {
...
     },
     onRemove(manager, oldParams) {
       if (oldParams) {
-        // if it was block, just unblock
+        // if it was blocked, just unblock
         unblockAboutPage(manager, "about:config");
         unsetAndUnlockPref("devtools.chrome.enabled");
+      } else {
+        blockAboutPage(manager, "about:config");
+        setAndLockPref("devtools.chrome.enabled", true);
```
jonathanmendez: This (and the other `onRemove` change) doesn't seem right to me? But also I'm wondering more broadly about how we should be handling `onRemove`...

From what I understood or can see from [the implementation](https://searchfox.org/enterprise-main/source/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#290-304), I think that `onRemove` is triggered when a policy is "unset". In that case I think we should revert to default behavior. I don't think the default behavior should depend on what the previous policy setting was, whereas this code either unblocks or blocks the page depending on what the previous setting was.

In the goal of reverting to default behavior, It may be that some policies may need to act differently depending on the previous policy setting, although I think we still have to be careful about that because you can have a sequence of ["Set A", "Set B", "Unset"] and in that case the `onRemove` would not have the context that A was ever set.

jonathanmendez (changes requested):

browser/components/enterprisepolicies/Policies.sys.mjs
lissyx: > This (and the other `onRemove` change) doesn't seem right to me? But also I'm wondering more broadly about how we should be handling `onRemove`...
> 
> From what I understood or can see from [the implementation](https://searchfox.org/enterprise-main/source/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#290-304), I think that `onRemove` is triggered when a policy is "unset". 

Yes, that is right

> In that case I think we should revert to default behavior. I don't think the default behavior should depend on what the previous policy setting was, whereas this code either unblocks or blocks the page depending on what the previous setting was.

Problem is there is no proper definition of "default behavior", so when we implemented this, we decided to settle for that compromise of keeping track of previous values and `onRemove` should go back to that.

> 
> In the goal of reverting to default behavior, It may be that some policies may need to act differently depending on the previous policy setting, although I think we still have to be careful about that because you can have a sequence of ["Set A", "Set B", "Unset"] and in that case the `onRemove` would not have the context that A was ever set.

Yes, we have to be careful, and in this specific case, I'm also not 100% sure on what exactly we want to do, this is why I commented there https://github.com/mozilla/enterprise-firefox/pull/626#issuecomment-4145252248 for the more general review of policies.

lissyx: So here the question we need to answer is:
> when the policy is removed and the previous value was `false` what is the behavior we want?

lissyx: @jonathanmendez Also keep in mind policies were designed with the assumption of being applied once at startup

lissyx: Like maybe in the case of `about:config` blocking, we want that the "default" behavior is it is not blocked, so in case of removal, we always unblock things?

I dont know if we have a proper definition of what is `default` behavior for any of the policies yet, maybe @mkaply knows more?

1rneh: > Like maybe in the case of about:config blocking, we want that the "default" behavior is it is not blocked, so in case of removal, we always unblock things?

Yes - exactly. For both policies, `BlockAboutConfig` and `DisableDeveloperTools`, removing them should fully restore the original behavior: no `about:` pages should remain blocked and no features should remain disabled.

> I dont know if we have a proper definition of what is default behavior for any of the policies yet, [...]

Since both policies rely on `setAndLockPref`, we can use `unsetAndUnlockPref` to revert the preference changes since we cached the previous default and user preference. For policies that modify preferences without using `PoliciesUtils.setDefaultPref`, I believe we currently don’t have a mechanism to restore the original default state.

1rneh (changes requested):

browser/components/enterprisepolicies/Policies.sys.mjs
1rneh: We want to reset unconditionally, otherwise we'll never unlock the preference if `oldParams == false`

1rneh: Same here.

1rneh (changes requested):

browser/components/enterprisepolicies/Policies.sys.mjs
lissyx: Sure

testing/enterprise/felt_tests.py
```diff
@@ -525,7 +532,7 @@ def setUp(self):
         self._manually_closed_child = False
         self.console_port = random.randrange(10000, 14999)
         self.sso_port = random.randrange(15000, 20000)
-        self.policy_block_about_config = Value("B", 1)
+        self.policy_block_about_config = Value("b", 1)
```
1rneh, line 527: Why are we doing this change?

testing/enterprise/felt_tests.py
```diff
@@ -199,8 +199,15 @@ def do_GET(self):
...
                 False if self.server.policy_block_about_config.value == 0 else True
             )
+
+            # Reflect the states:
+            #  - "Unset" is -1, no value is pushed
+            #  - "False" is 0
+            #  - "True" is 1
             policy_content.update(
-                {"BlockAboutConfig": policy_value} if policy_value else {}
+                {}
```
1rneh: If `self.server.policy_block_about_config.value == -1`, we shouldn't add an empty object to `policy_content` but just leave it out.

testing/enterprise/felt_tests.py
```diff
@@ -199,8 +199,15 @@ def do_GET(self):
...
+                if self.server.policy_block_about_config.value == -1
+                else {"BlockAboutConfig": policy_value}
             )
 
             policy_value = self.server.policy_extensions.value == 1
```
1rneh: Can we avoid reusing the same variable for multiple policies?

testing/enterprise/test_felt_browser_about_config_blocked.py
```diff
@@ -16,60 +16,83 @@
...
-                    self._logger.info(f"Policy update propagated at {url}!")
+                policy_value = self.get_server_value()
+                if policy_value is None or policy_value == new_value:
                     break
-                self._logger.info(f"Policy update not yet propagated at {url}")
-                time.sleep(0.5)
+                self._logger.info(
+                    f"Policy update not yet propagated: received {policy_value}, but expecting {new_value}"
+                )
+                time.sleep(1)
```
1rneh: Why do we need this loop?

testing/enterprise/test_felt_browser_about_config_blocked.py
1rneh:
```suggestion
        assert warning is not None, "Expected access to about:config to be allowed."
```

testing/enterprise/test_felt_browser_about_config_blocked.py
```diff
@@ -16,60 +16,83 @@
 
 class BrowserAboutConfigBlocked(FeltTests):
     def test_browser_about_config_blocked(self):
-        super().run_felt_base()
+        self.run_felt_base()
+        self.connect_child_browser()
+
+        self._logger.info("Default blocked, should block")
```
1rneh: This is not the default state, but rather that the initial set of policy has already been fetched and applied.

1rneh (commented):
Some questions, as I think the test implementation could be improved.

testing/enterprise/felt_tests.py
lissyx: `B` is usigned we want to be able to pass `-1`

testing/enterprise/test_felt_browser_about_config_blocked.py
lissyx: To make sure the server side reflects properly the value. I had to add it in the past to avoid raciness

testing/enterprise/test_felt_browser_about_config_blocked.py
lissyx: Yes, I need to change the message because it is confusing you're right

testing/enterprise/test_felt_browser_about_config_blocked.py
1rneh: You mean setting `self.policy_block_about_config.value` was racy?

lissyx: Yes, but it is all things dating from a long time now

testing/enterprise/felt_tests.py
```diff
@@ -196,27 +196,28 @@ def do_GET(self):
...
-                {
+            # Reflect the states:
+            #  - "Unset" is -1, no value is pushed
+            #  - "False" is 0
+            #  - "True" is 1
+            if self.server.policy_block_about_config.value >= 0:
+                policy_content.update({
+                    "BlockAboutConfig": True
+                    if self.server.policy_block_about_config.value == 1
+                    else False
```
1rneh: `True if self.server.policy_block_about_config.value == 1 else False` 
is 
`self.server.policy_block_about_config.value == 1` :)

testing/enterprise/test_felt_browser_about_config_blocked.py
```diff
@@ -16,16 +16,58 @@
...
+        r = requests.get(
+            f"http://localhost:{self.console_port}/api/browser/policies",
+            headers={"Authorization": f"Bearer {self.policy_access_token.value}"},
         )
+        j = r.json()
+
+        if not ("BlockAboutConfig" in j["policies"]):
+            return None
+
+        return int(j["policies"]["BlockAboutConfig"])
```
1rneh: `get_server_value` seems to be unused now?

testing/enterprise/test_felt_browser_about_config_blocked.py
```diff
@@ -16,16 +16,58 @@
...
+
+    def run_change_about_config_policy(self, new_value):
+        self._logger.info("Changing BlockAboutConfig policy")
+        self.policy_block_about_config.value = new_value
+
+        # Give time to make sure Policy got applied
+        time.sleep(2)
```
1rneh: Can we make this dependent on the polling interval + 1000ms?

testing/enterprise/test_felt_browser_about_config_blocked.py
lissyx: indeed

testing/enterprise/felt_tests.py
lissyx: yes

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: The default behavior should just be the way regular Firefox works. I agree with @1rneh.

lissyx: > The default behavior should just be the way regular Firefox works. I agree with @1rneh.

Thanks, that was my assumption as well, but I wanted we had the opportunity to make sure I'm not taking decisions on my own

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -540,15 +540,15 @@ export var Policies = {
...
         setAndLockPref("devtools.chrome.enabled", false);
-      }
-    },
-    onRemove(manager, oldParams) {
-      if (oldParams) {
-        // if it was block, just unblock
+      } else {
         unblockAboutPage(manager, "about:config");
-        unsetAndUnlockPref("devtools.chrome.enabled");
+        setAndLockPref("devtools.chrome.enabled", true);
```
jonathanmendez, line 545: This could conflict with line 1221 if `BlockAboutConfig: False` and `DisableDeveloperTools: True` (or the opposite cases can also conflict). I'm guessing there's no precedence, so the winner will be determined by the order the policies are processed?

1rneh (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: That's a pretty unlikely scenario, and yeah we've basically left people to manage this. 

If you're curious why we disable devtools

https://bugzilla.mozilla.org/show_bug.cgi?id=1429185#c38

Maybe at some point we can just fix.

PR #625, opened 2026-03-27
Bug 2019068 - When crash autosubmit policy is applied, set env var to trigger autosubmit in CrashReporter app

jonathanmendez:
Triggers the UI and behavior changes from #614 and #494

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -1076,6 +1076,7 @@ export var Policies = {
         );
         setAndLockPref("browser.tabs.crashReporting.sendReport", true);
         setAndLockPref("browser.tabs.crashReporting.includeURL", true);
+        Services.env.set("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT", "1");
       }
     },
     onRemove(_manager, _oldParam) {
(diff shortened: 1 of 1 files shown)


lissyx (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
lissyx, line 1071: @jonathanmendez What is the value is `false`? We should enforce it

jonathanmendez: See https://github.com/mozilla/enterprise-firefox/pull/530#discussion_r2947650460 and https://bugzilla.mozilla.org/show_bug.cgi?id=2026983
In short, we've deferred the false case for later.

However, I'm seeing that the `onRemove` code isn't structured correctly so I'll add that fix here.

lissyx: ok, i trust you, but just FYI I'm currently fixing https://bugzilla.mozilla.org/show_bug.cgi?id=2026214 where this is exactly the problem

jonathanmendez: Actually, given that we structured the policy differently to prepare for bug 2026983, you're right that we need to handle the false or unset case for the ForceAutoSubmit property of this policy. I've updated the PR to handle it for the current expectations.

jporter-dev (approved):

PR #626, opened 2026-03-27
Bug 2027104 - Enterprise: fix Policies application

lissyx:


lissyx:
@mkaply This was very mechanical of a work and made thinking about the states that the server can send us. Maybe there are a lot of cases where we should not care of some values being sent?

mkaply:
> @mkaply This was very mechanical of a work and made thinking about the states that the server can send us. Maybe there are a lot of cases where we should not care of some values being sent?

It's certainly possible that a lot of these we don't need to worry about "false" but we always need to worry about onRemove.

mkaply (approved):
This looks correct to me.

I think what we might want to build is some simple way to flip policies on and off quickly in product for testing. I'm going to investigate that.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -1973,9 +2017,7 @@ export var Policies = {
 
   ExtensionUpdate: {
     onBeforeAddons(manager, param) {
-      if (!param) {
-        setAndLockPref("extensions.update.enabled", param);
-      }
+      setAndLockPref("extensions.update.enabled", !param);
```
jonathanmendez, line 2020: This logic doesn't match what was previously there in the false case.
Previously:
if (param is false) setPref(false)
Now:
if (param is false) setPref(true)

(the previous logic didn't handle the true case but presumably it would not want to set it to false as the new logic does)

jonathanmendez: https://github.com/mozilla/enterprise-firefox/pull/685

PR #628, opened 2026-03-30
Enterprise release uplift 2026033000

lissyx:
~~1750a3291cb7 Bug 2012458 - Policies for AI Chat Sidebar (#623)~~
a3e29d7a4f09 Bug 2019068 - Fix onRemove and false case for CrashReportsSubmit policy
285193edbe54 Bug 2027073: Load startup image when it is visible(#624)
dcafa1e30151 Bug 2019068 - When crash autosubmit policy is applied, set env var to trigger autosubmit in CrashReporter app

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -1019,6 +1019,7 @@ export var Policies = {
         );
         setAndLockPref("browser.tabs.crashReporting.sendReport", true);
         setAndLockPref("browser.tabs.crashReporting.includeURL", true);
+        Services.env.set("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT", "1");
       }
     },
     onRemove(_manager, _oldParam) {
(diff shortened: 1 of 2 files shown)


gcp (changes requested):
https://bugzilla.mozilla.org/show_bug.cgi?id=2019983 changes how some of this works, including undoing the blocked state that looks like it won't work.

We'll have to pull this out until the above gets uplifted.

lissyx:
> https://bugzilla.mozilla.org/show_bug.cgi?id=2019983 changes how some of this works, including undoing the blocked state that looks like it won't work.
> 
> We'll have to pull this out until the above gets uplifted.

OK, this is on `enterprise-main` but not on release, so I think I need to include it

lissyx:
> > https://bugzilla.mozilla.org/show_bug.cgi?id=2019983 changes how some of this works, including undoing the blocked state that looks like it won't work.
> > We'll have to pull this out until the above gets uplifted.
> 
> OK, this is on `enterprise-main` but not on release, so I think I need to include it

Does not apply cleanly

gcp (approved):
rs+

PR #630, opened 2026-03-30
Bug 2027620: Reduce timeout on crash test to avoid 360 secs wait

victhorlopez:
Same as https://searchfox.org/firefox-main/source/testing/marionette/harness/marionette_harness/tests/unit/test_crash.py#101

### DIFF
--- a/testing/enterprise/felt_browser_crashes.py
+++ b/testing/enterprise/felt_browser_crashes.py
@@ -13,17 +13,22 @@ from felt_tests import FeltTests
 
 class BrowserCrashes(FeltTests):
     EXTRA_ENV = {"MOZ_GDB_SLEEP": "1"}
+    # Reduce the timeout for faster processing of the tests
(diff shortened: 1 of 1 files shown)


testing/enterprise/felt_browser_crashes.py
```diff
@@ -13,17 +13,22 @@
 
 class BrowserCrashes(FeltTests):
     EXTRA_ENV = {"MOZ_GDB_SLEEP": "1"}
+    # Reduce the timeout for faster processing of the tests
+    socket_timeout = 10
```
lissyx, line 17: Maybe 10 is too short in itself, but should at least allows to verify the theory here. Let's hammer taskcluster and if it is proven to work increase a bit, 30s may be a good compromise (I've seen things taking some time on CI on Windows on a regular basis)

lissyx (approved):

PR #631, opened 2026-03-30
Bug 2021009 - mach try

lissyx:


ahal (changes requested):
This isn't how to handle Mach settings. You should add a new entry to this dict instead:
https://searchfox.org/firefox-main/source/python/mach/mach/settings.py#98

Then access it like:
https://searchfox.org/firefox-main/source/tools/tryselect/mach_commands.py#49

ahal (approved):

PR #632, opened 2026-03-30
Bug 2026598 - Enterprise: remove felt profile on Windows

lissyx:


lissyx:
Not sure we can do anything for that on macOS

PR #638, opened 2026-03-31
Bug 2027744 - Enterprise: Ensure Policies.Sync proper alphabetical or…

lissyx:
…dering


1rneh (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -4276,16 +4286,6 @@ function processMIMEInfo(mimeInfo, realMIMEInfo) {
...
-  });
-  Policies.Sync = {
-    async onBeforeAddons(manager, param) {
-      await lazy.SyncPolicy.applySettings(manager, param);
-    },
-    async onRemove(manager, _) {
-      await lazy.SyncPolicy.restoreSettings(manager);
-    },
-  };
+if (!AppConstants.MOZ_ENTERPRISE) {
```
1rneh, line 4289: :( 
I didn't submit the comment: 

`Sync` is not the only policy that would need exclusion. The plan is to define the compatibility in the `policies-schema.json` (see https://bugzilla.mozilla.org/show_bug.cgi?id=2012084 and https://bugzilla.mozilla.org/show_bug.cgi?id=2008554#c3). So I think we can remove this block statement.

PR #639, opened 2026-03-31
Bug 2012442 - Closing Firefox logs out the user from Firefox Enterprise

gcp:


gcp:
The approach with async shutdown blockers that was proposed in the original patch is problematic because of: https://searchfox.org/firefox-main/rev/98bf4b92d3f5d7a9855281df4bf333210bcfbbc4/dom/xhr/XMLHttpRequestMainThread.cpp#1555

`eAttemptQuit` and `eForceQuit` use the shared code to run normal shutdown: https://searchfox.org/firefox-main/source/toolkit/components/startup/nsAppStartup.cpp#359

I need to manually test the edge cases on macOS still.

PR #640, opened 2026-03-31
Bug 2027777- Fetch Firefox configurations from api/browser/config

1rneh:
#### Changes
- Removes support for `api/browser/default` endpoint.
- Fetched Firefox configrations from `api/browser/config` endpoint.
- Sends configuration to Firefox via preferences through felt.

#### Tests
- [run_ensure_config_prefs_set_in_browser](https://searchfox.org/enterprise-main/rev/9042fdcca27a917ffddfd171644e103c676b4e58/testing/enterprise/felt_browser_starts.py#40-45) in `felt_browser_starts.py` tests that the config preferences are set in Firefox.

### DIFF
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -116,7 +116,7 @@ export const ConsoleClient = {
       SSO: "/sso/login",
       SIGNOUT: "/sso/logout",
       SSO_CALLBACK: "/sso/callback",
-      DEFAULT_PREFS: "/api/browser/hacks/default",
+      CONFIG: "/api/browser/config",
       REMOTE_POLICIES: "/api/browser/policies",
       KEY: "/api/browser/key",
       TOKEN: "/sso/token",
@@ -190,11 +190,13 @@ export const ConsoleClient = {
     return url.href + "?*";
   },
 
-  // prefs that do not need to be written and can be sent during runtime
-  // tbd: remove
-  async getDefaultPrefs() {
-    const payload = await this._get(this._paths.DEFAULT_PREFS);
-    return payload;
+  /**
+   * Fetches configurations for Firefox
+   *
+   * @returns {Promise<object>}
+   */
+  async getFirefoxConfigs() {
+    return this._get(this._paths.CONFIG);
   },
 
   /**
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -276,6 +276,75 @@ export class FeltProcessParent extends JSProcessActorParent {
     );
   }
 
+  /**
+   * Fetches the configurations for Firefox and sends
+   * each configuration point over to Firefox as preferences
+   */
+  async _applyFirefoxConfigs() {
+    const {
+      learn_more_url,
+      company_logo_url,
+      policies: { polling_frequency },
+      services: { push_url, remote_settings_url, tokenserver_url },
+      extra_prefs,
+    } = await lazy.ConsoleClient.getFirefoxConfigs();
+
+    Services.felt.sendStringPreference(
+      "enterprise.configs.learn_more_url",
+      learn_more_url
+    );
+    Services.felt.sendStringPreference(
+      "enterprise.configs.company_logo_url",
+      company_logo_url
+    );
+    Services.felt.sendIntPreference(
+      "browser.policies.live_polling.frequency",
+      polling_frequency
+    );
+
+    Services.felt.sendStringPreference(
+      "identity.sync.tokenserver.uri",
+      tokenserver_url
+    );
+    Services.felt.sendStringPreference(
+      "services.settings.server",
+      remote_settings_url
+    );
+    Services.felt.sendStringPreference("dom.push.serverURL", push_url);
+
+    extra_prefs.forEach(pref => {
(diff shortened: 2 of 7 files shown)


testing/enterprise/felt_consts.py
```diff
@@ -3,8 +3,28 @@
...
+        "pref_id": "identity.sync.tokenserver.uri",
+        "pref_value": "",
+    },
+    "push_url": {
+        "pref_id": "dom.push.serverURL",
+        "pref_value": "",
+    },
+    # Not checking remote settings url (services.settings.server),
+    # it's pre-populated in marionette test environemtns:
+    #  https://searchfox.org/firefox-main/rev/9a3317a65545e83f4e32b94fdf1f6860342423ef/remote/shared/RecommendedPreferences.sys.mjs#381-382
```
lissyx, line 29: maybe worth asserting in our tests anyway ?

1rneh: The pref `services.settings.server` didn't get overridden with the config value in our marionette test.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -190,11 +190,13 @@ export const ConsoleClient = {
...
-  async getDefaultPrefs() {
-    const payload = await this._get(this._paths.DEFAULT_PREFS);
-    return payload;
+  /**
+   * Fetches configurations for Firefox
+   *
+   * @returns {Promise<object>}
+   */
+  async getFirefoxConfigs() {
+    return this._get(this._paths.CONFIG);
```
jonathanmendez, line 199: nit: This now sticks out as being different than the others that assign the awaited result to a payload and then return. It would be nice if they had consistency.
I would speculate the assign/return pattern is used for easier debugging?

testing/enterprise/test_felt_browser_about_config_blocked.py
```diff
@@ -51,17 +51,7 @@ def run_change_about_config_policy(self, new_value):
         self.policy_block_about_config.value = new_value
 
         # Polling frequency + 1s, defaulting to 2s in total if missing pref
```
jonathanmendez: I don't think the "defaulting to 2s" part of this comment is true anymore with these changes?

testing/enterprise/felt_consts.py
```diff
@@ -3,8 +3,28 @@
...
+        "pref_value": 500,
+    },
+    "tokenserver_url": {
+        "pref_id": "identity.sync.tokenserver.uri",
+        "pref_value": "",
+    },
+    "push_url": {
+        "pref_id": "dom.push.serverURL",
+        "pref_value": "",
+    },
```
lissyx, line 26: how much more tests will break :)

lissyx (approved):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -276,6 +276,75 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+   * each configuration point over to Firefox as preferences
+   */
+  async _applyFirefoxConfigs() {
+    const {
+      learn_more_url,
+      company_logo_url,
+      policies: { polling_frequency },
+      services: { push_url, remote_settings_url, tokenserver_url },
+      extra_prefs,
+    } = await lazy.ConsoleClient.getFirefoxConfigs();
```
jonathanmendez, line 290: Looking at the [console response type](https://github.com/mozilla/enterprise-console-backend/blob/main/crates/ec_webserver/src/routes/browser/config/mod.rs#L16-L22), I think that it's possible to get "null" returned in the json for the `Option` fields. Do you know what happens if you get null and pass it along to e.g. `sendStringPreference`? Does it get gracefully handled as an empty string?

jonathanmendez (approved):
Approved assuming there's a reasonable answer to the "null" handling. The other comments are nits.

testing/enterprise/test_felt_browser_about_config_blocked.py
lissyx: true

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: All methods should return the promise directly instead of awaiting it only for the async function to wrap it again. I think when I wrote `ConsoleClient`, I was logging the responses and missed refactoring this afterward.

1rneh: I filed a follow-up bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2028206

browser/extensions/felt/content/FeltProcessParent.sys.mjs
1rneh: You're right. We'd be breaking the contract: https://searchfox.org/enterprise-main/rev/2196149e9a22890c3834e716fa1bc6bee7d21d5e/browser/extensions/felt/rust/nsIFelt.idl#32-39. I've updated `_applyFirefoxConfigs` to guard against `null` values. 

Note: Long-term goal is to add schema validation for console API calls: https://bugzilla.mozilla.org/show_bug.cgi?id=2000598

jonathanmendez: This works for now and we can revisit something possibly cleaner in the future when looking at schema validation etc.

victhorlopez (approved):

PR #643, opened 2026-03-31
Bug 2018561 - Support configurable learn more link

1rneh:
#### Changes

- Adds functions to validate https url and data url in `EntepriseHandler`
- Validated learn more url to be delivered over https (exception in test environment)
- Removes default `ConsoleClient.learnMoreURI` support

#### Notes
- ~Only commit [81f97ff](https://github.com/mozilla/enterprise-firefox/pull/643/commits/81f97ffee789bbbd425e7ffccaf7a1be578fe56e) is up for review in this PR~
- ~Commit [5ef0381](https://github.com/mozilla/enterprise-firefox/pull/643/commits/5ef03816254d0b682cfcfb1d8b4708ab39162926) will land as part of https://github.com/mozilla/enterprise-firefox/pull/640~


browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -128,25 +210,39 @@ export const EnterpriseHandler = {
...
-        e.preventDefault();
-
-        const panel = document
-          .getElementById("panelUI-enterprise")
-          .closest("panel");
-        win.PanelMultiView.hidePopup(panel);
-      });
+      if (!learnMoreUrl) {
+        lazy.log.warn("No learn more url available.");
+        return;
```
jonathanmendez: Do we really want to early return here and skip setting up the email part of the panel?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -26,6 +27,81 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+  }
+}
+
+/**
+ * Validate that the URL is HTTPS and hosted on the console host.
+ *
+ * @param {string} url - The URL string to validate.
+ * @returns {URL|null} A parsed `URL` object if validation succeeds, otherwise `null`.
+ */
+function validateHttpsUrl(url) {
```
jonathanmendez: If you have other changes to make too, probably rename this to something like `validateHttpsConsoleUrl`

browser/components/enterprise/EnterpriseHandler.sys.mjs
jonathanmendez: You could set up a helper `setupLearnMoreLink()` that can take advantage of early returns for that cleaner logic while not affecting the rest of the setup.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -244,39 +340,27 @@ export const EnterpriseHandler = {
...
       return;
     }
 
-    let validLogoUrl;
-    try {
-      validLogoUrl = new URL(logoUrl);
-    } catch {
-      throw new Error(`Invalid logo URL in pref: ${logoUrl}`);
-    }
+    const validLogoUrl = validateDataUrl(logoUrl);
```
jonathanmendez, line 368: Before this change, the logo url could be an `https:` url at the console. With this change, that is no longer the case (only a `data:` url, which does not need to be at the console). Is that an intentional change?

jonathanmendez (changes requested):

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: > Do we really want to early return here and skip setting up the email part of the panel?

Good catch! We don't want to skip the email setup part. I implemented the `setupLearnMoreLink` helper method.

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: The original "requirement" (sounds like a more formal and more official decision that it was in reality) for the company logo URL (and how the console currently provides it) was a data URL. It was implemented differently. I'd suggest we can reintroduce it if we decide to support HTTPS URLs as well?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -26,6 +27,81 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+function validateHttpsConsoleUrl(url) {
+  const parsedUrl = parseUrl(url);
+
+  if (!parsedUrl) {
+    return null;
+  }
+
+  const isLocalTest =
+    lazy.isTesting() &&
+    (parsedUrl.hostname === "localhost" || parsedUrl.hostname === "127.0.0.1");
```
lissyx, line 62: Maybe we could serve the mock console over https, i've already experimented with that and it was not so much changes

lissyx: (I still have the local diff I made for that, though it includes a bit more)

testing/enterprise/felt_tests.py
lissyx, line 518: To add https:

```suggestion
    cert_pem = os.path.join(os.path.dirname(__file__), ssl_cert_pem)
    print(f"Loading cert from {cert_pem}")
    key_pem = os.path.join(os.path.dirname(__file__), "key.pem")
    print(f"Loading key from {key_pem}")

    context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
    context.load_cert_chain(certfile=cert_pem, keyfile=key_pem)
    context.check_hostname = False

    httpd.socket = context.wrap_socket(httpd.socket, server_side=True)
    httpd.serve_forever()
```

lissyx: Of course we need cert/key

lissyx:
```
$  openssl genpkey -algorithm RSA -out testing/enterprise/key.pem -pkeyopt rsa_keygen_bits:2048
[...]
$ openssl req -x509 -new -key testing/enterprise/key.pem -days 30 -out testing/enterprise/cert.pem -subj "/CN=localhost"
```

And we can add them to the tree

1rneh:
@lissyx Would you be okay with porting the console to be served over HTTPS as a follow-up? This feels beyond the scope of this PR, and I suspect it may involve additional tweaks.

lissyx:
> @lissyx Would you be okay with porting the console to be served over HTTPS as a follow-up? This feels beyond the scope of this PR, and I suspect it may involve additional tweaks.

TBH it was a trivial change outside of updating URLs from `http` to `https`, but if you feel it's too much for this one, follow up it is :)

lissyx:
> @lissyx Would you be okay with porting the console to be served over HTTPS as a follow-up? This feels beyond the scope of this PR, and I suspect it may involve additional tweaks.

[bug 2028280](https://bugzilla.mozilla.org/show_bug.cgi?id=2028280)

browser/components/enterprise/EnterpriseHandler.sys.mjs
jonathanmendez: Sounds good. Just wanted to double-check it wasn't an accidental oversight.

jonathanmendez (approved):

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -124,22 +206,38 @@ export const EnterpriseHandler = {
...
+   * Retrieves, validates, and applies the learn more URL to the link element.
+   * Leaves the link unconfigured if missing or invalid.
+   *
+   * @param {Window} win - chrome window
+   * @returns {void}
+   */
+  _setupLearnMoreLink(win) {
+    const learnMoreUrl = Services.prefs.getStringPref(LEARN_MORE_URL_PREF);
+
+    if (!learnMoreUrl) {
```
gcp, line 219: @1rneh note the above throws if the pref isn't set because there is no default configured in the call, you won't get to this branch

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -148,6 +246,16 @@ export const EnterpriseHandler = {
...
       });
     }
+  },
+
+  openPanel(element, event) {
+    const win = element.ownerGlobal;
+    win.PanelUI.showSubView("panelUI-enterprise", element, event);
+    const document = element.ownerDocument;
+
+    if (!this._isLearnMoreLinkConfigured) {
```
gcp, line 256: this == `EnterpriseHandler`? Should be set on the window where the link lives instead?

gcp: Trying to test this by opening a new window but it doesn't even work for me in the original one :-/

jonathanmendez: The current url served by staging is `https://github.com/mozilla/enterprise-firefox/` which fails `validateHttpsConsoleUrl()` because it's on a different domain than the console.

jonathanmendez: Removed the console domain check in a local build and encountered the expected problem in a new window, i.e. the 'learn more' link works in the first window but not in the new/second window.

jonathanmendez: https://github.com/mozilla/enterprise-firefox/pull/662

1rneh: Thanks for catching the issues!
@fiji-flo Will the current learn more link (`https://github.com/mozilla/enterprise-firefox/`) be replaced by a console derived url? I implemented it as a requirement. Or should we loosen that requirement?

fiji-flo: I think this would point to the companies wiki explaining what policies they have in place. So external link. I talked with Brian and he recommended https://support.mozilla.org/kb/managed-browser-firefox as a fallback.

PR #646, opened 2026-03-31
Bug 2028166 - Downloaded updates from Felt notify Forced Update policy

gcp:


browser/modules/test/browser/browser_UpdatePolicyEnforcer.js
```diff
@@ -58,3 +62,38 @@ add_task(async function test_compulsoryRestartNotification() {
...
+    Assert.equal(1, win.gNotificationBox.allNotifications.length);
+    Assert.equal(
+      "COMPULSORY_RESTART_SCHEDULED",
+      win.gNotificationBox.allNotifications[0].getAttribute("value")
+    );
+  } finally {
+    testingOnly_resetTasks();
+    await popPrefs();
+  }
+});
```
lissyx, line 92: This test will not be visible until we green suites

lissyx: Are we fine with no `marionette-enterprise` test ?

browser/modules/UpdatePolicyEnforcer.sys.mjs
```diff
@@ -209,6 +211,7 @@ const observer = {
     switch (topic) {
       case "update-downloaded":
       case "update-staged":
+      case "felt-update-downloaded":
```
lissyx: Shouldn't we forward also `update-staged` ?

lissyx: So basically we forward the message from felt and the component here decides whether it should show the information?

browser/extensions/felt/rust/src/message.rs
```diff
@@ -89,6 +89,7 @@ pub enum FeltMessage {
     Logout(LogoutType),
     Exiting,
     UpdateReady,
+    UpdateDownloaded,
```
lissyx: I'm wondering if we should not just rely on `UpdateReady` that should already be sent

browser/modules/UpdatePolicyEnforcer.sys.mjs
```diff
@@ -218,5 +221,8 @@ export const UpdatePolicyEnforcer = {
   registerObservers() {
     Services.obs.addObserver(observer, "update-downloaded");
     Services.obs.addObserver(observer, "update-staged");
+    if (AppConstants.MOZ_ENTERPRISE) {
+      Services.obs.addObserver(observer, "felt-update-downloaded");
```
lissyx: `felt-update-ready` shouldn't already cover this?

testing/enterprise/test_felt_browser_compulsory_restart.py
```diff
@@ -0,0 +1,58 @@
...
+            """
+            Services.prefs.setStringPref(
+                "app.update.compulsory_restart",
+                JSON.stringify({
+                    NotificationPeriodHours: 0,
+                    RestartTimeOfDay: { Hour: 0, Minute: 30 },
+                })
+            );
+            """
+        )
```
lissyx: nit: `self._child_driver.using_prefs()` ?

https://searchfox.org/enterprise-main/rev/2196149e9a22890c3834e716fa1bc6bee7d21d5e/testing/marionette/client/marionette_driver/marionette.py#982-1005

testing/enterprise/test_felt_browser_compulsory_restart.py
```diff
@@ -0,0 +1,58 @@
...
+        self.connect_child_browser()
+
+        self._child_driver.set_context("chrome")
+
+        self._child_driver.execute_script(
+            """
+            Services.prefs.setStringPref(
+                "app.update.compulsory_restart",
+                JSON.stringify({
+                    NotificationPeriodHours: 0,
```
lissyx: so this will immediately show?

lissyx: looks so: https://searchfox.org/enterprise-main/rev/2196149e9a22890c3834e716fa1bc6bee7d21d5e/browser/modules/UpdatePolicyEnforcer.sys.mjs#165-182,189,193-197,199

lissyx (approved):

testing/enterprise/test_felt_browser_compulsory_restart.py
gcp: I'll make a test that checks before and after a delay but I'm worried that's going to end up a low-frequency intermittent.

lissyx: yes, exactly what i was worrying, but if it's immediate imho that's enough

gcp: Can't even do that as the minimum period is 1 hour.

testing/enterprise/test_felt_browser_compulsory_restart.py
```diff
@@ -0,0 +1,52 @@
...
+                    return null;
+                    """
+                )
+            )
+
+            assert notification_value == "COMPULSORY_RESTART_SCHEDULED", (
+                "felt-update-ready should trigger the compulsory restart notification"
+            )
+
+            self._child_driver.set_context("content")
```
lissyx: do we need this under the context manager? I've had cases where with code like that, the shutdown phase would trip on the context manager for NoSuchWindowException

PR #647, opened 2026-03-31
Bug 2028280 - Enterprise: use https for mock console

lissyx:


lissyx:
Context here: https://bugzilla.mozilla.org/show_bug.cgi?id=2018561 when reviewing this I suggested that instead of hacking to allow only localhost http for our mock console, we should just move the console to https. It works :)

jonathanmendez (commented):
I'm not super familiar with the various pieces you've included for the https setup. The changes to the various other files seem ok. But why is the setup done this way instead of as specified in https://firefox-source-docs.mozilla.org/build/buildsystem/test_certificates.html ?

lissyx:
> I'm not super familiar with the various pieces you've included for the https setup. The changes to the various other files seem ok. But why is the setup done this way instead of as specified in https://firefox-source-docs.mozilla.org/build/buildsystem/test_certificates.html ?

The PGO certificates thing depends on commands that may not be available at runtime on CI, `certutil` and others are not packages in the tests package we download

PR #648, opened 2026-03-31
Bug 2028303 - Update list of expected errors

lissyx:


browser/components/tests/marionette/test_no_errors_clean_profile.py
```diff
@@ -84,6 +84,9 @@
     {"message": "FeltExtension: _refreshSession()"},
     {"message": "Unable to update user icon in badge without user information"},
     {"message": "enterprise.logo_url pref is not set, skipping logo update"},
+    {
+        "message": "Unable to retrieve company logo url from: enterprise.configs.company_logo_url"
```
jonathanmendez, line 88: Maybe instead of having to add this, [the `warn` should be changed to an `info`](https://github.com/mozilla/enterprise-firefox/blob/enterprise-main/browser/components/enterprise/EnterpriseHandler.sys.mjs#L362)? I don't know if it's a valid decision for a company to not want/provide a logo and thus the pref would be validly empty. @1rneh ?

1rneh: > I don't know if it's a valid decision for a company to not want/provide a logo and thus the pref would be validly empty.

I think it's expected that every customer has a logo.

Also, what’s the benefit of listing errors in this context? Could we also make the mock console return a base64 encoded asset?

lissyx: Those tests are not running with the mock console at all, it's pure marionette that does not need to know about the console (and adding a step with FELT and the console would be a huge task)

lissyx: For more context, see the bug and the failure, this is get back `Mn` test suite as green, bug [2028166](https://bugzilla.mozilla.org/show_bug.cgi?id=2028166) broke it. Error is at https://treeherder.mozilla.org/logviewer?job_id=557144114&repo=enterprise-main&task=IRuGQViJQJi15ku4_VveTA.0&lineNumber=24887

lissyx:
This has somehow disappeared from treeherder, let's close

PR #660, opened 2026-04-01
Bug 2028602 - Enterprise: ensure macOS reads MAR channel IDs correctly

lissyx:
This is WIP from bug 2028575 to unblock updates. We will revert and pick the final version when it lands.


lissyx:
> This is WIP from bug 2028575 to unblock updates. We will revert and pick the final version when it lands.

I did verify it was working locally.

PR #662, opened 2026-04-02
Bug 2028723, Bug 2028555 - Persist learn more link to each window and hide if invalid

jonathanmendez:
Also fixes bug 2028555 by providing default value for `getStringPref()`

### DIFF
--- a/browser/components/enterprise/EnterpriseHandler.sys.mjs
+++ b/browser/components/enterprise/EnterpriseHandler.sys.mjs
@@ -45,12 +45,12 @@ function parseUrl(url) {
 }
 
 /**
- * Validate that the URL is HTTPS and hosted on the console host.
+ * Validate that the URL is HTTPS.
  *
  * @param {string} url - The URL string to validate.
  * @returns {URL|null} A parsed `URL` object if validation succeeds, otherwise `null`.
  */
-function validateHttpsConsoleUrl(url) {
+function validateHttpsUrl(url) {
   const parsedUrl = parseUrl(url);
 
   if (!parsedUrl) {
@@ -63,10 +63,6 @@ function validateHttpsConsoleUrl(url) {
 
   if (parsedUrl.protocol !== "https:" && !isLocalTest) {
     lazy.log.warn(`Expected HTTPS URL: ${url}`);
-    return null;
-  }
-  if (parsedUrl.hostname !== lazy.ConsoleClient.consoleBaseURI.hostname) {
-    lazy.log.warn(`Expected URL hosted by the console origin: ${url}`);
     return null;
   }
 
@@ -114,12 +110,6 @@ export const EnterpriseHandler = {
    * from the signed in user has been received from the console.
    */
   _isInitialized: false,
-
-  /**
-   * Whether the panel has been opened once,
-   * which populates the learn more link
-   */
-  _isLearnMoreLinkConfigured: false,
 
   /**
    * Handles the enterprise state for each new browser window.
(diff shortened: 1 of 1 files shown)


browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -246,23 +255,15 @@ export const EnterpriseHandler = {
...
-      lazy.log.warn("No learn more url available.");
-      return;
-    }
-
-    const validLearnMoreUrl = validateHttpsConsoleUrl(learnMoreUrl);
+    const validLearnMoreUrl = this._retrieveLearnMoreLink();
 
+    const document = win.document;
+    const learnMoreLink = document.getElementById("enterprise-learn-more-link");
+    learnMoreLink.classList.toggle("hidden", validLearnMoreUrl === null);
```
1rneh: Seems like what you are trying to do with the `hidden` class + setting `display:none` is to hide the link and remove it from the DOM? You can toggle `is-hidden` for that. I can follow-up on that.

1rneh (approved):

browser/components/enterprise/EnterpriseHandler.sys.mjs
gcp: ```learnMoreLink.hidden = validLearnMoreUrl === null;```

Shouldn't need new CSS. See https://searchfox.org/firefox-main/source/dom/webidl/HTMLElement.webidl#35 and https://searchfox.org/firefox-main/source/dom/html/nsGenericHTMLElement.cpp#1363

gcp (changes requested):

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: Nevermind, debunked by https://github.com/mozilla/enterprise-firefox/pull/662#discussion_r3026150106 :)

1rneh: Right, I believe the extra css can be removed: https://github.com/mozilla/enterprise-firefox/pull/662/changes#diff-a2ec492ea1a5f414a4c3df77d2a3b60b6a9ef954f5298e7777ec83d8ca71c139R2480-R2484

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -238,6 +232,21 @@ export const EnterpriseHandler = {
     wrapper.classList.remove("is-hidden");
   },
 
+  /**
+   * Retrieves and validates the learn more URL.
+   * Returns null if the url is invalid.
+   */
+  _retrieveLearnMoreLink() {
+    const learnMoreUrl = Services.prefs.getStringPref(LEARN_MORE_URL_PREF, "");
```
victhorlopez, line 236: this also fixes this test bug  https://bugzilla.mozilla.org/show_bug.cgi?id=2028555

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -239,53 +229,59 @@ export const EnterpriseHandler = {
...
+   * Retrieves, validates, and applies the learn more URL to the link element.
+   * Falls back to "https://support.mozilla.org/kb/managed-browser-firefox" is no valid URL provided.
+   *
+   * @param {Window} win - chrome window
+   * @returns {void}
+   */
+  _setupLearnMoreLink(win) {
+    const validLearnMoreUrl =
+      this._retrieveLearnMoreLink() ??
+      parseUrl("https://support.mozilla.org/kb/managed-browser-firefox");
```
1rneh, line 256: Could we inject the fallback url into the html `https://searchfox.org/enterprise-main/rev/00288e25bd40457e5a040b3333760f0fb6d9635e/browser/components/enterprise/content/enterprise-panel.inc.xhtml#19`?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -239,53 +229,59 @@ export const EnterpriseHandler = {
...
+    });
   },
 
   openPanel(element, event) {
     const win = element.ownerGlobal;
     win.PanelUI.showSubView("panelUI-enterprise", element, event);
     const document = element.ownerDocument;
 
-    if (!this._isLearnMoreLinkConfigured) {
+    if (!win._isLearnMoreLinkConfigured) {
```
1rneh: Can we add this on the panel element instead of the window?

1rneh (changes requested):
Thanks for taking on some extra scope!

browser/components/enterprise/EnterpriseHandler.sys.mjs
jonathanmendez: We discussed offline that we don't have the panel handy, but I'll use a more descriptive/scoped name on the win's property

browser/components/enterprise/EnterpriseHandler.sys.mjs
jonathanmendez: Done.

1rneh (approved):

browser/components/enterprise/EnterpriseHandler.sys.mjs
jonathanmendez: Thanks for the help pointing out the better way. This has become irrelevant in the recent changes, since there will always be a link via a fallback.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -239,53 +229,59 @@ export const EnterpriseHandler = {
...
-      });
     }
+
+    win._isEnterpriseLearnMoreLinkConfigured = true;
+    learnMoreLink.addEventListener("click", e => {
+      let where = lazy.BrowserUtils.whereToOpenLink(e, false, false);
+      if (where == "current") {
+        where = "tab";
+      }
+      win.openTrustedLinkIn(validLearnMoreUrl.href, where);
```
gcp, line 269: This is outside the `null` check above, which seems unintended.

gcp (changes requested):

browser/components/enterprise/EnterpriseHandler.sys.mjs
jonathanmendez: Revisited this after https://github.com/mozilla/enterprise-firefox/pull/662#discussion_r3033137215

When considering how to fix that, to keep the code logic more simple and scannable, I went back to putting the default within the js code.

browser/components/enterprise/EnterpriseHandler.sys.mjs
jonathanmendez: Thanks for catching that! Went back to an earlier implementation that avoids the issue.

gcp (approved):

PR #667, opened 2026-04-02
 Bug 2017606  Policies json schema extensions

argl:
This adds `description`, `x-category` and `examples` fields to the policies schema. 

Categories are sourced from [here](https://docs.google.com/spreadsheets/d/1h-UmrLEBVmLVEl9eIVQlNkrP-NkgGXfxpzN5tJWCWv4/edit?gid=1674198094#gid=1674198094)
Descriptions are sourced from the [enterprise policies repository](https://github.com/mozilla/enterprise-admin-reference)
Examples are sourced from there as well, but cleaned up to be valid JSON.

`x-category` is a custom extension of JSON-Schema that is compliant with the standard. `description` and `examples` are official keywords in JSON-Schema and used accordingly.


mkaply (changes requested):
Where are these descriptions from?

Some seem to come from here:

https://searchfox.org/firefox-main/source/browser/locales/en-US/browser/policies/policies-descriptions.ftl

Some seem completely original.

I would rather they were all consistent.

So we should either use the versions in the ftl or update all the FTL.

Also, are examples really necessary on the true/false items?

fiji-flo:
@mkaply descriptions come from https://github.com/mozilla/enterprise-admin-reference according to the PR description and a small sample I looked at seemed to be matching.

For now this will be only used in the console and we can revisit this when we upstream?

mkaply (approved):

PR #669, opened 2026-04-02
No bug: Drain ipc before exit

victhorlopez:


testing/enterprise/test_felt_browser_signout_race.py
```diff
@@ -0,0 +1,93 @@
...
+class BrowserSignoutRace(FeltTests):
+    """Regression test for the race between proc.exitPromise and logoutReported.
+
+    When Firefox calls performNormalLogout() followed immediately by quit(), the
+    felt-firefox-logout notification is dispatched from the Rust ipc_loop thread
+    to the JS main thread. proc.exitPromise also resolves via the subprocess
+    worker on the main thread. Without the fix, if proc.exitPromise.then() checks
+    logoutReported before logoutFirefox() runs, it sends FirefoxNormalExit.
+    logoutFirefox() then also sends FirefoxLogoutExit. Both call showWindow(),
+    producing 2 windows.
```
lissyx, line 26: We should throw in this case, to ease discovery of those bugs

testing/enterprise/test_felt_browser_signout_race.py
```diff
@@ -0,0 +1,93 @@
...
+        # window is still open. After run_felt_base() FELT transitions to
+        # background and closes its window, making execute_script fail.
+        with self._driver.using_context("chrome"):
+            self._driver.execute_script(
+                """
+                const env = Cc["@mozilla.org/process/environment;1"]
+                  .getService(Ci.nsIEnvironment);
+                env.set("FELT_TEST_LOGOUT_DELAY_MS", "500");
+                """
+            )
```
lissyx, line 59: Shouldn't we just `EXTRA_ENV = {"FELT_TEST_LOGOUT_DELAY_MS": "500"}` at the class level?


```suggestion

class BrowserSignoutRace(FeltTests):
    EXTRA_ENV = {
        "FELT_TEST_LOGOUT_DELAY_MS": "500"
    }
```

testing/enterprise/test_felt_browser_signout_race.py
```diff
@@ -0,0 +1,93 @@
...
+from felt_tests import FeltTests
+from marionette_driver.errors import UnexpectedAlertOpen
+from marionette_driver.wait import Wait
+
+
+class BrowserSignoutRace(FeltTests):
+    """Regression test for the race between proc.exitPromise and logoutReported.
+
+    When Firefox calls performNormalLogout() followed immediately by quit(), the
+    felt-firefox-logout notification is dispatched from the Rust ipc_loop thread
```
lissyx, line 21: To make it easier to follow, please clarify in the comment that the notification is dispatches from `felt_server::ipc_loop` on Felt side

browser/extensions/felt/rust/src/components.rs
```diff
@@ -442,6 +448,8 @@ impl FeltXPCOM {
                     }
                     trace!("FeltServerThread::felt_server::ipc_loop(): DONE");
                 }
+                // Fallback in case the loop was never entered (rx was None).
+                crate::utils::notify_observers("felt-firefox-ipc-disconnected".to_string());
```
lissyx, line 452: `rx` being `None` is not a good sign, maybe we should pass that information in the observer. Also worth adding a `debug!` or `trace!` statement here

browser/extensions/felt/rust/src/components.rs
```diff
@@ -427,6 +432,7 @@ impl FeltXPCOM {
                             },
                             Err(ipc_channel::ipc::IpcError::Disconnected) => {
                                 trace!("FeltServerThread::felt_server::ipc_loop(): DISCONNECTED");
+                                crate::utils::notify_observers("felt-firefox-ipc-disconnected".to_string());
```
lissyx, line 435: Isn't this one duplicated with the one after the end of the loop? When we `break` here, anyway, we end up closing the thread

PR #673, opened 2026-04-02
Bug 2012442 - Basic signout prompt UX plus signout on exit

gcp:


gcp:
Like #674 but with the changed signout prompts (but still not using fully custom UX to get dual checkboxes).

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -16,6 +16,12 @@ enterprise-panel-sign-out-btn =
     .label = Sign out…
 
 enterprise-signout-prompt-title = Sign out of { -brand-short-name }?
+# $tabCount (Number) - the number of open tabs
+enterprise-signout-prompt-title-tabs =
+    Sign out and close { $tabCount } { $tabCount ->
+        [one] tab
```
jporter-dev: Since we only show the tab closing text when multiple tabs are open, the "one tab" string is never used

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -318,67 +318,122 @@ export const EnterpriseHandler = {
...
-    // TODO: Bug 2001029 - Assert or force-enable session restore?
-
-    try {
-      await lazy.ConsoleClient.signoutUser();
-    } catch (e) {
-      console.error(`Unable to signout the user: ${e}`);
-    } finally {
-      Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
-    }
+    Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
```
jporter-dev: This was previously `Services.startup.quit(Ci.nsIAppStartup.eForceQuit);`, I'm not sure if this change was intentional or which one is preferred for our signout scenario. Just wanted to give it some thought.

Using `eAttemptQuit` allows other processes to interrupt shutdown/signout after our signout prompt is accepted. This seems to affect our exit procedure as well. Is it okay to click the "Sign Out" button then continue to get interrupted? I _think_ this is okay since signout does not truly occur until we've shutdown so we are never left in an odd state. 

I found a good example scenario for this. Websites with unsaved form data can prompt after our signout prompt is accepted.

Image

We can replicate this by pasting the following into the console, then click anywhere on the page to register an interaction, then try to quit or signout.

```
window.addEventListener('beforeunload', (event) => {
  event.preventDefault();
  event.returnValue = '';
});
```

I tested without our signout dialog and the behavior here matches the default so I believe this change is good to make. It goes:

1. Trigger quit
2. Get default quit prompt on MacOS (our custom prompt occurs at this stage as well)
3. Accept quit prompt to continue with shutdown
4. Tab prompts with `This page is asking you to confirm that you want to leave — information you’ve entered may not be saved.`
    1. Stay on page cancels shutdown
    2. Leave page shuts down as expected

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1543,6 +1550,22 @@ BrowserGlue.prototype = {
       return;
     }
 
+    // When Firefox was launched by FELT, show a signout confirmation prompt
+    // instead of the standard quit dialog.
+    if (
+      AppConstants.MOZ_ENTERPRISE &&
+      Services.felt?.isFeltBrowser() &&
+      lazy.EnterpriseHandler._isInitialized
```
jporter-dev: I'm not sure we need to check the `_isInitialized` private var here since `EnterpriseHandler.init()` is called once per window once a user is signed in.

We can check the `enterprise.promptOnSignout` pref here as well to preserve the default tab close and shortcut close prompts until we have the custom 2 checkbox signout and custom quit prompts implemented. 

```suggestion
      Services.prefs.getBoolPref("enterprise.promptOnSignout", true)
```

jporter-dev (commented):
This looks good, it's very similar to what I had in mind before venturing down the async blocker path. Left a couple of comments worth considering around the default prompt behavior.

jporter-dev:
Re-added `initiateShutdown` for the badge signout and its use when policy fetching fails in `EnterprisePoliciesParent`. 

This is the basic UX implementation, the full implementation is #569, which is currently being updated to work with this PR and #674

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -16,6 +16,11 @@ enterprise-panel-sign-out-btn =
     .label = Sign out…
 
 enterprise-signout-prompt-title = Sign out of { -brand-short-name }?
+# $tabCount (Number) - the number of open tabs
+enterprise-signout-prompt-title-tabs =
+    Sign out and close { $tabCount } { $tabCount ->
+       *[other] tabs
+    }?
```
lissyx: We need plural handling?

lissyx: https://firefox-source-docs.mozilla.org/l10n/fluent/tutorial.html#plural-rules

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1543,6 +1550,22 @@ BrowserGlue.prototype = {
       return;
     }
 
+    // When Firefox was launched by FELT, show a signout confirmation prompt
+    // instead of the standard quit dialog.
+    if (
+      AppConstants.MOZ_ENTERPRISE &&
+      Services.felt?.isFeltBrowser() &&
+      Services.prefs.getBoolPref("enterprise.promptOnSignout", true)
```
lissyx: `"enterprise.promptOnSignout"` should be `"enterprise.prompt_on_signout"`

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -314,54 +314,121 @@ export const EnterpriseHandler = {
...
+    for (let win of Services.wm.getEnumerator("navigator:browser")) {
+      if (!win.closed && win.gBrowser) {
+        tabCount += win.gBrowser.openTabs.length;
+      }
+    }
+
+    const titleId =
+      tabCount >= 2
+        ? { id: "enterprise-signout-prompt-title-tabs", args: { tabCount } }
+        : { id: "enterprise-signout-prompt-title" };
```
lissyx: Not sure how this is going to work with i18n plural handling

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -373,7 +440,7 @@ export const EnterpriseHandler = {
     } catch (e) {
       console.error(`Unable to signout the user: ${e}`);
     } finally {
-      Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
+      Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
```
lissyx: Should we use `eAttemptQuit` always? I'd guess we want to use it only when `Services.prefs.getBoolPref(PROMPT_ON_SIGNOUT_PREF, true)` returns `true` but if it's set to `false` we keep using `eForceQuit` ?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -427,7 +427,7 @@ export class FeltProcessParent extends JSProcessActorParent {
         );
 
         this.proc.exitPromise
-          .then(ev => {
+          .then(async ev => {
```
lissyx: pretty sure we said this was not a good idea

browser/components/BrowserGlue.sys.mjs
jporter-dev: This pref was already in use in `EnterpriseHandler`, I can change it, but are there any consequences of changing a preexisting pref I should know about?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -438,9 +438,19 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+                  await lazy.ConsoleClient._post(
+                    lazy.ConsoleClient._paths.SIGNOUT
+                  );
+                } catch (e) {
+                  console.error(`FeltExtension: Failed to post signout: ${e}`);
+                }
+                Services.cpmm.sendAsyncMessage(
+                  "FeltParent:FirefoxNormalExit",
+                  {}
+                );
```
lissyx:
```suggestion
                lazy.ConsoleClient._post(
                    lazy.ConsoleClient._paths.SIGNOUT
                ).catch(e => {
                  console.error(`FeltExtension: Failed to post signout: ${e}`);
                }).then(() => {
                  Services.cpmm.sendAsyncMessage(
                    "FeltParent:FirefoxNormalExit",
                    {}
                });
```

browser/components/BrowserGlue.sys.mjs
lissyx: we'd need to cover both, but TBH we should not have landed `enterprise.promptOnSignout` :(

lissyx: But in this case, it's not like this really did ship to anybody so we should take the opportunity to fix it right now

jporter-dev: Good to know, I'll go ahead and make the change now

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: I had a [similar thought](https://github.com/mozilla/enterprise-firefox/pull/673#discussion_r3032976646) when initially looking at this since `eAttemptQuit` allows other observers to interrupt shutdown. I'll take a look at this change a bit more.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
jporter-dev: I took a look at the other PR to look for the convo for your comment about the `async` not being a good idea, and noticed this change may be out of date?

it looks like the POST was removed here and is now called from the FELT browser so I'll just remove the changes here

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: I consolidated this to one title string and used fluent to handle the 1 or >1 tab logic 

- `<=1`: Sign out of Firefox Enterprise
- `>1`: Sign out and close X tabs

browser/locales/en-US/browser/enterprise/enterprise.ftl
jporter-dev: This was only handling plural, and using a separate Fluent string for singular, I updated it to use one string. I'll have to remember this in #569 when `browser.tabs.warnOnClose` is `false`, we'll need to show "Sign out of Firefox Enterprise" in this case regardless of tab count.

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Went ahead and kept this as `eForceQuit` to maintain the signout behavior we had originally. Our explicit signout path needs to operate differently than a browser quit so we have no need to `eAttemptQuit` here.

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1543,6 +1550,22 @@ BrowserGlue.prototype = {
       return;
     }
 
+    // When Firefox was launched by FELT, show a signout confirmation prompt
+    // instead of the standard quit dialog.
+    if (
+      AppConstants.MOZ_ENTERPRISE &&
+      Services.felt?.isFeltBrowser() &&
+      Services.prefs.getBoolPref("enterprise.prompt_on_signout", true)
```
1rneh: Can we import this preference from the `EnterpriseHandler` and reuse it here?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -314,54 +314,121 @@ export const EnterpriseHandler = {
...
+      checkLabelId: { id: "enterprise-signout-prompt-checkbox-label" },
+      signoutBtnLabelId: { id: "enterprise-signout-prompt-primary-btn-label" },
+      flags:
+        Services.prompt.BUTTON_TITLE_IS_STRING * Services.prompt.BUTTON_POS_0 +
+        Services.prompt.BUTTON_TITLE_CANCEL * Services.prompt.BUTTON_POS_1 +
+        Services.prompt.BUTTON_POS_0_DEFAULT,
+    };
+  },
+
+  _handleSignoutPromptResult(buttonPressed, checked) {
```
1rneh: Maybe we can pick a name that reflects both the boolean return value and the potential preference flip side effect e.g. `_confirmSignout`?

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -15,7 +15,13 @@ enterprise-panel-learn-more = Learn more
 enterprise-panel-sign-out-btn =
     .label = Sign out…
 
-enterprise-signout-prompt-title = Sign out of { -brand-short-name }?
+# $tabCount (Number) - the number of open tabs
+enterprise-signout-prompt-title =
+    { $tabCount ->
+        [zero] Sign out of { -brand-short-name }?
```
1rneh: Have you verified that we can use the key "zero" instead of "0"?
https://firefox-source-docs.mozilla.org/l10n/fluent/review.html#variants-and-plurals

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1543,6 +1550,22 @@ BrowserGlue.prototype = {
...
+    ) {
+      const promptWindow = lazy.BrowserWindowTracker.getTopWindow({
+        allowFromInactiveWorkspace: true,
+      });
+      if (!lazy.EnterpriseHandler.showSignoutPrompt(promptWindow)) {
+        aCancelQuit.QueryInterface(Ci.nsISupportsPRBool).data = true;
```
1rneh, line 1564: I guess we should also set `this._quitSource = "unknown"` to keep to the script :) 
https://searchfox.org/firefox-main/rev/1953976758c30482c9f06aefc837470e547bece2/browser/components/BrowserGlue.sys.mjs#1605-1607

1rneh (approved):
Looks good, thank you! Only minor suggestions. 

Note: I have some more suggestions regarding naming and JSDoc coverage in `EnterpriseHandler`, but I see that #569 is introducing a lot of changes as well. Probably makes sense to address them there.

browser/locales/en-US/browser/enterprise/enterprise.ftl
jporter-dev: Good catch, it in fact did not work. I didn't notice Fluent explicitly wants `0` for that key, while using `one` for 1.. 🙃

lissyx (approved):

PR #674, opened 2026-04-02
Bug 2012442 - Sign out session when Firefox exits normally

gcp:
Post /sso/logout from FELT when Firefox exits with code 0 and the exit was not a restart or explicit logout.

### DIFF
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -28,6 +28,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   BrowserWindowTracker: "resource:///modules/BrowserWindowTracker.sys.mjs",
   FeltStorage: "resource:///modules/FeltStorage.sys.mjs",
   PrivateBrowsingUtils: "resource://gre/modules/PrivateBrowsingUtils.sys.mjs",
+  ConsoleClient: "resource:///modules/enterprise/ConsoleClient.sys.mjs",
   isBlockingShutdown: "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
   shouldNotCloseWindow:
     "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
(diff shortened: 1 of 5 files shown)


gcp:
Trivial version of #673 with no UX changes.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -438,6 +438,15 @@ export class FeltProcessParent extends JSProcessActorParent {
...
                 this.abnormalExitFirstTime = 0;
+                // A clean exit that wasn't a restart or explicit logout
+                // means the user closed Firefox. Sign out the session.
+                try {
+                  await lazy.ConsoleClient._post(
+                    lazy.ConsoleClient._paths.SIGNOUT
+                  );
+                } catch (e) {
+                  console.error(`FeltExtension: Failed to post signout: ${e}`);
+                }
```
lissyx: If it's the case, why do we not do that in `FeltParent:FirefoxNormalExit` handling rather than here?

lissyx: We would have to pass some data into `FeltParent:FirefoxNormalExit` to distinguish this call from the one we trigger when he handle restart below: https://searchfox.org/enterprise-main/rev/00288e25bd40457e5a040b3333760f0fb6d9635e/browser/extensions/felt/content/FeltProcessParent.sys.mjs#470

like e.g. `{performLogout: true}`

And then in https://searchfox.org/enterprise-main/rev/00288e25bd40457e5a040b3333760f0fb6d9635e/browser/extensions/felt/api.js#245-248 we consume this and perform the logout when required?

lissyx (approved):
r=me with moving the signout into `FirefoxNormalExit` message handler (and making sure of proper promise chaining there to avoid `await` ?)

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -427,7 +427,7 @@ export class FeltProcessParent extends JSProcessActorParent {
         );
 
         this.proc.exitPromise
-          .then(ev => {
+          .then(async ev => {
```
lissyx: we dont need this anymore?

browser/extensions/felt/api.js
```diff
@@ -247,15 +248,27 @@ this.felt = class extends ExtensionAPI {
...
+            Services.startup.quit(
+              Ci.nsIAppStartup.eAttemptQuit | Ci.nsIAppStartup.eConsiderQuit
+            );
+          } else if (!this._win) {
+            Services.felt.makeBackgroundProcess(false);
+            this.showWindow();
+          }
+        };
+
+        if (message.data?.performLogout) {
```
lissyx: nit: `=== true`, I'm always worried about truthiness from JS

testing/enterprise/felt_tests.py
```diff
@@ -296,6 +296,8 @@ def do_GET(self):
             contentType = "text/xml"
 
         elif path == "/sso/callback":
+            self.server.policy_access_token.value = str(uuid.uuid4())
+            self.server.policy_refresh_token.value = str(uuid.uuid4())
```
lissyx, line 300: So we generate a new token each time? Maybe we should not pass any value at startup at all now, and thus assert (I think we do already) browser-side that we have received a non emptry token?

lissyx (approved):

testing/enterprise/felt_tests.py
gcp: We do assert this in the code, which is why the above was needed (the rest of the test clears it, if it's not reset, we won't relaunch correctly).

lissyx: https://searchfox.org/enterprise-main/rev/00288e25bd40457e5a040b3333760f0fb6d9635e/testing/enterprise/felt_tests.py#619-620

Maybe we should change this to not set any value and rely on the callback behavior, that gets us closer to real console's behavior?

testing/enterprise/felt_tests.py
```diff
@@ -616,8 +624,9 @@ def setUp(self):
         if hasattr(self, "EXTRA_PREFS"):
             self._extra_prefs.update(self.EXTRA_PREFS)
 
-        self.policy_access_token = SharedString(str(uuid.uuid4()))
-        self.policy_refresh_token = SharedString(str(uuid.uuid4()))
+        self.policy_access_token = SharedString("")
+        self.policy_refresh_token = SharedString("")
```
lissyx, line 628: this should work :)

lissyx (approved):

PR #678, opened 2026-04-03
Bug 2028771 - Fix test_felt_version.py to not error on release builds

jonathanmendez:
The test setup executes a script to return expected values. The setup was not returning the `is_nightly` key for non-nightly builds, and the check did not account for the missing key. I balanced the test setup to always include the key (with a false value for non-nightly builds).

This was found when doing the uplift of the felt version changes. I will include this in the uplift PR to verify that it fixes the issue.

### DIFF
--- a/testing/enterprise/test_felt_version.py
+++ b/testing/enterprise/test_felt_version.py
@@ -11,7 +11,7 @@ sys.path.append(os.path.dirname(__file__))
 from felt_tests import FeltTestsBase
 
 
-class FeltEmailInputFocus(FeltTestsBase):
+class FeltVersion(FeltTestsBase):
     """
     Tests the Firefox version in the Felt window
     """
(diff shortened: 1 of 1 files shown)


testing/enterprise/test_felt_version.py
lissyx: Can you fix the indentation here? I missed it the other day

lissyx, line 53: I think it may be cleaner to have a single return:


```suggestion
            const version = AppConstants.MOZ_APP_VERSION_DISPLAY;
            let l10n_id_value = "felt-version";
            let isodate_value = null;

            if (AppConstants.NIGHTLY_BUILD) {
                const buildID = Services.appinfo.appBuildID;
                const year = buildID.slice(0, 4);
                const month = buildID.slice(4, 6);
                const day = buildID.slice(6, 8);

                l10n_id_value = "felt-version-nightly",
                 isodate_value = `${year}-${month}-${day}`,
            }

            return {
                is_nightly: AppConstants.NIGHTLY_BUILD === true,
                l10n_id: l10n_id_value,
                version,
                isodate: isodate_value,
            };

```

lissyx (approved):
r=me with a single return code path

lissyx (approved):

PR #679, opened 2026-04-03
Bug 2029205 - Enterprise: remove default enterprise.console.address v…

lissyx:
…alue and rely only on distribution.ini

### DIFF
--- a/browser/branding/enterprise/distribution.ini
+++ b/browser/branding/enterprise/distribution.ini
@@ -4,7 +4,7 @@
 [Global]
 id=enterprise-local
 version=1.0
-about=Enterprise Local
+about=Mozilla Firefox Enterprise Local Build
 
 [Preferences]
 enterprise.console.address=https://stage.fx-enterprise.nonprod.webservices.mozgcp.net
--- a/browser/branding/enterprise/pref/firefox-enterprise.js
+++ b/browser/branding/enterprise/pref/firefox-enterprise.js
@@ -4,11 +4,6 @@
 
 /* global pref */
 
-pref(
-  "enterprise.console.address",
-  "https://stage.fx-enterprise.nonprod.webservices.mozgcp.net"
-);
-
 pref("browser.profiles.enabled", false);
 pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
 


jonathanmendez (approved):

browser/branding/enterprise/distribution.ini
```diff
@@ -1,5 +1,10 @@
...
-
-[Preferences]
-enterprise.console.address=https://stage.fx-enterprise.nonprod.webservices.mozgcp.net
+# Default partner distribution configuration file
+# Must be replaced as part of a repack for Enterprise builds
+
+[Global]
+id=enterprise-local
+version=1.0
+about=Enterprise Local
```
1rneh: `nit` since they get surfaced to the ui?
```suggestion
about=Mozilla Firefox Enterprise Local Build
```

lissyx: I think if we remove `id` and `version` then https://searchfox.org/enterprise-main/rev/e8c0742b4526a683d68073e2eed2ece46a5a7164/browser/components/distribution.sys.mjs#386-388 we exit early and we again do not apply `[Preferences]` from below

lissyx: We could remove `about` if we use `id=mozilla-XXX` according to https://searchfox.org/enterprise-main/rev/e8c0742b4526a683d68073e2eed2ece46a5a7164/browser/components/distribution.sys.mjs#390

lissyx: image

1rneh: Sorry, that code suggestion was only supposed to suggest a change to `about` and no other field removal.

1rneh (approved):

lissyx:
@1rneh @jonathanmendez do you both feel confident we should merge this now?

1rneh:
> @1rneh @jonathanmendez do you both feel confident we should merge this now?

Since this doesn’t change any actual preference state, we should be good to merge it.

1rneh:
> > @1rneh @jonathanmendez do you both feel confident we should merge this now?
> 
> Since this doesn’t change any actual preference state, we should be good to merge it.

Actually, running your PR and [removing sending the console address from Felt to the browser](https://searchfox.org/enterprise-main/rev/e8c0742b4526a683d68073e2eed2ece46a5a7164/browser/extensions/felt/content/FeltProcessParent.sys.mjs#267-270) (which I'm planning to get rid of in #479) leads to the browser not launching successfully. I would have assumed it's unaffected because the preference you set in `distribution.ini` is set in both Felt and the browser?

PR #680, opened 2026-04-03
Bug 2029205 - Enterprise: only construct appUpdateURL from preference…

lissyx:
… value during tests

### DIFF
--- a/toolkit/xre/nsAppRunner.cpp
+++ b/toolkit/xre/nsAppRunner.cpp
@@ -5834,11 +5834,19 @@ nsresult XREMain::XRE_mainRun() {
 
 #if defined(MOZ_ENTERPRISE)
     {
-      nsAutoCString consoleAddress;
-      rv =
-          Preferences::GetCString("enterprise.console.address", consoleAddress);
-      if (NS_SUCCEEDED(rv)) {
-        XRE_ParseEnterpriseServerURL(*mAppData, consoleAddress.get());
+      // When running tests we may override the appUpdateURL to a test value
+      bool readUpdateUrlFromPref = false;
+      rv = Preferences::GetBool(
+          "enterprise.felt_tests.read_update_url_from_prefs",
+          &readUpdateUrlFromPref);
+      if (NS_SUCCEEDED(rv) && readUpdateUrlFromPref) {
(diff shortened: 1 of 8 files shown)


toolkit/xre/nsAppRunner.cpp
```diff
@@ -5834,11 +5834,18 @@ nsresult XREMain::XRE_mainRun() {
...
 #if defined(MOZ_ENTERPRISE)
     {
-      nsAutoCString consoleAddress;
-      rv =
-          Preferences::GetCString("enterprise.console.address", consoleAddress);
+      bool readUpdateUrlFromPref = false;
+      rv = Preferences::GetBool(
+          "enterprise.felt_tests.read_update_url_from_prefs",
+          &readUpdateUrlFromPref);
       if (NS_SUCCEEDED(rv)) {
```
jonathanmendez: I imagine this should be
```suggestion
      if (NS_SUCCEEDED(rv) && readUpdateUrlFromPref) {
```
(Or I guess since you initialize readUpdateUrlFromPref to false it could be 
```suggestion
      if (readUpdateUrlFromPref) {
```
)

Either way, the current code doesn't use the value of `readUpdateUrlFromPref`

toolkit/xre/nsAppRunner.cpp
```diff
@@ -5834,11 +5834,18 @@ nsresult XREMain::XRE_mainRun() {
 
 #if defined(MOZ_ENTERPRISE)
     {
-      nsAutoCString consoleAddress;
-      rv =
-          Preferences::GetCString("enterprise.console.address", consoleAddress);
+      bool readUpdateUrlFromPref = false;
```
jonathanmendez, line 5838: Could use a comment to explain what this block of code is for
```suggestion
      // When running tests we may override the appUpdateURL to a test value
      bool readUpdateUrlFromPref = false;
```

jonathanmendez (changes requested):

toolkit/xre/nsAppRunner.cpp
lissyx: yes, i forgot to use it :(

jonathanmendez (approved):

PR #681, opened 2026-04-03
Bug 2029233 - Add missing items of expected format for distribution.ini

jonathanmendez:
Note: This also changes the line endings from CRLF to LF


lissyx (approved):

lissyx:
We need this because other wise the `distribution.ini` is not parsed by the code: https://searchfox.org/enterprise-main/rev/3b40675b46364276f994fe542c883e8ef26fd80d/browser/components/distribution.sys.mjs#381-393

It's not a big deal technically for us, but if we have people thinking they can use this file when not, it's getting confusing.

PR #683, opened 2026-04-06
Bug 2026242 - Improve access connector UX

jporter-dev:
## Summary

Bugzilla: Bug-2026242

Adds a dedicated urlbar button and panel for the access connector in enterprise builds, replacing the repurposed IPProtection toolbar widget. 

- Reverts UX changes from #562 
- Adds `access-connector-button` urlbar action and `panelUI-access-connector` panel view new FTL strings, consistent with Lockdown Mode UX                 
- Uses `addProgressListener` and `IPPProxyManager:StateChanged` to update the button when proxy state changes

## Links

- [Figma](https://www.figma.com/design/xYH6L8QrDz0Q16XFvdo7We/%F0%9F%A6%8A-Client-Design?node-id=1862-8748&m=dev&t=drdE7joEk1G5v99t-1)
- [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2026242)

- [x] Depends on Bug-2034102 landing
- [x] #562
- [x] #826

## Screenshots

access-connector

## Testing

* [ ] Verify test coverage resolved in #826 is still accurate
* [x] Added tests

1. _optional_: override Access Connector policy in console for your user
2. `mach build`
3. `mach run`
4. Navigate to `https://mozilla.org`
5. Observe access connector urlbar button and panel
6. Navigate to `https://firefox.com` and/or change tabs
7. No access connector button should be present


browser/components/enterprise/content/enterprise-urlbar-actions.css
```diff
@@ -6,15 +6,20 @@
...
-  #lockdown-mode-button .urlbar-icon {
+  #lockdown-mode-button .urlbar-icon,
+  #access-connector-button .urlbar-icon {
     -moz-context-properties: fill;
     fill: var(--icon-color-information);
   }
 }
 
-.lockdown-mode-subview {
+.enterprise-panel-subview {
```
1rneh: `nit:` To not confuse it with the enterprise panel - just because I was confused for a second :)
```suggestion
.enterprise-urlbar-panel-subview {
```

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -172,6 +179,54 @@ export const EnterpriseHandler = {
     });
   },
 
+  _initAccessConnectorButton(window) {
```
1rneh: Please add JSDocs for `_initAccessConnectorButton`?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -172,6 +179,54 @@ export const EnterpriseHandler = {
...
+          }) ?? false;
+      } catch (e) {
+        lazy.log.warn("Failed to check access connector state for URI: ", e);
+      }
+      button.hidden = !isProtected;
+    };
+
+    button.addEventListener("click", event => {
+      window.PanelUI.showSubView("panelUI-access-connector", button, event);
+    });
```
1rneh: We also need a listeners for the `keypress` event to support keyboard navigation to the access connector button.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -172,6 +179,54 @@ export const EnterpriseHandler = {
...
+    lazy.IPPProxyManager.addEventListener(
+      "IPPProxyManager:StateChanged",
+      handleProxyStateChange
+    );
+
+    window.addEventListener("unload", () => {
+      lazy.IPPProxyManager.removeEventListener(
+        "IPPProxyManager:StateChanged",
+        handleProxyStateChange
+      );
```
1rneh: Can we remove the progress listener as well?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -172,6 +179,54 @@ export const EnterpriseHandler = {
...
+    });
+
+    updateButtonState(window.gBrowser.selectedBrowser?.currentURI);
```
1rneh: This runs whenever a new chrome windows gets loaded, which also triggers a location change notification. Hence I believe we can remove calling `updateButtonState` here to avoid updating the button twice.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -172,6 +179,54 @@ export const EnterpriseHandler = {
...
+
+    button.addEventListener("click", event => {
+      window.PanelUI.showSubView("panelUI-access-connector", button, event);
+    });
+
+    window.gBrowser.addProgressListener({
+      onLocationChange(webProgress, _request, location) {
+        if (!webProgress.isTopLevel) {
+          return;
+        }
```
1rneh: Let's check the location change flag to avoid calling `updateButtonState` in cases like reloads, see https://searchfox.org/firefox-main/rev/07e27bfce5acf193bdb89c7ba2aa73451a3e43b4/uriloader/base/nsIWebProgressListener.idl#481-520.

1rneh (changes requested):
The changes look good overall. I only have a few suggested improvements.

1rneh (commented):
On another thought, instead of deriving whether any proxy filter is applied from the window location in `EnterpriseHandler` we could also let the feature notify us or call a method in `EnterpriseHandler`, that will toggle the button visibility
- when it applies any proxy filter: https://searchfox.org/enterprise-main/rev/6e2c2173cbd8f5fb0c7516a9a0ac06dd5e6b176a/toolkit/components/ipprotection/IPPChannelFilter.sys.mjs#266-269. 
- when it doesn't apply any proxy filter: https://searchfox.org/enterprise-main/rev/6e2c2173cbd8f5fb0c7516a9a0ac06dd5e6b176a/toolkit/components/ipprotection/IPPChannelFilter.sys.mjs#253-255

_(Those two code places are from a quick glimpse over the `ipprotection` component and need to be double-checked)._

It feels like a more stable solution as we don't decouple the decision to show the button entirely from the decision to apply any proxy filters. As this is a security feature, we should really make sure we avoid any mismatch.

@fiji-flo Do you have anything else to add, or does my comment cover what we have discussed?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -172,6 +191,78 @@ export const EnterpriseHandler = {
...
+  _initAccessConnectorButton(window) {
+    const button = window.document.getElementById("access-connector-button");
+
+    const updateButtonState = location => {
+      let isProtected = false;
+      try {
+        isProtected =
+          lazy.IPPProxyManager.channelFilter()?.shouldInclude({
+            URI: location,
+          }) ?? false;
```
fiji-flo: I'd love if we had a better way to determine if the proxy is active. @1rneh pointed us in a good direction. Maybe it's worth it to ping @strseb about this? But I see it's the same as before. So on second thought reaching out and asking the VPN folks seems like the right call.

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -149,6 +151,23 @@ export const EnterpriseHandler = {
...
   },
 
+  /**
+   * Initializes the enterprise-related urlbar buttons.
+   *
+   * @param {Window} window chrome window
+   */
+  _initUrlbarButtons(window) {
+    this._initLockdownModeButton(window);
+    this._initAccessConnectorButton(window);
```
fiji-flo: Not sure where to add this comment. But I did observe a lot of jumping in the awesomebar when browsing MDN behind the connector. The issue is that reader mode gets detected only after page load but displayed to the right of he access connector. Can we change the position?

Image

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -45,16 +45,15 @@ lockdown-mode-button =
 lockdown-mode-popup-header = Viewing with restrictions applied
 lockdown-mode-popup-message = You’re viewing this page with extra security protections applied by your organization. Some features may be limited to help reduce security risks.
 
+access-connector-button =
+    .aria-label = Access connector enabled
+    .tooltiptext = Access connector enabled
+access-connector-popup-header = Access connector enabled
+access-connector-popup-message = Connections to this site use additional authentication methods and are routed through a secure enterprise proxy. Page loading times may vary.
```
fiji-flo: I'd remove the `Page loading times may vary.` Part unless someone has strong feelings there.

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: I can move our enterprise urlbar buttons (currently just Access Connector and Lockdown Mode) to show after reader mode since that one gets toggled on/off regularly. I'll take a look at what else shows up there and find a better placement. 

Maybe move it to show to the right of Zoom as well

image

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: I reached out on Slack to get some input. 

I'm taking a closer look at the [IPProtectionToolbarButton](https://searchfox.org/firefox-main/source/browser/components/ipprotection/IPProtectionToolbarButton.sys.mjs#193-272) to at least better handle the states and errors of the IPP service here, but not sure this will solve the problem of not decoupling the status indication from the _actual_ proxy state.

jporter-dev:
@1rneh could I get a re-review 😄 

I've moved the Access Connector button out of `EnterpriseHandler` into its own component as it's grown quite a bit and will continue to grow in complexity once I handle Bug-2040381. I've updated the PR to use the upcoming `IPPPrincipalRules` API from Bug-2034102 as well once that lands (assuming it lands in its [current state](https://phabricator.services.mozilla.com/D302528)). I figure this can be reviewed without that, and should it require changes I can update the usage here to reflect it.

1rneh (approved):
Looks really good to me! Thank you!

PR #685, opened 2026-04-06
Bug 2027104, Bug 2029685 - Further fixes to live policy application

jonathanmendez:
I had noticed a bug in https://github.com/mozilla/enterprise-firefox/pull/626 and decided to review all its changes. In doing so I found several small things to fix up.
I also discovered bug 2029685, which is fixed in this PR.

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -296,12 +296,16 @@ export var Policies = {
       // true, we disallow turning off auto updating, and visa versa.
       if (param) {
         manager.disallowFeature("app-auto-updates-off");
+        manager.allowFeature("app-auto-updates-on");
       } else {
         manager.disallowFeature("app-auto-updates-on");
+        manager.allowFeature("app-auto-updates-off");
       }
     },
     onRemove(manager, _oldParams) {
+      // In a default unpolicied state, it is allowed to turn auto updating on or off
       manager.allowFeature("app-auto-updates-on");
+      manager.allowFeature("app-auto-updates-off");
     },
   },
 
@@ -506,14 +510,20 @@ export var Policies = {
 
   BackgroundAppUpdate: {
     onBeforeAddons(manager, param) {
+      // Logic feels a bit reversed here, but it's correct. If BackgroundAppUpdate is
+      // true, we disallow turning off background updating, and visa versa.
       if (param) {
         manager.disallowFeature("app-background-update-off");
+        manager.allowFeature("app-background-update-on");
       } else {
         manager.disallowFeature("app-background-update-on");
+        manager.allowFeature("app-background-update-off");
       }
     },
     onRemove(manager, _oldParams) {
+      // In a default unpolicied state, it is allowed to turn background updating on or off
       manager.allowFeature("app-background-update-on");
+      manager.allowFeature("app-background-update-off");
     },
   },
 
@@ -570,6 +580,7 @@ export var Policies = {
       }
     },
     onRemove(manager, _oldParams) {
+      manager.allowFeature("profileManagement");
       unblockAboutPage(manager, "about:profiles");
       unblockAboutPage(manager, "about:profilemanager");
       unblockAboutPage(manager, "about:editprofile");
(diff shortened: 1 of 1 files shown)


browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -2017,7 +2021,7 @@ export var Policies = {
 
   ExtensionUpdate: {
     onBeforeAddons(manager, param) {
-      setAndLockPref("extensions.update.enabled", !param);
+      setAndLockPref("extensions.update.enabled", param);
```
jonathanmendez, line 2032: https://github.com/mozilla/enterprise-firefox/pull/626#discussion_r3034448986

lissyx: My bad yes, confirms we need to run those tests asap

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -302,6 +302,7 @@ export var Policies = {
     },
     onRemove(manager, _oldParams) {
       manager.allowFeature("app-auto-updates-on");
+      manager.allowFeature("app-auto-updates-off");
```
lissyx, line 308: This looks contradictory ?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -514,6 +515,7 @@ export var Policies = {
     },
     onRemove(manager, _oldParams) {
       manager.allowFeature("app-background-update-on");
+      manager.allowFeature("app-background-update-off");
```
lissyx, line 526: Same here

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -570,6 +572,7 @@ export var Policies = {
       }
     },
     onRemove(manager, _oldParams) {
+      manager.allowFeature("profileManagement");
```
lissyx, line 583: LGTM

lissyx (approved):
LGTM but check with @mkaply for the `app-auto-updates-off` and `app-background-update-off` thing

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: See:

https://searchfox.org/firefox-main/source/toolkit/modules/UpdateUtils.sys.mjs#557

We actually need to have something that removes a Feature.

browser/components/enterprisepolicies/Policies.sys.mjs
mkaply: See:

https://searchfox.org/firefox-main/source/toolkit/modules/UpdateUtils.sys.mjs#573

We actually need to have something that removes a Feature.

lissyx: OK, so maybe it's worth a comment @jonathanmendez

browser/components/enterprisepolicies/Policies.sys.mjs
jonathanmendez: Yes, this looks weird, but I believe it has the right effect. Basically when either of these are marked with `disallowFeature()` then it gets stores in a `DisallowedFeatures` dict. Existing within that dict controls behavior (in [this code path](https://searchfox.org/firefox-main/source/toolkit/modules/UpdateUtils.sys.mjs#551-585)).

Calling `allowFeature()` removes the items from the `DisallowedFeatures` dict, reverting to "default behavior if this policy didn't get activated at all".

Actually arguably we have a bug and should be calling the opposing `allowFeature()` in `onBeforeUIStartup` because otherwise setting the policy to true then to false will disallow both.

mkaply: Yeah, you're right. I should have looked at the allowFeature impl.

And agree with your other statement.

jonathanmendez: Added the balancing `allowFeatures()` and an extra comment.

browser/components/enterprisepolicies/Policies.sys.mjs
jonathanmendez: Added the balancing `allowFeatures()` and comments to match AppAutoUpdate

PR #687, opened 2026-04-07
Bug 2006564: Put app as active on startup(macos) and focus email on windows focus r=lissyx

victhorlopez:
…indows focus

The app currently is not set as active if you start it from the dock or desktop in macos, I used this to verify(or you can see it is just behind apps)

` while true; do osascript -e 'tell application "System Events" to get name of (first application process whose frontmost is true)'; sleep 0.5; done`

### DIFF
--- a/browser/base/content/nonbrowser-mac.js
+++ b/browser/base/content/nonbrowser-mac.js
@@ -157,6 +157,11 @@ var NonBrowserWindow = {
       privateWindowItem.setAttribute("disabled", "true");
     }
 
+    // bug 2006564
+    // make sure that when application starts from dock it enforces windows'focus via activateApplication
+    // https://searchfox.org/enterprise-main/rev/4b4e7c59db50500302fa0e437ee07a84d92aa076/widget/nsIMacDockSupport.idl#36-45
+    this.dockSupport.activateApplication(true);
+
     waitForFeltFirefoxWindowReady().then(() => {
       if (newWindowItem) {
         newWindowItem.removeAttribute("disabled");
--- a/browser/extensions/felt/content/window.js
+++ b/browser/extensions/felt/content/window.js
@@ -480,15 +480,19 @@ function focusEmailOnLoginVisible() {
   const loginPane = document.querySelector(".felt-login");
   const emailInput = document.getElementById("felt-form__email");
 
-  new MutationObserver(() => {
+  function maybeFocusEmail() {
     if (!loginPane.classList.contains("is-hidden")) {
       emailInput?.focus();
     }
-  }).observe(loginPane, { attributeFilter: ["class"] });
-
-  if (!loginPane.classList.contains("is-hidden")) {
-    emailInput?.focus();
   }
+
+  new MutationObserver(maybeFocusEmail).observe(loginPane, {
+    attributeFilter: ["class"],
+  });
+
+  window.addEventListener("focus", maybeFocusEmail);
+
+  maybeFocusEmail();
 }
 
 /**


browser/base/content/nonbrowser-mac.js
```diff
@@ -129,6 +129,7 @@ var NonBrowserWindow = {
       // In Felt mode, disable dock menu items until Firefox is ready
       if (Services.felt?.isFeltUI()) {
         this.setupFeltDockMenuState();
+        this.dockSupport.activateApplication(true);
```
lissyx: Why do we need this ? We already do that in https://searchfox.org/enterprise-main/rev/4b4e7c59db50500302fa0e437ee07a84d92aa076/browser/base/content/nonbrowser-mac.js#23-38

victhorlopez: as far as I understand for the feltUI it goes inside the if, also that is only when you click the icon with the app running, not when it starts up.

Can you check also? If you have it dock and you open the app, it is actually it does not go in the foreground

lissyx: > as far as I understand for the feltUI it goes inside the if, also that is only when you click the icon with the app running, not when it starts up.
> 
> Can you check also? If you have it dock and you open the app, it is actually it does not go in the foreground

Sorry but it's not clear. What is the state that you describe? Having `Firefox Enterprise` in the dock, but not started? And then clicking on it to start it, and it does not focus? Or is this fixing the application being brought to foreground and a byproduct of that is this allows the code to handle focusing the email field to work properly ?

victhorlopez: The second one "Or is this fixing the application being brought to foreground and a byproduct of that is this allows the code to handle focusing the email field to work properly ?"

lissyx: OK so with the STRs outlined in https://bugzilla.mozilla.org/show_bug.cgi?id=2006564#c12 I can see that we only hit `load` event https://searchfox.org/enterprise-main/rev/4b4e7c59db50500302fa0e437ee07a84d92aa076/browser/base/content/nonbrowser-mac.js#201-202

browser/base/content/nonbrowser-mac.js
lissyx, line 173: I think we should rather plug into there:

```suggestion
// Disable dock menu items until Firefox is ready
if (newWindowItem) {
  newWindowItem.setAttribute("disabled", "true");
}
if (privateWindowItem) {
  privateWindowItem.setAttribute("disabled", "true");
}

this.dockSupport.activateApplication(true);

waitForFeltFirefoxWindowReady().then(() => {
  if (newWindowItem) {
    newWindowItem.removeAttribute("disabled");
  }
  if (privateWindowItem) {
    privateWindowItem.removeAttribute("disabled");
  }
});
```

victhorlopez: Done, are you also 100% sure about  ```this.dockSupport.activateApplication(false);```

lissyx: Looks like this works and I think it's more aligned with how things should work ?

lissyx: not really sure, I suggested that before reading https://searchfox.org/enterprise-main/rev/4b4e7c59db50500302fa0e437ee07a84d92aa076/widget/nsIMacDockSupport.idl#36-45 I'm unsure now

lissyx, line 142: Update this comment and make sure it states that this also enforces windows' focus via `activateApplication` https://searchfox.org/enterprise-main/rev/4b4e7c59db50500302fa0e437ee07a84d92aa076/widget/nsIMacDockSupport.idl#36-45

victhorlopez: I just know there was quite some work regarding opening urls and so on

lissyx: Yeah no reading in details and checking also with claude comes to the same conclusion, we dont want `activateApplication(false)` because as documented in the IDL it's still a way to request focus, which we do not want once the launched firefox is ready

browser/base/content/nonbrowser-mac.js
```diff
@@ -158,6 +163,7 @@ var NonBrowserWindow = {
     }
 
     waitForFeltFirefoxWindowReady().then(() => {
+      this.dockSupport.activateApplication(false);
```
lissyx: so we remove this one

browser/base/content/nonbrowser-mac.js
victhorlopez: removing it then

lissyx (approved):
r=me with updated comments, removed `this.dockSupport.activateApplication(false)`

PR #688, opened 2026-04-07
Bug 2030411: move enterprise mock console also for mochitest

victhorlopez:
Extract the enterprise console server into its own module (enterprise_server.py) and instantiate it in the mochitest harness (runtests.py) so it   starts automatically when running enterprise mochitests.

### DIFF
--- a/browser/extensions/felt/rust/src/lib.rs
+++ b/browser/extensions/felt/rust/src/lib.rs
@@ -77,6 +77,31 @@ pub extern "C" fn felt_init() {
     trace!("felt_init(): is_felt_safe_mode={}", is_felt_safe_mode);
     IS_FELT_SAFE_MODE.store(is_felt_safe_mode, Ordering::Relaxed);
 
+    // Allow tests to inject tokens via environment variables so they are
+    // available before any JS runs (e.g. before ConsoleClient tries to refresh).
+    if !self::is_felt_ui() && !self::is_felt_browser() {
+        if let (Ok(access), Ok(refresh)) = (
+            env::var("TEST_ENTERPRISE_ACCESS_TOKEN"),
+            env::var("TEST_ENTERPRISE_REFRESH_TOKEN"),
+        ) {
+            if !access.is_empty() && !refresh.is_empty() {
+                let expires_at = std::time::SystemTime::now()
+                    .duration_since(std::time::UNIX_EPOCH)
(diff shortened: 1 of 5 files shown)


browser/extensions/felt/rust/src/lib.rs
```diff
@@ -77,6 +77,29 @@ pub extern "C" fn felt_init() {
     trace!("felt_init(): is_felt_safe_mode={}", is_felt_safe_mode);
     IS_FELT_SAFE_MODE.store(is_felt_safe_mode, Ordering::Relaxed);
 
+    // Allow tests to inject tokens via environment variables so they are
+    // available before any JS runs (e.g. before ConsoleClient tries to refresh).
+    if let (Ok(access), Ok(refresh)) = (
+        env::var("ENTERPRISE_MOCK_ACCESS_TOKEN"),
+        env::var("ENTERPRISE_MOCK_REFRESH_TOKEN"),
+    ) {
```
lissyx: If it's limited to mochitests, we should make sure it is properly gated: neither `is_felt_ui()` nor `is_felt_browser()` should be allowed

testing/mochitest/runtests.py
```diff
@@ -1534,6 +1534,35 @@ def startServers(self, options, debuggerInfo, public=None):
...
+            return
+
+        port = self.findFreePort(socket.SOCK_STREAM)
+        server = EnterpriseConsoleServer(port)
+        if not server.start():
+            self.log.error("Enterprise mock server failed to start")
+            return
+
+        self.enterpriseServer = server
+        self.log.info(f"Enterprise mock server started at http://localhost:{port}")
```
lissyx, line 1564: All of those changes could benefit from some testing peer, maybe @jmaher can share a feedback on that and a better way for us to integrate to reduce risks of conflicts on further merges?

lissyx (changes requested):

browser/extensions/felt/rust/src/lib.rs
victhorlopez: I put it behind ```if bypass_env``` cause it is the only case we care, or do you want the other if?

testing/mochitest/runtests.py
victhorlopez: closing it, no need to review

PR #689, opened 2026-04-07
Release merge for Firefox 149.0.2

jonathanmendez:
Of note:
- Manually resolved versions from 149.0.2 to 149.2.0 per our versioning scheme
- Resolved conflicts for the addition of AIControls and IPProtectionAvailable policies
  - IPProtectionAvailable's policy application uses one of the same prefs as the AccessConnector policy (`browser.ipProtection.enabled`), which can conflict
- I re-applied the previously uplifted e51a0014c2a8 Bug 2017606 - Policies json schema extensions (#667)
  - This was in order to get the previously intentionally left-out AIControls and IPProtectionAvailable policies


toolkit/locales/en-US/toolkit/neterror/netError.ftl
```diff
@@ -200,6 +200,9 @@ fp-neterror-offline-body-title = Looks like there’s a problem with your intern
 
 fp-neterror-connection-intro = { -brand-short-name } can’t create a secure connection to the server at { $hostname }.
 fp-neterror-offline-intro = { -brand-short-name } can’t connect to the server at <strong>{ $hostname }</strong>
+# Variables:
+#   $hostname (String) - Hostname of the website to which the user was trying to connect.
+fp-neterror-http-error-page-intro = <strong>{ $hostname }</strong> sent back an error.
```
lissyx, line 205: @jonathanmendez since there are new strings we should update the l10n changeset I think?

lissyx (approved):
r=me, but make sure we do not need an l10n changeset update?

toolkit/locales/en-US/toolkit/neterror/netError.ftl
jonathanmendez: Do you know what to set the changeset to? This string text looks to be isolated to 149/150 while 151/nightly has `The server at <strong>{ $hostname }</strong> sent back an error: { $responsestatus } { $responsestatustext }`. And I thought in general we wanted the changeset from main/nightly because of the enterprise-specific strings we've been adding?

lissyx: They updated with https://bugzilla.mozilla.org/show_bug.cgi?id=2028335. I think if we update to our current changeset on main, that should be fine, the string `fp-neterror-http-error-page-intro` is supposed to be there, right?

jonathanmendez: Looks like `enterprise-main` has [the string](https://github.com/mozilla/enterprise-firefox/blob/enterprise-main/toolkit/locales/en-US/toolkit/neterror/netError.ftl#L217) (the version from nightly) and the new strings in policies-descriptions.ftl.

And [enterprise-main's l10n-changesets.json](https://github.com/mozilla/enterprise-firefox/blob/enterprise-main/browser/locales/l10n-changesets.json#L18) matches [enterprise-release's enterprise-l10n-changesets.json](https://github.com/mozilla/enterprise-firefox/blob/enterprise-release/browser/locales/enterprise-l10n-changesets.json#L22) (which I did check as part of creating this merge).

So I think we're good

PR #695, opened 2026-04-08
Bug 2030388: improve flakyness of tests by using proper prefs

victhorlopez:


testing/enterprise/base_test.py
```diff
@@ -33,16 +33,19 @@ def setUp(self):
 
         super().setUp()
 
+        if hasattr(self, "_extra_prefs"):
+            self._saved_required_prefs = deepcopy(
+                self.marionette.instance.required_prefs
```
lissyx: likely this should live in `FeltTestsBase.setUp` where we define `self._extra_prefs` ?

testing/enterprise/base_test.py
```diff
@@ -33,16 +33,19 @@ def setUp(self):
...
+
         if hasattr(self, "_extra_cli_args"):
             self._saved_cli_args = deepcopy(self.marionette.instance.app_args)
             self.marionette.instance.app_args += self._extra_cli_args
 
         self.marionette.quit(in_app=False, clean=True)
         self.marionette.start_session(timeout=60)
 
-        if hasattr(self, "_extra_prefs"):
-            self.marionette.enforce_gecko_prefs(self._extra_prefs)
```
lissyx, line 44: i would be very cautious and triple check, this was required to make sure the instance was started with the correct prefs because  `enforce_gecko_prefs` makes sure of that, so it sounds kinda risky

testing/enterprise/base_test.py
lissyx: and it should be IMHO the other way around: we `self._extra_prefs.update(self.marionette.instance.required_prefs)` and continue to use `self.marionette.enforce_gecko_prefs`

testing/enterprise/felt_tests.py
```diff
@@ -729,6 +737,30 @@ def teardown(self):
...
+            }.items()
+            # Drop prefs with %-style format placeholders (e.g. "%(server)s")
+            # as they require interpolation that isn't performed here.
+            if not isinstance(v, str) or "%" not in v
+        }
+        if extra_prefs:
+            prefs.update(extra_prefs)
+        lines = [f'user_pref("{k}", {pref_value_js(v)});\n' for k, v in prefs.items()]
+        with open(os.path.join(profile_path, "user.js"), "w") as f:
+            f.writelines(lines)
```
lissyx: we need to make sure this is not overwritten by felt, it was done at some point. I think it was removed, but we should guard against that by making sure the prefs are still there when we start testing IMHO

testing/enterprise/test_felt_browser_signout.py
```diff
@@ -66,14 +67,18 @@ def run_perform_signout(self):
...
+            try:
+                driver.switch_to_alert().accept()
+                self._logger.info(
+                    "Signing out the user by clicking the Signout button in the dialog"
+                )
+                return True
+            except NoAlertPresentException:
+                return False
+
+        self._waiter(self._child_driver).until(wait_for_and_accept_alert)
```
lissyx, line 74: On the recorded video there was a gray area so I was unsure whether we had a race or another message than the one we were expecting

lissyx (changes requested):

testing/enterprise/base_test.py
```diff
@@ -31,20 +31,27 @@ def setUp(self):
 
         self._logger = self.logger
 
-        super().setUp()
+        marionette = self._marionette_weakref()
```
lissyx, line 67: you may end up where i wanted to avoid :)
https://searchfox.org/enterprise-main/rev/f1803e33aa50eee6e3b1f4d0c5c460045cdefbd3/toolkit/xre/test/marionette/xdg_config/xdg_config_home_test_case.py#13-66

testing/enterprise/base_test.py
```diff
@@ -116,7 +123,8 @@ def get_marionette_port(self, max_try=100):
         while (not found_marionette_port) and (tries < max_try):
             tries += 1
             found_marionette_port = os.path.isfile(marionette_port_file)
-            time.sleep(0.5)
+            if not found_marionette_port:
+                time.sleep(0.5)
```
lissyx: or move the sleep at the start of the function, we already check for `not found_marionette_port` in `while`

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -0,0 +1,92 @@
...
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from marionette_driver.geckoinstance import DesktopInstance, GeckoInstance
+
+from test_felt_browser_signout import BaseBrowserSignout
+
+
+class HarnessPrefs(BaseBrowserSignout):
```
lissyx, line 16: We should also have tests covering the `os.environ` to make sure we set what is required and **we dont leak** between runs

victhorlopez: Will extend

testing/enterprise/base_test.py
victhorlopez: yeah.. I don't want to make it too hacky.. but I think getting rid of the extra restarts between tests might be worth, I will try to clean up.. I will put this back to draft and see

victhorlopez:
I am closing this one, I separated a bunch of fixes into other PRS. The prevention of restarting processes to much is also working in here, but for now as it is a nice to have I will see if I put it later on on another PR

PR #697, opened 2026-04-08
Bug 2030297 - update ipc-channel

lissyx:


browser/extensions/felt/rust/src/client.rs
```diff
@@ -422,22 +422,15 @@ impl FeltClientThread {
...
-                                        ipc_channel::ErrorKind::Io(err) => {
-                                            trace!("FeltClientThread::felt_client::ipc_loop(): MESSAGE I/O ERROR: {}", err);
-                                        },
-                                        err => {
-                                            trace!("FeltClientThread::felt_client::ipc_loop(): MESSAGE OTHER ERROR: {}", err);
-                                        },
-                                    }
-                                }
+                                Err(serde_error) => {
+				    trace!("FeltClientThread::felt_client::ipc_loop(): MESSAGE SERDE ERROR: {}", serde_error);
```
gcp, line 411: `./mach format`

(Why didn't this go red on lints?)

python/mozbuild/mozbuild/vendor/vendor_rust.py
```diff
@@ -713,17 +713,34 @@ def vendor(self, ignore_modified=False, force=False):
...
-                        {
-                            "crate": name,
-                            "num": num,
-                        },
-                        "There are {num} different versions of crate {crate}. "
-                        "Please avoid the extra duplication.",
-                    )
-                    failed = True
+                    if (
+                        name == "embedded-io"
```
gcp: Is this the first time we add an exception to this? I'm tempted to say this will be hard no then?

supply-chain/audits.toml
```diff
@@ -3817,7 +3817,13 @@ notes = "Crate update"
 [[audits.ipc-channel]]
 who = "Alexandre Lissy <lissyx+mozillians@lissyx.dyndns.org>"
 criteria = "safe-to-deploy"
-delta = "0.20.2 -> 0.20.2@git:f2b5aed36a60f07589579d31aaace2f90d2e5e9e"
+delta = "0.20.2 -> 0.21.0"
+notes = "New upstream release"
```
gcp: That's not a suitable audit summary. If you didn't actually review this that's one more reason for this to be a very hard no. The whole point of cargo-vet is that we don't blindly do these kind of updates.

gcp (changes requested):

browser/extensions/felt/rust/src/client.rs
lissyx: My bad, I keep running `mach lint` but, I've reported that `mach lint` was not working well with rust a couple of months ago: https://bugzilla.mozilla.org/show_bug.cgi?id=2010931 maybe I need to push again, I'll fix that

python/mozbuild/mozbuild/vendor/vendor_rust.py
lissyx: It was permitted until a couple of months ago on m-c to have duplicates, but it was removed. Problem here is that this is coming from `postcard` dependency that `ipc-channel` moved to. We may be able to vendor and patch `postcard` and make `ipc-channel` depends on that vendored version to remove it, maybe, but is this a road we want to go? I think I hacked it locally and it was working, I can try resurrecting the patch if you prefer we take this path rather than allowing this specific dupe.

There are other crates we vendor that depends on postcard, but they dont have the same problem because either:
 - they depend on an earlier version where there is only `embedded-io` once
 - they depend on newer version but its a dev dependency, which are not vendored, so the problem do not surface

supply-chain/audits.toml
lissyx: I had a decent look, but by the time I did multiple rebase and re-vendor etc, I had lost the audit summary (but we need one to just be able to build). I'll make sure we have a decent and deeper one

python/mozbuild/mozbuild/vendor/vendor_rust.py
gcp: I wouldn't take this without approval from build peers, as it may block us from merging back to `main`. They should at least advise how to proceed in this kind of situation.

lissyx: I did ask for help on that matter back then on #build on matrix but to no feedback, I'll push again

lissyx: Let's see if I get more feedback this time

lissyx: Ok, no feedback but vendoring postcard is not a big deal technically

lissyx: ok it's green

supply-chain/audits.toml
lissyx: I'll update the audit now that things seems to sail properly with `postcard`

supply-chain/audits.toml
```diff
@@ -3911,8 +3911,25 @@ notes = "Crate update"
 [[audits.ipc-channel]]
 who = "Alexandre Lissy <lissyx+mozillians@lissyx.dyndns.org>"
 criteria = "safe-to-deploy"
-delta = "0.20.2 -> 0.20.2@git:f2b5aed36a60f07589579d31aaace2f90d2e5e9e"
+delta = "0.20.2 -> 0.21.0"
+notes = """
+This is a new upstream release that contains fixes and improvements on error
+handling, notably fixing risks of ipc communication getting blocked.
+It also drops bincode in favor of postcard (already used in other dependencies).
```
gcp, line 3910: Doesn't this mean the wire format is incompatible? If so we need an IPC version bump.

gcp (approved):

supply-chain/audits.toml
lissyx: Indeed, I forgot to update the version.

PR #701, opened 2026-04-09
Bug 2030628: Do not put private properties in win + remove async from…

victhorlopez:
… cmds

Fixes the two issues that make the tests fail 
```MOZ_BYPASS_FELT=1 ./mach mochitest   browser/base/content/test/popupNotifications/browser_popupNotification_no_anchors.js``` 
```MOZ_BYPASS_FELT=1 ./mach mochitest  toolkit/components/printing/tests/browser_modal_print.js```

```test left unexpected property on window: _isEnterpriseLearnMoreLinkConfigured```

```
resource://testing-common/PromiseTestUtils.sys.mjs:assertNoUncaughtRejections:266
chrome://mochikit/content/browser-test.js:handleTask:1408
chrome://mochikit/content/browser-test.js:_runTaskBasedTest:1466
chrome://mochikit/content/browser-test.js:Tester_execTest:1617
chrome://mochikit/content/browser-test.js:nextTest/<:1356
chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:SimpleTest.waitForFocus/<:1060
  FAIL testPrintMultiple - A promise chain failed to handle a rejection: Tab-modal print UI already open - stack: _openTabModalPrint@chrome://global/content/printUtils.js:181:13
 ```
 
 I also checked manually that when I open a new window the "learn more" button in the enterprise badge has the link ```github.com/mozilla/enterprise-firefox/```

### DIFF
--- a/browser/base/content/browser-sets.js
+++ b/browser/base/content/browser-sets.js
@@ -15,7 +15,7 @@ document.addEventListener(
     document
       .getElementById("mainCommandSet")
       // eslint-disable-next-line complexity
-      .addEventListener("command", async event => {
+      .addEventListener("command", event => {
         switch (event.target.id) {
           case "cmd_newNavigator":
             OpenBrowserWindow();
@@ -202,7 +202,7 @@ document.addEventListener(
             gGestureSupport.rotateEnd();
             break;
           case "cmd_signoutEnterpriseUser":
-            await EnterpriseHandler.onSignOut(window);
+            EnterpriseHandler.onSignOut(window).catch(console.error);
             break;
           case "Browser:OpenLocation":
             openLocation(event);
--- a/browser/components/enterprise/EnterpriseHandler.sys.mjs
+++ b/browser/components/enterprise/EnterpriseHandler.sys.mjs
@@ -260,7 +260,6 @@ export const EnterpriseHandler = {
     lazy.log.debug(`Setting learn more uri to ${validLearnMoreUrl.href}`);
     learnMoreLink.setAttribute("href", validLearnMoreUrl.href);
 
-    win._isEnterpriseLearnMoreLinkConfigured = true;
     learnMoreLink.addEventListener("click", e => {
       let where = lazy.BrowserUtils.whereToOpenLink(e, false, false);
       if (where == "current") {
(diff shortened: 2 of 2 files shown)


browser/base/content/browser-sets.js
```diff
@@ -15,7 +15,7 @@ document.addEventListener(
     document
       .getElementById("mainCommandSet")
       // eslint-disable-next-line complexity
-      .addEventListener("command", async event => {
```
victhorlopez, line 18: The async was added in here https://searchfox.org/enterprise-main/rev/e8c0742b4526a683d68073e2eed2ece46a5a7164/browser/base/content/browser-sets.js#13 on enterprise

Somebody in m-c also added an async function as fire and forget so I followed same pattern, I do not think there is any impact in our code if we do this? 
https://searchfox.org/firefox-main/rev/62b11cf9d978f8c6f8144156feebf42af5d71cf9/browser/base/content/browser-sets.js#131

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -281,7 +283,7 @@ export const EnterpriseHandler = {
     win.PanelUI.showSubView("panelUI-enterprise", element, event);
     const document = element.ownerDocument;
 
-    if (!win._isEnterpriseLearnMoreLinkConfigured) {
+    if (!_learnMoreConfiguredWindows.has(win)) {
       this._setupLearnMoreLink(win);
     }
```
1rneh: We could also set it on `element` (the enterprise badge). To avoid maintaining the windows in a WeakSet. What do you think?

```suggestion
    if (!element._isEnterpriseLearnMoreLinkConfigured) {
      this._setupLearnMoreLink(win);
      element._isEnterpriseLearnMoreLinkConfigured = true;
    }
```

victhorlopez: Done, I assume the element cannot change/be recreated right?

1rneh (approved):
Lgtm, thank you!

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: The enterprise badge `element` is a unique, stable per-window element, which is exactly what we need here :)

browser/base/content/browser-sets.js
jonathanmendez: I reasoned through this myself also. I think both structures amount to a fire and forget for the specific `onSignOut` call, since the [callback is expected to return nothing](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback) (so a returned Promise will never be observed). So removing the `async` from the callback allows the other non-async cases to have their throw errors observed/caught.

jonathanmendez (approved):

PR #702, opened 2026-04-09
Bug 2030640: Fix testPageBlockedByPolicy to use setupPolicyEngineWith…

victhorlopez:
…Json wrapper

Fixes ```browser/components/enterprisepolicies/tests/browser/browser_policy_block_about.js```

```TEST-UNEXPECTED-FAIL | browser/components/enterprisepolicies/tests/browser/browser_policy_block_about.js | Cleanup function threw an exception - at resource://testing-common/EnterprisePolicyTesting.sys.mjs:88 - TypeError: registerCleanupFunction is not a function ```

testPageBlockedByPolicy was bypassing the head.js wrapper and calling EnterprisePolicyTesting.setupPolicyEngineWithJson directly this left _httpd uninitialized so the cleanup called servePolicyWithJson without registerCleanupFunction on a fresh server, causing a TypeError, leaked setInterval, and 404 spam
```
:13.90 GECKO(66642) console.warn: "RemotePoliciesProvider performPolling() with frequency 250 caused error Error: Fetch GET /api/browser/policies failed (404): <html>                    <head><title>404 Not Found</title></head>                    <body>                      <h1>404 Not Found</h1>                      <p>                        <span style='font-family: monospace;'>&#47;&#97;&#112;&#105;&#47;&#98;&#114;&#111;&#119;&#115;&#101;&#114;&#47;&#112;&#111;&#108;&#105;&#99;&#105;&#101;&#115;</span> was not found.                      </p>                    </body>                  </html>"
```

### DIFF
--- a/browser/components/enterprisepolicies/tests/browser/head.js
+++ b/browser/components/enterprisepolicies/tests/browser/head.js
@@ -15,7 +15,7 @@ ChromeUtils.defineESModuleGetters(this, {
 
 PoliciesPrefTracker.start();
 
-async function setupPolicyEngineWithJson(json, customSchema, shutdown = false) {
+async function setupPolicyEngineWithJson(json, customSchema) {
   PoliciesPrefTracker.restoreDefaultValues();
   if (!Services.prefs.getBoolPref("browser.policies.testUseHttp", false)) {
     if (typeof json != "object") {
@@ -35,14 +35,11 @@ async function setupPolicyEngineWithJson(json, customSchema, shutdown = false) {
     );
   }
   if (JSON.stringify(json) === "{}") {
-    if (!shutdown) {
-      return EnterprisePolicyTesting.servePolicyWithJson(
-        { policies: {} },
-        {},
-        registerCleanupFunction
-      );
-    }
-    return EnterprisePolicyTesting.servePolicyWithJson({ policies: {} }, {});
+    return EnterprisePolicyTesting.servePolicyWithJson(
+      { policies: {} },
+      {},
+      registerCleanupFunction
+    );
   }
   return EnterprisePolicyTesting.servePolicyWithJson(
     json,

### CODE REVIEW

**Comment by lissyx** [comment]
This looks like on purpose, we should fix it on m-c

**Review by lissyx — CHANGES_REQUESTED** [review]
I dont think it's the right fix here. This code has been there for years, it's likely we are breaking it with our changes to `EnterprisePolicyTesting.setupPolicyEngineWithJson`

**Comment by lissyx** [comment]
> …Json wrapper
> 
> Fixes `browser/components/enterprisepolicies/tests/browser/browser_policy_block_about.js`
> 
> `TEST-UNEXPECTED-FAIL | browser/components/enterprisepolicies/tests/browser/browser_policy_block_about.js | Cleanup function threw an exception - at resource://testing-common/EnterprisePolicyTesting.sys.mjs:88 - TypeError: registerCleanupFunction is not a function `
> 
> testPageBlockedByPolicy was bypassing the head.js wrapper and calling EnterprisePolicyTesting.setupPolicyEngineWithJson directly this left _httpd uninitialized so the cleanup called servePolicyWithJson without registerCleanupFunction on a fresh server, causing a TypeError, leaked setInterval, and 404 spam
> 
> ```
> :13.90 GECKO(66642) console.warn: "RemotePoliciesProvider performPolling() with frequency 250 caused error Error: Fetch GET /api/browser/policies failed (404): <html>                    <head><title>404 Not Found</title></head>                    <body>                      <h1>404 Not Found</h1>                      <p>                        <span style='font-family: monospace;'>&#47;&#97;&#112;&#105;&#47;&#98;&#114;&#111;&#119;&#115;&#101;&#114;&#47;&#112;&#111;&#108;&#105;&#99;&#105;&#101;&#115;</span> was not found.                      </p>                    </body>                  </html>"
> ```

The problem here is that we are bypassing `head.js`, this is not what we want: https://searchfox.org/enterprise-main/rev/f1803e33aa50eee6e3b1f4d0c5c460045cdefbd3/browser/components/enterprisepolicies/tests/browser/live/browser.toml#9,49

Edit: Though I dont get how we bypass `head.js` if we call `testPageBlockedByPolicy` correctly which is inside it?

**Comment by lissyx** [comment]
@victhorlopez isn't the problem that `servePolicyWithJson` https://searchfox.org/enterprise-main/rev/f1803e33aa50eee6e3b1f4d0c5c460045cdefbd3/toolkit/components/enterprisepolicies/tests/EnterprisePolicyTesting.sys.mjs#57 expects a `registerCleanupFunction` that it calls unconditionnally https://searchfox.org/enterprise-main/rev/f1803e33aa50eee6e3b1f4d0c5c460045cdefbd3/toolkit/components/enterprisepolicies/tests/EnterprisePolicyTesting.sys.mjs#88 

but looking at https://searchfox.org/enterprise-main/rev/f1803e33aa50eee6e3b1f4d0c5c460045cdefbd3/browser/components/enterprisepolicies/tests/browser/head.js#39-43,45,47-51 we may miss in one case to pass it?

File: toolkit/components/enterprisepolicies/tests/EnterprisePolicyTesting.sys.mjs [hunk 1]
```diff
@@ -85,7 +85,7 @@ export var EnterprisePolicyTesting = {
         tokenData.expires_at
       );
 
-      registerCleanupFunction(async () => {
+      registerCleanupFunction?.(async () => {
```
lissyx: That should avoid the failure, but I am not sure it is not hiding a problem, because we would still keed to unregister things?

victhorlopez: yeah I pushed too early, are you sure you want to unregister? the original code was done on purpose with the shutdown flag, if we add it there the shutdown flag will do nothing (i.e both branches will do the same)

lissyx: > yeah I pushed too early, are you sure you want to unregister? the original code was done on purpose with the shutdown flag, if we add it there the shutdown flag will do nothing (i.e both branches will do the same)

Never forget that the original code we're talking about (`serveEnginePolicyWithJson`) is something I hacked to verify manually, but it is not something we had running properly (not running when we were on github actions, hidden in other failures for months since we are on taskcluster), so it may be broken as well.

I really wonder if you can just add `registerCleanupFunction` to the caller where it is missing to fix it? Did you tried that locally?

victhorlopez: that fixes it but then the shutdown flag is not used, I am checking the other tests for usages of that.. and we need to check if we break something

lissyx: My point being, if there's a bug, it's much more likely in this code I wrote to try to tie together things that were not really made for it, than in the historical policies testing code

victhorlopez: got it, I will close the PR as Janika is refactoring also this

PR #708, opened 2026-04-09
Bug 2030737 - Simulate user signed out of enterprise account

1rneh:


victhorlopez (approved):
lgtm

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
```diff
@@ -61,7 +59,18 @@ export class EnterpriseStorageManager {
...
-    const data = await this.#getAccountDataPromise;
+    let data;
+    try {
+      data = await this.#getAccountDataPromise;
+    } catch (e) {
+      if (Services.felt.isFeltBrowser()) {
+        lazy.log.error("Critical! We're not signed into the console?");
+      } else {
+        // We're in felt or we're running m-c tests,
+        // hence we expect to not be signed-in to the console
```
lissyx, line 70: Should we maybe throw/error like above in case we are in Felt? I dont expect us to hit this code in felt, so this may be an indication that something is wrong.

And for tests we can rely on the fact that we are both not `isFeltBrowser()` and `isFeltUI()`, where it is OK we are not signed in

lissyx (approved):
r=me just make sure if my comment makes sense or not

browser/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
1rneh: Sorry, I merged too quickly. 

We do expect to hit this path - both in m-c tests and in Felt (see the change above in `initialize`). Because of that, I only log an error when `Services.felt.isFeltBrowser()` is true, since that’s the only case where hitting this would be unexpected.

lissyx: There's the expected caller that you mention from `initialize`, I was just worrying there may be other coming from else where, now or later, and whether we'd want to know it or not, though I'm unsure of the consequences in fact, so if you think it's good this way, I trust your judgement :)

PR #710, opened 2026-04-09
Bug 2012690 - Enterprise: limit new windows on FELT

lissyx:
While running in FELT, it is not desirable that the user can escape the login process to browse the web as they like. During the SSO flow this can be triggered with one of the page offering a window.open() call, e.g. the Duo documentation. Let's limit the new windows opened in FELT mode to very limited ability to interact: no location bar, no toolbar, no menu.

### DIFF
--- a/browser/base/content/browser-init.js
+++ b/browser/base/content/browser-init.js
@@ -197,6 +197,19 @@ var gBrowserInit = {
     // Call this after we set attributes that might change toolbars' computed
     // text color.
     ToolbarIconColor.init(window);
+
+    // Windows opened when running as Felt should not allow the user to escape
+    // hence remove toolbar, location bar, menubar
+    if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI()) {
+      [
+        "tabbrowser-tabs",
+        "toolbar-menubar",
+        "nav-bar",
+        "PersonalToolbar",
+      ].forEach(id => {
+        document.getElementById(id)?.setAttribute("hide-in-felt", "true");
+      });
+    }
   },
 
   onDOMContentLoaded() {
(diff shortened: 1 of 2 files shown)


lissyx:
There is an uplift ready in #711

Mossop (changes requested):
Rather than making the changes to the features we pass to windows it is probably easier (and safer) to set a top-level attribute on browser windows in the felt UI that lets us hide parts of the UI via CSS. `taskbartab` is a prior example of this in tree.

browser/base/content/browser-init.js
```diff
@@ -197,6 +197,15 @@ var gBrowserInit = {
...
     // text color.
     ToolbarIconColor.init(window);
+
+    // Windows opened when running as Felt should not allow the user to escape
+    // hence remove toolbar, location bar, menubar
+    if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI()) {
+      document
+        .getElementById("toolbar-menubar")
+        ?.setAttribute("hide-in-felt", "true");
+      document.getElementById("nav-bar")?.setAttribute("hide-in-felt", "true");
```
Mossop: You may also want to consider `PersonalToolbar` and `TabsToolbar` for completeness, but I think we hide those by default at the moment anyway

toolkit/themes/shared/global-shared.css
```diff
@@ -421,3 +421,9 @@ button.text-link .button-text {
 #ContentSelectDropdown menuitem[indented] > .menu-text {
   padding-inline-start: var(--space-xxlarge);
 }
+
+/* When in Enterprise and running in Felt mode, some parts of browser should
+ * not be accessible */
+*[hide-in-felt="true"] {
```
Mossop: I think the `*` is not needed.

Mossop (approved):

browser/base/content/browser-init.js
lissyx: Done

toolkit/themes/shared/global-shared.css
lissyx: Yes, done

PR #711, opened 2026-04-09
Bug 2012690 - Enterprise: limit new windows on FELT

lissyx:
While running in FELT, it is not desirable that the user can escape the login process to browse the web as they like. During the SSO flow this can be triggered with one of the page offering a window.open() call, e.g. the Duo documentation. Let's limit the new windows opened in FELT mode to very limited ability to interact: no location bar, no toolbar, no menu.


lissyx:
Setting review here for the uplift of #710

Mossop:
See the comment in #710.

Mossop (approved):
Approved with the same comments as #710

PR #713, opened 2026-04-10
NO BUG - Add PR template

1rneh:


.github/pull_request_template.md
```diff
@@ -0,0 +1,44 @@
+<!-- NICE PR! Thank you for putting in the work!
```
jporter-dev: I like the positivity 😄

.github/pull_request_template.md
```diff
@@ -0,0 +1,44 @@
...
+
+## CHECKLIST FOR AUTHOR (in addition to filling out the template)
+- [ ] Ensure the linked Bugzilla item is up to date
+- [ ] Provide sufficient implementation details in commit messages when necessary  
+
+This information helps reviewers get a quick understanding of your changes. -->
+
+#### Description
+Please include a quick summary of the changes and add the related Bugzilla item ID.
```
jporter-dev: I think this should be in a comment like the other section descriptions?

```suggestion
<!-- Please include a quick summary of the changes and add the related Bugzilla item ID. -->
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,44 @@
...
+
+Bugzilla: bug-<BUG_ID>
```
lissyx: nit: `Bug-<BUG_ID>`

.github/pull_request_template.md
```diff
@@ -0,0 +1,44 @@
...
+
+---
+
+#### Related Issues / Dependencies
+<!-- Link related PRs or issues -->
+
+- Depends on:
+- Blocks:
+- Related:
```
jonathanmendez: I think we want to distinguish between bugzilla and PR here, and I wouldn't want to have to duplicate the information in bugzilla here. So IMO this part should be focused only on PRs (e.g. two PRs that are a 'stack' addressing a single bug). From that perspective I would prioritize:
1. Depends on - this is the most useful for practical decision making: "please review the other one first, please don't merge this unless the other one has been merged".
2. Blocks - this is less likely to affect decision making but can add urgency and context ("it's useful if you prioritize reviewing this because other work is waiting on it and might need to change based on any review comments here")
3. Related - to me this is not likely to affect reviewer decision making, but it can add context.

Based on those priorities I would remove "Related" at least and consider removing "Blocks" (partly because I've found it uncommon to see stacked PRs on our team vs stacked commits within a single PR).

jonathanmendez: Sorry, to clarify I have two suggestions in my comment:
1. Make this section clear that it's only about PR relationships, not bugzilla
2. Remove Related and possibly Blocks

.github/pull_request_template.md
```diff
@@ -0,0 +1,44 @@
+<!-- NICE PR! Thank you for putting in the work!
+
+## CHECKLIST FOR AUTHOR (in addition to filling out the template)
```
jonathanmendez: Do you know how the template is rendered overall for the person creating the PR? Do they see the full raw text in the description editor?

jporter-dev: The comments will show up in the editor, but not when rendered once saved (or previewed)

.github/pull_request_template.md
jonathanmendez: I think there's maybe a meaningful distinction to be made between "hidden instructions to the PR creator" and "placeholder text that marks where replacement text is expected to eventually exist". I think this text is an example of the latter: there _should_ be a summary of the changes in the end, so this is placeholder text that should be replaced.

Maybe a simple scheme can be used to more directly mark such text: `<REPLACE>` (and `<REPLACE_OR_DELETE>` for optional sections) or something.

.github/pull_request_template.md
jonathanmendez:
```suggestion
Bugzilla: bug-<BUG_ID>

Please include a quick summary of the changes and add the related Bugzilla item ID.
```
It would be nice for the bug link to be a) the first thing I see and b) in a consistent location on screen for muscle memory.

.github/pull_request_template.md
```diff
@@ -0,0 +1,44 @@
...
+- [ ] Ensure the linked Bugzilla item is up to date
+- [ ] Provide sufficient implementation details in commit messages when necessary  
+
+This information helps reviewers get a quick understanding of your changes. -->
+
+#### Description
```
jonathanmendez:
```suggestion
### Description
```
(For this and all other sections) I think when rendered on my browser at least, the H3 text stands out better than H4 (which is barely distinguishable from the rest of the text and so doesn't function well as a header).

.github/pull_request_template.md
jporter-dev: If we want to use this as placeholder text, we should have a newline between this and the header. But the common pattern I've seen is using a comment for this guidance as well just to have one less thing to remove/replace when creating a PR (`mozilla/webservices-infra` does this in their template).

.github/pull_request_template.md
victhorlopez: Agree with this one, let's try to have a proper understanding of what is supposed to be on PR level and what goes on bugzilla.

.github/pull_request_template.md
victhorlopez: Does this get autopopulated from the title?

.github/pull_request_template.md
jonathanmendez: It should be fine to not use placeholder text generally, it just seemed like that was the pattern being used here.

I'm also fine with fewer guidance comments if it's just general instructions like "include a description" vs something less obvious or more specific like "_if your PR includes UI changes_, please include _before and after_ screenshots" (emphasis mine to highlight the more specific aspects).

.github/pull_request_template.md
jonathanmendez: No, I believe this template just gets pasted directly into the PR description field when creating a PR. It's possible that a github action/bot could do some sort of autopopulation but that's a different scope of work.

.github/pull_request_template.md
1rneh: Good point about avoiding duplication of information that should live in Bugzilla.

I’m fine with removing both `Blocks` and `Related`. That said, I do think `Related` can be useful, but we can leave it out for now. If it ends up being used more often than expected, we can always add it back to the template.

.github/pull_request_template.md
1rneh: Good point about having to remove/replace as little as possible! I've shortened and clarified comments and instructions. I've also added `REQUIRED`/`OPTIONAL` comments to the section titles - let me know what you think about that. My assumption is that the author would remove the entire section when it's not applicable.

.github/pull_request_template.md
1rneh: Do you think the small checklist is adding too much instruction noise?

lissyx (approved):

jporter-dev (approved):

jonathanmendez (approved):

victhorlopez (approved):

PR #714, opened 2026-04-10
More fixes for enterprise-firefox-try / local debugging

ahal:


taskcluster/gecko_taskgraph/parameters.py
```diff
@@ -155,7 +149,7 @@ def get_defaults(repo_root=None):
         "next_version": None,
         "optimize_strategies": None,
         "phabricator_diff": None,
-        "project": "mozilla-central",
+        "project": "enterprise-firefox",
```
lissyx, line 152: we dont have a better solution than this?

lissyx (approved):

PR #715, opened 2026-04-10
Bug 2030963:  Add isEnterpriseManagedPrimaryPassword to LoginHelper mock in browser_primaryPassword test

victhorlopez:
Fixes the test by adding missing method in mock (pretending we are not enterprise as it is m-c test)
```
MOZ_BYPASS_FELT=1 ./mach mochitest browser/components/preferences/tests/credentials/browser_primaryPassword.js           
```

Issue
```
JavaScript Error: "TypeError: LoginHelper.isEnterpriseManagedPrimaryPassword is not a function"
```


1rneh (approved):
`nit:` Could you add a comment explaining that we are pretending to be a non-enterprise browser in m-c tests?

PR #718, opened 2026-04-13
Bug 2030392: handle exception if browser closes faster than marionette request

victhorlopez:
[Bug-2030392](https://bugzilla.mozilla.org/show_bug.cgi?id=2030392)
The browser is closing faster than marionette in the CI sometimes, this makes send_async_script request from marionette fail.
https://searchfox.org/enterprise-main/rev/8ad7bae3daf8c761d6611955fb03602820ef04a3/testing/enterprise/test_felt_browser_token_refresh.py#44
https://searchfox.org/enterprise-main/rev/8ad7bae3daf8c761d6611955fb03602820ef04a3/testing/enterprise/base_test.py#169

To prevent that we just add a try-catch

### DIFF
--- a/testing/enterprise/test_felt_browser_token_refresh.py
+++ b/testing/enterprise/test_felt_browser_token_refresh.py
@@ -41,7 +41,16 @@ class BrowserTokenRefresh(FeltTests):
         with self.felt_logout_checker.assert_browser_logouts_with(
             "console-forced-logout"
         ):
-            self.get_logged_in_user_info(env=Environment.FIREFOX)
+            self._child_driver.set_context("chrome")
+            # Run synchronously without awaiting the promise — the only purpose is
+            # to trigger a 401/403 response which causes the browser to shut down.
+            self._child_driver.execute_script("""
+                const { ConsoleClient } = ChromeUtils.importESModule(
+                    "resource:///modules/enterprise/ConsoleClient.sys.mjs"
+                );
+                ConsoleClient.getLoggedInUserInfo();
+            """)
+
         self._manually_closed_child = True
         self.assert_child_browser_closed()
 


testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -41,7 +42,10 @@ def assert_browser_closes_on_401(self):
         with self.felt_logout_checker.assert_browser_logouts_with(
             "console-forced-logout"
         ):
-            self.get_logged_in_user_info(env=Environment.FIREFOX)
+            try:
+                self.get_logged_in_user_info(env=Environment.FIREFOX)
+            except (NoSuchWindowException, OSError):
+                pass
```
1rneh: In respect to the errors we expect and catch in `assert_child_browser_closed` (see [here](https://searchfox.org/enterprise-main/rev/74d9e8ce44e87b187286e3bae4ac60962218a497/testing/enterprise/base_test.py#206-209)), we'll probably need to cover those here as well?

That being that, I was wondering instead of keep racing and handling the racy error, we can simply execute the request without awaiting the promise result since the purpose of the request is simply to run into the the `401`/`403` to trigger the browser shutdown.

```
  driver.execute_script("""                                                                                                                                                                                                                                       
      const { ConsoleClient } = ChromeUtils.importESModule(                                                                                                                                                                                                       
          "resource:///modules/enterprise/ConsoleClient.sys.mjs"                                                                                                                                                                                                  
      );                                                                                                                                                                                                                                                          
      ConsoleClient.getLoggedInUserInfo();                                                                                                                                                                                                                        
  """)    
```

victhorlopez: Regarding 
```
  driver.execute_script("""                                                                                                                                                                                                                                       
      const { ConsoleClient } = ChromeUtils.importESModule(                                                                                                                                                                                                       
          "resource:///modules/enterprise/ConsoleClient.sys.mjs"                                                                                                                                                                                                  
      );                                                                                                                                                                                                                                                          
      ConsoleClient.getLoggedInUserInfo();                                                                                                                                                                                                                        
  """)    
  ```
  
  I am not 100% sure it will still not produce the same error (i.e if firefox gets closed from the very same callstack from ```getLoggedInUserInfo``` or if for whatever reason marionette gets slow to process I think we could still get exception.
  
 Regarding exceptions... I can add them but I am not sure we can get these 2
 ```
 errors.InvalidSessionIdException, errors.TimeoutException
 ```

Comment (1rneh) [reply testing/enterprise/test_felt_browser_token_refresh.py]: > I am not 100% sure it will still not produce the same error (i.e if firefox gets closed from the very same callstack from getLoggedInUserInfo or if for whatever reason marionette gets slow to process I think we could still get exception.

From what I understand, the reason why we encounter these errors is because the browser shuts down before `getLoggedInUserInfo` returns a promise result in our async marionette script.

What my suggestion changes: `execute_script` runs the script to completion synchronously and sends back the result to python so we're done on the python side. On the browser side the event loop picks up the XHR request which kicks off a browser shutdown. As we are not awaiting the promise result, we won't encounter the race that this bug is trying to fix.

Comment (victhorlopez) [reply testing/enterprise/test_felt_browser_token_refresh.py]: done, i left it inline as in most cases the user is actually needed i.e
```
user = self.get_logged_in_user_info(env=Environment.FIREFOX)
```

No point in restoring ```self._child_driver.set_context("chrome")``` back to what it was as t hen we will run into the same problen

Comment (1rneh) [reply testing/enterprise/test_felt_browser_token_refresh.py]: > No point in restoring self._child_driver.set_context("chrome") back to what it was as t hen we will run into the same problen

Can you elaborate?

File: testing/enterprise/test_felt_browser_token_refresh.py [hunk 2]
```diff
@@ -10,6 +10,7 @@
 
 from base_test import Environment
 from felt_tests import FeltLogoutChecker, FeltTests
+from marionette_driver import errors
```
1rneh:
```suggestion
```

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: if we try to use the driver (which contacts firefox) it will show the same exceptions

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -41,7 +42,14 @@ def assert_browser_closes_on_401(self):
...
             "console-forced-logout"
         ):
-            self.get_logged_in_user_info(env=Environment.FIREFOX)
+            self._child_driver.set_context("chrome")
+            self._child_driver.execute_script("""
+                const { ConsoleClient } = ChromeUtils.importESModule(
+                    "resource:///modules/enterprise/ConsoleClient.sys.mjs"
+                );
+                ConsoleClient.getLoggedInUserInfo();
+            """)
```
1rneh, line 52: Could you add a comment that we are running this synchronous script and not caring about awaiting the promise result, because its only purpose is to get a `401`/`403` which should trigger the browser shutdown.

1rneh (approved):

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: I will reopen it and write the exception :(. 

Unless you know what happens in here other than python gets some really slow scheduling and by the time firefox is closed then it finishes the send script part?

https://treeherder.mozilla.org/logviewer?job_id=560615527&repo=enterprise-firefox-pr&task=NKbsXjbdQtOkkvigksEy_A.0&lineNumber=21448

1rneh: Looks like we have another race: https://treeherder.mozilla.org/logviewer?job_id=560615527&repo=enterprise-firefox-pr&task=NKbsXjbdQtOkkvigksEy_A.0&lineNumber=21386

We collide with an intermittent remote policies fetch that results in a `401` which triggers a shutdown and the browsing context is discarded before `execute_script` can even run.

PR #719, opened 2026-04-13
Bug 2030397: Do not use port 10080 for mock console as it is blocked by firefox

victhorlopez:
The port 10080 is blocked on firefox, every once in a very while we get a CI failure because of this

https://searchfox.org/firefox-main/rev/6925ce3cf2987bec22cae8a32e1e8a57e49efd3b/netwerk/base/nsIOService.cpp#202

### DIFF
--- a/testing/enterprise/felt_tests.py
+++ b/testing/enterprise/felt_tests.py
@@ -7,13 +7,14 @@ import ctypes
 import datetime
 import json
 import os
-import random
 import shutil
+import socket
 import sys
 import tempfile
 import time
 import urllib.parse
 import uuid
+from contextlib import closing
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from multiprocessing import Array, Process, Value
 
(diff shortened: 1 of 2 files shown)


1rneh (commented):
Hm, looking through the`gBadPortList`, it doesn't seem to list any other ports in that range but I think it might be more robust to look for a free port explicitly? 

```
  import socket                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                  
  def get_free_port():
      with socket.socket() as s:
          s.bind(('', 0))                                                                                                                                                                                                                                         
          return s.getsockname()[1]
                                                                                                                                                                                                                                                                  
  self.console_port = get_free_port()
```

What do you think?

victhorlopez:
> Hm, looking through the`gBadPortList`, it doesn't seem to list any other ports in that range but I think it might be more robust to look for a free port explicitly?
> 
> ```
>   import socket                                                                                                                                                                                                                                                   
>                                                                                                                                                                                                                                                                   
>   def get_free_port():
>       with socket.socket() as s:
>           s.bind(('', 0))                                                                                                                                                                                                                                         
>           return s.getsockname()[1]
>                                                                                                                                                                                                                                                                   
>   self.console_port = get_free_port()
> ```
> 
> What do you think?

done, in theory that has a race between finding the port number and using it, but I guess in practice is almost 0.00% chance?

1rneh:
> done, in theory that has a race between finding the port number and using it, but I guess in practice is almost 0.00% chance?

It's in theory a possible race. But I think the collision risk with the random port selection is "more real" and already happened (hence this bug).

1rneh (approved):
Nice, thanks for extending this to also improve how we choose the other ports!

PR #720, opened 2026-04-13
Bug 2031238:  Use about:buildconfig for restart checks and reduce connect_child_browser wait

victhorlopez:
Navigate (instead of new tab) for speed in CI, use different page because the other one is loading dlls when querying for info

### DIFF
--- a/testing/enterprise/base_test.py
+++ b/testing/enterprise/base_test.py
@@ -106,7 +106,7 @@ class EnterpriseTestsBase(MarionetteTestCase):
     def open_tab_child(self, url):
         return self._open_tab(url, self._child_driver)
 
-    def get_marionette_port(self, max_try=100):
+    def get_marionette_port(self, max_try):
         marionette_port_file = os.path.join(
             self._child_profile_path, "MarionetteActivePort"
         )
@@ -124,8 +124,10 @@ class EnterpriseTestsBase(MarionetteTestCase):
 
         return (marionette_port, marionette_port_file)
 
-    def connect_child_browser(self, capabilities=None):
-        (marionette_port, marionette_port_file) = self.get_marionette_port()
+    def connect_child_browser(self, capabilities=None, max_try=100):
+        (marionette_port, marionette_port_file) = self.get_marionette_port(
+            max_try=max_try
+        )
         assert marionette_port > 0, "Valid marionette port"
         self._logger.info(f"Marionette PORT: {marionette_port}")
 
--- a/testing/enterprise/felt_browser_crashes.py
+++ b/testing/enterprise/felt_browser_crashes.py
@@ -24,7 +24,8 @@ class BrowserCrashes(FeltTests):
         try:
             # This is going to trigger exception for sure
             self._logger.info("Crashing main process")
-            self.open_tab_child("about:crashparent")
+            self._child_driver.set_context("content")
+            self._child_driver.navigate("about:crashparent")
         except Exception as ex:
             self._logger.info(f"Caught exception {ex}")
         finally:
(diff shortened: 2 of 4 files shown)


testing/enterprise/felt_browser_crashes.py
```diff
@@ -24,7 +24,8 @@ def crash_parent(self):
         try:
             # This is going to trigger exception for sure
             self._logger.info("Crashing main process")
-            self.open_tab_child("about:crashparent")
+            self._child_driver.set_context("content")
```
victhorlopez, line 27: no need to restore context here cause it will close/crash (and then we need to try catch it)

testing/enterprise/test_felt_browser_shutdown_crash_exit.py
```diff
@@ -55,7 +55,7 @@ def run_force_quit(self):
         self.wait_process_exit(browser_pid)
         try:
             self._logger.info("Trying to connect to new browser")
-            self.connect_child_browser()
+            self.connect_child_browser(max_try=40)
```
lissyx, line 58: We already have a default value of `max_try=100` which combined with the `time.sleep()` call gives us 50s. Why are we reducing here? From experience with CI on Windows, it's not unusual that things may suddenly be slow and I worry this may in fact **increase** intermittents here

testing/enterprise/test_felt_browser_crash_restart.py
```diff
@@ -25,13 +25,4 @@ def test_browser_crash_abort_restart(self):
...
-        self.wait_process_exit(self._browser_pid)
-        self._logger.info("Connecting to new browser")
-        self.connect_child_browser()
-        self._browser_pid = self._child_driver.session_capabilities["moz:processID"]
-        self._logger.info(f"Connected to {self._browser_pid}")
-        self.open_tab_child("about:support")
-
-        version_box = self.get_elem_child("#version-box")
-        self._child_wait.until(lambda d: len(version_box.text) > 0)
+        self.run_felt_proper_restart()
```
lissyx, line 28: that was a missed opportunity, thanks for fixing :)

testing/enterprise/test_felt_browser_shutdown_crash_exit.py
victhorlopez: we waste 60 seconds on CI, but up to you, can leave it as is

Just to clarify: this connect is only here to ensure that firefox is NOT there, so it is a guaranteed 60 seconds always

testing/enterprise/felt_browser_crashes.py
```diff
@@ -45,10 +46,11 @@ def run_felt_proper_restart(self):
         self.connect_child_browser()
         self._browser_pid = self._child_driver.session_capabilities["moz:processID"]
         self._logger.info(f"Connected to {self._browser_pid}")
-        self.open_tab_child("about:support")
+        with self._child_driver.using_context("content"):
+            self._child_driver.navigate("about:buildconfig")
```
lissyx, line 50: We still have a couple of other `about:support`, do we want to get rid of them as well, except for the one actually checking values? (and do we still want?/need? https://searchfox.org/enterprise-main/rev/8ad7bae3daf8c761d6611955fb03602820ef04a3/testing/enterprise/test_firefox_start.py#23 that is a resurgence of the snap tests I used as a basis to build those marionette, and on Snap packaging we had issues where a wrong version happened at some point so the test made sense)

testing/enterprise/test_felt_browser_shutdown_crash_exit.py
lissyx: GitHub Review hide so much context I missed that :-(

testing/enterprise/felt_browser_crashes.py
victhorlopez: I think as long as we don't crash we will not run into issues.
It looks like it is actually doing some version checks, so maybe we can leave it?
https://searchfox.org/enterprise-main/rev/8ad7bae3daf8c761d6611955fb03602820ef04a3/testing/enterprise/test_firefox_start.py#35

testing/enterprise/base_test.py
```diff
@@ -124,8 +124,10 @@ def get_marionette_port(self, max_try=100):
 
         return (marionette_port, marionette_port_file)
 
-    def connect_child_browser(self, capabilities=None):
-        (marionette_port, marionette_port_file) = self.get_marionette_port()
+    def connect_child_browser(self, capabilities=None, max_try=100):
```
lissyx, line 127: nit: there is already a default value defined on `get_marionette_port`, given it is only called from here, we should keep the default value on `connect_child_browser()` and remove it from `get_marionette_port()`

testing/enterprise/felt_browser_crashes.py
lissyx: Given that `about:support` may be fragile, we should make sure in a follow up to only keep it where it is required. As I said this version check made sense on snap packaging testing, I am not sure it is really relevant here. It might be better to in fact assert we are not `Services.felt?.isFeltUI()` not `Services.felt?.isFeltBrowser()` now that I think about it.

lissyx (approved):

PR #721, opened 2026-04-13
Bug 2031252: fix reload_chrome_window race condition with page marker r=lissyx

victhorlopez:


testing/enterprise/felt_tests.py
```diff
@@ -897,6 +897,39 @@ def maybe_save_screenshot(
...
 
 class FeltTests(FeltTestsBase):
+    def reload_chrome_window(self):
+        # We set a marker before reloading so we can reliably detect when the
+        # new page is ready. A simple readyState == "complete" check is not
+        # sufficient because the old page may still report "complete" for a
+        # brief window after window.location.reload() is called, causing a
+        # false positive. The marker is absent on the new page, so we can
+        # unambiguously tell old page (marker matches), mid-reload (exception),
+        # and new page loading (marker gone, readyState != "complete") apart.
```
lissyx, line 907: 👍

testing/enterprise/felt_updater_errors.py
```diff
@@ -34,11 +34,12 @@ def reload_chrome_window(self):
...
               const { Updates } = ChromeUtils.importESModule("resource:///modules/enterprise/Updates.sys.mjs");
               Updates.uninit();
-              window.location.reload();
-            }).finally(() => callback());
+              callback();
+            });
             """
         )
         self._driver.set_context("content")
+        super().reload_chrome_window()
```
lissyx, line 42: I dont think it is correct, we probably want to rather force the reload of `nsIUpdateManager` after the chrome window reloaod has happened, to be certain there is nothing being done by loading code of the window that may interact and report us an incorrect state

lissyx (approved):
r=me with moving `super().reload_chrome_window()` at the start instead of the end of the method

testing/enterprise/felt_updater_errors.py
victhorlopez: isn't the behaviour that was there?
```
const { Updates } = ChromeUtils.importESModule("resource:///modules/enterprise/Updates.sys.mjs");
              Updates.uninit();
              window.location.reload();
            }).finally(() => callback());
              callback();
            });
```

Uninit and then page reload?

lissyx: yes, but since you are changing it I was wondering if we should not be protective here

testing/enterprise/felt_tests.py
```diff
@@ -897,6 +897,39 @@ def maybe_save_screenshot(
...
+                    if current == marker:
+                        return False
+                    return (
+                        self._driver.execute_script("return document.readyState")
+                        == "complete"
+                    )
+                except Exception:
+                    return False
+
+            self._wait.until(new_page_loaded)
```
lissyx, line 935: Reading that again, is it solid? `MOZ_CHAOSMODE=0` etc? Because when I was iterating around, this kind of construct would not work well over time. Especially `return window.__felt_reload_marker;` after the `window.location.reload()` would usually end up in exception

victhorlopez: why exception? there is

```
self._wait.until
```

which keeps polling the function.
- it will wait when the exception is raised
- it will wait when the value is still the same
- it will succed after it is loaded the for onready.

What is MOZ_CHAOSMODE=0?

testing/enterprise/felt_updater_errors.py
victhorlopez: You mean that if we force reload and Updates.uninit(); as async, it might not finish it we reload after, right?

lissyx: That or just that the code we have in `Updates.sys.mjs` and others is taking to the `UpdateService` and it may mess with the state we expected to reload

testing/enterprise/felt_tests.py
lissyx: I have seen cases where the `window.location.reload()` interaction with marionette ended up in `NoSuchWindow` and other exception.

MOZ_CHAOSMODE=0` enabled chaosmode in a specific mode, that changes how timers are callback and promises are resolved/triggered, and it may ease/improve repro of race conditions.
https://searchfox.org/enterprise-main/rev/8ad7bae3daf8c761d6611955fb03602820ef04a3/toolkit/xre/nsAppRunner.cpp#4213-4223

victhorlopez: I see... I think this is wrong and should be set from outside

```
window.__felt_reload_marker = Math.random();
window.location.reload();
return window.__felt_reload_marker;
```
to and probably try catch it as you said

```
window.__felt_reload_marker = arg_from_python;
window.location.reload();
```

lissyx: Yep, likely less risky

victhorlopez:
I need to recheck.. looks like the code changes made the test fail

testing/enterprise/felt_updater_errors.py
victhorlopez: I tried and does not work, but this code looks harmless (i.e nothing async). 
https://searchfox.org/enterprise-main/rev/8ad7bae3daf8c761d6611955fb03602820ef04a3/browser/components/enterprise/modules/Updates.sys.mjs#63

I assume the check only happens early and we do return here
https://searchfox.org/enterprise-main/rev/8ad7bae3daf8c761d6611955fb03602820ef04a3/browser/components/enterprise/modules/Updates.sys.mjs#30

lissyx (approved):

PR #722, opened 2026-04-13
Bug 2031236: ensure child browser has harness prefs set before launch

victhorlopez:
Issue:  Currently the managed firefox is instanted with its own prefs, which differ from the marionette setup. This is causing issues in some tests (mainly signout verify), so closing the browser might get stuck for 10-30 seconds due to glean as far as I understand

Proposal-solution, prepopulate the same prefs we have for marionette tests before FX child is spawned. 

This fixes mainly the following test, to proper shutdown
```
https://searchfox.org/enterprise-main/rev/74d9e8ce44e87b187286e3bae4ac60962218a497/testing/enterprise/test_felt_browser_signout_verify.py#14
```

### DIFF
--- a/testing/enterprise/felt_tests.py
+++ b/testing/enterprise/felt_tests.py
@@ -23,6 +23,8 @@ from base_test import EnterpriseTestsBase
 from felt_consts import firefox_config
 from marionette_driver import expected
 from marionette_driver.by import By
+from marionette_driver.geckoinstance import DesktopInstance, GeckoInstance
+from mozprofile.prefs import Preferences
 
 
 class SharedString:
@@ -696,6 +698,7 @@ class FeltTestsBase(EnterpriseTestsBase):
             name="enterprise-tests-browser"
         )
         self._logger.info(f"Using browser profile at {self._child_profile_path}")
+        self._apply_marionette_and_local_prefs_to_child_profile()
 
         # Pref does not like passing '\' ?
         if sys.platform == "win32":
@@ -736,6 +739,22 @@ class FeltTestsBase(EnterpriseTestsBase):
         if hasattr(self, "_child_profile_path"):
             self._logger.info(f"Removing browser profile at {self._child_profile_path}")
             shutil.rmtree(self._child_profile_path, ignore_errors=True)
+
+    def _apply_marionette_and_local_prefs_to_child_profile(self):
+        prefs = {
+            k: v
+            for k, v in {
+                **GeckoInstance.required_prefs,
+                **DesktopInstance.desktop_prefs,
+            }.items()
+            # Drop prefs with %-style format placeholders (e.g. "%(server)s")
+            # as they require interpolation that isn't performed here.
+            if not isinstance(v, str) or "%" not in v
+        }
+        prefs.update({"enterprise.is_testing": True, "enterprise.log_level": "Debug"})
+        if hasattr(self, "EXTRA_CHILD_PREFS"):
+            prefs.update(self.EXTRA_CHILD_PREFS)
+        Preferences.write(os.path.join(self._child_profile_path, "user.js"), prefs)
 
     def set_string_pref(self, pref_name, pref_value):
         self._logger.info(f"Setting {pref_name} to {pref_value}")
--- a/testing/enterprise/test_felt_browser_restart_works.py
+++ b/testing/enterprise/test_felt_browser_restart_works.py
@@ -37,6 +37,10 @@ class BrowserRestartWorks(FeltTests):
             self._logger.info("Received expected UnknownException")
         except NoSuchWindowException:
             self._logger.info("Received expected NoSuchWindowException")
+        except OSError:
+            self._logger.info(
+                "Firefox quit before execute_script returned, no data received over Marionette socket"
+            )
         finally:
             self._logger.info(
                 f"Issued restartecting quit underway, checking PID {self._browser_pid}"
--- a/testing/enterprise/test_felt_browser_signout.py
+++ b/testing/enterprise/test_felt_browser_signout.py
@@ -12,6 +12,7 @@ sys.path.append(os.path.dirname(__file__))
 from base_test import Environment
 from felt_tests import FeltTests
 from marionette_driver.errors import (
+    NoAlertPresentException,
     UnexpectedAlertOpen,
 )
 
@@ -33,14 +34,7 @@ class BaseBrowserSignout(FeltTests):
         self._driver.set_context("content")
         return private_cookies
 
-    def run_perform_signout(self):
-        self.connect_child_browser(
-            capabilities={
-                # Do not auto-handle prompts.
-                "unhandledPromptBehavior": "ignore"
-            }
-        )
-
+    def _do_signout(self):
         self.assert_user_signed_in(env=Environment.FIREFOX)
         # Cache email for later use in prefilled email input field assertion
         user = self.get_logged_in_user_info(env=Environment.FIREFOX)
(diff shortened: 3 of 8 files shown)


testing/enterprise/test_felt_browser_restart_works.py
```diff
@@ -37,6 +37,8 @@ def run_felt_perform_restart(self):
             self._logger.info("Received expected UnknownException")
         except NoSuchWindowException:
             self._logger.info("Received expected NoSuchWindowException")
+        except OSError:
```
victhorlopez, line 40: all this OSError come from 
```
"browser.tabs.remote.unloadDelayMs": 0
```

testing/enterprise/felt_tests.py
```diff
@@ -729,6 +732,24 @@ def teardown(self):
             self._logger.info(f"Removing browser profile at {self._child_profile_path}")
             shutil.rmtree(self._child_profile_path, ignore_errors=True)
 
+    def _apply_prefs_to_child_profile(self):
```
lissyx: nit: 
```suggestion
    def _apply_marionette_and_local_prefs_to_child_profile(self):
```

testing/enterprise/test_felt_browser_restart_works.py
lissyx: Can you be more specific?

testing/enterprise/test_felt_browser_signout.py
```diff
@@ -66,14 +60,18 @@ def run_perform_signout(self):
...
+            try:
+                driver.switch_to_alert().accept()
+                self._logger.info(
+                    "Signing out the user by clicking the Signout button in the dialog"
+                )
+                return True
+            except NoAlertPresentException:
+                return False
+
+        self._waiter(self._child_driver).until(wait_for_and_accept_alert)
```
lissyx, line 74: That's on me when I did the migration from `Selenium` to `Marionette`, I even left the `Wait()` call in comment, I was not sure how marionette behaved there :)

testing/enterprise/test_felt_browser_signout_on_exit.py
```diff
@@ -50,5 +50,8 @@ def perform_quit(self):
             Services.startup.quit(Ci.nsIAppStartup.eForceQuit);
             """,
         )
-        driver.set_context("content")
+        try:
+            driver.set_context("content")
+        except OSError:
+            pass
```
lissyx: Why do we have this one? Related to the pref you mentionned? Can you share a bit of context to make sure we are not hiding a legit issue?

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -0,0 +1,142 @@
...
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from marionette_driver.geckoinstance import DesktopInstance, GeckoInstance
+from test_felt_browser_signout import BaseBrowserSignout
+
+
+class HarnessPrefs(BaseBrowserSignout):
```
lissyx, line 15: Marionette has this notion of `harness_unit`, I'm wondering if we should not have one for us now? https://searchfox.org/enterprise-main/source/testing/marionette/harness/marionette_harness/tests/harness_unit/

lissyx (approved):

testing/enterprise/test_felt_browser_signout_on_exit.py
victhorlopez: same, because it closes f aster now, you are not allowed to run  driver.set_context("content") if you closed the browser

testing/enterprise/test_felt_browser_restart_works.py
victhorlopez: yeah I can, actually I was wrong, the setting is 

"browser.startup.page": 0


If you set it to 0 you get OSError cause the browser closes faster, but you can correct me here, I just did some try removing all settings and just setting that one to 0 and 1

testing/enterprise/test_felt_harness_prefs.py
victhorlopez: we can create a bug if we see value on this, not sure how much value it brings now if this test already covers us from the issues we might face.


What we should fix imo is screenshot generation on child browser and crash dumps gathering on child.

testing/enterprise/test_felt_browser_signout_on_exit.py
lissyx: Make sure we have a comment to explain

testing/enterprise/test_felt_harness_prefs.py
lissyx: Yes, a follow up would be good enough. I'm not sure which issue you have in mind about screenshot generation?

testing/enterprise/test_felt_browser_restart_works.py
lissyx: Let's just make sure we have explanation why we catch, I'm always a bit worried when we catch such a generic error

testing/enterprise/test_felt_harness_prefs.py
victhorlopez: when a test fails, marionette makes a screenshot on the driver, In here you can see where a Felt screenshot is done in the report

https://firefoxci.taskcluster-artifacts.net/YGZvo8yZTUm5NH4v70iA5w/0/public/test_info/report.html

this only works currently when the issue is in felt, if the test fails on firefox, you see nothing.

lissyx: Do you have a link with such a case?

victhorlopez: This one ( but you will never see a screenshot from firefox, I didnt check in detail, of course it is not expected to see anything if the browser closed, but for timeouts on get_element you might see it)

https://firefoxci.taskcluster-artifacts.net/Q36PCoNoTNqI6NRLyFmpHQ/0/public/test_info/report.html

Just to clarify, we have 2 windows, 2 drivers. 

The current approach only works on the driver,

If you check some test report from Mn you can see
https://firefoxci.taskcluster-artifacts.net/UDg53xi0RF2SFfOpHwBAUg/0/public/test_info/report.html


See 
https://searchfox.org/enterprise-main/rev/74d9e8ce44e87b187286e3bae4ac60962218a497/testing/marionette/harness/marionette_harness/runner/base.py#722

lissyx: > This one ( but you will never see a screenshot from firefox, I didnt check in detail, of course it is not expected to see anything if the browser closed, but for timeouts on get_element you might see it)
> 
> https://firefoxci.taskcluster-artifacts.net/Q36PCoNoTNqI6NRLyFmpHQ/0/public/test_info/report.html
> 
> Just to clarify, we have 2 windows, 2 drivers.
> 
> The current approach only works on the driver,
> 
> If you check some test report from Mn you can see https://firefoxci.taskcluster-artifacts.net/UDg53xi0RF2SFfOpHwBAUg/0/public/test_info/report.html

I'm lost here, what you share shows it is working. One is an assertion failure the other is an exception timeout, I remember having to manually screenshot in the case of exception timeout, so I'm not sure if there is anything broken (that is fixable) here.

> 
> See https://searchfox.org/enterprise-main/rev/74d9e8ce44e87b187286e3bae4ac60962218a497/testing/marionette/harness/marionette_harness/runner/base.py#722

victhorlopez: What I am trying to show here is that we will never get a screenshot of a child browser in a report in case the test happens to fail on the child and not on the felt

```
If you check some test report from Mn you can see
https://firefoxci.taskcluster-artifacts.net/UDg53xi0RF2SFfOpHwBAUg/0/public/test_info/report.html
```
The link here was from Mn not MnEnt

lissyx: > What I am trying to show here is that we will never get a screenshot of a child browser in a report in case the test happens to fail on the child and not on the felt
> 
> ```
> If you check some test report from Mn you can see
> https://firefoxci.taskcluster-artifacts.net/UDg53xi0RF2SFfOpHwBAUg/0/public/test_info/report.html
> ```
> 
> The link here was from Mn not MnEnt

OK I see your point here, I was missing the fact that it happens when we have both and it 100% makes sense now reading the marionette code. I worry we will have to do it ourselves

1rneh:
Removed myself as reviewer as @lissyx has already signed off of the changes and I don't want to block the landing (as I wouldn't be able to take a look at it today.)

PR #724, opened 2026-04-13
Bug 2013821, Bug 2013823, Bug 2031742 - Set expected default theme in test when running in Enterprise

jonathanmendez:
### Description 

Bugzilla:
Bug-2013821
Bug-2013823
Bug-2031742

The tests were failing because they expect the `default-theme@mozilla.org` to be installed. On Firefox Enterprise the default theme is instead `firefox-enterprise-light@mozilla.org`, but in the test environment the theme doesn't get loaded. This was apparently encountered also in Firefox Desktop DevEdition builds, in which case `default-theme@mozilla.org` is enabled for the tests. This change does the same for Enterprise builds.

Bug 2013823 captured a side effect of the failure in 2013821. When the test fails an assertion, the test is aborted and does not fully clean up. Fixing the assertion allows the cleanup to occur.

---

### Testing 

* [x] This is a fix to an existing test.

### DIFF
--- a/toolkit/mozapps/extensions/test/xpcshell/test_amo_stats_telemetry.js
+++ b/toolkit/mozapps/extensions/test/xpcshell/test_amo_stats_telemetry.js
@@ -40,7 +40,10 @@ add_setup(async () => {
   // xpcshell test environment and so here we force enable the default theme
   // (which is installed by disabled) to let this test to run across all builds
   // included the DevEdition beta builds.
-  if (AppConstants.MOZ_DEV_EDITION) {
+  // Similarly in Firefox Enterprise the default theme is expected to be
+  // firefox-enterprise-light@mozilla.org, but it isn't loaded in the minimal
+  // xpcshell test environment.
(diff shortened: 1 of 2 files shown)


1rneh (approved):

1rneh:
I'm curious, how did you only run these three tasks?

jonathanmendez:
> I'm curious, how did you only run these three tasks?

@1rneh I am also curious about that. I did not do it intentionally. My guess is that it happened because I only touched an xpshell test file? More builds did actually run but definitely not the usual amount
image

jonathanmendez:
In double-checking my fix had passed the tests, I found another remarkably similar test failure, so I've fixed it the same way.

1rneh:
> @1rneh I am also curious about that. I did not do it intentionally. My guess is that it happened because I only touched an xpshell test file? More builds did actually run but definitely not the usual amount

That explanation sounds reasonable!

1rneh (approved):

PR #731, opened 2026-04-15
Bug 2031939 - Enterprise: Add code-review for all marionette and mochitest-plain

lissyx:
Those suites should be stable green now on Enterprise branches, and only have the same intermittents as m-c, so report them on GitHub Checks. It is OK to merge with those reporting failures as long as it was verified it is known m-c issue, or existing intermittent.

### Description 

Bugzilla: Bug-2031939


taskcluster/gecko_taskgraph/transforms/test/other.py
```diff
@@ -652,12 +652,15 @@ def handle_tier(config, tasks):
...
                 "linux64/debug",
+                "linux64-enterprise/opt",
+                "linux64-enterprise/debug",
                 "linux64-shippable/opt",
                 "linux64-devedition/opt",
                 "linux64-asan/opt",
                 "linux64-qr/opt",
                 "linux64-qr/debug",
                 "linux64-shippable-qr/opt",
+                "linux64-enterprise-shippable/opt",
```
jonathanmendez: nit; Thinking of potential future merge conflicts, I think it would be nice if the grouping of these matches the grouping for the other platforms, i.e. put `linux64-enterprise-shippable/opt` just after `linux64-enterprise/debug`.

taskcluster/gecko_taskgraph/transforms/test/other.py
```diff
@@ -693,6 +702,9 @@ def handle_tier(config, tasks):
                 "macosx1400-64-qr/debug",
                 "macosx1500-64-shippable/opt",
                 "macosx1500-64/debug",
+                "macosx1500-64-enterprise-shippable/opt",
+                "macosx1500-64-enterprise/opt",
+                "macosx1500-64-enterprise/debug",
```
jonathanmendez: Even more nit: the order of these doesn't match the order for the other platforms.
```suggestion
                "macosx1500-64-enterprise/opt",
                "macosx1500-64-enterprise/debug",
                "macosx1500-64-enterprise-shippable/opt",
```

jonathanmendez (approved):
A couple of optional nits

taskcluster/gecko_taskgraph/transforms/test/other.py
lissyx: i was groupping with shippable but your argument holds

lissyx: (but i was inconsistent doing it ...)

taskcluster/gecko_taskgraph/transforms/test/other.py
lissyx: yes, thanks

lissyx:
I am going to merge this. If it's proving to be too problematic, we'll revert and wait for adding some filtering from orangefactor or similar to help

PR #732, opened 2026-04-15
Bug 2031964 - Enterprise: consider sleep events for force restart policy

lissyx:
The scheduled time for a restart may happen during device sleep, in which case no restart at all will be performed. Monitoring the wake_notification observer to check when device resumes whether the scheduled restart should have been done, and if that is the case, cancel the deferred task, remove the notification and prepare for a new forced restart in a couple of minutes.

### Description 

Bugzilla: Bug-2031964

---

### Testing 

Confere https://bugzilla.mozilla.org/show_bug.cgi?id=2028364#c0

<!-- 
**Steps to verify changes:**
1. Send force restart policy from console `RelaunchRequired` (at least 1h after the current time otherwise it will schedule next day)
2. Restart
3. In browser console, `Services.obs.notifyObservers(null, "update-staged")`
4. Notification shows
5. Put device to sleep to cover the restart time
6. Put device back from sleep

**Expected result:**

Notification should be updated, new scheduled time being now +5min

### DIFF
--- a/browser/modules/UpdatePolicyEnforcer.sys.mjs
+++ b/browser/modules/UpdatePolicyEnforcer.sys.mjs
@@ -5,6 +5,7 @@
 const lazy = {};
 const PREF_APP_UPDATE_COMPULSORY_RESTART = "app.update.compulsory_restart";
 let deferredRestartTasks = null;
+let notification = null;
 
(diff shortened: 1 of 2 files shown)


lissyx:
This should land directly on m-c for now

PR #733, opened 2026-04-15
Bug 2032016 - Control enterprise log level via preference enterprise.logLevel

1rneh:
### Description 

Bugzilla: Bug-2032016

Converts all usages to console Web API to use a customized console instance to enable controlling enterprise specific logs via the preference `enterprise.logLevel`.

### DIFF
--- a/browser/components/enterprise/EnterpriseCommon.sys.mjs
+++ b/browser/components/enterprise/EnterpriseCommon.sys.mjs
@@ -23,7 +23,15 @@ export const shouldNotCloseWindow = () => {
   return Services.prefs.getBoolPref(SHOULD_NOT_CLOSE_WINDOW, false);
 };
 
+const ENTERPRISE_LOGLEVEL_PREF = "enterprise.loglevel";
+
 export const EnterpriseCommon = {
   ENTERPRISE_DEVICE_ID_PREF: "enterprise.sync.device_id",
-  ENTERPRISE_LOGLEVEL_PREF: "enterprise.loglevel",
 };
+
+export function createEnterpriseLogger(logPrefix) {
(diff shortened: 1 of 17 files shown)


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -102,7 +101,7 @@ export const ConsoleClient = {
     try {
       consoleURI = Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
     } catch (e) {
-      console.error("Critial misconfiguration: Missing console URI.");
+      lazy.log.error("Critial misconfiguration: Missing console URI.");
```
lissyx, line 104: This will show up by default, right, no specific value required of the logging pref ?

browser/components/enterprise/EnterpriseCommon.sys.mjs
```diff
@@ -23,7 +23,15 @@ export const shouldNotCloseWindow = () => {
   return Services.prefs.getBoolPref(SHOULD_NOT_CLOSE_WINDOW, false);
 };
 
+const ENTERPRISE_LOGLEVEL_PREF = "enterprise.logLevel";
```
lissyx: nit: I know in a couple of other places they use `logLevel` but given all our prefs I'd prefer `log_level`

lissyx (approved):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Yes, error messages are logged to the console by default.

PR #734, opened 2026-04-15
Bug 2026004 - Add support for future SitePolicies in Lockdown Mode UX

jporter-dev:
### Description 

This PR builds upon #554 which relied on the JIT-related policy check to toggle the Lockdown Mode urlbar button using `isAllowedForURI("jit")` which will not work as additional policies are added. The addition of `hasSitePoliciesForURI` checks for the existence of any policy on the given URI independent of which features are configured. 

Bugzilla: Bug-2026004

---

* Depends on: #554 

---

### Screenshots 

No visual changes, screenshots should appear identical as they are on #554.

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. `mach build`
2. Ensure policy is set with `SitePolicies`

```json
{
  "policies": {
    "SitePolicies": [
      {
        "Match": ["*.mozilla.org"],
        "Policies": {
          "DisableJit": true
        }
      }
    ]
  }
}
```

3. `mach run`
4. Verify policy is applied `about:policies`
5. Navigate to https://mozilla.org
6. Observe icon in URL bar and click for subpanel
7. Navigate to any other page or tab for icon to be removed

**Expected result:**

Lockdown mode UI displays for configured url patterns, and does not display for others.

### DIFF
--- a/toolkit/components/enterprisepolicies/SitePolicyUtils.sys.mjs
+++ b/toolkit/components/enterprisepolicies/SitePolicyUtils.sys.mjs
@@ -43,9 +43,10 @@ export const SitePolicyUtils = {
       }
 
       if (
-        policies.match.matches(uri) ||
-        // This is necessary to correctly match moz-nullprincipal URIs.
-        policies.match.matchesAllWebUrls
+        (policies.match.matches(uri) ||
+          // This is necessary to correctly match moz-nullprincipal URIs.
+          policies.match.matchesAllWebUrls) &&
+        Object.keys(policies.features).length
       ) {
         return true;
       }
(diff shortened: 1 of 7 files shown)


toolkit/components/enterprisepolicies/SitePolicyUtils.sys.mjs
```diff
@@ -31,4 +31,26 @@ export const SitePolicyUtils = {
...
+      ) {
+        continue;
+      }
+
+      if (
+        policies.match.matches(uri) ||
+        // This is necessary to correctly match moz-nullprincipal URIs.
+        policies.match.matchesAllWebUrls
+      ) {
+        return true;
```
Mossop, line 51: So this doesn't check whether there are actually any policies. Seems unlikely an admin would add matches but no policies but it is possible. Do we care about catching this case?

Mossop (approved):
Looks good unless you want to cover the case I commented on.

toolkit/components/enterprisepolicies/SitePolicyUtils.sys.mjs
jporter-dev: I went ahead and added this as a defensive guard to not convey the wrong message with the lockdown mode indicator

PR #735, opened 2026-04-15
Bug 2005719 - Enterprise: Force Felt window focus when queueing pending links

lissyx:
When the user triggers a new start of FELT by opening a link from a third party application, make sure to force focusing the FELT window for UX awareness of the action.

### Description 

Bugzilla: Bug-2005719

---

### Testing 

* [x] Added tests

**Steps to verify changes:**
1. Start Felt
2. Make sure it is not foreground
3. Open link from external application

**Expected result:**
Felt gets the focus

### DIFF
--- a/testing/enterprise/test_felt_external_link.py
+++ b/testing/enterprise/test_felt_external_link.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import os
+import subprocess
+import sys
+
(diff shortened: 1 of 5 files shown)


testing/enterprise/test_felt_external_link.py
```diff
@@ -0,0 +1,59 @@
...
+
+            console.debug(`FeltTests: run_felt_open_external_link(): window.focus() override`);
+            window.focus = function(...args) {
+              console.debug(`FeltTests: run_felt_open_external_link(): window.focus() override calls original`);
+              window.focusWasRequested = true;
+              return originalFocus(...args);
+            };
+            """
+        )
+        assert not self.get_focus_was_requested(), "Window did not request focus"
```
gcp: Assertion failure would be "Window requested focus" (the opposite).

gcp (approved):

PR #738, opened 2026-04-16
Bug 2032161: fix double HTTP response when check_auth fails in mock console r=lissyx

victhorlopez:
…onsole

### Description 

Bugzilla: Bug-2032161

check_auth() was sending a 401 reply but returning None, so callers in do_GET continued executing and sent a second response on the same connection. This caused a ConnectionAbortedError on Windows when the client closed the socket after the 401 before the server finished writing the duplicate response body.

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [ ] Added tests
* [x] Run current tests


lissyx (approved):
Ah thanks, this is something I had fix on a PR and I think never landed for reasons.

PR #740, opened 2026-04-16
Bug 2032305 - Enterprise: wait for distribution completion before getting enterprise console address

lissyx:
### Description 

Bugzilla: Bug-2032305

### Summary

Expecting preferences from `distribution.ini` to be populated in time without any further verification is wrong as of now, we need to observe the `DistributionCustomizer` if the preference value does not exists.

Follow up have been filed:
 - improving this dependency: Bug-2032453
 - adding the pref specific observer topic: Bug-2032453

### DIFF
--- a/browser/components/distribution.sys.mjs
+++ b/browser/components/distribution.sys.mjs
@@ -4,6 +4,8 @@
 
 const DISTRIBUTION_CUSTOMIZATION_COMPLETE_TOPIC =
   "distribution-customization-complete";
+const DISTRIBUTION_PREFERENCES_COMPLETE_TOPIC =
+  "distribution-preferences-complete";
 
 const PREF_CACHED_FILE_EXISTENCE = "distribution.iniFile.exists.value";
 const PREF_CACHED_FILE_APPVERSION = "distribution.iniFile.exists.appversion";
@@ -618,15 +620,18 @@ DistributionCustomizer.prototype = {
     }
 
     let prefDefaultsApplied = this._prefDefaultsApplied || !this._ini;
-    if (
-      this._customizationsApplied &&
-      this._bookmarksApplied &&
-      prefDefaultsApplied
-    ) {
+    if (this._customizationsApplied && prefDefaultsApplied) {
       Services.obs.notifyObservers(
         null,
-        DISTRIBUTION_CUSTOMIZATION_COMPLETE_TOPIC
+        DISTRIBUTION_PREFERENCES_COMPLETE_TOPIC
       );
+
+      if (this._bookmarksApplied) {
+        Services.obs.notifyObservers(
+          null,
+          DISTRIBUTION_CUSTOMIZATION_COMPLETE_TOPIC
+        );
+      }
     }
   },
 };
--- a/browser/components/enterprise/EnterpriseEndpoints.sys.mjs
+++ b/browser/components/enterprise/EnterpriseEndpoints.sys.mjs
@@ -53,10 +53,10 @@ export const BASE_CONSOLE_URI_PREFS = new Set([
 ]);
 
 export const EnterpriseEndpoints = {
-  init() {
+  async init() {
     lazy.log.info("Setting and locking enterprise endpoints");
 
-    const consoleBaseURI = lazy.ConsoleClient.consoleBaseURI;
+    const consoleBaseURI = await lazy.ConsoleClient.consoleBaseURI;
 
     const defaultBranch = Services.prefs.getDefaultBranch("");
 
--- a/browser/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/browser/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -81,6 +81,7 @@ class InvalidAuthError extends Error {
  */
 export const ConsoleClient = {
   _refreshPromise: null,
+  _consoleUriReadyPromise: null,
 
   /**
    * This is Felt-specific: While the browser is running, it has ownership of the
@@ -94,17 +95,57 @@ export const ConsoleClient = {
   /**
    * Base URL of the remote enterprise console
    *
-   * @returns {URL}
+   * @throws {Error}
+   * @returns {Promise<URL>}
    */
   get consoleBaseURI() {
-    let consoleURI;
-    try {
-      consoleURI = Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
-    } catch (e) {
-      lazy.log.error("Critial misconfiguration: Missing console URI.");
-      throw e;
+    if (!this._consoleUriReadyPromise) {
+      this._consoleUriReadyPromise = new Promise((resolve, reject) => {
+        try {
+          const consoleURI = Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
+          resolve(consoleURI);
+        } catch (e) {
+          lazy.log.warn(
+            `Missing console URI. Waiting on distribution customization to complete.`
+          );
+          const kDistributionPreferencesCompleteTopic =
+            "distribution-preferences-complete";
+          const distributionCompleteObserver = {
+            observe(_aSubject, aTopic, _aData) {
+              Services.obs.removeObserver(
+                distributionCompleteObserver,
+                "xpcom-shutdown"
+              );
+              Services.obs.removeObserver(
+                distributionCompleteObserver,
+                kDistributionPreferencesCompleteTopic
+              );
+              if (aTopic === kDistributionPreferencesCompleteTopic) {
+                try {
+                  const consoleURI =
+                    Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
+                  resolve(consoleURI);
+                } catch (ex) {
+                  lazy.log.error(
+                    `Critical misconfiguration: Missing console URI`
+                  );
+                  reject(ex);
+                }
+              }
+            },
+          };
+          Services.obs.addObserver(
+            distributionCompleteObserver,
+            kDistributionPreferencesCompleteTopic
+          );
+          Services.obs.addObserver(
+            distributionCompleteObserver,
+            "xpcom-shutdown"
+          );
+        }
+      });
     }
-    return new URL(consoleURI);
+    return this._consoleUriReadyPromise.then(url => new URL(url));
   },
 
   /**
@@ -131,8 +172,8 @@ export const ConsoleClient = {
    * @param {string} path
    * @returns {string} Absolute URL string.
    */
-  constructURI(path) {
-    const url = this.consoleBaseURI;
+  async constructURI(path) {
+    const url = await this.consoleBaseURI;
     url.pathname = path;
     return url.href;
   },
@@ -144,9 +185,9 @@ export const ConsoleClient = {
    * @param {string} devicePostureToken - Token received for device posture
    * @returns {nsIURI}
    */
-  constructSsoLoginURI(email, devicePostureToken) {
+  async constructSsoLoginURI(email, devicePostureToken) {
     const deviceId = lazy.FeltStorage.getDeviceId();
-    const url = this.consoleBaseURI;
+    const url = await this.consoleBaseURI;
     url.pathname = this._paths.SSO;
     url.searchParams.set("target", "browser");
     url.searchParams.set("email", email);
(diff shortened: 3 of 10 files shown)


browser/components/distribution.sys.mjs
```diff
@@ -4,6 +4,8 @@
 
 const DISTRIBUTION_CUSTOMIZATION_COMPLETE_TOPIC =
   "distribution-customization-complete";
+const DISTRIBUTION_PREFERENCES_COMPLETE_TOPIC =
+  "distribution-preferences-complete";
```
lissyx, line 8: @mkaply If that is OK for you I'll upstream as https://bugzilla.mozilla.org/show_bug.cgi?id=2032450

lissyx:
@1rneh Tagging for review mostly for the `ConsoleClient` and its dependencies changes where we need to make use of promises
@victhorlopez Tagging for review on the test harness part
@mkaply Tagging  for review on the `DistributionCustomizer` part

browser/components/distribution.sys.mjs
mkaply: Now that I look at this, I'm confused. Are you saying that you weren't getting CUSTOMIZATION_COMPLETE at all?

Why did you add a new topic instead of just listening for CUSTOMIZATON_COMPLETE?

lissyx: Yes, this is exactly what I was talking about on Slack: no observer `distribution-customization-complete`
 fired at all

lissyx: firing it depends on bookmarks applied: https://searchfox.org/enterprise-main/rev/8e6c06bf3c573ee2682ef7233a8a04bfe32a7c07/browser/components/distribution.sys.mjs#623
set at https://searchfox.org/enterprise-main/rev/8e6c06bf3c573ee2682ef7233a8a04bfe32a7c07/browser/components/distribution.sys.mjs#309
called from https://searchfox.org/enterprise-main/rev/8e6c06bf3c573ee2682ef7233a8a04bfe32a7c07/browser/components/distribution.sys.mjs#687-690

lissyx: And it's called by https://searchfox.org/enterprise-main/rev/8e6c06bf3c573ee2682ef7233a8a04bfe32a7c07/browser/components/places/PlacesBrowserStartup.sys.mjs#190,234

Since we have no bookmarks here ...

testing/enterprise/base_test.py
```diff
@@ -33,16 +35,49 @@ def setUp(self):
...
         super().setUp()
 
+        # Some test may need a non modified distribution.ini, so respect it and
+        # and do not overwrite it when requested.
+        # Also some tests may not define a console port, so skip this for them.
+        # All this needs to happen before process is started to avoid race
+        # conditions, but requires self.marionette that is setup by
+        # super().setUp() right above.
+        if hasattr(self, "console_port") and not hasattr(self, "KEEP_DISTRIBUTION_INI"):
+            self.distribution_ini_path = self.get_distribution_ini(self.marionette)
```
victhorlopez: Does it make sense to put this in another function to have the setup readable ( which is already hard to read) or you leaft it in here for a reason

```
            self.distribution_ini_path = self.get_distribution_ini(self.marionette)
            self.distribution_ini_orig_path = os.path.join(
                os.path.dirname(self.distribution_ini_path), "distribution.ini.orig"
            )
            assert os.path.isfile(self.distribution_ini_path)
            self._logger.info(
                f"Backup {self.distribution_ini_path} as {self.distribution_ini_orig_path}"
            )
            shutil.copy(self.distribution_ini_path, self.distribution_ini_orig_path)

            self._logger.info(f"Writing console pref in {self.distribution_ini_path}")
            with open(self.distribution_ini_path, "w") as dist_ini:
                dist_ini.write(f"""# Test specific distribution.ini file
[Global]
id=enterprise-test
version=1.0
about=Mozilla Firefox Enterprise Test Build

[Preferences]
enterprise.console.address=http://localhost:{self.console_port}
""")
```

browser/components/distribution.sys.mjs
mkaply: That should have broken so many other things.

I'm ok with this fix for now, but I want to dig into this more on release.

lissyx: > That should have broken so many other things.

My changes ? `distribution-customization-complete` still fires under the same conditions as before here

lissyx: > I'm ok with this fix for now, but I want to dig into this more on release.

I had opened https://bugzilla.mozilla.org/show_bug.cgi?id=2032450

testing/enterprise/base_test.py
```diff
@@ -62,10 +97,38 @@ def tearDown(self):
             self.marionette.instance.app_args = deepcopy(self._saved_cli_args)
             del self._saved_cli_args
 
+        if hasattr(self, "_extra_prefs"):
+            self.marionette.instance.prefs = None
```
victhorlopez, line 82: can you write a comment why we do this

testing/enterprise/base_test.py
```diff
@@ -62,10 +97,38 @@ def tearDown(self):
...
+
         del os.environ["MOZ_DISABLE_NONLOCAL_CONNECTIONS"]
 
         self.marionette.quit(in_app=False, clean=True)
 
+        if hasattr(self, "console_port") and not hasattr(self, "KEEP_DISTRIBUTION_INI"):
+            assert os.path.isfile(self.distribution_ini_path)
```
victhorlopez, line 99: does this behave good if a test fails?(i.e I am not sure about asserting on teardown)

testing/enterprise/base_test.py
lissyx: why not yes

testing/enterprise/base_test.py
lissyx: sure, let me know what you think about this one

testing/enterprise/base_test.py
lissyx: ok let's simplify

testing/enterprise/base_test.py
victhorlopez: I would prefer if you move it in a function as I said to not clutter the setup method (i.e create_ini() or whatever you want to call it) but fine if you want to have it here

victhorlopez (approved):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -94,17 +94,38 @@ export const ConsoleClient = {
   /**
    * Base URL of the remote enterprise console
    *
-   * @returns {URL}
+   * @returns {Promise<{URL: string}>}
```
1rneh: The promise result remains to be an URL, not a string.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -94,17 +94,44 @@ export const ConsoleClient = {
...
   get consoleBaseURI() {
-    let consoleURI;
-    try {
-      consoleURI = Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
-    } catch (e) {
-      lazy.log.error("Critial misconfiguration: Missing console URI.");
-      throw e;
-    }
-    return new URL(consoleURI);
+    const consolePromise = new Promise((resolve, reject) => {
```
1rneh: This creates a new promise on every call and, in the slow-path, registers a new observer each time. Let's create it only once and memoize (`this._consoleUriReadyPromise`) and return it.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -94,17 +94,44 @@ export const ConsoleClient = {
...
+              Services.obs.removeObserver(
+                distributionCompleteObserver,
+                kDistributionPreferencesCompleteTopic
+              );
+              try {
+                const consoleURI =
+                  Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
+                resolve(new URL(consoleURI));
+              } catch (ex) {
+                lazy.log.error(`Critial misconfiguration: Missing console URI`);
```
1rneh: Could you fix the typo in "Criti**c**al"

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh, line 204: This currently returns a promise. It can remain a string if adapt it to my suggestion below.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -168,10 +195,11 @@ export const ConsoleClient = {
...
     // not happen.
-    const url = this.consoleBaseURI;
-    url.pathname = this._paths.SSO_CALLBACK;
-    url.port = "";
-    return url.href + "?*";
+    return this.consoleBaseURI.then(url => {
+      url.pathname = this._paths.SSO_CALLBACK;
+      url.port = "";
+      return url.href + "?*";
+    });
```
1rneh, line 219: Can we stay consistent in this file, unless there is a reason to not use `await`.
```suggestion
    const url = await this.consoleBaseURI;
    url.pathname = this._paths.SSO_CALLBACK;
    url.port = "";
    return url.href + "?*";
```

browser/components/enterprise/EnterpriseEndpoints.sys.mjs
```diff
@@ -56,19 +56,19 @@ export const EnterpriseEndpoints = {
...
-
-    for (const pref of BASE_CONSOLE_URI_PREFS) {
-      defaultBranch.setStringPref(pref, consoleBaseURI.href);
-      Services.prefs.lockPref(pref);
-    }
+      for (const pref of BASE_CONSOLE_URI_PREFS) {
+        defaultBranch.setStringPref(pref, consoleBaseURI.href);
+        Services.prefs.lockPref(pref);
+      }
+    });
```
1rneh: Can we also add a `catch` block? Otherwise this will fail silently.

browser/components/enterprise/EnterpriseEndpoints.sys.mjs
```diff
@@ -56,19 +56,19 @@ export const EnterpriseEndpoints = {
   init() {
```
1rneh: Just double-checked, the caller of `EnterpriseEndpoints.init()` can await the promise: 
https://searchfox.org/enterprise-main/rev/8411df1e6f57ed9261bae54ebcfc96cf8a3986fb/toolkit/modules/BrowserUtils.sys.mjs#684

1rneh (changes requested):

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: When I was doing this in my first version, the returned url was somehow shared and tests were failing because some code was changing the value while another expected a new one, I need to check again

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Is `object` enough?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: sure, done :)

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Ah, it's because it is a getter, we cannot mark it `async` and having `await` in the body trips a syntax error: `(new SyntaxError("await is a reserved identifier", "resource:///modules/enterprise/ConsoleClient.sys.mjs", 198))`

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: Yes this is because we were previously returning a new `URL` object each time when now we share the same object within the promise

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Why not keep `URL`? In just noticed that the docs are missing also `@throws`. Could you help add it?

lissyx: `Promise<{URL: object}>` ?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: OK we should be fine  now

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: But it's the built-in `URL` class, not just any object.

lissyx: Sorry

lissyx: This makes me think we should https://firefox-source-docs.mozilla.org/tools/moztreedocs/jsdoc-support.html ?

browser/components/enterprise/EnterpriseEndpoints.sys.mjs
lissyx: I believe this one is outdated now, since we moved to `async/await`

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Not sure if our code should really be documented at this point 😅

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -94,17 +95,49 @@ export const ConsoleClient = {
   /**
    * Base URL of the remote enterprise console
    *
-   * @returns {URL}
+   * @throws {Error}
+   * @returns {Promise<{URL}>}
```
```suggestion
   * @returns {Promise<URL>}
```

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -94,17 +95,49 @@ export const ConsoleClient = {
...
+                  );
+                  reject(ex);
+                }
+              }
+            },
+          };
+          Services.obs.addObserver(
+            distributionCompleteObserver,
+            kDistributionPreferencesCompleteTopic
+          );
```
1rneh, line 140: Let's make sure we remove the observer on shutdown.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
1rneh: Since we're sticking to returning the derived `.then()` promise, the JSDoc still needs adjustment.

1rneh (approved):
r+ after comments are addressed.

PR #742, opened 2026-04-16
Bug 2030392: Trigger a fetch policies request with invalid tokens expecting a forced logout on test

victhorlopez:
### Description 

Bugzilla: Bug-2030392

The recurring remote fetch of policies that the felt browser has is currently affecting the test, as it can also get a 401 and close firefox when the tokens are invalidated. The changes remove the extra request that was used and uses directly the fetch remote policies request to trigger the browser forced signout

* Depends on:

---

### Screenshots 

---

### Testing 

* [x] Added tests

### DIFF
--- a/testing/enterprise/test_felt_browser_token_refresh.py
+++ b/testing/enterprise/test_felt_browser_token_refresh.py
@@ -14,6 +14,9 @@ from marionette_driver.by import By
 
 
 class BrowserTokenRefresh(FeltTests):
+    # make explicit as it is used to trigger 401 to close the browser
+    EXTRA_CHILD_PREFS = {"browser.policies.live_polling.frequency": 500}
+
     def setup(self):
         super().setup()
         self.felt_logout_checker = FeltLogoutChecker(self)
@@ -34,24 +37,17 @@ class BrowserTokenRefresh(FeltTests):
     def assert_browser_closes_on_401(self):
         old_access_token = self.policy_access_token.value
         old_refresh_token = self.policy_refresh_token.value
-        self.policy_access_token.value = ""
-        self.policy_refresh_token.value = ""
-
-        # Trigger an auth request with invalid tokens, expecting a forced logout.
+        self._manually_closed_child = True
+        # waits until firefox exits and gives the proper logout type
         with self.felt_logout_checker.assert_browser_logouts_with(
             "console-forced-logout"
         ):
-            self._child_driver.set_context("chrome")
-            # Run synchronously without awaiting the promise — the only purpose is
-            # to trigger a 401/403 response which causes the browser to shut down.
-            self._child_driver.execute_script("""
-                const { ConsoleClient } = ChromeUtils.importESModule(
-                    "resource:///modules/enterprise/ConsoleClient.sys.mjs"
-                );
-                ConsoleClient.getLoggedInUserInfo();
-            """)
+            # The context manager is listening for logout events. Once we invalidate
+            # the tokens, the next remote policies fetch
+            # will fail and trigger a "console-forced-logout".
+            self.policy_refresh_token.value = ""
+            self.policy_access_token.value = ""
 
-        self._manually_closed_child = True
         self.assert_child_browser_closed()
 
         self.await_felt_auth_window()


testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -41,15 +42,15 @@ def assert_browser_closes_on_401(self):
...
+                self.get_logged_in_user_info(env=Environment.FIREFOX)
+            except (
+                errors.InvalidSessionIdException,
+                errors.NoSuchWindowException,
+                errors.TimeoutException,
+                OSError,
+            ):
+                pass
 
         self._manually_closed_child = True
```
lissyx: shouldn't we move this one earlier? I guess we would still need to catch the exception, but that would be more correct

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -41,15 +42,15 @@ def assert_browser_closes_on_401(self):
...
-                );
-                ConsoleClient.getLoggedInUserInfo();
-            """)
+            try:
+                self.get_logged_in_user_info(env=Environment.FIREFOX)
+            except (
+                errors.InvalidSessionIdException,
+                errors.NoSuchWindowException,
+                errors.TimeoutException,
+                OSError,
```
lissyx: have you observed all those being required? I'm worried catching some wide ones may still hide other issues

victhorlopez: I would say OSError and NoSuchWindow are the ones possible. in here, I added thos because of this comment in here https://github.com/mozilla/enterprise-firefox/pull/718#discussion_r3074069857

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: where do you want to put/ firefox will close either way? try/catch/finally, in the finally?

testing/enterprise/test_felt_browser_token_refresh.py
lissyx: I'd say as soon as we know things will not need to be closed, so it could be just before the `with`, because that is when you asks for logout so at that moment you know it will be the case

victhorlopez: @1rneh  found the root cause, the remote fetch policies is also causing 401 and closing firefox

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -34,24 +37,14 @@ def _setup_beforeunload_tab(self):
     def assert_browser_closes_on_401(self):
         old_access_token = self.policy_access_token.value
         old_refresh_token = self.policy_refresh_token.value
-        self.policy_access_token.value = ""
-        self.policy_refresh_token.value = ""
-
-        # Trigger an auth request with invalid tokens, expecting a forced logout.
+        self._manually_closed_child = True
+        # Trigger a fetch policies request with invalid tokens, expecting a forced logout.
```
lissyx: This comment is more descriptive that then commit title :)

victhorlopez: changed it, can change also the commit

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -34,24 +37,14 @@ def _setup_beforeunload_tab(self):
...
-            # Run synchronously without awaiting the promise — the only purpose is
-            # to trigger a 401/403 response which causes the browser to shut down.
-            self._child_driver.execute_script("""
-                const { ConsoleClient } = ChromeUtils.importESModule(
-                    "resource:///modules/enterprise/ConsoleClient.sys.mjs"
-                );
-                ConsoleClient.getLoggedInUserInfo();
-            """)
+            self.policy_access_token.value = ""
+            self.policy_refresh_token.value = ""
```
lissyx: why do we do that **after** asserting that we logged out? Shouldn't we do it before so that the policy fetch fails and forces the logout?

testing/enterprise/test_felt_browser_token_refresh.py
lissyx: yes it would be cool

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: I am not 100% following, the asserting is done on the context manager (after we left the scope).

One concern though would be that the order of the tokens is incorrect, and first we should invalidate refresh token

lissyx: Ah, I forgot how it works. Then if it's on exit it should be fine, but we should be explicit here to make sure people are not tricked like I was

victhorlopez: Just to clarify, the context manager has a wait

https://searchfox.org/enterprise-main/rev/8e6c06bf3c573ee2682ef7233a8a04bfe32a7c07/testing/enterprise/felt_tests.py#590

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -14,6 +14,9 @@
 
 
 class BrowserTokenRefresh(FeltTests):
+    # make explicit as it is used to trigger 401 to close the browser
+    EXTRA_CHILD_PREF = {"browser.policies.live_polling.frequency": 500}
```
1rneh: Correct me if I am wrong, but this is not changing the value, as the console already returns 500ms as the interval.

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -34,24 +37,15 @@ def _setup_beforeunload_tab(self):
...
-            # to trigger a 401/403 response which causes the browser to shut down.
-            self._child_driver.execute_script("""
-                const { ConsoleClient } = ChromeUtils.importESModule(
-                    "resource:///modules/enterprise/ConsoleClient.sys.mjs"
-                );
-                ConsoleClient.getLoggedInUserInfo();
-            """)
+            # Trigger a fetch policies request with invalid tokens, expecting a forced logout.
+            self.policy_refresh_token.value = ""
+            self.policy_access_token.value = ""
```
1rneh: Can we enhance the comment to include that the logout checker is waiting for the next remote policies fetch to trigger the logout with the message "console-forced-logout", which should happen within the next 500ms, as this is the expected polling interval?

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: you are correct, this is to ensure if at some point we change that we do not break this code, as we have a direct dependency

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: let me do that, I also updated the PR body description

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -34,24 +37,16 @@ def _setup_beforeunload_tab(self):
...
-            # Run synchronously without awaiting the promise — the only purpose is
-            # to trigger a 401/403 response which causes the browser to shut down.
-            self._child_driver.execute_script("""
-                const { ConsoleClient } = ChromeUtils.importESModule(
-                    "resource:///modules/enterprise/ConsoleClient.sys.mjs"
-                );
-                ConsoleClient.getLoggedInUserInfo();
-            """)
+            # Invalidate the tokens so the next remote policies fetch (within ~500ms, the expected
+            # polling interval) triggers a "console-forced-logout" via the logout checker.
```
lissyx: Reading only the code here, it's non trivial to know that this will be properly respected. Are we checking only the very next remote policies fetch, or just expect that given we have the 500ms polling we will hit the logout quickly enough?

I think this comment should clearly states the behavior here, otherwise either we are introducing a source of future race condition, OR this may be suspected in case of race condition and make us loose debugging time.

testing/enterprise/test_felt_browser_token_refresh.py
lissyx: If we force it, maybe we should even lower it for the purpose of this test?

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: The policies fetch, and the test are in desync, meaning it could be in 1ms or 499ms after you invalidate the tokens.

Does something like this make more sense? (100ms will be new time?)

```
            # The context manager is listening for logout events. Once we invalidate
            # the tokens, the next remote policies fetch (within ~100ms, the polling interval)
            # will fail and trigger a "console-forced-logout".
```

lissyx: > The policies fetch, and the test are in desync, meaning it could be in 1ms or 499ms after you invalidate the tokens.
> 
> Does something like this make more sense? (100ms will be new time?)
> 
> ```
>             # The context manager is listening for logout events. Once we invalidate
>             # the tokens, the next remote policies fetch (within ~100ms, the polling interval)
>             # will fail and trigger a "console-forced-logout".
> ```

I think we should not mention any time at all and maybe should word "the next remote policies fetch is expected to fail and trigger a console-forced-logout"

testing/enterprise/test_felt_browser_token_refresh.py
```diff
@@ -34,24 +37,17 @@ def _setup_beforeunload_tab(self):
     def assert_browser_closes_on_401(self):
         old_access_token = self.policy_access_token.value
         old_refresh_token = self.policy_refresh_token.value
-        self.policy_access_token.value = ""
-        self.policy_refresh_token.value = ""
-
-        # Trigger an auth request with invalid tokens, expecting a forced logout.
+        self._manually_closed_child = True
+        # waits until firefox exists and gives the proper logout type
```
lissyx:
```suggestion
        # waits until firefox exits and gives the proper logout type
```

lissyx (approved):

testing/enterprise/test_felt_browser_token_refresh.py
1rneh:
```suggestion
    EXTRA_CHILD_PREFS = {"browser.policies.live_polling.frequency": 500}
```

1rneh (approved):
r+ if variable name is fixed. Thank you!

1rneh:
I filed [Bug 2033492](https://bugzilla.mozilla.org/show_bug.cgi?id=2033492) to revisit any other potential tests that are effected by potential races due to uncontrolled remote policy polls.

testing/enterprise/test_felt_browser_token_refresh.py
victhorlopez: :( , thanks for catching

PR #745, opened 2026-04-17
Bug 2019348 - Enterprise: fix test_no_window_update_restart.py

lissyx:
### Description 

Bugzilla: Bug-2019348

This re-enables the test, which was likely fixed in fact by Bug-2024587: there in `AppShutdown::MaybeDoRestart` profile path was being saved to an environment variable, but we were not handling that right.

---

### Screenshots 

---

### Testing 

* [x] Added tests

### DIFF
--- a/toolkit/mozapps/update/tests/marionette/manifest.toml
+++ b/toolkit/mozapps/update/tests/marionette/manifest.toml
@@ -4,6 +4,6 @@ tags = "os_integration"
 
 ["test_no_window_update_restart.py"]
 run-if = [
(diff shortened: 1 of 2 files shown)


toolkit/mozapps/update/tests/marionette/manifest.toml
```diff
@@ -4,6 +4,6 @@ tags = "os_integration"
 
 ["test_no_window_update_restart.py"]
 run-if = [
-  "os == 'mac' && !enterprise", # Bug 2019348 - Test is failing and there is a "profile missing" window visible
```
lissyx, line 7: And this missing profile window may likely have been fixed by bug 2024587

PR #747, opened 2026-04-20
Bug 2033414 - Enterprise: add update channel in profile name

lissyx:
Add the update channel to the name of the profile being used both for Felt and for browser, to avoid running into "profile downgrade" errors.

### Description 

Bugzilla: Bug-2033414

---

### Testing 

* [ ] Added tests

**Steps to verify changes:**
1. Download/build
2. Run felt
3. Connect and launch browser

**Expected result:**
1. (before step 3) Felt profile path (in your temporary directory) should be created including the update channel, e.g. `felt-nightly-enterprise`)
2. (after step 3) Browser to `about:support` and verify that the profile path includes the same update channel key

### DIFF
--- a/browser/extensions/felt/content/FeltCommon.sys.mjs
+++ b/browser/extensions/felt/content/FeltCommon.sys.mjs
@@ -2,8 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
 export const FeltCommon = {
   PRIVATE_BROWSING_ID: 1,
-  ENTERPRISE_PROFILE: "enterprise-profile",
+  ENTERPRISE_PROFILE: `enterprise-profile-${AppConstants.MOZ_UPDATE_CHANNEL}`,
   POLICY_POLLING_FREQUENCY: 60_000,
 };
--- a/toolkit/xre/nsAppRunner.cpp
+++ b/toolkit/xre/nsAppRunner.cpp
@@ -3150,11 +3150,11 @@ static nsresult SelectProfile(nsToolkitProfileService* aProfileSvc,
     const char* xreProfileLocalPath = PR_GetEnv("XRE_PROFILE_LOCAL_PATH");
     auto savedLocalProfile = xreProfileLocalPath && *xreProfileLocalPath;
 
-    // When running as FELT UI use a hardcoded profile named "felt" and living
-    // in the OS' temporary directory. There are a few special cases where it is
-    // required to use a profile that is being given from different means:
-    //  - CLI argument "-profile ..."
-    //  - environment variables "XRE_PROFILE_PATH" / "XRE_PROFILE_LOCAL_PATH"
+    // When running as FELT UI use a hardcoded profile named "felt-XXX" with
+    // XXX being the update channel and living in the OS' temporary directory.
+    // There are a few special cases where it is required to use a profile that
+    // is being given from different means: - CLI argument "-profile ..." -
+    // environment variables "XRE_PROFILE_PATH" / "XRE_PROFILE_LOCAL_PATH"
     //
     // If either of those are present, the following code is skipped and profile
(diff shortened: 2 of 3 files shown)


browser/extensions/felt/content/FeltCommon.sys.mjs
```diff
@@ -2,8 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
 export const FeltCommon = {
   PRIVATE_BROWSING_ID: 1,
-  ENTERPRISE_PROFILE: "enterprise-profile",
+  ENTERPRISE_PROFILE: `enterprise-profile-${AppConstants.MOZ_UPDATE_CHANNEL}`,
```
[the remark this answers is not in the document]
lissyx: no we dont care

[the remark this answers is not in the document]
gcp: It's dead code after your change, so you should at least remove it?

toolkit/xre/nsAppRunner.cpp
```diff
@@ -3164,7 +3164,13 @@ static nsresult SelectProfile(nsToolkitProfileService* aProfileSvc,
       nsCOMPtr<nsIFile> file;
       MOZ_TRY(GetSpecialSystemDirectory(OS_TemporaryDirectory,
                                         getter_AddRefs(file)));
-      MOZ_TRY(file->AppendNative("felt"_ns));
+      nsCString profileName;
+      profileName = "felt"_ns;
+#  if defined(MOZ_UPDATE_CHANNEL)
```
gcp: That define seems to be always set.

gcp: It's guaranteed to be, see: https://searchfox.org/firefox-main/rev/e374045fac6f19383a04d087d2930ced5c1551fe/build/moz.configure/init.configure#1192

gcp (changes requested):

toolkit/xre/nsAppRunner.cpp
lissyx: OK, there were many places checking for it anyway, so I assumed there were cases where it may not be defined

gcp: Wrong thread. Also: 

>OK, there were many places checking for it anyway, so I assumed there were cases where it may not be defined

There are many places checking the *contents*, but there are none that check if it exists (even before checking against the contents).

browser/extensions/felt/content/FeltCommon.sys.mjs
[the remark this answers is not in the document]
lissyx: In fact maybe we care a bit :)

gcp (changes requested):

toolkit/xre/nsAppRunner.cpp
gcp: nit: not really true now

toolkit/xre/nsAppRunner.cpp
```diff
@@ -3164,7 +3164,9 @@ static nsresult SelectProfile(nsToolkitProfileService* aProfileSvc,
       nsCOMPtr<nsIFile> file;
       MOZ_TRY(GetSpecialSystemDirectory(OS_TemporaryDirectory,
                                         getter_AddRefs(file)));
-      MOZ_TRY(file->AppendNative("felt"_ns));
+      nsCString profileName =
+          nsPrintfCString("felt-%s", MOZ_STRINGIFY(MOZ_UPDATE_CHANNEL));
```
gcp: You can avoid the printf as you have 2 string literals of the right format: `MOZ_TRY(file->AppendNative("felt-" MOZ_STRINGIFY(MOZ_UPDATE_CHANNEL) ""_ns));`

toolkit/xre/nsAppRunner.cpp
lissyx: fixed

gcp (approved):

PR #748, opened 2026-04-20
Bug 2032092 - Enterprise: persist pending URLs to disk

lissyx:
If Felt is being restarted without authentication being completed (e.g. updates being applied), then the pending URLs that may have opened it are going to be lost on next restart. This change ensures those are persisted on disk and reloaded.

### Description 

Bugzilla: Bug-2032092

### Testing 

* [x] Added tests

### DIFF
--- a/browser/components/FeltURLHandler.sys.mjs
+++ b/browser/components/FeltURLHandler.sys.mjs
@@ -14,6 +14,10 @@ ChromeUtils.defineLazyGetter(
     )
 );
 
+ChromeUtils.defineESModuleGetters(lazy, {
+  JSONFile: "resource://gre/modules/JSONFile.sys.mjs",
+});
+
 export const FELT_OPEN_WINDOW_DISPOSITION = {
   DEFAULT: 0,
   NEW_WINDOW: 1,
@@ -23,7 +27,78 @@ export const FELT_OPEN_WINDOW_DISPOSITION = {
 // Queue for Felt external link handling
 // URL requests are stored here when they arrive via command line (before Felt extension loads)
 // FeltProcessParent imports this module and manages forwarding from this queue
-export let gFeltPendingURLs = [];
+export const gFeltPendingURLs = {
+  _pendingURLs: [],
+  _pendingFilePath: PathUtils.join(
+    Services.dirsvc.get("ProfD", Ci.nsIFile).path,
+    "pendingURLs.json"
+  ),
+  _ready: false,
+  _initPromise: null,
+
+  async init() {
+    if (this._ready) {
+      return;
+    }
+
+    if (this._initPromise) {
+      await this._initPromise;
+      return;
+    }
+
+    this._initPromise = (async () => {
+      if (this._ready) {
+        return;
+      }
+
+      const storage = new lazy.JSONFile({
+        path: this._pendingFilePath,
+      });
+
+      await storage.load();
+
+      if (storage.data?.pendingURLs) {
+        this._pendingURLs = storage.data.pendingURLs;
+      }
+
+      this._storage = storage;
+      this._ready = true;
+    })();
+    try {
+      await this._initPromise;
+    } finally {
+      this._initPromise = null;
+    }
+  },
+
+  [Symbol.iterator]() {
+    return this._pendingURLs[Symbol.iterator]();
+  },
+
+  async push(payload) {
+    await this.init();
+    const rv = this._pendingURLs.push(payload);
+    this._storage.data.pendingURLs = this._pendingURLs;
+    this._storage.saveSoon();
+    return rv;
+  },
+
+  get length() {
+    return this._pendingURLs.length;
+  },
+
+  clear() {
+    this._pendingURLs = [];
+    this._storage.data.pendingURLs = [];
+    this._storage.saveSoon();
+  },
+};
+
+if (Services.felt?.isFeltUI()) {
+  gFeltPendingURLs.init().catch(error => {
+    console.error(`Failed to initialize Felt pending URL storage: ${error}`);
+  });
+}
 
 let lastNotificationShown = 0;
 let gFeltFirefoxReadyNotified = false;
@@ -82,7 +157,9 @@ export function queueFeltURL(payload) {
       `Retrying to queue url ${payload.url} after initial failure:`,
       e
     );
-    gFeltPendingURLs.push(payload);
+    gFeltPendingURLs.push(payload).catch(err => {
+      console.error("Failed to persist pending Felt URL", err);
+    });
     Services.cpmm.sendAsyncMessage("FeltParent:ForceFeltFocus", {});
   }
 
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -43,7 +43,9 @@ export function queueURL(payload) {
     Services.felt.makeBackgroundProcess(true);
   } else {
     // Queue at module level until ready
-    gFeltPendingURLs.push(payload);
+    gFeltPendingURLs.push(payload).catch(err => {
+      lazy.log.error("Failed to persist pending Felt URL", err);
+    });
     Services.cpmm.sendAsyncMessage("FeltParent:ForceFeltFocus", {});
   }
 }
@@ -238,7 +240,9 @@ export class FeltProcessParent extends JSProcessActorParent {
           case "felt-extension-ready": {
             if (gFeltProcessParentInstance) {
               gFeltProcessParentInstance.extensionReady = true;
-              gFeltProcessParentInstance.forwardPendingURLs();
+              gFeltProcessParentInstance.forwardPendingURLs().catch(err => {
+                lazy.log.error("Failed to forward pending URLs", err);
+              });
               notifyFirefoxReady();
             }
             break;
@@ -478,7 +482,7 @@ export class FeltProcessParent extends JSProcessActorParent {
         this.firefoxReady = true;
 
         // Try to forward pending URLs now (will only forward if extension is also ready)
-        this.forwardPendingURLs();
+        await this.forwardPendingURLs();
         notifyFirefoxReady();
       })
       .then(() => {
@@ -686,7 +690,9 @@ export class FeltProcessParent extends JSProcessActorParent {
   /**
    * Forward all pending URLs to Firefox
    */
-  forwardPendingURLs() {
+  async forwardPendingURLs() {
+    await gFeltPendingURLs.init();
+
     if (gFeltPendingURLs.length === 0) {
       return;
     }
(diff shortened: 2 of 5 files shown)


browser/components/FeltURLHandler.sys.mjs
```diff
@@ -23,7 +27,58 @@ export const FELT_OPEN_WINDOW_DISPOSITION = {
 // Queue for Felt external link handling
 // URL requests are stored here when they arrive via command line (before Felt extension loads)
 // FeltProcessParent imports this module and manages forwarding from this queue
-export let gFeltPendingURLs = [];
+export const gFeltPendingURLs = {
+  _pendingURLs: [],
+  _pendingFilePath: PathUtils.join(
+    Services.dirsvc.get("ProfD", Ci.nsIFile).path,
```
gcp, line 33: Just a santity check that this isn't writable in the sandbox, as we do allow some parts of the profile? At least read? Do we allow write anywhere?

browser/components/FeltURLHandler.sys.mjs
```diff
@@ -23,7 +27,58 @@ export const FELT_OPEN_WINDOW_DISPOSITION = {
...
+    }
+
+    this._ready = true;
+  },
+
+  [Symbol.iterator]() {
+    return this._pendingURLs[Symbol.iterator]();
+  },
+
+  push(payload) {
```
gcp: This needs to await `init()`.

browser/components/FeltURLHandler.sys.mjs
```diff
@@ -23,7 +27,58 @@ export const FELT_OPEN_WINDOW_DISPOSITION = {
...
+    this._storage.data.pendingURLs = this._pendingURLs;
+    this._storage.saveSoon();
+    return rv;
+  },
+
+  get length() {
+    return this._pendingURLs.length;
+  },
+
+  set length(v) {
```
gcp: This is a weird function. Why not just make a `clear` what's the use case for a non-zero set?

gcp (changes requested):

browser/components/FeltURLHandler.sys.mjs
lissyx: I am not sure I get your point here

browser/components/FeltURLHandler.sys.mjs
lissyx: I was just following how it is used, there's a clear-like call made by setting `length = 0` https://searchfox.org/enterprise-main/source/browser/extensions/felt/content/FeltProcessParent.sys.mjs#673

But if we are fine changing those call, all good

browser/components/FeltURLHandler.sys.mjs
lissyx: Yes, but then it made `push` being `async` and I was not sure it was fine, trying to proxy `Array` this way

browser/components/FeltURLHandler.sys.mjs
lissyx: sandbox as in content process sandbox?

gcp: The idea is that this shouldn't allow a compromised content process to read the URLs that are being loaded in other processes (breaking site isolation).

lissyx: I've added sandbox tests to cover

lissyx:
try: https://firefox-ci-tc.services.mozilla.com/tasks/groups/PDH2HVNiSvKVHTKsCnUK_A

lissyx:
> try: https://firefox-ci-tc.services.mozilla.com/tasks/groups/PDH2HVNiSvKVHTKsCnUK_A

It's properly passing (blocked) on linux and windows, but there's a (unrelated?) failure on macOS and I dont have confirmation we properly block there.

testing/enterprise/test_felt_browser_external_link.py
```diff
@@ -62,3 +67,95 @@ def run_felt_open_external_link(self):
...
+                        },
+                    ]
+                })
+            )
+        self._logger.info(f"New pending URLs file: {urls_file}")
+
+        self._driver.start_session(timeout=60)
+        new_urls_file = os.path.join(self._driver.profile, "pendingURLs.json")
+        assert os.path.exists(new_urls_file), (
+            f"Pending URLs file should exists: {new_urls_file}"
```
gcp, line 153: nit: typo

security/sandbox/test/browser_content_sandbox_fs_tests.js
```diff
@@ -117,6 +117,30 @@ async function testFileAccessAllPlatforms() {
...
+    });
+
+    const feltPendingURLs = GetProfileEntry("pendingURLs.json");
+    tests.push({
+      desc: "felt pending urls", // description
+      ok: false, // expected to succeed?
+      browser: webBrowser, // browser to run test in
+      file: feltPendingURLs, // nsIFile object
+      minLevel: minProfileReadSandboxLevel(), // min level to enable test
+      func: readFile,
```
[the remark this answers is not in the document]
gcp: Can be fixed by copying the guard idea from https://github.com/mozilla/enterprise-firefox/pull/748/changes#diff-8949d29421035bb8294887c13a8464840c2956589e871fbbcd164ee42a29bdf7R193

browser/components/FeltURLHandler.sys.mjs
```diff
@@ -23,7 +27,57 @@ export const FELT_OPEN_WINDOW_DISPOSITION = {
...
+      return;
+    }
+
+    this._storage = new lazy.JSONFile({
+      path: this._pendingFilePath,
+    });
+    await this._storage.load();
+
+    if (this._storage.data?.pendingURLs) {
+      this._pendingURLs = this._storage.data?.pendingURLs;
```
gcp: nit: You `data?` in the `if` above so this one isn't needed.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
gcp: This probably needs to be `async` and then have an `await gFeltPendingURLs.init()` before the check below. That would address the race Copilot is complaining about above.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -725,7 +725,7 @@ export class FeltProcessParent extends JSProcessActorParent {
     }
 
     // Clear the queue
-    gFeltPendingURLs.length = 0;
+    gFeltPendingURLs.clear();
```
gcp, line 726: This would also throw without the `await` above because the `._storage` fields it accesses need to be initialized.

testing/enterprise/test_felt_browser_external_link.py
gcp: Ok I missed the one above :P

gcp (changes requested):
Some of the race concerns seem valid to me especially as disk IO is involved?

testing/enterprise/test_felt_browser_external_link.py
```diff
@@ -62,3 +67,95 @@ def run_felt_open_external_link(self):
...
+        self.check_has_external_link_tab()
+
+        self._logger.info(f"Waiting for pending URLs file to be cleared: {urls_file}")
+
+        def has_no_pending_url(file):
+            with open(urls_file) as pending:
+                parsed = json.loads(pending.read())
+                return len(parsed["pendingURLs"]) == 0
+
+        self._wait.until(lambda mn: has_no_pending_url(urls_file))
```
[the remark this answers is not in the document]
lissyx: That's wrong, the `file` parameter should be passed, using `urls_file` was a mistake

browser/components/FeltURLHandler.sys.mjs
```diff
@@ -23,7 +27,57 @@ export const FELT_OPEN_WINDOW_DISPOSITION = {
...
+
+  clear() {
+    this._pendingURLs = [];
+    this._storage.data.pendingURLs = [];
+    this._storage.saveSoon();
+  },
+};
+
+if (Services.felt?.isFeltUI()) {
+  gFeltPendingURLs.init();
```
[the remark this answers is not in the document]
lissyx: This is top level `await` which is not going to work.

security/sandbox/test/browser_content_sandbox_fs_tests.js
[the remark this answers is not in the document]
lissyx: I am not sure what guard idea you are talking about here?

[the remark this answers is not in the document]
lissyx: I've just forced writing the files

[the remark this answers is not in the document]
gcp: >I am not sure what guard idea you are talking about here?

I guess the link doesn't work, but it's around line 193 in the same file.

browser/components/FeltURLHandler.sys.mjs
```diff
@@ -23,7 +27,59 @@ export const FELT_OPEN_WINDOW_DISPOSITION = {
...
+    this._storage = new lazy.JSONFile({
+      path: this._pendingFilePath,
+    });
+    await this._storage.load();
+
+    if (this._storage.data?.pendingURLs) {
+      this._pendingURLs = this._storage.data.pendingURLs;
+    }
+
+    this._ready = true;
```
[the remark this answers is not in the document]
lissyx: It may be valid.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -43,7 +43,9 @@ export function queueURL(payload) {
     Services.felt.makeBackgroundProcess(true);
   } else {
     // Queue at module level until ready
-    gFeltPendingURLs.push(payload);
+    (async () => {
+      await gFeltPendingURLs.push(payload);
+    })();
```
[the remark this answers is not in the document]
lissyx: It's better indeed

browser/components/FeltURLHandler.sys.mjs
```diff
@@ -23,7 +27,59 @@ export const FELT_OPEN_WINDOW_DISPOSITION = {
...
+    this._pendingURLs = [];
+    this._storage.data.pendingURLs = [];
+    this._storage.saveSoon();
+  },
+};
+
+if (Services.felt?.isFeltUI()) {
+  (async () => {
+    await gFeltPendingURLs.init();
+  })();
```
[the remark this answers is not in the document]
lissyx: valid.

gcp:
Note there's two more instances of 
```    
(async () => {
      await gFeltPendingURLs.push(payload);
    })();
``` 
in your code that can be fixed the same way, but the bot didn't display them inline due to low confidence.

lissyx:
> Note there's two more instances of
> 
> ```
> (async () => {
>       await gFeltPendingURLs.push(payload);
>     })();
> ```
> 
> in your code that can be fixed the same way, but the bot didn't display them inline due to low confidence.

Thanks, fixed

gcp (approved):

PR #750, opened 2026-04-21
Bug 2032400: Reduce number of restarted processes

victhorlopez:
### Description 

Bugzilla: Bug-2032400

The goal of the PR is to improve the start/stop sequence firefox is having on marionette, current sequence is as follows

**General marionette starts =>**

- firefox starts on harness ([link](https://searchfox.org/enterprise-main/rev/8411df1e6f57ed9261bae54ebcfc96cf8a3986fb/testing/marionette/client/marionette_driver/marionette.py#636))
  - test 1 starts
      - we close firefox                                                                                                                                                                                
      - we start firefox
      - we enforce prefs (can also restart firefox)                                                                                                                                                     
  - test 1 finished                                                                                                                                                                                   
      - we close firefox                                                                                                                                                                                
  - test 2 starts 
      - firefox is started by harness
      - we close firefox
      - we start firefox
      - we enforce prefs (can also restart firefox)                                                                                                                                                     
  - test 2 finished
      - we close firefox

After this PR:

**General marionette starts =>**
  - firefox starts on harness ([link](https://searchfox.org/enterprise-main/rev/8411df1e6f57ed9261bae54ebcfc96cf8a3986fb/testing/marionette/client/marionette_driver/marionette.py#636))
  - test 1 starts                                                                                                                                                                                       
      - we close firefox                                                                                                                                                     
      - we enforce prefs (can also restart firefox, but should never really happen)                                                                                                                                              
  - test 1 finished                                                                                                                                                                                     
      - we close firefox                                                                                                                                                                                
  - test 2 starts                                                                                                                                                                                       
      - we set prefs before harness starts firefox                                                                                                                                                      
      - firefox is started by harness                                                                                                                                                                   
      - we enforce prefs as security (should not restart)
  - test 2 finished                                                                                                                                                                                     
      - we close firefox

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [x] Added tests

### DIFF
--- a/testing/enterprise/base_test.py
+++ b/testing/enterprise/base_test.py
@@ -18,6 +18,27 @@ from marionette_driver.wait import Wait
 from marionette_harness import MarionetteTestCase
 
 
+def patch_process_runner_args(instance):
+    # It looks like on windows we are hitting bug 1906191, even after
+    # all processes exit, the pipe write end may not close, causing mozprocess's
+    # reader thread to block indefinitely on readline(). ignore_children makes
+    # mozprocess skip joining the reader thread, allowing shutdown to proceed.
+    original = instance._get_runner_args
+
+    def _patched():
+        args = original()
+        # ignore_children is only supported by mozprocess (not the macOS psutil-based handler)
+        if not sys.platform.startswith("darwin"):
+            args["process_args"]["ignore_children"] = True
+        return args
+
+    def unpatch():
+        instance._get_runner_args = original
+
+    instance._get_runner_args = _patched
+    return unpatch
+
+
 class Environment(Enum):
     FELT = "felt"
     FIREFOX = "Firefox"
@@ -27,29 +48,38 @@ class EnterpriseTestsBase(MarionetteTestCase):
     def setUp(self):
         os.environ.update({"MOZ_DISABLE_NONLOCAL_CONNECTIONS": "0"})
 
-        if hasattr(self, "EXTRA_ENV"):
+        if getattr(self, "EXTRA_ENV", None):
             self._saved_env = deepcopy(os.environ)
             os.environ.update(self.EXTRA_ENV)
 
         self._logger = self.logger
 
-        super().setUp()
+        marionette = self._marionette_weakref()
+
+        if hasattr(self, "_extra_cli_args"):
+            self._saved_cli_args = deepcopy(marionette.instance.app_args)
+            marionette.instance.app_args += self._extra_cli_args
+
+        self._unpatch_process_runner_args = patch_process_runner_args(
+            marionette.instance
+        )
+
+        # On the first test the harness has Firefox already running, so we stop
+        # it here. On subsequent tests tearDown already quit it, so this
+        # will only discard and recreate the profile.
+        marionette.instance.close(clean=True)
+
+        if hasattr(self, "_apply_prefs_for_instance"):
+            self._apply_prefs_for_instance()
 
         # All this needs to happen before process is started to avoid race
         # conditions, but requires self.marionette that is setup by
         # super().setUp() right above.
-        self.overwrite_distribution_ini()
+        self.overwrite_distribution_ini(marionette)
 
-        if hasattr(self, "_extra_cli_args"):
-            self._saved_cli_args = deepcopy(self.marionette.instance.app_args)
-            self.marionette.instance.app_args += self._extra_cli_args
+        super().setUp()
 
-        self._logger.info("Marionette force quit with cleanup.")
-        self.marionette.quit(in_app=False, clean=True)
-        self._logger.info("Marionette start new session, new instance expected.")
-        self.marionette.start_session(timeout=60)
-
-        if hasattr(self, "_extra_prefs"):
+        if getattr(self, "_extra_prefs", None):
             self._logger.info("Marionette enforcing gecko prefs")
             self.marionette.enforce_gecko_prefs(self._extra_prefs)
 
@@ -73,6 +103,8 @@ class EnterpriseTestsBase(MarionetteTestCase):
             self.marionette.instance.app_args = deepcopy(self._saved_cli_args)
             del self._saved_cli_args
 
+        self._unpatch_process_runner_args()
+
         # If there were prefs forced during setUp(), marionette's geckoinstance
         # does cache them and on the next execution of enforce_gecko_pref(), if
         # those prefs are not there anymore in self._extra_prefs, marionette code
@@ -87,12 +119,12 @@ class EnterpriseTestsBase(MarionetteTestCase):
 
         self.restore_distribution_ini()
 
-    def overwrite_distribution_ini(self):
+    def overwrite_distribution_ini(self, marionette):
         # Some test may need a non modified distribution.ini, so respect it and
         # and do not overwrite it when requested.
         # Also some tests may not define a console port, so skip this for them.
         if hasattr(self, "console_port") and not hasattr(self, "KEEP_DISTRIBUTION_INI"):
-            self.distribution_ini_path = self.get_distribution_ini(self.marionette)
+            self.distribution_ini_path = self.get_distribution_ini(marionette)
             self.distribution_ini_orig_path = os.path.join(
                 os.path.dirname(self.distribution_ini_path), "distribution.ini.orig"
             )
--- a/testing/enterprise/felt_tests.py
+++ b/testing/enterprise/felt_tests.py
@@ -630,14 +630,6 @@ class FeltTestsBase(EnterpriseTestsBase):
         self.device_posture_reply_forbidden = Value("B", 0)
         """
 
-        self._extra_prefs = {
-            "enterprise.is_testing": True,
-            "enterprise.log_level": "Debug",
-        }  # + test_prefs
-
-        if hasattr(self, "EXTRA_PREFS"):
-            self._extra_prefs.update(self.EXTRA_PREFS)
-
         self.policy_access_token = SharedString("")
         self.policy_refresh_token = SharedString("")
         self.signout_count = Value("i", 0)
@@ -684,6 +676,18 @@ class FeltTestsBase(EnterpriseTestsBase):
         self._logger.info(f"Starting console server: {self.console_port}")
         self._logger.info(f"Starting SSO server: {self.sso_port}")
 
+    def _apply_prefs_for_instance(self):
+        self._extra_prefs = {
+            "enterprise.is_testing": True,
+            "enterprise.log_level": "Debug",
+        }  # + test_prefs
+
+        if hasattr(self, "EXTRA_PREFS"):
+            self._extra_prefs.update(self.EXTRA_PREFS)
+
+        marionette = self._marionette_weakref()
+        marionette.instance.profile.set_preferences(self._extra_prefs)
+
     def setup(self):
         console_addr = f"http://localhost:{self.console_port}"
 
@@ -722,9 +726,11 @@ class FeltTestsBase(EnterpriseTestsBase):
         if not self._manually_closed_child:
             self._logger.info("Closing browser")
             self._child_driver.set_context("chrome")
+            pid = self._child_driver.session_capabilities["moz:processID"]
             self._child_driver.execute_script(
                 "Services.startup.quit(Ci.nsIAppStartup.eForceQuit);"
             )
+            self.wait_process_exit(pid)
             self._logger.info("Closed browser")
         else:
             self._logger.info("Browser was already manually closed.")
--- a/testing/enterprise/test_felt_harness_prefs.py
+++ b/testing/enterprise/test_felt_harness_prefs.py
@@ -13,12 +13,30 @@ from test_felt_browser_signout import BaseBrowserSignout
 
 
 class HarnessPrefs(BaseBrowserSignout):
-    EXTRA_PREFS = {
-        "enterprise.felt_tests.harness_sets_pref": True,
-    }
-    EXTRA_CHILD_PREFS = {
-        "enterprise.felt_tests.harness_sets_child_pref": True,
-    }
+    # Marionette runs each test_* method through its own setUp()/tearDown()
+    # cycle, which starts a new session each time. We count session starts to
+    # verify sequential execution (e.g., test_4 expects exactly 4 starts).
+    # Subclasses that need cumulative restart tracking must define
+    # start_session_count = 0 to get their own independent counter.
+    start_session_count = 0
+
+    def setUp(self):
+        marionette = self._marionette_weakref()
+        self._original_start_session = marionette.start_session
+
+        def counting_start_session(*args, **kwargs):
+            type(self).start_session_count += 1
+            return self._original_start_session(*args, **kwargs)
+
+        marionette.start_session = counting_start_session
+
+        super().setUp()
+
+        self._manually_closed_child = True
+
+    def tearDown(self):
+        super().tearDown()
+        self.marionette.start_session = self._original_start_session
 
     def _check_prefs(self, driver, expected_prefs, label):
         failures = []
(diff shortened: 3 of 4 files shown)


testing/enterprise/felt_tests.py
```diff
@@ -684,6 +678,37 @@ def setUp(self):
...
+            "enterprise.log_level": "Debug",
+        }  # + test_prefs
+
+        if hasattr(self, "EXTRA_PREFS"):
+            self._extra_prefs.update(self.EXTRA_PREFS)
+
+        # self.marionette is only set after super().setUp(), so we use the weak
+        # reference to access the instance here before the browser starts.
+        marionette = self._marionette_weakref()
+        # Two complementary paths to ensure enterprise prefs reach Firefox:
```
victhorlopez: Removing any of the 2 will make the restart count check fail https://github.com/mozilla/enterprise-firefox/pull/750/changes#diff-48da781c1298d003bb51a11de1bf794594ac16be0499c38c7bdb1bfa4cabf64bR190

testing/enterprise/base_test.py
```diff
@@ -27,27 +27,32 @@ class EnterpriseTestsBase(MarionetteTestCase):
     def setUp(self):
         os.environ.update({"MOZ_DISABLE_NONLOCAL_CONNECTIONS": "0"})
 
-        if hasattr(self, "EXTRA_ENV"):
+        if getattr(self, "EXTRA_ENV", None):
```
lissyx, line 51: this does not change the behavior, is this a leftover of previous debugging? other `hasattr` are still here so I'm wondering

testing/enterprise/base_test.py
```diff
@@ -27,27 +27,32 @@ class EnterpriseTestsBase(MarionetteTestCase):
...
-        super().setUp()
+        marionette = self._marionette_weakref()
+
+        if hasattr(self, "_extra_cli_args"):
+            self._saved_cli_args = deepcopy(marionette.instance.app_args)
+            marionette.instance.app_args += self._extra_cli_args
+
+        # Env vars and CLI args require a fresh process to take effect.
+        # On the first test the harness has Firefox already running, so we stop
+        # it here. On subsequent tests tearDown already quit it, so this is a no-op.
```
lissyx: Should we really rely on `tearDown` to have cleaned up? I'd prefer being defensive here

testing/enterprise/base_test.py
```diff
@@ -27,27 +27,32 @@ class EnterpriseTestsBase(MarionetteTestCase):
...
+        if (
+            getattr(self, "EXTRA_ENV", None) or hasattr(self, "_extra_cli_args")
+        ) and self._is_first_run(marionette):
```
lissyx: maybe we should close each time it makes sense

```suggestion
        ):
```

testing/enterprise/felt_tests.py
```diff
@@ -684,6 +678,37 @@ def setUp(self):
...
+        # 1. instance.prefs: used on the first test, where base_test.setUp() calls
+        #    close(clean=True) which creates a fresh profile from profile_args (which
+        #    reads instance.prefs). The profile that exists here belongs to the harness
+        #    Firefox and gets discarded by that close().
+        #
+        # 2. set_preferences: used on subsequent tests, where tearDown has already
+        #    called quit(clean=True). That call clears instance.prefs first, then
+        #    quit(clean=True) pre-creates a new profile — without enterprise prefs.
+        #    instance.start() reuses that profile without re-reading instance.prefs,
+        #    so we patch user.js in-place instead.
```
lissyx: This is confusing to me, in both case we have a clean profile without enterprise prefs? So why two paths?

testing/enterprise/base_test.py
```diff
@@ -27,27 +27,32 @@ class EnterpriseTestsBase(MarionetteTestCase):
...
         self._logger = self.logger
 
-        super().setUp()
+        marionette = self._marionette_weakref()
+
+        if hasattr(self, "_extra_cli_args"):
+            self._saved_cli_args = deepcopy(marionette.instance.app_args)
+            marionette.instance.app_args += self._extra_cli_args
+
+        # Env vars and CLI args require a fresh process to take effect.
```
lissyx: I do not see in the PR where you detect changing env/cli args that is mentionned in `we close only if env or cli args are different` of the description

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -55,8 +55,21 @@ def _check_env(self, expected, excluded):
...
+        failures = []
+        for pref_name in excluded_prefs:
+            actual = driver.get_pref(pref_name)
+            if actual is not None:
+                failures.append(f"  {pref_name}: expected absent, got {actual!r}")
+        assert not failures, f"[{label}] Prefs should be absent:\n" + "\n".join(
+            failures
+        )
+
     def run_check_felt_prefs(self):
```
lissyx:
```suggestion
    def run_check_felt_set_prefs_absent_prefs_and_env(self):
```

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -110,19 +123,59 @@ class HarnessPrefsTest(HarnessPrefs):
         "test_2_harness_prefs_no_leakage": {"ENTERPRISE_HARNESS_TEST_2": "second"},
         "test_3_harness_prefs_no_leakage": {"ENTERPRISE_HARNESS_TEST_3": "third"},
     }
```
lissyx, line 226: This may now be a problem: I think we do not handle the case of an env var being there in both tests, but having different values. At least we do not cover checking this in the test ?

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -110,19 +123,59 @@ class HarnessPrefsTest(HarnessPrefs):
...
+    _PREFS_BY_TEST = {
+        "test_1_harness_prefs": {"enterprise.felt_tests.harness_sets_pref_1": True},
+        "test_2_harness_prefs_no_leakage": {
+            "enterprise.felt_tests.harness_sets_pref_2": True
+        },
+        "test_3_harness_prefs_no_leakage": {
+            "enterprise.felt_tests.harness_sets_pref_3": True
+        },
+    }
```
lissyx, line 244: Similar, we should harden to make sure a pref with the same name properly behaves

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -55,8 +55,21 @@ def _check_env(self, expected, excluded):
...
+                failures.append(f"  {pref_name}: expected absent, got {actual!r}")
+        assert not failures, f"[{label}] Prefs should be absent:\n" + "\n".join(
+            failures
+        )
+
     def run_check_felt_prefs(self):
         self._check_prefs(self._driver, self._extra_prefs, "FELT UI")
+        self._check_prefs_absent(
+            self._driver, getattr(self, "EXCLUDED_PREFS", set()), "FELT UI"
+        )
```
lissyx: I think this should properly make sure we have exactly the prefs and values that we want.

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -55,8 +55,21 @@ def _check_env(self, expected, excluded):
                     failures.append(f"  {key}: expected absent in Firefox process")
```
lissyx:
```suggestion
                    failures.append(f"  {key}: expected absent in FELT process")
```

lissyx (commented):
LGTM but I'd like those comments to be addressed, and please provide a try link with many retries :)

testing/enterprise/base_test.py
victhorlopez: I removed the if, cause the first restart is unavoidable, and the next calls will be no-op, unless we had an issue on teardown

testing/enterprise/test_felt_harness_prefs.py
victhorlopez: I added the same check than child browser
```
        felt_prefs = {
            **gecko_prefs,
            **DesktopInstance.desktop_prefs,
            **enterprise_prefs,
            **self._extra_prefs,
        }
```

testing/enterprise/test_felt_harness_prefs.py
victhorlopez: Not yet, adding it (for pref and env)

testing/enterprise/felt_tests.py
victhorlopez: I changed the setup to have it "cleaner". It only calls now
```
        marionette.instance.profile.set_preferences(self._extra_prefs)
```

testing/enterprise/base_test.py
victhorlopez: So I added some more tests for more cases and I found that actually... 
We have this prefs here, that will always force a restart on the very very first firefox
https://searchfox.org/enterprise-main/rev/b38a1b389382b925dd5983afd824ba6f1fca6529/testing/enterprise/felt_tests.py#633

testing/enterprise/base_test.py
victhorlopez: so.. in this case you are right but the reason why I use getattr is becase if you define 
```EXTRA_ENV = {}``` it will go inside the if vs getattr it will not go inside the if.

In this case you are right it is correct, in some other cases (like if you restart or something) it is then not correct

**Comment by victhorlopez** [comment]
> LGTM but I'd like those comments to be addressed, and please provide a try link with many retries :)

Here I ran with rebuild 50

from 300, 7 failed

https://firefox-ci-tc.services.mozilla.com/tasks/groups/HZwEIIf1Sy2MHLWNJwlh0A#marionette

I will monkey patch the ignore-children flag now, also I need to rebase cause the focus line is not there

**Comment by victhorlopez** [comment]
> LGTM but I'd like those comments to be addressed, and please provide a try link with many retries :)

Here one cleaner 

https://firefox-ci-tc.services.mozilla.com/tasks/groups/fqH0XRMsQ8qqGrS8fc2blg#marionette

File: testing/enterprise/base_test.py (hunk 2 above)
Comment (lissyx) [reply testing/enterprise/base_test.py:51]: OK, I was just curious about the reasonning

File: testing/enterprise/test_felt_harness_prefs.py [hunk 12]
```diff
@@ -77,55 +95,176 @@ def run_check_child_prefs(self):
...
+        }
+        current_prefs = self._PREFS_BY_TEST[self._testMethodName]
+        self.EXCLUDED_PREFS = {
+            key
+            for test_name, prefs in self._PREFS_BY_TEST.items()
+            if test_name != self._testMethodName
+            for key in prefs
+            if key not in current_prefs
         }
+
```
lissyx, line 270:
```suggestion
```

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -77,55 +95,176 @@ def run_check_child_prefs(self):
...
+    start_session_count = 0
+
     def setUp(self):
-        # Inject the env var for this test and exclude all env vars from other
-        # tests, so run_check_felt_prefs can assert presence and absence.
-        self.EXTRA_ENV = self._ENV_BY_TEST.get(self._testMethodName, {})
+        current_env = self._ENV_BY_TEST.get(self._testMethodName, {})
+        self.EXTRA_ENV = current_env
         self.EXCLUDED_ENV = {
             key
```
lissyx, line 252: How much do you think it may be valuable to also assert that there is no overlap in those sets' keys?
(same for prefs)

victhorlopez:
> LGTM but I'd like those comments to be addressed, and please provide a try link with many retries :)

Here another one, (only one failure regarding something else)

https://firefox-ci-tc.services.mozilla.com/tasks/groups/MZZ8pkeQTE6X_7dl-RXUSw#marionette

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -77,55 +95,176 @@ def run_check_child_prefs(self):
...
     }
 
+    start_session_count = 0
+
     def setUp(self):
-        # Inject the env var for this test and exclude all env vars from other
-        # tests, so run_check_felt_prefs can assert presence and absence.
-        self.EXTRA_ENV = self._ENV_BY_TEST.get(self._testMethodName, {})
+        current_env = self._ENV_BY_TEST.get(self._testMethodName, {})
+        self.EXTRA_ENV = current_env
```
lissyx: This could be removed right?

```suggestion
        self.EXTRA_ENV = self._ENV_BY_TEST.get(self._testMethodName, {})
```

And below we update `if key not in current_env` as `if key not in self.EXTRA_ENV`

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -77,55 +95,176 @@ def run_check_child_prefs(self):
...
             if test_name != self._testMethodName
             for key in env
+            if key not in current_env
+        }
+
+        self.EXTRA_PREFS = {
+            **HarnessPrefsTest.EXTRA_PREFS,
+            **self._PREFS_BY_TEST[self._testMethodName],
+        }
+        current_prefs = self._PREFS_BY_TEST[self._testMethodName]
```
lissyx: I'm a bit lost why we do this for prefs and not for env.

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -77,55 +95,176 @@ def run_check_child_prefs(self):
...
+        self.EXCLUDED_PREFS = {
+            key
+            for test_name, prefs in self._PREFS_BY_TEST.items()
+            if test_name != self._testMethodName
+            for key in prefs
+            if key not in current_prefs
         }
```
lissyx: this could we factorized between env and pref right? That kind of logic would benefit from it I guess

lissyx:
> > LGTM but I'd like those comments to be addressed, and please provide a try link with many retries :)
> 
> Here another one, (only one failure regarding something else)
> 
> https://firefox-ci-tc.services.mozilla.com/tasks/groups/MZZ8pkeQTE6X_7dl-RXUSw#marionette

No deadlock? So it may confirm our deadlocking is just a side effect of the mess I introduced when moving from Selenium to Marionette?

testing/enterprise/base_test.py
```diff
@@ -18,6 +18,27 @@
 from marionette_harness import MarionetteTestCase
 
 
+def patch_process_runner_args(instance):
```
lissyx, line 21: I assume you found no better way to transmit the information from our code down to the runner via marionette?

victhorlopez: no, it's not possible as far as I saw.. without modifying framework parts of course

testing/enterprise/felt_tests.py
```diff
@@ -722,9 +726,11 @@ def teardown(self):
         if not self._manually_closed_child:
             self._logger.info("Closing browser")
             self._child_driver.set_context("chrome")
+            pid = self._child_driver.session_capabilities["moz:processID"]
             self._child_driver.execute_script(
                 "Services.startup.quit(Ci.nsIAppStartup.eForceQuit);"
             )
+            self.wait_process_exit(pid)
```
lissyx, line 733: I missed the `ignore_children=True` patch from above and I am wondering how much the cleaner try you shared can be justified by either those changes. Have you ran then separately?

I would think that maybe doing this `self.wait_process_exit()` call here may help and we do not need the `ignore_children=True` patching?

testing/enterprise/test_felt_harness_prefs.py
victhorlopez: done

testing/enterprise/felt_tests.py
victhorlopez: I wrote it in the bug, it looks both could avoid also the main issue

testing/enterprise/test_felt_harness_prefs.py
victhorlopez: done

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -77,55 +95,189 @@ def run_check_child_prefs(self):
...
+    def setUp(self):
+        self.EXTRA_ENV = {
+            **HarnessPrefsTest.EXTRA_ENV,
+            **self._ENV_BY_TEST.get(self._testMethodName, {}),
+        }
+        self.EXCLUDED_ENV = self._excluded_keys(
+            self._ENV_BY_TEST, self._testMethodName, self.EXTRA_ENV
+        )
+
+        current_prefs = self._PREFS_BY_TEST[self._testMethodName]
```
lissyx: same, I dont get why we need this intermediate variable

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -77,55 +95,189 @@ def run_check_child_prefs(self):
...
+        self.EXTRA_PREFS = {
+            **HarnessPrefsTest.EXTRA_PREFS,
+            **current_prefs,
         }
+        self.EXCLUDED_PREFS = self._excluded_keys(
+            self._PREFS_BY_TEST, self._testMethodName, current_prefs
+        )
```
lissyx: `self._excluded_keys` is nice, but we should refactor a bit more, there is still the same processing done for both here, assuming we do not need `current_pref` (and I dont see a justification for its existence)

testing/enterprise/test_felt_harness_prefs.py
```diff
@@ -134,9 +286,24 @@ def _run(self):
...
         self._run()
+        assert HarnessPrefsTest.start_session_count == 3, (
+            f"Expected 3 session starts, got {HarnessPrefsTest.start_session_count}"
+        )
+
+    def test_4_harness_prefs_no_leakage(self):
+        self._run()
+        assert HarnessPrefsTest.start_session_count == 4, (
+            f"Expected 4 session starts, got {HarnessPrefsTest.start_session_count}"
+        )
```
lissyx, line 308: Four because it is the fourth eexecution of a test on the class, right?

I'd like the comment somewhere on those classes to be clear about that fact, because it's not obvious if you dont know a bit down of the details of execution from marionette that each `test_...` method will trigger a cycle of `setUp()` and `tearDown()`, hence count as one session start

lissyx (approved):
LGTM, please fix the comment on session start counts at least if you feel the refactor is not really required

testing/enterprise/test_felt_harness_prefs.py
victhorlopez: done, put a comment
 ```
 # Each test triggers a setUp()/tearDown() cycle, incrementing
    # start_session_count by one. The cumulative count confirms that
    # tests run in order and each gets exactly one new session.
    
 ```

PR #751, opened 2026-04-21
Enterprise release merge 2026042100

lissyx:
Merging from `release`


browser/locales/enterprise-l10n-changesets.json
```diff
@@ -19,6 +19,6 @@
             "win64-aarch64-devedition",
             "win64-devedition"
         ],
-        "revision": "fd59448016825ef627cda5bfd37899b033f0ab64"
+        "revision": "ccc4a1c76137398d7091571f05cb6d04ee7582d6"
```
jonathanmendez: Same concern about getting the right revision instead of the one from beta. This one might be more important to fix now given we might not has as much traffic in enterprise-release.

jonathanmendez (changes requested):
I also want to ensure people know how to grab the 150 build before we bump to 150.1 (see slack thread). But that should be resolved today.

jonathanmendez (approved):
Ok, I think there's been sufficient notice on the build for 150, let's move forward with release merges

PR #754, opened 2026-04-21
Enterprise beta merge 2026042100

lissyx:
No conflict merge

### DIFF
--- a/browser/locales/enterprise-l10n-changesets.json
+++ b/browser/locales/enterprise-l10n-changesets.json
@@ -19,6 +19,6 @@
             "win64-aarch64-devedition",
             "win64-devedition"
         ],
-        "revision": "ea7d8bd174908602cd662c73dcfadf135b4fa886"
+        "revision": "2c662982e9d0d1172214e319e8f678100e70ca8d"
     }
 }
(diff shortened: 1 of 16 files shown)


browser/locales/enterprise-l10n-changesets.json
```diff
@@ -19,6 +19,6 @@
             "win64-aarch64-devedition",
             "win64-devedition"
         ],
-        "revision": "ea7d8bd174908602cd662c73dcfadf135b4fa886"
+        "revision": "ccc4a1c76137398d7091571f05cb6d04ee7582d6"
```
jonathanmendez: This should technically be the revision from enterprise-main (I guess ideally after merging enterprise-main first, so I'll add a note to our instructions). If you want to catch it with tomorrow's merge that should be fine.

jonathanmendez (approved):

browser/locales/enterprise-l10n-changesets.json
lissyx: you're right, I'm doing too many things in parallel here, it's from `main` and not `enterprise-main`

PR #757, opened 2026-04-21
Bug 2033795: put locked properties on expected schema for enterprise

victhorlopez:
### Description 

Bugzilla: Bug-2033795

The test uses Troubleshoot.snapshot() to check all information that goes on about:support, for the preferences it is failing as it is checking which ones we have locked, to fix this I just added to the schema our prefs

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [ ] Added tests

[here](https://treeherder.mozilla.org/logviewer?job_id=561472635&repo=enterprise-firefox-pr&task=RKNoLdv_RsexQM-TwcDNFw.0&lineNumber=76416) the test passes

### DIFF
--- a/toolkit/modules/tests/browser/browser_Troubleshoot.js
+++ b/toolkit/modules/tests/browser/browser_Troubleshoot.js
@@ -489,6 +489,20 @@ const SNAPSHOT_SCHEMA = {
           required: false,
           type: "boolean",
         },
+        ...(AppConstants.MOZ_ENTERPRISE && {
+          "network.connectivity-service.IPv4.url": {
+            required: false,
+            type: "string",
+          },
+          "network.connectivity-service.IPv6.url": {
+            required: false,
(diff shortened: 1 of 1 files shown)


1rneh (commented):
Do you know why it's just the three preferences that need to be added? Looking at `EnterpriseEndpoints.sys.mjs` I would have assumed that we need to list all of the locked preferences?

victhorlopez:
> Do you know why it's just the three preferences that need to be added? Looking at `EnterpriseEndpoints.sys.mjs` I would have assumed that we need to list all of the locked preferences?

They are filtering for certain domains etc

https://searchfox.org/firefox-main/rev/a78cfde410d7bb58c143d46f3dea0af8ec0181f3/toolkit/modules/Troubleshoot.sys.mjs#87

1rneh (approved):

PR #764, opened 2026-04-22
Bug 2032305 - Enterprise: wait for distribution completion before getting enterprise console address (release uplift)

lissyx:
Bugzilla: Bug-2032305

Uplifting of PR #740 to release


lissyx:
We still have the pref on `enterprise-release` so that PR is not needed (and this is the reason it failed tests).

PR #768, opened 2026-04-22
Uplift Bug 2033812 - Common taskcluster setup for enterprise-beta

jonathanmendez:
### Description 

Bugzilla: Bug-2033812

Uplift of https://github.com/mozilla/enterprise-firefox/pull/761 to release.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -411,13 +411,13 @@ tasks:
                   in:
                       $if: >
                         eventType in ["action", "cron"]
-                        || (eventType == "push" && shortRef == "enterprise-release")
+                        || (eventType == "push" && shortRef in ["enterprise-release", "enterprise-beta"])
                         || (eventType == "push" && project == "enterprise-firefox-try")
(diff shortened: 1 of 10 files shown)


lissyx (commented):
I'm not sure there is much value in uplifting this beta-scoped change to release

jonathanmendez:
> I'm not sure there is much value in uplifting this beta-scoped change to release

The value is just in keeping these files in sync to reduce mental load. For example when 151 goes to release we will merge enterprise-beta to enterprise-release. These changes will be part of the merge and we'll have to think about it then. Or we just go ahead and uplift it now so they're in sync.

lissyx:
> > I'm not sure there is much value in uplifting this beta-scoped change to release
> 
> The value is just in keeping these files in sync to reduce mental load. For example when 151 goes to release we will merge enterprise-beta to enterprise-release. These changes will be part of the merge and we'll have to think about it then. Or we just go ahead and uplift it now so they're in sync.

Hm I missed the fact it would generate a merge conflict if we dont have them

lissyx (approved):

PR #780, opened 2026-04-23
Bug 2034430 - Disable Firefox Relay feature in enterprise builds

victhorlopez:
### Description 

Bugzilla: Bug-2034430

This PR does the following:
- Sets the signon.firefoxRelay.feature pref default to "disabled" (instead of "available") in enterprise builds via #ifdef MOZ_ENTERPRISE in firefox.js
  - Skips the existing Relay browser tests (browser_relay_signup_flow.js, browser_relay_signup_flow_showToAllBrowsers.js, browser_relay_telemetry.js,
  browser_relay_use.js) on enterprise builds since they assume the feature is available
  - Adds a new enterprise-only test (browser_relay_disabled_enterprise.js) that verifies:
    - The default pref value is "disabled"
    - FirefoxRelayUtils.relayIsAvailableOrEnabled() returns false
    - The Relay autocomplete item does not appear in the password popup

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [x] Added tests

[See issue](https://treeherder.mozilla.org/logviewer?job_id=561992871&repo=enterprise-firefox-pr&task=Hw735vkiTPyRoyuwAWb97g.0&lineNumber=196211)
[Here with the changes, passing](https://treeherder.mozilla.org/logviewer?job_id=562024511&repo=enterprise-firefox-pr&task=OyOP9BifRG6eMaQTnO2Uzw.0&lineNumber=187484)

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -2974,7 +2974,12 @@ pref("browser.migrate.preferences-entrypoint.enabled", true);
 // "offered"        - we have offered feature to user and they have not yet made a decision.
 // "enabled"        - user opted in to the feature.
 // "disabled"       - user opted out of the feature.
+#ifdef MOZ_ENTERPRISE
+// Firefox Relay is not supported in enterprise builds
+pref("signon.firefoxRelay.feature", "unavailable");
+#else
 pref("signon.firefoxRelay.feature", "available");
+#endif
 pref("signon.management.page.breach-alerts.enabled", true);
 pref("signon.management.page.vulnerable-passwords.enabled", true);
 pref("signon.management.page.sort", "name");
--- a/toolkit/components/passwordmgr/test/browser/browser.toml
+++ b/toolkit/components/passwordmgr/test/browser/browser.toml
@@ -282,17 +282,25 @@ skip-if = [
   "os == 'win' && os_version == '11.26200' && verify-standalone",
 ]
 
+["browser_relay_disabled_enterprise.js"]
+support-files = ["browser_relay_utils.js"]
+run-if = ["enterprise"]
+
 ["browser_relay_signup_flow.js"]
 support-files = ["browser_relay_utils.js"]
+skip-if = ["enterprise"]
 
 ["browser_relay_signup_flow_showToAllBrowsers.js"]
 support-files = ["browser_relay_utils.js"]
+skip-if = ["enterprise"]
 
 ["browser_relay_telemetry.js"]
 support-files = ["browser_relay_utils.js"]
+skip-if = ["enterprise"]
 
 ["browser_relay_use.js"]
 support-files = ["browser_relay_utils.js"]
+skip-if = ["enterprise"]
 
 ["browser_telemetry_SignUpFormRuleset.js"]
 
--- a/toolkit/components/passwordmgr/test/browser/browser_relay_disabled_enterprise.js
+++ b/toolkit/components/passwordmgr/test/browser/browser_relay_disabled_enterprise.js
@@ -1,0 +1,50 @@
+const TEST_URL_PATH = `https://example.org${DIRECTORY_PATH}form_basic_signup.html`;
+
+Services.scriptloader.loadSubScript(
+  "chrome://mochitests/content/browser/toolkit/components/passwordmgr/test/browser/browser_relay_utils.js",
+  this
+);
+
+add_task(async function test_enterprise_relay_feature_default_disabled() {
+  Assert.equal(
+    Services.prefs
+      .getDefaultBranch("")
+      .getStringPref("signon.firefoxRelay.feature"),
+    "unavailable",
+    "signon.firefoxRelay.feature default should be 'unavailable' in enterprise builds"
+  );
+});
+
+add_task(async function test_enterprise_relay_utils_reports_disabled() {
+  const { FirefoxRelayUtils } = ChromeUtils.importESModule(
+    "resource://gre/modules/FirefoxRelayUtils.sys.mjs"
+  );
+
+  Assert.ok(
+    !FirefoxRelayUtils.relayIsAvailableOrEnabled(),
+    "FirefoxRelayUtils.relayIsAvailableOrEnabled() should return false when feature is disabled"
+  );
+});
+
+add_task(
+  async function test_enterprise_relay_autocomplete_not_shown_when_disabled() {
+    const rsSandbox = await stubRemoteSettingsAllowList();
+    await BrowserTestUtils.withNewTab(
+      {
+        gBrowser,
+        url: TEST_URL_PATH,
+      },
+      async function (browser) {
+        const popup = document.getElementById("PopupAutoComplete");
+        await openACPopup(popup, browser, "#form-basic-username");
+
+        const relayItem = getRelayItemFromACPopup(popup);
+        Assert.ok(
+          !relayItem,
+          "Relay item SHOULD NOT be present in the autocomplete popup when the feature is disabled"
+        );
+      }
+    );
+    rsSandbox.restore();
+  }
+);


lissyx (approved):

1rneh (commented):
My first thought was that this test can be skipped because Relay is a feature coupled to Firefox Accounts which is not supported in enterprise builds, hence this test should also not pass with an unlocked preference because Relay shouldn't be available/surfaced in enterprise builds in the first place.

lissyx:
> My first thought was that this test can be skipped because Relay is a feature coupled to Firefox Accounts which is not supported in enterprise builds, hence this test should also not pass with an unlocked preference because Relay shouldn't be available/surfaced in enterprise builds in the first place.

You're raising a good point, maybe the test should run on enterprise build to assert relay is properly blocked ?

victhorlopez:
> > My first thought was that this test can be skipped because Relay is a feature coupled to Firefox Accounts which is not supported in enterprise builds, hence this test should also not pass with an unlocked preference because Relay shouldn't be available/surfaced in enterprise builds in the first place.
> 
> You're raising a good point, maybe the test should run on enterprise build to assert relay is properly blocked ?

Should we disable this pref?
https://searchfox.org/firefox-main/rev/a358f48a49a4d72df7b214805c4b2d2049d34a5a/browser/app/profile/firefox.js#2770

lissyx:
> > > My first thought was that this test can be skipped because Relay is a feature coupled to Firefox Accounts which is not supported in enterprise builds, hence this test should also not pass with an unlocked preference because Relay shouldn't be available/surfaced in enterprise builds in the first place.
> > 
> > 
> > You're raising a good point, maybe the test should run on enterprise build to assert relay is properly blocked ?
> 
> Should we disable this pref? https://searchfox.org/firefox-main/rev/a358f48a49a4d72df7b214805c4b2d2049d34a5a/browser/app/profile/firefox.js#2770

At first that may be a good idea, but maybe we want `--disable-relay` at build time? Would it be feasible?

1rneh:
> At first that may be a good idea, but maybe we want --disable-relay at build time? Would it be feasible?

The need is not so clear to me why relay should be disabled at build time since that's not what we do for other non-enterprise functionalities. Setting the preference to `signon.firefoxRelay.feature=disabled` as Victor suggested seems sufficient. 

That what we do with `browser.profiles.enabled` in `firefox-enterprise.js` as well. That makes me think, how we override defaults is quite scattered. We disable profiles in firefox-enterprise.js but clear a bunch of preferences next to where they are originally set e.g. in firefox.js. Should we align our approaches and always override enterprise defaults next to where the non-enterprise defaults are defined?

1rneh:
> You're raising a good point, maybe the test should run on enterprise build to assert relay is properly blocked ?

I think we can still skip this one. But also add a test that's only run in enterprise builds that asserts that the expected relay item is not surfaced? That approach seems less invasive and less prone to merge conflicts?

victhorlopez:
Okay so this is what I will do

- Close this PR, this test will be disabled
- Create a new test (in the same .toml enterprise only) that ensures the pref is disabled and that relay is not working
- action item for us, we need to clarify longterm if we want to disable features at compile time?

1rneh:
> Okay so this is what I will do
> 
>     * Close this PR, this test will be disabled

You can also simply update this PR and leave a comment on the Bugzilla item explaining why we are skipping the test in enterprise builds. This way the contents of the conversation in this PR are not becoming obsolete.

>     * Create a new test (in the same .toml enterprise only) that ensures the pref is disabled and that relay is not working
> 

Yes!

>     * action item for us, we need to clarify longterm if we want to disable features at compile time?

Maybe @lissyx can elaborate here a bit?

lissyx:
> > At first that may be a good idea, but maybe we want --disable-relay at build time? Would it be feasible?
> 
> The need is not so clear to me why relay should be disabled at build time since that's not what we do for other non-enterprise functionalities. Setting the preference to `signon.firefoxRelay.feature=disabled` as Victor suggested seems sufficient.

Well, any feature we absolutely do not use, I think it's better the code is just not there to reduce risks, and it's something we would have on m-c, assuming relay would never be something we care about (which may not be true, I dont know).

Setting the preference may work, but i'm just unsure how all the rest of the code will behave when reading this: https://searchfox.org/firefox-main/rev/a358f48a49a4d72df7b214805c4b2d2049d34a5a/toolkit/components/satchel/integrations/FirefoxRelayUtils.sys.mjs#16-22

Probably easier to just ask to the people working on it what is the best course of action here

> 
> That what we do with `browser.profiles.enabled` in `firefox-enterprise.js` as well. That makes me think, how we override defaults is quite scattered. We disable profiles in firefox-enterprise.js but clear a bunch of preferences next to where they are originally set e.g. in firefox.js. Should we align our approaches and always override enterprise defaults next to where the non-enterprise defaults are defined?

I dont have an opinion yet on how we should do it, but I agree the current state is hard to track

lissyx:
> > Okay so this is what I will do
> > ```
> > * Close this PR, this test will be disabled
> > ```
> 
> You can also simply update this PR and leave a comment on the Bugzilla item explaining why we are skipping the test in enterprise builds. This way the contents of the conversation in this PR are not becoming obsolete.

Yep, I agree on that

> 
> > ```
> > * Create a new test (in the same .toml enterprise only) that ensures the pref is disabled and that relay is not working
> > ```
> 
> Yes!

:+1: 

> 
> > ```
> > * action item for us, we need to clarify longterm if we want to disable features at compile time?
> > ```
> 
> Maybe @lissyx can elaborate here a bit?

I'm not sure what I can elaborate more, I was just sharing the idea: if the feature is completely disabled and no code is compiled, then tests would not have to run as well

1rneh:
> > > At first that may be a good idea, but maybe we want --disable-relay at build time? Would it be feasible?
> > 
> > 
> > The need is not so clear to me why relay should be disabled at build time since that's not what we do for other non-enterprise functionalities. Setting the preference to `signon.firefoxRelay.feature=disabled` as Victor suggested seems sufficient.
> 
> Well, any feature we absolutely do not use, I think it's better the code is just not there to reduce risks, and it's something we would have on m-c, assuming relay would never be something we care about (which may not be true, I dont know).
> 

Generally I agree, but so far it just hasn't been anything we explored for other non-enterprise functionalities unless for simple `#ifdef` exclusions at build time. So it's maybe worth coming up with a strategy here.

So I think this highlights a broader question about which services we want to offer in Firefox Enterprise. What do we do about Relay, Monitor and Mozilla VPN (Note: That's not Firefox VPN) in the enterprise builds? Do we disable all, only some, none? Let's discuss this with the team. 

> Setting the preference may work, but i'm just unsure how all the rest of the code will behave when reading this: https://searchfox.org/firefox-main/rev/a358f48a49a4d72df7b214805c4b2d2049d34a5a/toolkit/components/satchel/integrations/FirefoxRelayUtils.sys.mjs#16-22
> 
> Probably easier to just ask to the people working on it what is the best course of action here
> 

Yes, 100%. I can help out with that, as I've worked on parts of the Relay integration into the password manager in my previous team :)

victhorlopez:
> > > > At first that may be a good idea, but maybe we want --disable-relay at build time? Would it be feasible?
> > > 
> > > 
> > > The need is not so clear to me why relay should be disabled at build time since that's not what we do for other non-enterprise functionalities. Setting the preference to `signon.firefoxRelay.feature=disabled` as Victor suggested seems sufficient.
> > 
> > 
> > Well, any feature we absolutely do not use, I think it's better the code is just not there to reduce risks, and it's something we would have on m-c, assuming relay would never be something we care about (which may not be true, I dont know).
> 
> Generally I agree, but so far it just hasn't been anything we explored for other non-enterprise functionalities unless for simple `#ifdef` exclusions at build time. So it's maybe worth coming up with a strategy here.
> 
> So I think this highlights a broader question about which services we want to offer in Firefox Enterprise. What do we do about Relay, Monitor and Mozilla VPN (Note: That's not Firefox VPN) in the enterprise builds? Do we disable all, only some, none? Let's discuss this with the team.
> 
> > Setting the preference may work, but i'm just unsure how all the rest of the code will behave when reading this: https://searchfox.org/firefox-main/rev/a358f48a49a4d72df7b214805c4b2d2049d34a5a/toolkit/components/satchel/integrations/FirefoxRelayUtils.sys.mjs#16-22
> > Probably easier to just ask to the people working on it what is the best course of action here
> 
> Yes, 100%. I can help out with that, as I've worked on parts of the Relay integration into the password manager in my previous team :)

I asked here, https://matrix.to/#/!MkUoZtetjdyykPqtIH:mozilla.org/$sCcfL-iBcwQPFgHz1Cs-FZd--QFy88o0qyK_IdOjzd4?via=mozilla.org&via=matrix.org&via=vovo.id.au 

Regarding the general strategy about those features let's discuss later then

lissyx:
> > > > At first that may be a good idea, but maybe we want --disable-relay at build time? Would it be feasible?
> > > 
> > > 
> > > The need is not so clear to me why relay should be disabled at build time since that's not what we do for other non-enterprise functionalities. Setting the preference to `signon.firefoxRelay.feature=disabled` as Victor suggested seems sufficient.
> > 
> > 
> > Well, any feature we absolutely do not use, I think it's better the code is just not there to reduce risks, and it's something we would have on m-c, assuming relay would never be something we care about (which may not be true, I dont know).
> 
> Generally I agree, but so far it just hasn't been anything we explored for other non-enterprise functionalities unless for simple `#ifdef` exclusions at build time. So it's maybe worth coming up with a strategy here.

Not for non enterprise features, but it's not unheard off in other context, I can remember of `--disable` being added for B2G builds where feature were either not required or could even collide. We definitively need to think this and come up with a strategy, and if we do this, it should be on m-c and not just on our builds.

> 
> So I think this highlights a broader question about which services we want to offer in Firefox Enterprise. What do we do about Relay, Monitor and Mozilla VPN (Note: That's not Firefox VPN) in the enterprise builds? Do we disable all, only some, none? Let's discuss this with the team.

:+1: 

> 
> > Setting the preference may work, but i'm just unsure how all the rest of the code will behave when reading this: https://searchfox.org/firefox-main/rev/a358f48a49a4d72df7b214805c4b2d2049d34a5a/toolkit/components/satchel/integrations/FirefoxRelayUtils.sys.mjs#16-22
> > Probably easier to just ask to the people working on it what is the best course of action here
> 
> Yes, 100%. I can help out with that, as I've worked on parts of the Relay integration into the password manager in my previous team :)

Ah, cool!

browser/app/profile/firefox.js
```diff
@@ -2880,7 +2868,11 @@ pref("browser.migrate.preferences-entrypoint.enabled", true);
 // "offered"        - we have offered feature to user and they have not yet made a decision.
 // "enabled"        - user opted in to the feature.
 // "disabled"       - user opted out of the feature.
+#ifdef MOZ_ENTERPRISE
+pref("signon.firefoxRelay.feature", "disabled");
+#else
 pref("signon.firefoxRelay.feature", "available");
+#endif
```
lissyx, line 2982: So I think this is something that was discussed at some point, and we should likely make sure that kind of change is properly scoped to our preferences file, and not disseminated. @1rneh I think you reported this right? But I cannot remember if we had a follow up bug to fix those.

So here I think this should rather move to `firefox-enterprise.js` from our branding directory

toolkit/components/passwordmgr/test/browser/browser.toml
lissyx:
```suggestion
run-if = ["enterprise"]
```

Otherwise with the other below it's confusing

lissyx (changes requested):

browser/app/profile/firefox.js
victhorlopez: It was discussed somewhere that it was scattered currently. Mainly it comes from https://github.com/mozilla/enterprise-firefox/commit/937f94b78c4c759f3b2a06c8ac52c5accb688c31

Is the goal to have them all in firefox-enterprise.js?

lissyx: > It was discussed somewhere that it was scattered currently. Mainly it comes from [937f94b](https://github.com/mozilla/enterprise-firefox/commit/937f94b78c4c759f3b2a06c8ac52c5accb688c31)
> 
> Is the goal to have them all in firefox-enterprise.js?

I'm not sure we have established where yet

1rneh: I remember advocating for exactly the opposite 😅

I mentioned that I'd prefer to not scatter the overrides of enterprise preferences. By that I meant setting them in the same place where the non-enterprise preference are set (as we did in 937f94b78c4c759f3b2a06c8ac52c5accb688c31 as @victhorlopez mentioned), hence whenever possible we should avoid setting them in `firefox-enterprise.js`, because that is just an extra file to look for overrides.

victhorlopez: I put it back in the other file

lissyx: > I remember advocating for exactly the opposite 😅
> 
> I mentioned that I'd prefer to not scatter the overrides of enterprise preferences. By that I meant setting them in the same place where the non-enterprise preference are set (as we did in [937f94b](https://github.com/mozilla/enterprise-firefox/commit/937f94b78c4c759f3b2a06c8ac52c5accb688c31) as @victhorlopez mentioned), hence whenever possible we should avoid setting them in `firefox-enterprise.js`, because that is just an extra file to look for overrides.

Thanks for confirming, did we ever open a follow up for that? If not we should now and make it blocking MVP, I think it's a rather non invasive refactoring that would help making things more clear.

lissyx (approved):

browser/app/profile/firefox.js
```diff
@@ -2880,7 +2880,12 @@ pref("browser.migrate.preferences-entrypoint.enabled", true);
 // "offered"        - we have offered feature to user and they have not yet made a decision.
 // "enabled"        - user opted in to the feature.
 // "disabled"       - user opted out of the feature.
+#ifdef MOZ_ENTERPRISE
+// Firefox Relay is not supported in enterprise builds
```
1rneh, line 2978: `nit:` "not supported" might be misleading here. We support it but the default for enterprise builds is that it's disabled but can be enabled by a policy (which doesn't exist as of now).

1rneh (approved):

browser/app/profile/firefox.js
victhorlopez: If we support it (I thought we don't based on the excel), then the tests should be enabled and set "signon.firefoxRelay.feature", "available" for those tests for enterprise. (together with the original unlock of url). Do you see it like this or have a diferent opinion?

1rneh: My comment was intended as a nit and solely about the phrasing of the comment. But I have to correct my previous statement. An idea was to have a policy to disable Relay. But now that I'm revisiting this, that policy won't be applicable for Firefox Enterprise builds. 

Let's wait for a product decision, because if we disable the Relay feature entirely, we also need to make sure it's not surfaced and toggleable from anywhere e.g. about:preferences, ideally with test coverage.

victhorlopez: okay, is my understand correct now, that we **do not** support relay?

I wrote a test already using their utils and similar logic, but based on your comment we need to test more (i.e about preferences and so on).

1rneh: Yes, as of now we do not support Relay, hence the user shouldn't be able to turn it on. I checked and setting the preference `signon.firefoxRelay.feature` to `unavailable` instead of `disabled`, will make it not show up e.g. in `about:preferences#privacy`. So let's set the preference value to `unavailable`? Relay then still shows up on `about:preferences#moreFromMozilla`, but that section can be controled via the policy `UserMessaging`.

victhorlopez: Where did you get this `unavailable` from?  

In the prefs it is not stated, and for this check they use this. Does it behave different than disabled?

[link](https://searchfox.org/firefox-main/rev/e28b34ab33dbf49364999070168cbb7e11e8e5bd/toolkit/components/satchel/integrations/FirefoxRelayUtils.sys.mjs#21)

edit: I just tried and with `unavailable` it disappears ... nice

victhorlopez: Based on this test I guess this is equivalent to undefined?
[test](https://searchfox.org/firefox-main/rev/e28b34ab33dbf49364999070168cbb7e11e8e5bd/browser/components/preferences/tests/privacy/browser_privacy_relayIntegration.js#40)

edit: ok I am leaving it as `unavailable`, reason is to get `undefined` we should not set the pref, but that seems quite easy that it gets broken

1rneh: I had a vague memory that when we integrated relay into the password manager that we made a distinction between whether the pref was set to `unavailable` or `disabled`. So I tried it out and then assumed preemptively that `unavailable` was the correct value. But I think you are pointing it out correctly. It can be any arbitrary value except `"available", "offered", "enabled" or "disabled"`.

1rneh: I've [reached out to Relay folks](https://matrix.to/#/!MkUoZtetjdyykPqtIH:mozilla.org/$C5fbXD_XlIjdLqrrIzskFknhCl223eWQ7pgJfhqx7nY) to make sure this doesn't cause unexpected breakage.

jporter-dev:
Closing this for #1224 since this PR no longer receives a decision task on new pushes. I force-pushed to this and it failed to run (creating #1224 with the same commit hashes made it run).

PR #783, opened 2026-04-23
Bug 2034765: Expose glean telemetry dictionary in a standardized format

michaelvanstraten:
### Description

Bugzilla: Bug-2034765

To support client side telemetry reporting to generic [Open Telemetry]([url](https://opentelemetry.io/)) compatible endpoint we need to convert the glean telemetry dictionary information into a [schema that is standardized and well understood](https://opentelemetry.io/docs/concepts/semantic-conventions/). 

This patch sets up a task that generates a [semconv schema](https://github.com/open-telemetry/weaver/blob/173fed70c05bf9c7833f59f80aae09642ac05114/schemas/semconv-syntax.v2.md) from all `metrics.yaml` files and uploads it as an artifact.

You can run the generator manually by running:

```bash
mach glean semconv
```

### DIFF
--- a/taskcluster/docs/kinds.rst
+++ b/taskcluster/docs/kinds.rst
@@ -977,3 +977,10 @@ Mac notarization of customized versions of releases for enterprises.
 enterprise-repack-mac-signing
 ----------------------------------
 Mac signature of customized versions of releases for enterprises.
+
+glean
+-----
+Tasks that run ``./mach glean`` utilities. The ``semconv`` task produces the
+resolved OTel semantic convention registry schema from Firefox Glean metric
+definitions, consumed by the ``moa generate release``
+tool.
--- a/taskcluster/kinds/glean/kind.yml
+++ b/taskcluster/kinds/glean/kind.yml
@@ -1,0 +1,48 @@
+---
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - gecko_taskgraph.transforms.job:transforms
+    - gecko_taskgraph.transforms.task:transforms
+
+kind-dependencies:
+    - toolchain
+    - fetch
+
+tasks:
+    semconv:
+        description: >-
+            Generate the resolved OTel semantic convention registry schema from
+            Firefox Glean metric definitions. The resulting resolved.yaml is
+            consumed by the `moa generate release` tool, which embeds it as a
+            semconv referrer artifact on Harbor.
+        run-on-projects: ["enterprise-firefox"]
+        run-on-repo-type: ["git"]
+        worker-type: b-linux
+        worker:
+            docker-image: {in-tree: "lint"}
+            max-run-time: 1800
+            artifacts:
+                - type: file
+                  name: public/build/resolved.yaml
+                  path: /builds/worker/artifacts/resolved.yaml
+                - type: file
+                  name: public/build/manifest.yaml
+                  path: /builds/worker/artifacts/manifest.yaml
+        run:
+            using: mach
+            mach: "glean semconv --output-dir /builds/worker/artifacts"
+            use-caches: [checkout, pip, uv]
+        attributes:
+            shippable: true
+        shipping-phase: build
+        shipping-product: firefox-enterprise
+        index:
+            product: firefox
+            job-name: glean-semconv
+            type: shippable
+        treeherder:
+            symbol: Glean(semconv)
+            kind: build
+            platform: linux64/opt
+            tier: 1
--- a/toolkit/components/glean/build_scripts/semconv/__init__.py
+++ b/toolkit/components/glean/build_scripts/semconv/__init__.py
@@ -6,7 +6,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 
 import yaml
-from buildconfig import config, topsrcdir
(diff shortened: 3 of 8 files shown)


michaelvanstraten:
Hi @jonathanmendez, I heard that you started to get familiar with task cluster, this might be the perfect PR to review for you :)?

toolkit/components/glean/build_scripts/semconv/__init__.py
```diff
@@ -0,0 +1,323 @@
...
+            f,
+            Dumper=_LiteralBlockDumper,
+            sort_keys=False,
+            allow_unicode=True,
+        )
+
+    with open(manifest_path, "w") as f:
+        yaml.dump(manifest, f, sort_keys=False, allow_unicode=True)
+
+    manifest_errors = validate_against_schema(manifest, MANIFEST_SCHEMA_PATH)
```
jonathanmendez, line 195: I would think this validation should happen before writing the file(s).

I would expect either:
```
if validate_resolved():
  print errors
  exit
write_resolved()
print("Wrote resolved")

if validate_manifest():
  print errors
  exit
write_manifest()
print("Wrote manifest")
```

or:
```
if validate_resolved():
  print errors
  exit
if validate_manifest():
  print errors
  exit

write_resolved()
write_manifest()

print("Wrote resolved")
print("Wrote manifest")
```

taskcluster/kinds/glean/kind.yml
```diff
@@ -0,0 +1,48 @@
...
+                  path: /builds/worker/artifacts/resolved.yaml
+                - type: file
+                  name: public/build/manifest.yaml
+                  path: /builds/worker/artifacts/manifest.yaml
+        run:
+            using: mach
+            mach: "glean semconv --output-dir /builds/worker/artifacts"
+            use-caches: [checkout, pip, uv]
+        attributes:
+            shippable: true
```
jonathanmendez, line 37: Why is this marked shippable?

taskcluster/kinds/glean/kind.yml
```diff
@@ -0,0 +1,48 @@
...
+        shipping-phase: build
+        shipping-product: firefox-enterprise
+        index:
+            product: firefox
+            job-name: glean-semconv
+            type: shippable
```
jonathanmendez, line 43: As above: why is this marked shippable?

taskcluster/docs/kinds.rst
```diff
@@ -977,3 +977,10 @@ Mac notarization of customized versions of releases for enterprises.
 enterprise-repack-mac-signing
 ----------------------------------
 Mac signature of customized versions of releases for enterprises.
+
+glean
+-----
+Tasks that run ``./mach glean`` utilities. The ``semconv`` task produces the
+resolved OTel semantic convention registry schema from Firefox Glean metric
+definitions, consumed by the ``moa generate release``
+tool.
```
jonathanmendez, line 986: Is there a way to provide better context for what any of this means? I barely know what OTel means (using "OpenTelemetry" would be nicer), but would have no idea what a "resolved semantic convention registry schema" is nor what the `moa generate release` tool is or why we want to use the `moa generate release` tool.

Here's an attempt:

```suggestion
Tasks that run ``./mach glean`` utilities.

The ``semconv`` task transforms Firefox Glean metric definitions into
a standardized OpenTelemetry format known as "Semantic Conventions".
This format can be consumed by the [[tooling explained by purpose]].
```

taskcluster/kinds/glean/kind.yml
```diff
@@ -0,0 +1,48 @@
...
+    - gecko_taskgraph.transforms.task:transforms
+
+kind-dependencies:
+    - toolchain
+    - fetch
+
+tasks:
+    semconv:
+        description: >-
+            Generate the resolved OTel semantic convention registry schema from
```
jonathanmendez, line 15: Same concerns on the description as above in kinds.rst

toolkit/components/glean/build_scripts/semconv/schemas/publication-manifest.v2.json
```diff
@@ -0,0 +1,124 @@
+{
```
jonathanmendez, line 1: There's a process for vendoring components: https://firefox-source-docs.mozilla.org/mozbuild/vendor/index.html

I don't know if it works differently when there's only a couple of schema files, but it does feel odd for them to just be here in the source without any indication that they are third party.

jonathanmendez (changes requested):
I haven't had time to review all of the python, but I have some comments on the rest so far.

I don't feel like I am enough to review this on my own. I would add @ahal for extra taskcluster coverage at least.

toolkit/components/glean/build_scripts/semconv/schemas/publication-manifest.v2.json
michaelvanstraten: Okay, valid point. I will add a `moz.yaml`

toolkit/components/glean/build_scripts/semconv/__init__.py
michaelvanstraten: That is a very good point, thank you.

PR #784, opened 2026-04-23
Bug 2034400 - Lock browser.ipProtection.openedPanelWithLocation with AccessConnector policy to prevent transient VPN button with dot from showing

jporter-dev:
### Description 

This PR adds an additional pref to the AccessConnector policy to force disable the new transient "attention dot" on the VPN button which is currently being repurposed for AccessConnector. The "attention dot" implementation causes the VPN button to show until this pref is toggled (which happens when a user opens the panel).

> Note: this does _*not*_ fix the AccessConnector connectivity/initialization bug, just the button.

Bugzilla: Bug-2034400

---

* Change that caused this: https://phabricator.services.mozilla.com/D295063
* Blocks: #785 

---

### Screenshots 

image

_Before locking the pref on a fresh profile_

image

_After locking the pref on a fresh profile_

---

### Testing 

**Steps to verify changes:**

1. Start with a fresh profile
2. Open the browser
3. Observe a visible disconnected VPN icon with a blue status dot
4. Click the VPN button to show the panel (this toggles the pref)
5. Quit and reopen Firefox Enterprise
6. Observe the original AccessConnector button

**Expected result:**

VPN button is not visible, AccessConnector button is visible.

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -174,6 +174,14 @@ export var Policies = {
         );
       }
 
+      // Must be set before enabled so the badge dot doesn't flash when
+      // the pref observer triggers IPProtectionService.init().
+      PoliciesUtils.setDefaultPref(
+        "browser.ipProtection.openedPanelWithLocation",
+        true,
+        locked
+      );
+
       // Set enabled last so that all other prefs are in place when
       // the pref observer triggers IPProtectionService.init().
       PoliciesUtils.setDefaultPref(
@@ -188,6 +196,7 @@ export var Policies = {
       unsetAndUnlockPref("browser.ipProtection.autoStartEnabled");
       unsetAndUnlockPref("browser.ipProtection.mode");
       unsetAndUnlockPref("browser.ipProtection.override.serverlist");
+      unsetAndUnlockPref("browser.ipProtection.openedPanelWithLocation");
       if ("MatchPatterns" in oldParams) {
         unsetAndUnlockPref("browser.ipProtection.inclusion.match_patterns");
       }


toolkit/components/ipprotection/fxa/IPPSignInWatcher.sys.mjs
```diff
@@ -26,13 +28,20 @@ class IPPSignInWatcherSingleton extends EventTarget {
...
     this.#signedIn = Services.prefs.prefHasUserValue("services.sync.username");
   }
 
   /**
    * Adds an observer for the FxA sign-in state, only when the browser is fully started.
    */
   async initOnStartupCompleted() {
+    if (AppConstants.MOZ_ENTERPRISE) {
+      return;
+    }
```
lissyx: So from what I was debugging `initOnStartupCompleted` was never called at all, because https://searchfox.org/enterprise-main/rev/11acd910851157e4c161e4eb32623217ab1d0f0a/toolkit/components/ipprotection/IPProtectionService.sys.mjs#138-142 was never called :(

toolkit/components/ipprotection/fxa/IPPEnrollAndEntitleManager.sys.mjs
```diff
@@ -181,6 +191,19 @@ class IPPEnrollAndEntitleManagerSingleton extends EventTarget {
...
+      // For enterprise, FxA enrollment is not required. Fetch entitlement
+      // directly using the Felt access token.
+      const { entitlement, error } =
+        await lazy.IPProtectionService.authProvider.getEntitlement();
+      if (!entitlement) {
+        this.#setEntitlement(null);
+        return { isEnrolledAndEntitled: false, error };
+      }
+      this.#setEntitlement(entitlement);
+      return { isEnrolledAndEntitled: true };
```
lissyx: this may be the correct fix

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -181,13 +181,22 @@ export var Policies = {
...
         locked
       );
+
+      // Pref to track if user has opened the VPN panel since location controls were introduced
+      // The implementation of the status dot conflicted with the Access Connector button
+      PoliciesUtils.setDefaultPref(
+        "browser.ipProtection.openedPanelWithLocation",
+        true,
+        locked
+      );
```
lissyx: While this works, I'm unsure we should use that like this. That's where the UX of that component is unclear to me, but I would have assumed the `#updateBadge()` code that depends on that pref would update an icon we **removed**, so a no-op. Are we removing it or just hiding?

toolkit/components/ipprotection/fxa/IPPSignInWatcher.sys.mjs
```diff
@@ -26,13 +28,20 @@ class IPPSignInWatcherSingleton extends EventTarget {
   }
 
   init() {
+    if (AppConstants.MOZ_ENTERPRISE) {
+      this.#signedIn = true;
+      return;
+    }
     this.#signedIn = Services.prefs.prefHasUserValue("services.sync.username");
```
lissyx, line 29: Should this work? I dont see this value ever set

toolkit/components/ipprotection/IPPStartupCache.sys.mjs
```diff
@@ -113,7 +113,9 @@ class IPPStartupCacheSingleton {
     this.#stateFromCache = null;
 
     if (
-      lazy.IPProtectionService.state !== lazy.IPProtectionStates.UNINITIALIZED
+      lazy.IPProtectionService.state !==
+        lazy.IPProtectionStates.UNINITIALIZED ||
+      lazy.IPProtectionService.featureEnabled
```
lissyx: This looks really important, it is starting to unblock a lot of things

jporter-dev: Yeah, I think we need more tracing to find where this state gets changed that we're not hitting. Think this hack should be removed

lissyx: @jporter-dev without this, nothing can happen.

toolkit/components/ipprotection/fxa/IPPSignInWatcher.sys.mjs
lissyx: This one seems not to be needed

toolkit/components/ipprotection/fxa/IPPSignInWatcher.sys.mjs
```diff
@@ -26,13 +28,20 @@ class IPPSignInWatcherSingleton extends EventTarget {
   }
 
   init() {
+    if (AppConstants.MOZ_ENTERPRISE) {
+      this.#signedIn = true;
+      return;
+    }
```
lissyx: This is required, but I think the pref below should be populated and it seems not to be?

toolkit/components/ipprotection/fxa/IPPEnrollAndEntitleManager.sys.mjs
```diff
@@ -204,6 +227,10 @@ class IPPEnrollAndEntitleManagerSingleton extends EventTarget {
    * @returns {string} [status.error] - Error message if entitlement fetch failed.
    */
   async #entitle(forceRefetch = false) {
+    if (AppConstants.MOZ_ENTERPRISE) {
+      return { isEnrolledAndEntitled: true };
+    }
+
```
lissyx: This one can be removed, I get the feature working correctly without

toolkit/components/ipprotection/fxa/IPPEnrollAndEntitleManager.sys.mjs
```diff
@@ -61,6 +63,14 @@ class IPPEnrollAndEntitleManagerSingleton extends EventTarget {
...
     // This bit must be async because we want to trigger the updateState at
     // the end of the rest of the initialization.
+    if (AppConstants.MOZ_ENTERPRISE) {
+      // For enterprise, use maybeEnrollAndEntitle so that isEnrolling is true
+      // synchronously during startup. This prevents IPPEarlyStartupFilter from
+      // cancelling the channel filter when isReady is checked during the
+      // async entitlement fetch.
+      this.maybeEnrollAndEntitle();
+      return;
+    }
```
lissyx: Maybe not needed, things are fine without for me

toolkit/components/ipprotection/IPPStartupCache.sys.mjs
jporter-dev: Turns out this is the only change that's needed. I removed the changes in `IPPEnrollAndEntitleManager` and things worked, and both `IPProtectionService.state` and `IPPSignInWatcher.isSignedIn` reflected the proper states in the browser console. Startup cache just skips initializing things entirely if there's no cached state regardless of the feature enablement (which I guess makes sense for normal browser users and the standard VPN feature with manual enabling)

lissyx: AH so I was not wrong, I made so many tests I was not sure anymore at some point

lissyx (approved):

lissyx:
This is a temporary measure until the UX changes lands

lissyx:
I'll include it as an uplift in beta merge because it was uplifted to beta on m-c: https://bugzilla.mozilla.org/show_bug.cgi?id=2032605#c7

PR #785, opened 2026-04-23
Bug 2034487 - Initialize IPProtectionService on startup when feature is enabled but cache state is empty

jporter-dev:
### Description 

In addition to the Access Connector icon not being displayed when launching with a fresh profile, the feature is entirely broken on first run. This is due to a conditional that was added in `IPPStartupCache.sys.mjs` which prevents the `IPProtectionService` from being initialized.

Bugzilla: Bug-2034487

---

* Depends on: #784

---

### Testing 

**Steps to verify changes:**

1. Start with a fresh profile
2. Open the browser
3. Navigate to a protected page
4. Observe an inactive Access Connector

**Expected result:**

Access connector is visible and active

### DIFF
--- a/toolkit/components/ipprotection/IPPStartupCache.sys.mjs
+++ b/toolkit/components/ipprotection/IPPStartupCache.sys.mjs
@@ -113,7 +113,9 @@ class IPPStartupCacheSingleton {
     this.#stateFromCache = null;
 
     if (
-      lazy.IPProtectionService.state !== lazy.IPProtectionStates.UNINITIALIZED
+      lazy.IPProtectionService.state !==
+        lazy.IPProtectionStates.UNINITIALIZED ||
+      lazy.IPProtectionService.featureEnabled
     ) {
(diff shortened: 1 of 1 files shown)


lissyx:
> ### Description
> 
> In addition to the Access Connector icon not being displayed when launching with a fresh profile, the feature is entirely broken on first run. This is due to a conditional that was added in `IPPStartupCache.sys.mjs` which prevents the `IPProtectionService` from being initialized.

For tracking and later archeology, can you please identify the bug that regressed this and put it in `regressed by` field on bugzilla ? It looks like you already tracked what regressed this so you already know it ?

lissyx (approved):
Let's make sure we file a follow up (blocking the MVP meta bug) if there is no existing bug to cover some testing within our marionnette enterprise suite, there is enough specific behavior of us to justify it

lissyx:
No need to uplift: https://bugzilla.mozilla.org/show_bug.cgi?id=2034487#c6

PR #788, opened 2026-04-24
Bug 2034591 - Enterprise: make sure to properly react to policy changes for Access Connectors

lissyx:
Starting the browser with policy disabled and then live enable, or starting with the policy enabled, then live disable and then live enable again would result in the Access Connector feature not working anymore.

Second set of steps is handled by making sure isSignedIn() always returns true on enterprise builds.

First set of steps is handled by making sure enterprise builds do not rely on remote settings serverlist but just rely on prefs, and then also making sure there is no race condition between the values of preferences that controls the featuure enabling and the server list: this last one is getting modified at the same time but the code checks for the IPProtectionServerlist's ingestion of the pref, that delays things a bit. Rely on its event to ensure things are processed correctly.

### Description 

Bugzilla: Bug-2034591

---

* Depends on:

---

### Screenshots 

---

### Testing 

**Steps to verify changes:**
1. Disable policy on staging server
2. Start browser
3. Check Access Connector is disabled
4. Enable policy on server
5. Check Access Connector is enabled
6. Disable policy again
7. Check Access Connector is disabled
8. Enable policy again
9. Check Access Connector is enabled

### DIFF
--- a/toolkit/components/ipprotection/IPPAutoStart.sys.mjs
+++ b/toolkit/components/ipprotection/IPPAutoStart.sys.mjs
@@ -30,6 +30,7 @@ const AUTOSTART_PREF = "browser.ipProtection.autoStartEnabled";
  */
 class IPPAutoStartSingleton {
   #shouldStartWhenReady = false;
+  #handleListChanged = null;
 
   constructor() {
     XPCOMUtils.defineLazyPreferenceGetter(
@@ -52,6 +53,22 @@ class IPPAutoStartSingleton {
       AUTOSTART_FEATURE_ENABLE_PREF,
       false
     );
+
+    this.#handleListChanged = () => this.init();
+    lazy.IPProtectionServerlist.addEventListener(
+      "IPProtectionServerlist:ListChanged",
+      this.#handleListChanged
+    );
+  }
+
+  destroy() {
+    if (this.#handleListChanged) {
+      lazy.IPProtectionServerlist.removeEventListener(
+        "IPProtectionServerlist:ListChanged",
+        this.#handleListChanged
+      );
+      this.#handleListChanged = null;
+    }
   }
 
   init() {
--- a/toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
+++ b/toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
@@ -437,6 +437,9 @@ export class PrefServerList extends IPProtectionServerlistBase {
     );
   }
   static get prefValue() {
+    if (!PrefServerList.hasPrefValue) {
+      return null;
+    }
     try {
       const value = Services.prefs.getStringPref(this.PREF_NAME);
       return JSON.parse(value);
--- a/toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
+++ b/toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
 /**
  * This file contains functions that work on top of the RemoteSettings
  * Bucket for the IP Protection server list.
@@ -451,6 +453,9 @@ export class PrefServerList extends IPProtectionServerlistBase {
  * @returns {IPProtectionServerlistBase} - The appropriate serverlist implementation.
  */
 export function IPProtectionServerlistFactory() {
+  if (AppConstants.MOZ_ENTERPRISE) {
+    return new PrefServerList();
+  }
   return PrefServerList.hasPrefValue
     ? new PrefServerList()
     : new RemoteSettingsServerlist();
--- a/toolkit/components/ipprotection/fxa/IPPSignInWatcher.sys.mjs
+++ b/toolkit/components/ipprotection/fxa/IPPSignInWatcher.sys.mjs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
 const lazy = {};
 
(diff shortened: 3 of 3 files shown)


toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -451,6 +453,9 @@ export class PrefServerList extends IPProtectionServerlistBase {
  * @returns {IPProtectionServerlistBase} - The appropriate serverlist implementation.
  */
 export function IPProtectionServerlistFactory() {
+  if (AppConstants.MOZ_ENTERPRISE) {
+    return new PrefServerList();
+  }
```
lissyx, line 461: I am not sure if we want to only rely on preferences pushed  by policy on enterprise builds or if relying on remote settings is OK?

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
lissyx, line 424: I think this should be fixed to verify the list has changed, right now it pushed events constantly, but this may be a follow up?

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
```diff
@@ -52,6 +52,8 @@ class IPPAutoStartSingleton {
       AUTOSTART_FEATURE_ENABLE_PREF,
       false
     );
+
+    lazy.IPProtectionServerlist.addEventListener("IPProtectionServerlist:ListChanged", this.init.bind(this));
```
lissyx: maybe we want to do it a bit nicer

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
jporter-dev, line 94: Do we need to remove the event listener here too? May need to define `#handleListChanged = () => this.init()` on the class to do this.

```suggestion

      lazy.IPProtectionServerlist.removeEventListener(
        "IPProtectionServerlist:ListChanged",
        this.#handleListChanged
      );
```

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -414,10 +417,12 @@ export class PrefServerList extends IPProtectionServerlistBase {
     );
   }
   maybeFetchList(_forceUpdate = false) {
-    this.__list = IPProtectionServerlistBase.dataToList(
-      PrefServerList.prefValue
-    );
-    this.dispatchEvent(new Event(LIST_CHANGED_EVENT));
+    const newListValue = PrefServerList.prefValue;
+    if (newListValue != this._previousListValue) {
```
jporter-dev: This always evaluates to true since we're comparing two JSON objects. Maybe we can compare the raw string values instead using the string pref instead of PrefServerList.prefValue since that converts to JSON.

lissyx: > This always evaluates to true since we're comparing two JSON objects. Maybe we can compare the raw string values instead using the string pref instead of PrefServerList.prefValue since that converts to JSON.

Ah my bad, I missed that it returns an object, I thought it was returning the string

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
lissyx: we can, I was not sure about the approach in itself

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
jporter-dev: I'm also not sure on this one but it seems like this code uses `#` to prefix private vars instead of `_`. I could be wrong, maybe `#` signifies something else. Couldn't find detail on it in the style guide.

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
jporter-dev: I guess this is called in the constructor and not `init()` so maybe we dont need to remove it?

lissyx: most of the time that kind of thing ends up biting in back with something leaking :p

lissyx: Ah but you're right I think.

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
lissyx: you're right, I'll fix

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
lissyx: Verified this with romain, it's what we want

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
```diff
@@ -52,6 +53,22 @@ class IPPAutoStartSingleton {
...
+    this.#handleListChanged = () => this.init();
+    lazy.IPProtectionServerlist.addEventListener(
+      "IPProtectionServerlist:ListChanged",
+      this.#handleListChanged
+    );
+  }
+
+  destroy() {
+    if (this.#handleListChanged) {
+      lazy.IPProtectionServerlist.addEventListener(
```
jporter-dev:
```suggestion
      lazy.IPProtectionServerlist.removeEventListener(
```

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -414,13 +417,29 @@ export class PrefServerList extends IPProtectionServerlistBase {
...
+      this.__list = IPProtectionServerlistBase.dataToList(prefValue);
+      this.dispatchEvent(new Event(LIST_CHANGED_EVENT));
+    }
     return Promise.resolve();
   }
 
+  getPrefValueIfNew() {
+    try {
+      const value = PrefServerList.prefValue;
+      if (value != this.#previousValue) {
```
jporter-dev: nit: prefer strict inequality `!==`

```suggestion
      if (value !== this.#previousValue) {
```

jporter-dev (changes requested):

jporter-dev (approved):

PR #789, opened 2026-04-24
Bug 2034591 - Enterprise: make sure to properly react to policy changes for Access Connectors (beta uplift)

lissyx:
### Description 

Bugzilla: Bug-2034591

Beta uplift with specific changes

---

* Depends on: #788 since it is an uplift

### DIFF
--- a/toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
+++ b/toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
 /**
  * This file contains functions that work on top of the RemoteSettings
  * Bucket for the IP Protection server list.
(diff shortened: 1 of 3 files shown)


toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -371,9 +380,12 @@ export class PrefServerList extends IPProtectionServerlistBase {
     );
   }
   maybeFetchList(_forceUpdate = false) {
-    this.__list = IPProtectionServerlistBase.dataToList(
-      PrefServerList.prefValue
-    );
+    const newListValue = PrefServerList.prefValue;
+    if (newListValue != this._previousListValue) {
```
jporter-dev: This always evaluates to `true` since we're comparing two JSON objects. Maybe we can compare the raw string values instead using the string pref instead of `PrefServerList.prefValue` since that converts to JSON.

jporter-dev: Made this comment in the wrong PR 😅 moving it to the other one since thats the source

lissyx: will be fixed

jporter-dev (approved):

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -391,6 +400,9 @@ export class PrefServerList extends IPProtectionServerlistBase {
     );
   }
   static get prefValue() {
+    if (!PrefServerList.hasPrefValue) {
```
jporter-dev, line 403: Cherry picked this to the beta uplift as well since the error was logged in the browser console. No test failures since the addition of those tests was not uplifted here.

PR #790, opened 2026-04-24
Bug 2027006,2036743: Fetch remote policies on Felt UI before launching Firefox

victhorlopez:
### Description 

Bugzilla: Bug-2027006

 - Moves remote policy fetching from Firefox startup to the Felt UI login flow, so policies are available before Firefox launches, and we can avoid extra startups in case of failures                                                                      
  - Fetches user info and remote policies in parallel after SSO authentication completes
  - Passes policies to Firefox via a new IPC message
  -  If either fetch fails, signs the user out, clears SSO cookies, signouts the user, and resets the Felt UI to the login page with a connection error 

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [x] Added tests
- I put a throw with a counter (to make it fail only the first time), and see how the sso works but fails on fetch

### DIFF
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -296,7 +296,7 @@ this.felt = class extends ExtensionAPI {
 
       case "FeltParent:FirefoxLaunchFailure": {
         Services.felt.makeBackgroundProcess(false);
-        this.showWindow("felt-browser-error-launch-failure");
+        this.resetToLoginPage("felt-browser-error-launch-failure");
         break;
       }
 
@@ -368,6 +368,14 @@ this.felt = class extends ExtensionAPI {
     this._win.close();
     this._win = null;
     this._winObserver = null;
+  }
+
+  resetToLoginPage(errorMessage = "") {
+    if (!this._win) {
+      this.showWindow(errorMessage);
+      return;
+    }
+    this._win.resetToLoginPage(errorMessage);
   }
 
   showWindow(errorMessage = "") {
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -437,6 +437,16 @@ export class FeltProcessParent extends JSProcessActorParent {
     }
   }
 
+  _handleFirefoxLaunchFailure() {
+    Services.felt.clearTokens();
+    Services.cookies.removeCookiesWithOriginAttributes(
+      JSON.stringify({
+        privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID,
+      })
+    );
+    Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure");
+  }
+
   async startFirefox(startReason, ssoCollectedCookies = []) {
     this.restartReported = false;
     this.logoutReported = false;
@@ -445,6 +455,17 @@ export class FeltProcessParent extends JSProcessActorParent {
     this.extensionReady = false;
     resetFeltFirefoxWindowReady();
     gFeltFirefoxReadyNotified = false;
+
+    let initialPoliciesPayload;
+    try {
+      initialPoliciesPayload = JSON.stringify(
+        await lazy.ConsoleClient.getRemotePolicies()
+      );
+    } catch (err) {
+      lazy.log.error(`Failed to fetch remote policies: ${err}`);
+      this._handleFirefoxLaunchFailure();
+      return;
+    }
 
     // There is no message being sent to the message listener on restart phases
     // whether it is a requested restart from the browser or from a crash.
@@ -474,6 +495,9 @@ export class FeltProcessParent extends JSProcessActorParent {
         await this._applyFirefoxConfigs();
 
         Services.felt.sendCookies(ssoCollectedCookies);
+
+        Services.felt.sendPolicies(initialPoliciesPayload);
+
         Services.felt.sendReady();
         this.firefoxReady = true;
 
@@ -511,7 +535,7 @@ export class FeltProcessParent extends JSProcessActorParent {
         lazy.log.error(
           `Firefox launch failure (${err.result} / ${err.name}): ${err.message}`
         );
-        Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure");
+        this._handleFirefoxLaunchFailure();
       });
   }
 
--- a/browser/extensions/felt/content/window.js
+++ b/browser/extensions/felt/content/window.js
@@ -99,6 +99,22 @@ const ErrorReport = {
   },
 };
 
+/**
+ * Hides the SSO pane, reveals the email login form, and displays an error if provided.
+ *
+ * @param {string} errorType - CSS class name identifying the error type to display.
+ * @param {string|null} details - Additional error details.
+ * @param {string|null} cause - The underlying cause of the error.
+ */
+function resetToLogin(errorType, details = null, cause = null) {
+  document.querySelector(".felt-login__sso").classList.add("is-hidden");
+  document
+    .querySelector(".felt-login__email-pane")
+    .classList.remove("is-hidden");
+  ErrorReport.update(errorType, details, cause);
+  document.getElementById("browser").setAttribute("src", "about:blank");
+}
+
 async function connectToConsole(email) {
   let posture;
   try {
@@ -199,11 +215,7 @@ async function connectToConsole(email) {
         browser.removeProgressListener(progressListener);
       } catch (_) {}
     }
-    document.querySelector(".felt-login__sso").classList.add("is-hidden");
-    document
-      .querySelector(".felt-login__email-pane")
-      .classList.remove("is-hidden");
-    ErrorReport.update(errorType, details, cause);
+    resetToLogin(errorType, details, cause);
   }
 
   let ssoTimeout = setTimeout(() => {
@@ -545,6 +557,7 @@ window.addEventListener(
   () => {
     setBuildVersion();
     ErrorReport.init();
+    window.resetToLoginPage = resetToLogin;
     lazy.Updates.init(document, ErrorReport);
     setupMarionetteEnvironment();
     setupPopupNotifications();
--- a/browser/extensions/felt/rust/nsIFelt.idl
+++ b/browser/extensions/felt/rust/nsIFelt.idl
@@ -76,6 +76,12 @@ interface nsIFelt : nsISupports {
   void openURL(in ACString url, in long disposition);
 
   [must_use]
+  void sendPolicies(in ACString jsonPayload);
+
+  [must_use]
+  ACString getStartupPolicies();
+
+  [must_use]
   ACString getConsoleUrl();
 
   [must_use]
--- a/browser/extensions/felt/rust/src/client.rs
+++ b/browser/extensions/felt/rust/src/client.rs
@@ -395,6 +395,10 @@ impl FeltClientThread {
                                     trace!("FeltClientThread::felt_client::ipc_loop(): Shutdown");
                                     utils::notify_observers("felt-firefox-shutdown".to_string());
                                 },
+                                Ok(FeltMessage::Policies(json_payload)) => {
+                                    trace!("FeltClientThread::felt_client::ipc_loop(): Policies({} bytes)", json_payload.len());
+                                    let _ = utils::set_startup_policies(json_payload);
+                                },
                                 Ok(FeltMessage::OpenURL((url, disposition, focus_hint))) => {
                                     trace!(
                                         "FeltClientThread::felt_client::ipc_loop(): OpenURL({}, {}, {:?})",
--- a/browser/extensions/felt/rust/src/components.rs
+++ b/browser/extensions/felt/rust/src/components.rs
@@ -22,7 +22,7 @@ use log::{error, trace};
 use crate::message::{FeltMessage, FELT_IPC_VERSION};
 #[cfg(target_os = "linux")]
 use crate::utils;
-use crate::utils::{Tokens, CONSOLE_URL, TOKENS, TOKEN_EXPIRY_SKEW};
+use crate::utils::{get_startup_policies, Tokens, CONSOLE_URL, TOKENS, TOKEN_EXPIRY_SKEW};
 
 #[xpcom(implement(nsIFelt), atomic)]
 pub struct FeltXPCOM {
@@ -253,6 +253,24 @@ impl FeltXPCOM {
         } else {
             trace!("FeltXPCOM::SendExtensionReady: not in browser, ignoring");
             NS_OK
+        }
+    }
+
+    fn SendPolicies(&self, json_payload: *const nsACString) -> nserror::nsresult {
+        let payload = unsafe { (*json_payload).to_string() };
+        trace!("FeltXPCOM::SendPolicies");
+        self.send(FeltMessage::Policies(payload))
+    }
+
+    fn GetStartupPolicies(&self, retval: *mut nsACString) -> nserror::nsresult {
+        match get_startup_policies() {
+            Ok(policies) => {
+                unsafe {
+                    (*retval).assign(policies.as_str());
+                }
+                NS_OK
+            }
+            Err(()) => NS_ERROR_FAILURE,
         }
     }
 
--- a/browser/extensions/felt/rust/src/utils.rs
+++ b/browser/extensions/felt/rust/src/utils.rs
@@ -5,12 +5,12 @@
 use nserror::NS_OK;
 use nsstring::nsCString;
 use serde::{Deserialize, Serialize};
-use std::sync::{Arc, LazyLock, OnceLock, RwLock};
+use std::sync::{Arc, LazyLock, Mutex, OnceLock, RwLock};
 use std::{ffi::CString, future::Future};
 use xpcom::interfaces::{nsICookie, nsICookieManager, nsIObserverService, nsIPrefBranch};
 use xpcom::RefPtr;
 
-use log::trace;
+use log::{trace, warn};
 #[cfg(target_os = "linux")]
 use std::os::raw::c_char;
 
@@ -62,6 +62,29 @@ pub static TOKEN_EXPIRY_SKEW: i64 = 5 * 60;
 
 pub static TOKENS: LazyLock<Arc<RwLock<Tokens>>> =
     LazyLock::new(|| Arc::new(RwLock::new(Default::default())));
+pub static INITIAL_POLICIES: OnceLock<Mutex<Option<String>>> = OnceLock::new();
+
+pub fn get_startup_policies() -> Result<String, ()> {
+    let mutex = INITIAL_POLICIES
+        .get()
+        .ok_or(())
+        .inspect_err(|_| warn!("Startup policies not initialized"))?;
+    mutex
+        .lock()
+        .unwrap()
+        .take()
+        .ok_or(())
+        .inspect_err(|_| warn!("Startup policies already taken"))
+}
+
+pub fn set_startup_policies(policies: String) -> Result<(), ()> {
+    INITIAL_POLICIES
+        .set(Mutex::new(Some(policies)))
+        .map_err(|_| {
+            warn!("Startup policies already set");
+        })
+}
+
 pub static CONSOLE_URL: OnceLock<Arc<String>> = OnceLock::new();
 
 pub fn inject_one_cookie(cookie: nsICookieWrapper) {
--- a/testing/enterprise/felt_tests.py
+++ b/testing/enterprise/felt_tests.py
@@ -117,6 +117,11 @@ class LocalHttpRequestHandler(BaseHTTPRequestHandler):
 
 
 class SsoHttpHandler(LocalHttpRequestHandler):
+    def has_session_cookie(self):
+        cookies = self.headers.get("Cookie", "")
+        expected = f"{self.server.cookie_name.value}={self.server.cookie_value.value}"
+        return expected in cookies
+
     def do_GET(self):
         print("GET", self.path)
         m = None
@@ -126,6 +131,19 @@ class SsoHttpHandler(LocalHttpRequestHandler):
         print("path: ", path)
 
         if path == "/sso_url":
+            # Simulate real SSO behavior: if the session cookie from a
+            # previous login is present, skip the login form and redirect
+            # straight to the callback.
+            if self.has_session_cookie():
+                location = (
+                    f"http://localhost:{self.server.console_port}/sso/callback?foo"
+                )
+                self.send_response(302, "Found")
+                self.send_header("Location", location)
+                self.send_header("Content-Length", "0")
+                self.end_headers()
+                return
+
             # Dummy sso login page
             m = """
 <html>
@@ -218,7 +236,9 @@ class ConsoleHttpHandler(LocalHttpRequestHandler):
             m = json.dumps({
                 "learn_more_url": firefox_config["learn_more_url"]["pref_value"],
                 "company_logo_url": "",
-                "policies": {"polling_frequency": 500},
+                "policies": {
+                    "polling_frequency": self.server.config_polling_frequency.value
+                },
                 "services": {
                     "push_url": "",
                     "remote_settings_url": "",
@@ -553,6 +573,7 @@ def serve(
     policy_access_connector=None,
     policies_fail_request=None,
     signout_count=None,
+    config_polling_frequency=None,
     # TODO: Behavior is not yet clearly defined
     # device_posture_reply_forbidden=None,
 ):
@@ -582,6 +603,7 @@ def serve(
         policies_fail_request if policies_fail_request is not None else Value("B", 0)
     )
     httpd.signout_count = signout_count if signout_count is not None else Value("i", 0)
+    httpd.config_polling_frequency = config_polling_frequency
     httpd.serve_updates = False
     httpd.serve_updates_version = ""
     httpd.serve_forced_updates_count = 0
@@ -697,6 +719,7 @@ class FeltTestsBase(ConsoleSSOPortMixin, EnterpriseTestsBase):
         self.policy_access_token = SharedString("")
         self.policy_refresh_token = SharedString("")
         self.signout_count = Value("i", 0)
+        self.config_polling_frequency = Value("i", 500)
 
         self.console_httpd = Process(
             target=serve,
@@ -712,6 +735,7 @@ class FeltTestsBase(ConsoleSSOPortMixin, EnterpriseTestsBase):
                 policy_refresh_token=self.policy_refresh_token,
                 policies_fail_request=self.policies_fail_request,
                 signout_count=self.signout_count,
+                config_polling_frequency=self.config_polling_frequency,
                 # TODO: Behavior is not yet clearly defined
                 # device_posture_reply_forbidden=self.device_posture_reply_forbidden,
             ),
--- a/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
+++ b/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
@@ -10,8 +10,6 @@ ChromeUtils.defineESModuleGetters(lazy, {
   JsonSchemaValidator:
     "resource://gre/modules/components-utils/JsonSchemaValidator.sys.mjs",
   // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
-  EnterpriseHandler: "resource:///modules/enterprise/EnterpriseHandler.sys.mjs",
-  // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
   Policies: "resource:///modules/policies/Policies.sys.mjs",
   WindowsGPOParser: "resource://gre/modules/policies/WindowsGPOParser.sys.mjs",
   macOSPoliciesParser:
@@ -113,7 +111,7 @@ EnterprisePoliciesManager.prototype = {
     }
   },
 
-  async _initialize() {
+  _initialize() {
     this._cleanupPolicies();
 
     const changesHandler = provider => {
@@ -155,7 +153,7 @@ EnterprisePoliciesManager.prototype = {
     this._status = Ci.nsIEnterprisePolicies.INACTIVE;
     Services.prefs.setBoolPref(PREF_POLICIES_APPLIED, false);
 
-    let provider = await this._chooseProvider(changesHandler);
+    let provider = this._chooseProvider(changesHandler);
     if (provider.failed) {
       this._status = Ci.nsIEnterprisePolicies.FAILED;
     }
@@ -166,7 +164,7 @@ EnterprisePoliciesManager.prototype = {
     Glean.policies.isEnterprise.set(this.isEnterprise);
   },
 
-  async _chooseProvider(handler) {
+  _chooseProvider(handler) {
     let platformProvider = null;
     if (AppConstants.platform == "win" && AppConstants.MOZ_SYSTEM_POLICIES) {
       platformProvider = new WindowsGPOPoliciesProvider();
@@ -183,14 +181,6 @@ EnterprisePoliciesManager.prototype = {
     jsonProvider.onPoliciesChanges(handler);
 
     let remoteProvider = RemotePoliciesProvider.createInstance();
-    // Fetch first set of remote policies during the
-    // initialization of the policy engine
-    await remoteProvider.fetchPoliciesOnStartup();
-    if (Services.felt?.isFeltBrowser() && remoteProvider.failed) {
-      // bug 2027006 will move the fetching of policies to felt
-      // and not shutdown will be needed then
-      lazy.EnterpriseHandler.initiateShutdown();
-    }
     remoteProvider.onPoliciesChanges(handler);
 
     if (platformProvider && platformProvider.hasPolicies) {
@@ -446,12 +436,7 @@ EnterprisePoliciesManager.prototype = {
 
     switch (topic) {
       case "policies-startup": {
-        const initializedPromise = this._initialize();
-        // _initialize() does async work (fetching remote policies).
-        // We spin a nested event loop until the promise resolves so
-        // this observer doesn't return before initialization completes.
-        // This keeps startup behavior effectively synchronous.
-        this.spinResolve(initializedPromise);
+        this._initialize();
         this._runPoliciesCallbacks("onBeforeAddons");
         break;
       }
@@ -501,44 +486,6 @@ EnterprisePoliciesManager.prototype = {
         );
 
         break;
-    }
-  },
-
-  /**
-   * Spin the event loop until the passed promise resolves.
-   *
-   * This is used to await the response when fetching remote
-   * policies during the initialization of the policy engine.
-   *
-   * @param {Promise} promise
-   * @returns {any} Result of the resolved promise
-   */
-  spinResolve(promise) {
-    if (!(promise instanceof Promise)) {
-      return promise;
-    }
-    let done = false;
-    let result = null;
-    let error = null;
-    promise
-      .catch(e => {
-        error = e;
-      })
-      .then(r => {
-        result = r;
-        done = true;
-      });
-
-    Services.tm.spinEventLoopUntil(
-      "EnterprisePoliciesManager.sys.mjs:_initialize",
-      () => done
-    );
-    if (!done) {
-      throw new Error("Forcefully exited event loop.");
-    } else if (error) {
-      throw error;
-    } else {
-      return result;
     }
   },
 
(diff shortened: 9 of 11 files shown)


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -718,9 +727,29 @@ export class FeltProcessParent extends JSProcessActorParent {
...
-          // TODO: Bug 2003001 - Pass user info from Felt to Firefox to avoid network request on startup
-          this.loggedInUserInfo =
-            await lazy.ConsoleClient.getLoggedInUserInfo();
+          let userInfo, remotePolicies;
+          try {
+            // TODO: Bug 2003001 - Pass user info from Felt to Firefox to avoid network request on startup
+            [userInfo, remotePolicies] = await Promise.all([
+              lazy.ConsoleClient.getLoggedInUserInfo(),
+              lazy.ConsoleClient.getRemotePolicies(),
+            ]);
```
victhorlopez: should we add here a check for 
```remotePolicies == "" or == "{}"```

and count it as invalid (i.e throw), otherwise we will open firefox and throw there

File: browser/extensions/felt/content/FeltProcessParent.sys.mjs [hunk 2]
```diff
@@ -718,9 +727,29 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+          } catch (e) {
+            lazy.log.error(`Failed to fetch startup data: ${e}`);
+            await lazy.ConsoleClient.postSignout().catch(err =>
```
victhorlopez: in theory in this step, we actually managed to log in but either getRemotePolicies or getLoggedInUserInfo failed.  I assumed we want to clean up everything?

browser/extensions/felt/content/window.js
```diff
@@ -199,6 +199,14 @@ async function connectToConsole(email) {
     ErrorReport.update(errorType, details, cause);
   }
 
+  window.addEventListener(
```
victhorlopez: I had different ideas how to do this, basically I wanted to reuse 
[this](https://searchfox.org/enterprise-main/rev/0eb957a7186719a0f626a6827bc728347b5b86af/browser/extensions/felt/content/window.js#187)

browser/extensions/felt/rust/src/utils.rs
```diff
@@ -62,6 +62,7 @@ pub static TOKEN_EXPIRY_SKEW: i64 = 5 * 60;
 
 pub static TOKENS: LazyLock<Arc<RwLock<Tokens>>> =
     LazyLock::new(|| Arc::new(RwLock::new(Default::default())));
+pub static POLICIES: OnceLock<Arc<String>> = OnceLock::new();
```
victhorlopez: I assumed we write only once?

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -712,6 +712,13 @@ export const ConsoleClient = {
     Services.felt.clearTokens(allowMirror);
   },
 
+  /**
+   * Post a signout request to the console.
+   */
+  async postSignout() {
+    return this._post(this._paths.SIGNOUT);
+  },
```
lissyx: There is already `signoutUser` https://searchfox.org/enterprise-main/rev/60774249204178398f72d33d03a1134470c20d03/browser/components/enterprise/modules/ConsoleClient.sys.mjs#715-739

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -389,6 +391,9 @@ export class FeltProcessParent extends JSProcessActorParent {
     resetFeltFirefoxWindowReady();
     gFeltFirefoxReadyNotified = false;
 
+    this.lastRemotePolicies = remotePolicies;
```
lissyx: Since we persist here, do we want to persist also on disk? Would it make any sense? @1rneh what is your opinion?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: What is the justification for the `try` block?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: I defer to @1rneh for that

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -718,9 +727,29 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+            ]);
+          } catch (e) {
+            lazy.log.error(`Failed to fetch startup data: ${e}`);
+            await lazy.ConsoleClient.postSignout().catch(err =>
+              lazy.log.error(`Failed to post signout: ${err}`)
+            );
+            lazy.ConsoleClient.clearTokenData();
+            // Clear SSO cookies so the provider does not auto-authenticate
+            // on the next login attempt.
+            this.removeAllCookies();
```
lissyx: I think we should do that earlier.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -718,9 +727,29 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+            Services.cpmm.sendAsyncMessage("FeltParent:ShowLogin", {
+              reason: "felt-browser-error-connection",
```
lissyx: Likely we want a more focused error message, but this is UX we should discuss with Matthew ?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -718,9 +727,29 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+            });
+            break;
+          }
+
+          this.loggedInUserInfo = userInfo;
```
lissyx: Why do we add this now?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -718,9 +727,29 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+            await lazy.ConsoleClient.postSignout().catch(err =>
+              lazy.log.error(`Failed to post signout: ${err}`)
+            );
+            lazy.ConsoleClient.clearTokenData();
+            // Clear SSO cookies so the provider does not auto-authenticate
+            // on the next login attempt.
+            this.removeAllCookies();
+            Services.cpmm.sendAsyncMessage("FeltParent:ShowLogin", {
+              reason: "felt-browser-error-connection",
+            });
```
lissyx: That makes us having to round-trip. Can't we leverage JS Actors here?

This actor here is created via `FeltWindowChild` maybe we should just communicate back directly to that `FeltWindowChild` instead of going over the asyncMessage and then issue a custom event?

https://searchfox.org/enterprise-main/rev/60774249204178398f72d33d03a1134470c20d03/browser/extensions/felt/content/FeltWindowChild.sys.mjs#22-24

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -1065,13 +1058,15 @@ class RemotePoliciesProvider {
 
     let res;
     try {
-      res = await lazy.ConsoleClient.getRemotePolicies();
+      if (Services.felt?.isFeltBrowser()) {
+        res = JSON.parse(Services.felt.getStartupPolicies());
+      }
```
lissyx: WOo, that is rather intrusive, I dont know how it collides with the work @1rneh is doing on policies

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: those are network requests, if they throw an exception you end up with a "blank page" in the sso

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: This is related to your question about what we type of exceptions we can get.. 401 should be handled already(would be weird to get it in this step but you never know), so I would expect here to get a no connection error/ server errors etc (or no policies setup on the console).

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: I would need to investigate, but the problem I had is the resetToLoginPage is in closure

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: Because this changed in here
```
         - this.loggedInUserInfo =
         -  await lazy.ConsoleClient.getLoggedInUserInfo();
         + let userInfo, remotePolicies;
         + try {
         +  // TODO: Bug 2003001 - Pass user info from Felt to Firefox to avoid network request on startup
         +   [userInfo, remotePolicies] = await Promise.all([
         +    lazy.ConsoleClient.getLoggedInUserInfo(),
         +     lazy.ConsoleClient.getRemotePolicies(),
         +  ]);
```

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: just to add context on why I am saving this.. it is because if firefox crashes, we need to restart it instantly ( and it uses another path), but.... now I have another question, is this well behaved if the policies changed during runtime? Maybe 1rneh had another idea, to populate also the policies storage from the browser and populate the storage, cause otherwise we might start with old policies?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Also I'm wondering if we should not handle that just like others in https://searchfox.org/enterprise-main/rev/753b9cbaa2529ceb14c22f0ec69fd5f81294e7a4/browser/extensions/felt/api.js#296-300

browser/components/enterprise/modules/ConsoleClient.sys.mjs
victhorlopez: I put it in a different function because there was an if there preventing felt_ui, but okay, I just put it now together

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: Can you elaborate what would be optimal?

As of now here the cookies are deleted after trying to signout, we did successfully SSO but something failed in whoami/policies

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: I did some changes based on your comment, I hope I went in the right direction.

Regarding like the others, the other they "reset the window", in this case the form is always visible, so we should not call show window cause it opens another window.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: thanks, your comment was helpful to understand better the flow

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: technically since it's private browsing, even quickly after the `<browser>` is dismissed it's already gone, so I think reloading the window would already be enough

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: I think the actor solution is better

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: Let me know what would you do in here, (also to have an optimal user experience), cause I tried some things and did not manage to make it work properly until I cleared the cookies (i.e I had problems if I was trying to login a second time)

victhorlopez:
@github-copilot review

browser/extensions/felt/rust/src/client.rs
```diff
@@ -410,6 +410,17 @@ impl FeltClientThread {
...
                                 }
+                                Ok(FeltMessage::Policies(json_payload)) => {
+                                    trace!("FeltClientThread::felt_client::ipc_loop(): Policies({})", json_payload);
+                                    match utils::POLICIES.set(Arc::new(json_payload)) {
+                                        Ok(()) => {
+                                            trace!("FeltClientThread::felt_client::ipc_loop(): Policies stored");
+                                        }
+                                        Err(_) => {
+                                            trace!("FeltClientThread::felt_client::ipc_loop(): Policies already set, ignoring");
+                                        }
```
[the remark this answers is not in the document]
fiji-flo: I'd argue this is an `error!` and we should maybe even crash. This must not happen.

browser/extensions/felt/rust/src/utils.rs
fiji-flo: Could we use 
```rust
pub static INITIAL_POLICIES: LazyLock<Arc<Mutex<InitialPolicies>>> = LazyLock::new(|| Mutex::new(Arc::new(Default::default())));
```
here.

With 

```rust
#[derive(Default)]
pub enum InitialPolicies {
  Uninitialized,
  Policies(String),
  Taken,
}
```

And then consume it when we read it. That allows us to output a warning on consecutive reads.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
fiji-flo: We can fetch the policies here in stead of the event handler. That would solve this. We really should not apply old policies.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -1080,20 +1030,26 @@ class RemotePoliciesProvider {
     }
   }
 
-  async fetchPoliciesOnStartup() {
+  fetchPoliciesOnStartup() {
```
1rneh: As this is not a fetch anymore can we rename the method to something like `loadRemotePolicies`?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -1080,20 +1030,26 @@ class RemotePoliciesProvider {
...
     }
 
     let res;
     try {
-      res = await lazy.ConsoleClient.getRemotePolicies();
+      if (
+        AppConstants.MOZ_ENTERPRISE &&
+        !AppConstants.MOZ_BACKGROUNDTASKS &&
+        Services.felt?.isFeltBrowser()
+      ) {
```
1rneh: `EnterprisePoliciesParent` is currently not clean on excluding enterprise code. I'm taking care of that in my policy engine refactor, as this enterprise-specific code path will need to be excluded earlier.

I'm still curious why we'd need `AppConstants.MOZ_BACKGROUNDTASKS`?

testing/enterprise/felt_tests.py
```diff
@@ -126,6 +131,19 @@ def do_GET(self):
...
+            if self.has_session_cookie():
+                location = (
+                    f"http://localhost:{self.server.console_port}/sso/callback?foo"
+                )
+                self.send_response(302, "Found")
+                self.send_header("Location", location)
+                self.send_header("Content-Length", "0")
+                self.end_headers()
+                return
+
```
1rneh, line 146: Actually, felt shouldn't allow preserving the sso session cookie as it's loaded in a private window.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -446,6 +446,25 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      );
+      Services.felt.clearTokens();
+      Services.cookies.removeCookiesWithOriginAttributes(
+        JSON.stringify({
+          privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID,
+        })
+      );
+      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure");
+      return;
+    }
```
1rneh: It makes sense to clear the tokens and remove any cookies stored for the private window since the purpose of sending the message`FeltParent:FirefoxLaunchFailure` is to reset to the login page with a clean state. But I noticed that we don't do that [for every launch failure](https://searchfox.org/enterprise-main/rev/9b9a912beab0f98702e5828fbad230b40ffac88b/browser/extensions/felt/content/FeltProcessParent.sys.mjs#510-515), which I believe is not intentional? Can we centralize the operations and e.g. in a new method `handleBrowserLaunchFailure` and have the two places where launch failures are expected call that method instead?

browser/extensions/felt/content/window.js
```diff
@@ -99,6 +99,28 @@ const ErrorReport = {
...
+  ErrorReport.update(errorType, details, cause);
+}
+
+// Resets the window to the login page, clearing the SSO browser content.
+// Called from api.js to recover after a launch failure without reopening the window.
+function resetToLogin(errorType, details = null, cause = null) {
+  showLoginForm(errorType, details, cause);
+  document.getElementById("browser").setAttribute("src", "about:blank");
+}
+
```
1rneh, line 117: I think we'll always want to to reset the embedded browser when restoring the login form so the two methods can be unified to one.

browser/extensions/felt/rust/src/utils.rs
```diff
@@ -62,6 +62,16 @@ pub static TOKEN_EXPIRY_SKEW: i64 = 5 * 60;
 
 pub static TOKENS: LazyLock<Arc<RwLock<Tokens>>> =
     LazyLock::new(|| Arc::new(RwLock::new(Default::default())));
+#[derive(Default)]
+pub enum InitialPolicies {
+    #[default]
+    Uninitialized,
+    Policies(String),
+    Taken,
+}
```
1rneh: @fiji-flo I saw you suggested this enum. I'm not sure it's necessary, since we have a deterministic ordering for each browser lifecycle:
1. The felt side launches the browser; browser startup [blocks on `firefox_felt_is_startup_complete()`](https://searchfox.org/enterprise-main/rev/9b9a912beab0f98702e5828fbad230b40ffac88b/toolkit/xre/nsAppRunner.cpp#5888-5890).
2. The felt side sends config (including policies).
3. The felt side sends `StartupReady`, which release the `startup_ready` barrier.
4. The startup resumes (the policies message is already drained by that time) and the policy engine reads the policies exactly once.

So the policies shouldn't be read before they are initialized nor taken at any point... What do you think?

1rneh (changes requested):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
victhorlopez: This is for [this](https://treeherder.mozilla.org/logviewer?job_id=567130632&repo=enterprise-firefox-pr&task=c30q6hF4SWm8GuRHtnlJJg.0&lineNumber=15949)
Because it gets instantiated [in here](https://searchfox.org/firefox-main/rev/a249a43995d7634e6e4a56bfd6496039031648bc/toolkit/xre/nsXREDirProvider.cpp#651)

It was [discussed in here](https://github.com/mozilla/enterprise-firefox/pull/833), basically I added the backgroundtasks for the test.

I am not sure if your refactor is changing that part somehow.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: It depends, it was a question for you originally(but got lost with push force), in my implementation I did clear the cookies, and expect that user then needs to reauth. In some FirefoxLaunchFailure maybe we want another behaviour. The question is do we want always that the user needs to relog if we hit this?

browser/extensions/felt/content/window.js
victhorlopez: okay, I did it like this because the internal closure was not doing this line, can unify if you are 100% sure

browser/extensions/felt/rust/src/utils.rs
victhorlopez: I do see value in having the tri-state to be able to catch improper usage, even if on paper it looks this would never happen.

testing/enterprise/felt_tests.py
victhorlopez: As far as I understand as long as we do not close the window, the cookie will be there if we were able to login once and we do not close it. I added this based on manually testing and seeing different behaviour on the test vs fastly.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
victhorlopez: Can you also comment on the following,

This code would not go in the catch for {} or "", do you want to be able to start firefox with that or should we handle it (I know it was a point recently)

browser/extensions/felt/rust/src/utils.rs
victhorlopez: waiting for @fiji-flo to clarify if this is a hard requirement

fiji-flo: I'd still do this. Shouldn't happen does not mean can not happen.

I thinks that's fine. In an ideal state this would be a `OnceRead` (which does not exist). But that needs quite some Rust skills. But the enum makes it explicit what happens.

fiji-flo: Feel free to run it with an LLM. Maybe there's a easy solution to make it so it's asserted by Rust that it's only written once and read once.

victhorlopez: I had a look, writing this if it is only for this use case might be overkill
```
pub struct OnceRead<T> {
      value: Mutex<Option<T>>,
      read_count: AtomicUsize,
  }
 ```
 
 This is the same I did but without being able to distinguish between uninitialized and consumed 
 ```
LazyLock<Mutex<Option<String>>>
 ```
 
 So I would leave it as is right now unless we see value in changing this

File: browser/extensions/felt/content/FeltProcessParent.sys.mjs (hunk 6 above)
Comment (victhorlopez) [reply browser/extensions/felt/content/FeltProcessParent.sys.mjs]: Solved, we are no longer storing (as Florian suggested), so we do not need to deal with policies out of date. We fetch always fresh

File: browser/extensions/felt/content/FeltProcessParent.sys.mjs (hunk 8 above)
Comment (victhorlopez) [reply browser/extensions/felt/content/FeltProcessParent.sys.mjs]: As this call was moved to the firefox startup, I went back to the original approach.

File: toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs (hunk 14 above)
Comment (victhorlopez) [reply toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs]: I did some testing and changed this, cause `AppConstants.MOZ_BACKGROUNDTASKS` is true in my computer when running ./mach run??
I am using this, but would like to see then the mochitest to see if this works. 
`!lazy.BackgroundTasks?.isBackgroundTaskMode`

I also added some test on the marionette to check the policies are there

Edit: I checked and is just set `AppConstants.MOZ_BACKGROUNDTASKS` if you compile with enabled background tasks, the other one is. if you launch it as background task.

I changed the test a bit, if you comment
```#res = JSON.parse(Services.felt.getStartupPolicies());``` it will fail

File: browser/extensions/felt/rust/src/components.rs [hunk 19]
```diff
@@ -256,6 +258,34 @@ impl FeltXPCOM {
...
 
+    fn SendPolicies(&self, json_payload: *const nsACString) -> nserror::nsresult {
+        let payload = unsafe { (*json_payload).to_string() };
+        trace!("FeltXPCOM::SendPolicies");
+        self.send(FeltMessage::Policies(payload))
+    }
+
+    fn GetStartupPolicies(&self, retval: *mut nsACString) -> nserror::nsresult {
+        let mut guard = INITIAL_POLICIES.lock().unwrap();
+        match std::mem::take(&mut *guard) {
```
fiji-flo: This line sets `INITIAL_POLICIES` to uninitialized. Please use std::mem::replace

victhorlopez: isn't it then reset in 274 and 283? and left unitialiazed for the case `InitialPolicies::Uninitialized`?

edit: okay well, will adapt if that is the correct way to do it

browser/extensions/felt/rust/src/utils.rs
fiji-flo: Sorry, Looking at this again the cleanest way is probably (plus nicer messages):

```rust
static INITIAL_POLICIES: OnceLock<Mutex<Option<String>>> = OnceLock::new();

...

pub fn get_startup_policies() -> Result<String, ()> {
    let mutex = INITIAL_POLICIES
        .get()
        .ok_or(())
        .inspect_err(|_| println!("not initialized"))?;
    mutex
        .lock()
        .unwrap()
        .take()
        .ok_or(())
        .inspect_err(|_| println!("already taken"))
}

pub fn set_startup_policies(policies: String) -> Result<(), ()> {
    INITIAL_POLICIES
        .set(Mutex::new(Some(policies)))
        .map_err(|_| {
            println!("already set");
            ()
        })
}

```

victhorlopez: okay, I will change it then, thanks for the input

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -185,12 +191,7 @@ EnterprisePoliciesManager.prototype = {
     let remoteProvider = RemotePoliciesProvider.createInstance();
     // Fetch first set of remote policies during the
     // initialization of the policy engine
-    await remoteProvider.fetchPoliciesOnStartup();
-    if (Services.felt?.isFeltBrowser() && remoteProvider.failed) {
-      // bug 2027006 will move the fetching of policies to felt
-      // and not shutdown will be needed then
-      lazy.EnterpriseHandler.initiateShutdown();
-    }
+    remoteProvider.loadRemotePolicies();
```
1rneh: Can we load the policies in `RemotePoliciesProvider`'s constructor?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -437,6 +437,19 @@ export class FeltProcessParent extends JSProcessActorParent {
     }
   }
 
+  _handleFirefoxLaunchFailure(err) {
```
1rneh: Can the logging move out of `_handleFirefoxLaunchFailure`? Then we can have a more tailored log message when catching the error + we don't need to pass the error.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Sorry, I'm not able to follow the reasoning on why we need `AppConstants.MOZ_BACKGROUNDTASKS` here.

> This is for [this](https://treeherder.mozilla.org/logviewer?job_id=567130632&repo=enterprise-firefox-pr&task=c30q6hF4SWm8GuRHtnlJJg.0&lineNumber=15949) Because it gets instantiated [in here](https://searchfox.org/firefox-main/rev/a249a43995d7634e6e4a56bfd6496039031648bc/toolkit/xre/nsXREDirProvider.cpp#651)
> 
> It was [discussed in here](https://github.com/mozilla/enterprise-firefox/pull/833), basically I added the backgroundtasks for the test.
> 
> I am not sure if your refactor is changing that part somehow.

What was discussed in those comments is that currently the unexpected extra task manager gets created when spinning the event loop during the policy engine initialization. Fetching remote policies from felt will remove the need to spin the event loop. Maybe I'm missing some context on how that is related to `AppConstants.MOZ_BACKGROUNDTASKS`?

testing/enterprise/felt_tests.py
1rneh: The cookies are removed in `_handleFirefoxLaunchFailure`:
```
Services.cookies.removeCookiesWithOriginAttributes(
      JSON.stringify({
        privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID,
      })
    );
```

1rneh (commented):
Round of review No 1.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
victhorlopez: The `AppConstants.MOZ_BACKGROUNDTASKS ` was changed to `lazy.BackgroundTasks?.isBackgroundTaskMode` cause `AppConstants.MOZ_BACKGROUNDTASKS` just says whether the app was compiled with support for background tasks.

Regarding the general question on why we need that, here is my reasoning. Firefox has a mode to run as background task.

For instance a dummy one ```./mach run -- --backgroundtask success```

The test `toolkit/components/backgroundtasks/tests/browser/browser_xpcom_graph_wait.js` checks which services are instantiated in background mode. My assumption based on the other discussion is our code should not be hit on backgroundmode.

https://firefox-source-docs.mozilla.org/toolkit/components/backgroundtasks/index.html

Just to extend on this, as soon as you do `Services.felt`, it instantiates a singleton.

testing/enterprise/felt_tests.py
victhorlopez: Yes, cookies are being removed from the client side. The changes made here are on the backend server to replicate Fastly's behavior. Specifically, during testing, I noticed that when cookies where not deleted the sso page was just bypassed (even though i was restarting sso browser etc)

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Ohh, thanks for clarifying! I read through #833 and I'm afraid I didn't understand the entire discussion and only addressed half of what was asked. I left a correcting comment there. We'll probably need to revisit #833. We'd likely want to skip initializing the entire policy engine in background mode, not just skip loading the remote policies.

victhorlopez: Removed `lazy.BackgroundTasks?.isBackgroundTaskMode` , I assume we need to put a check in m-c for background task in [here](https://searchfox.org/firefox-main/rev/a249a43995d7634e6e4a56bfd6496039031648bc/toolkit/xre/nsXREDirProvider.cpp#651)

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -441,6 +437,16 @@ export class FeltProcessParent extends JSProcessActorParent {
...
 
+  _handleFirefoxLaunchFailure() {
+    Services.felt.clearTokens();
+    Services.cookies.removeCookiesWithOriginAttributes(
+      JSON.stringify({
+        privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID,
+      })
+    );
+    Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure");
+  }
```
lissyx, line 448: This is not related to fetching the remote policies ? Not directly. Can we move that to a different commit in this PR for ease of review ?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -450,6 +456,17 @@ export class FeltProcessParent extends JSProcessActorParent {
     resetFeltFirefoxWindowReady();
     gFeltFirefoxReadyNotified = false;
 
+    let initialPoliciesPayload;
+    try {
+      initialPoliciesPayload = JSON.stringify(
+        await lazy.ConsoleClient.getRemotePolicies()
+      );
+    } catch (err) {
+      lazy.log.error(`Failed to fetch remote policies: ${err}`);
```
lissyx, line 465: We need to be extremely precise here and ensure any failure ends up is **actionable** logging

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -450,6 +456,17 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      this._handleFirefoxLaunchFailure();
```
lissyx, line 466: We should augment this to make sure we can display correct error message to the user, since this method can handle different cases of failures now

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx, line 477: So if policies fails it is before we hide the window

browser/extensions/felt/content/window.js
```diff
@@ -99,6 +99,22 @@ const ErrorReport = {
...
+ * @param {string|null} cause - The underlying cause of the error.
+ */
+function resetToLogin(errorType, details = null, cause = null) {
+  document.querySelector(".felt-login__sso").classList.add("is-hidden");
+  document
+    .querySelector(".felt-login__email-pane")
+    .classList.remove("is-hidden");
+  ErrorReport.update(errorType, details, cause);
+  document.getElementById("browser").setAttribute("src", "about:blank");
+}
```
lissyx, line 116: We already have a `resetToLoginPage` why adding a new one with a close name?

browser/extensions/felt/api.js
```diff
@@ -296,7 +296,7 @@ this.felt = class extends ExtensionAPI {
 
       case "FeltParent:FirefoxLaunchFailure": {
         Services.felt.makeBackgroundProcess(false);
-        this.showWindow("felt-browser-error-launch-failure");
+        this.resetToLoginPage("felt-browser-error-launch-failure");
```
lissyx, line 299: I'm afraid this change by how it's implemented after may hide broken states

browser/extensions/felt/rust/nsIFelt.idl
```diff
@@ -75,6 +75,12 @@ interface nsIFelt : nsISupports {
   [must_use]
   void openURL(in ACString url, in long disposition);
 
+  [must_use]
+  void sendPolicies(in ACString jsonPayload);
```
lissyx, line 79: Make sure to provide accurate documentation.

browser/extensions/felt/rust/nsIFelt.idl
```diff
@@ -75,6 +75,12 @@ interface nsIFelt : nsISupports {
...
+
+  [must_use]
+  ACString getStartupPolicies();
```
lissyx, line 82: Same

browser/extensions/felt/rust/nsIFelt.idl
lissyx, line 1: While I look at this, can you file a follow up blocking mvp so we can add proper documentation to all content here?

browser/extensions/felt/rust/src/components.rs
```diff
@@ -256,6 +256,24 @@ impl FeltXPCOM {
...
+
+    fn GetStartupPolicies(&self, retval: *mut nsACString) -> nserror::nsresult {
+        match get_startup_policies() {
+            Ok(policies) => {
+                unsafe {
+                    (*retval).assign(policies.as_str());
+                }
+                NS_OK
+            }
+            Err(()) => NS_ERROR_FAILURE,
```
lissyx, line 273: wondering if we should not panic, given we want to anyway block the startup.

browser/extensions/felt/rust/src/components.rs
```diff
@@ -256,6 +256,24 @@ impl FeltXPCOM {
         }
     }
 
+    fn SendPolicies(&self, json_payload: *const nsACString) -> nserror::nsresult {
```
lissyx, line 259: Maybe we should assert it's called from Felt and not browser

browser/extensions/felt/rust/src/components.rs
```diff
@@ -256,6 +256,24 @@ impl FeltXPCOM {
...
+        let payload = unsafe { (*json_payload).to_string() };
+        trace!("FeltXPCOM::SendPolicies");
+        self.send(FeltMessage::Policies(payload))
+    }
+
+    fn GetStartupPolicies(&self, retval: *mut nsACString) -> nserror::nsresult {
```
lissyx, line 265: Similar asserting that it is called from browser and not felt

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx, line 1037: This assumes the code starts after startup policies have been populated rust-side. Maybe we should have a barrier to ensure those were sent over IPC when we reach this area?

fiji-flo:
@gcp We should also send the browser_key

PR #792, opened 2026-04-24
Add Github workflow to automatically create uplift PRs

jonathanmendez:
### Description 

Bugzilla: Bug-TODO
Adds a workflow that:
- triggers when a PR is merged
- checks whether labels `uplift-to-beta` or `uplift-to-release` are present in the PR and if the PR's target branch is one we would uplift from (`enterprise-main` for beta or release, and `enterprise-beta` for release)
- if so, cherry-picks the commit(s) from the merge into the `enterprise-beta`/`enterprise-release` branch(es) and creates uplift PR(s)
   - if the PR was done as a merge, cherry-picks the individual commits
   - if the PR was done as a squash, cherry-picks the squash commit
- if a conflict is encountered during a cherry-pick, the created uplift PR is changed to a Draft, and the original author is added as a reviewer with a comment for resolution instructions

---

### Testing 

TODO??
* [ ] Added tests

### DIFF
--- a/.github/workflows/uplift.yml
+++ b/.github/workflows/uplift.yml
@@ -18,6 +18,7 @@ jobs:
   build-matrix:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
+    permissions: {}
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
@@ -59,7 +60,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
 
@@ -188,15 +189,17 @@ jobs:
         if: steps.uplift.outputs.branch != ''
         env:
           GH_TOKEN: ${{ github.token }}
+          COMMIT_SUMMARY: ${{ steps.uplift.outputs.summary }}
         run: |
           BRANCH="${{ steps.uplift.outputs.branch }}"
 
-          cat <<'EOF' > pr_body.txt
+          DELIM="PR_BODY_$(uuidgen)"
+          cat <<$DELIM > pr_body.txt
           Uplift from #${{ env.PR_NUMBER }}
 
           ## Commits
-          ${{ steps.uplift.outputs.summary }}
-          EOF
+          $COMMIT_SUMMARY
+          $DELIM
 
           PR_URL=$(gh pr create \
             --base "${{ matrix.target }}" \
--- a/.github/workflows/uplift.yml
+++ b/.github/workflows/uplift.yml
@@ -103,7 +103,7 @@ jobs:
             COMMITS=$SHA
           fi
 
-          APPLIED=0
+          HAS_APPLIED_COMMITS=0
 
           for commit in $COMMITS; do
             SHORT=$(git rev-parse --short $commit)
@@ -136,10 +136,10 @@ jobs:
               echo "- [$SHORT]($LINK) $MSG" >> $SUMMARY_FILE
             fi
 
-            APPLIED=1
+            HAS_APPLIED_COMMITS=1
           done
 
-          if [ "$APPLIED" -eq 0 ]; then
+          if [ "$HAS_APPLIED_COMMITS" -eq 0 ]; then
             echo "No new commits applied"
             echo "summary=All commits already exist in $TARGET." >> $GITHUB_OUTPUT
             echo "branch=" >> $GITHUB_OUTPUT
--- a/.github/workflows/uplift.yml
+++ b/.github/workflows/uplift.yml
@@ -191,7 +191,7 @@ jobs:
         run: |
           BRANCH="${{ steps.uplift.outputs.branch }}"
 
-          cat <<EOF > pr_body.txt
+          cat <<'EOF' > pr_body.txt
           Uplift from #${{ env.PR_NUMBER }}
 
           ## Commits
@@ -209,15 +209,15 @@ jobs:
           gh pr edit "$PR_URL" --add-reviewer "mozilla/enterprise-firefox-engineering"
 
           if [ "${{ steps.uplift.outputs.has_conflicts }}" = "true" ]; then
-            cat <<EOF >> comment_body.txt
+            cat <<'EOF' >> comment_body.txt
             ⚠️ Commits marked with conflicts require manual resolution, by rebasing locally to edit the commits or by recreating the branch. **DO NOT** fix up by adding an extra commit, only resolve conflicts directly on the cherry-pick commit(s).
 
             To recreate this branch manually:
-            \`\`\`
+            ```
             git fetch enterprise-firefox ${{ matrix.target }}
             git checkout -b ${{ steps.uplift.outputs.branch }} enterprise-firefox/${{ matrix.target }}
             git cherry-pick ${{ steps.uplift.outputs.commits }}
-            \`\`\`
+            ```
             EOF
 
             gh pr comment "$PR_URL" --body-file comment_body.txt
--- a/.github/workflows/uplift.yml
+++ b/.github/workflows/uplift.yml
@@ -222,7 +222,8 @@ jobs:
 
           gh pr edit "$PR_URL" --add-reviewer "mozilla/enterprise-firefox-engineering"
 
-          # Add original author if conflicts happened, since manual resolution is required
+          # If there are conflicts, add original author and mark PR as draft since manual resolution is required
           if [ "${{ steps.uplift.outputs.has_conflicts }}" = "true" ]; then
             gh pr edit "$PR_URL" --add-reviewer "${{ github.event.pull_request.user.login }}"
+            gh pr ready "$PR_URL" --undo
           fi
--- a/.github/workflows/uplift.yml
+++ b/.github/workflows/uplift.yml
@@ -5,8 +5,7 @@ on:
   # of the source PR [0], as it could risk leaking the GH_TOKENs.
   #
   # In this case, we do it as the job needs to run within the context of the
-  # target repo, so it can get a GH_TOKEN which it can use to comment on and
-  # update the PR.
+  # target repo, so it can get a GH_TOKEN which it can use to create a PR.
   #
   # Crucially, no external code is loaded or run as part of this workflow.
   #
@@ -19,7 +18,8 @@ jobs:
   uplift:
     if: >
       github.event.pull_request.merged == true &&
-      github.event.pull_request.base.ref == 'enterprise-main'
+      github.event.pull_request.base.ref == matrix.source &&
+      contains(github.event.pull_request.labels.*.name, matrix.trigger_label)
 
     runs-on: ubuntu-latest
 
@@ -31,36 +31,37 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - target: enterprise-beta
-            label_check: uplift-to-beta
-            label: branch:beta
-          - target: enterprise-release
-            label_check: uplift-to-release
-            label: branch:release
+          - source: enterprise-main
+            target: enterprise-beta
+            trigger_label: uplift-to-beta
+            target_label: branch:beta
+          - source: enterprise-main
+            target: enterprise-release
+            trigger_label: uplift-to-release
+            target_label: branch:release
+          - source: enterprise-beta
+            target: enterprise-release
+            trigger_label: uplift-to-release
+            target_label: branch:release
 
     env:
       MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
       PR_NUMBER: ${{ github.event.pull_request.number }}
       PR_TITLE: ${{ github.event.pull_request.title }}
       REPO_URL: https://github.com/${{ github.repository }}
-      TARGET: ${{ matrix.target }}
-      PR_LABELS: ${{ join(github.event.pull_request.labels.*.name, ',') }}
 
     steps:
       - name: Checkout
-        if: contains(env.PR_LABELS, matrix.label_check)
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
       - name: Configure git
-        if: contains(env.PR_LABELS, matrix.label_check)
         run: |
           git config user.name "github-actions"
           git config user.email "github-actions@github.com"
 
       - name: Create uplift script
-        if: contains(env.PR_LABELS, matrix.label_check)
         run: |
           cat << 'EOF' > uplift.sh
           set -e
@@ -165,14 +166,11 @@ jobs:
 
       - name: Uplift
         id: uplift
-        if: contains(env.PR_LABELS, matrix.label_check)
         run: |
-          bash uplift.sh $TARGET $MERGE_SHA
+          bash uplift.sh ${{ matrix.target }} $MERGE_SHA
 
       - name: Create PR
-        if: >
-          contains(env.PR_LABELS, matrix.label_check) &&
-          steps.uplift.outputs.branch != ''
+        if: steps.uplift.outputs.branch != ''
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
@@ -188,10 +186,10 @@ jobs:
           EOF
 
           gh pr create \
-            --base "$TARGET" \
+            --base "${{ matrix.target }}" \
             --head "$BRANCH" \
-            --title "[$TARGET uplift] ${{ env.PR_TITLE }}" \
+            --title "[${{ matrix.target }} uplift] $PR_TITLE" \
             --body-file pr_body.txt \
-            --label "${{ matrix.label }}" \
+            --label "${{ matrix.target_label }}" \
             --label "uplift" \
             --reviewer "mozilla/enterprise-firefox-engineering"
--- a/.github/workflows/uplift.yml
+++ b/.github/workflows/uplift.yml
@@ -109,7 +109,7 @@ jobs:
               fi
 
               set +e
-              git cherry-pick -x -X theirs $commit
+              git cherry-pick -x $commit
               STATUS=$?
               set -e
 
@@ -142,7 +142,7 @@ jobs:
             LINK="$REPO_URL/commit/$SHA"
 
             set +e
-            git cherry-pick -x -X theirs $SHA
+            git cherry-pick -x $SHA
             STATUS=$?
             set -e
 
--- a/.github/workflows/uplift.yml
+++ b/.github/workflows/uplift.yml
@@ -95,58 +95,36 @@ jobs:
 
           if [ "$COUNT" -eq 3 ]; then
             echo "Merge commit → multi-commit mode"
-
             COMMITS=$(git log --reverse --no-merges --pretty=format:"%H" $SHA^1..$SHA^2)
-
-            for commit in $COMMITS; do
-              SHORT=$(git rev-parse --short $commit)
-              MSG=$(git log -1 --pretty=format:"%s" $commit)
-              LINK="$REPO_URL/commit/$commit"
-
-              if git branch -r --contains $commit | grep -q "origin/$TARGET"; then
-                echo "- [$SHORT]($LINK) $MSG (already in branch)" >> $SUMMARY_FILE
-                continue
-              fi
-
-              set +e
-              git cherry-pick -x $commit
-              STATUS=$?
-              set -e
-
-              if [ $STATUS -ne 0 ]; then
-                # Check if it's an empty cherry-pick
-                if git diff --cached --quiet && git diff --quiet; then
-                  echo "Empty cherry-pick for $commit → skipping"
-
-                  git cherry-pick --skip
-
-                  echo "- [$SHORT]($LINK) $MSG (already applied)" >> $SUMMARY_FILE
-                  continue
-                fi
-
-                git add -A
-                git cherry-pick --continue
-                echo "- [$SHORT]($LINK) $MSG (⚠️ conflict)" >> $SUMMARY_FILE
-              else
-                echo "- [$SHORT]($LINK) $MSG" >> $SUMMARY_FILE
-              fi
-
-              APPLIED=1
-            done
-
           else
             echo "Single commit mode"
+            COMMITS=$SHA
+          fi
 
-            SHORT=$(git rev-parse --short $SHA)
-            MSG=$(git log -1 --pretty=format:"%s" $SHA)
-            LINK="$REPO_URL/commit/$SHA"
+          for commit in $COMMITS; do
+            SHORT=$(git rev-parse --short $commit)
+            MSG=$(git log -1 --pretty=format:"%s" $commit)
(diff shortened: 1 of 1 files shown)


.github/workflows/uplift.yml
```diff
@@ -0,0 +1,216 @@
...
+          else
+            echo "Single commit mode"
+
+            SHORT=$(git rev-parse --short $SHA)
+            MSG=$(git log -1 --pretty=format:"%s" $SHA)
+            LINK="$REPO_URL/commit/$SHA"
+
+            set +e
+            git cherry-pick -x -X theirs $SHA
+            STATUS=$?
```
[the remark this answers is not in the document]
jonathanmendez: @lissyx I wonder if we should disallow the 'rebase and merge' strategy (for general reasons, not just for this). AFAIK we don't use it and wouldn't want to use it?

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,227 @@
...
+  # [0] https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target:~:text=Warning-,Running,websitehttps://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target:~:text=Warning-,Running,website
+  #
+  pull_request_target:
+    types: [closed]
+
+jobs:
+  uplift:
+    if: >
+      github.event.pull_request.merged == true &&
+      github.event.pull_request.base.ref == 'enterprise-main'
```
lissyx: nit: do we want to care about beta-only fixes that may need an uplift?

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,227 @@
...
+
+    permissions:
+      contents: write
+      pull-requests: write
+
+    env:
+      MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
+      PR_NUMBER: ${{ github.event.pull_request.number }}
+      PR_TITLE: ${{ github.event.pull_request.title }}
+      REPO_URL: https://github.com/${{ github.repository }}
```
lissyx, line 59: I am not sure it is worth duplicating those

.github/workflows/uplift.yml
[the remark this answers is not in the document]
lissyx: rebase and merge is mostly what i always do manually :'( :'(

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,227 @@
...
+          fi
+
+          git push origin $BRANCH
+
+          echo "branch=$BRANCH" >> $GITHUB_OUTPUT
+
+          DELIM="GITHUB_OUTPUT_$(uuidgen)"
+          echo "summary<<$DELIM" >> $GITHUB_OUTPUT
+          cat $SUMMARY_FILE >> $GITHUB_OUTPUT
+          echo "$DELIM" >> $GITHUB_OUTPUT
```
lissyx, line 180: This should be a file in the repo, and not on the action, would make it easier to review as well

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,227 @@
...
+      # ----------------------------------------
+      - name: Uplift to enterprise-release
+        if: steps.labels.outputs.release == 'true'
+        id: rel_uplift
+        run: |
+          bash uplift.sh enterprise-release $MERGE_SHA
+
+      - name: Create PR (enterprise-release)
+        if: steps.labels.outputs.release == 'true' && steps.rel_uplift.outputs.branch != ''
+        uses: peter-evans/create-pull-request@v6
```
lissyx: is this vetted by github admins?

lissyx: Can't we just write a quick action in shell, that basically does:
 - `git checkout <branch>`
 - `git pull`
 - `git merge` (accept and commit any unsolved merges conflicts, it will be fixable after)
 - `git push HEAD:...`

I would feel safer than introducing a third party dep, and as much as I can tell there's an official action from github you can depend on to get `gh cli?`

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,227 @@
...
+          title: "[enterprise-release uplift] ${{ env.PR_TITLE }}"
+          labels: branch:release,uplift
+          body: |
+            Uplift from #${{ env.PR_NUMBER }}
+
+            ## Commits
+            ${{ steps.rel_uplift.outputs.summary }}
+
+            ⚠️ Commits marked with conflicts require manual resolution.
+          team-reviewers: enterprise-firefox-engineering
```
lissyx: I suspect this could be rewritten as a `matrix` over beta, release.

.github/workflows/uplift.yml
lissyx: > Can't we just write a quick action in shell, that basically does:
> 
>     * `git checkout <branch>`
> 
>     * `git pull`
> 
>     * `git merge` (accept and commit any unsolved merges conflicts, it will be fixable after)
> 
>     * `git push HEAD:...`
> 
> 
> I would feel safer than introducing a third party dep, and as much as I can tell there's an official action from github you can depend on to get `gh cli?`

It's already what your scripts partly does so I really dont get the extra dep

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,227 @@
...
+          git checkout -b $BRANCH origin/$TARGET
+
+          touch $SUMMARY_FILE
+
+          PARENTS=$(git rev-list --parents -n 1 $SHA)
+          COUNT=$(echo $PARENTS | wc -w)
+
+          APPLIED=0
+
+          if [ "$COUNT" -eq 3 ]; then
```
lissyx: Why 3 ? Why distinction between "multi commits" and single commit?

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,227 @@
...
+                fi
+
+                git add -A
+                git cherry-pick --continue
+                echo "- [$SHORT]($LINK) $MSG (⚠️ conflict)" >> $SUMMARY_FILE
+              else
+                echo "- [$SHORT]($LINK) $MSG" >> $SUMMARY_FILE
+              fi
+
+              APPLIED=1
```
lissyx: Shouldn't we count how many we applied?

.github/workflows/uplift.yml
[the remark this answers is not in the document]
jonathanmendez: Well I'm only talking about the PR completion action, not any manual rebasing/merging. And I haven't put a lot of thought into "we shouldn't use the rebase-and-merge PR completion", I just know that we've previously said among ourselves to use merge if our commit history is good or squash if not, so I thought we'd implicitly ruled it out.

.github/workflows/uplift.yml
jonathanmendez: I think it would work to add a check for that here, yes.

.github/workflows/uplift.yml
jonathanmendez: I intentionally avoided making a script because it is generally seen as a security vulnerability when using `on: pull_request_target` that executes with privilege. The idea of the security vulnerability is that while the existing/committed version of the workflow will be what runs (even if the PR contains changes to it), having the workflow call a script could in theory run a (maliciously) modified version of the script in the PR.

It may be enough mitigation that I am only running this code when the PR is closed and merged, but I guess I'll let the github admins weigh in on that.

.github/workflows/uplift.yml
jonathanmendez: I should be able to replace with `gh` CLI usage. I think this particular action does have widespread usage (I've seen it before and previously thought "well that's weird to see some random person's name in this enterprise workflow"), but it looks like a `gh` CLI replacement wouldn't be too bad.

.github/workflows/uplift.yml
jonathanmendez: > Why 3 ?

`git rev-list --parents -n 1` returns the SHAs of the current commit and parent commit(s). For a non-merge commit, it returns 2 SHAs (current and parent), while for merges it returns more than 2 (current and each parent).

I think `if [ "$COUNT" -gt 2 ]; then` would be technically better (inclusive of octopus merges).

I'll also add a comment making that clearer.

jonathanmendez: > Why distinction between "multi commits" and single commit?

In the case of a single commit (squash), we want to cherry-pick the single squash commit. In the case of multiple commits (merge), we want to cherry-pick each of the commits from the merged branch (and not the merge commit).

I think the script could be simplified, however, to only distinguish between multi/single commits when initializing `COMMITS` and then use a single code path to loop over `$COMMITS` and apply.

.github/workflows/uplift.yml
jonathanmendez: This is just used as a boolean check later. Probably can use true/false in some way instead.

jonathanmendez: After reading up on options in bash, I think 0/1 is safer but I've improved the variable name to be more clear that it's treated as a boolean.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,228 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          BRANCH="${{ steps.uplift.outputs.branch }}"
+
+          cat <<EOF > pr_body.txt
+          Uplift from #${{ env.PR_NUMBER }}
+
+          ## Commits
+          ${{ steps.uplift.outputs.summary }}
```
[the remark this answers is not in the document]
jonathanmendez: Updated to use a quoted heredoc delimiter to avoid shell expansion.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,228 @@
...
+                echo "Empty cherry-pick for $commit → skipping"
+                git cherry-pick --skip
+                echo "- [$SHORT]($LINK) $MSG (already applied)" >> $SUMMARY_FILE
+                continue
+              fi
+
+              git add -A
+              git cherry-pick --continue
+              echo "- [$SHORT]($LINK) $MSG (⚠️ conflict)" >> $SUMMARY_FILE
+              HAS_CONFLICTS=1
```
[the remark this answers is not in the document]
jonathanmendez: Instead of aborting we intentionally continue so that the full set of work can be documented, including any later commits. The PR is marked as a draft and the author needs to manually re-create or rebase to fix the issues.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,228 @@
...
+          # <commit> <parent1> [parent2 ...]
+          # So:
+          # 2 fields → normal commit (1 parent)
+          # >2 fields → merge commit (2+ parents)
+          PARENTS=$(git rev-list --parents -n 1 $SHA)
+          COUNT=$(echo $PARENTS | wc -w)
+          if [ "$COUNT" -gt 2 ]; then
+            echo "Merge commit → multi-commit mode"
+            COMMITS=$(git rev-list --reverse --no-merges $SHA^1..$SHA)
+          else
```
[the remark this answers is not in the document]
jonathanmendez: Duplicate comment from above.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,228 @@
...
+    steps:
+      - name: Set matrix
+        id: set-matrix
+        run: |
+          # Define all possible uplift paths
+          ALL_ENTRIES='[
+            {"source": "enterprise-main", "target": "enterprise-beta", "trigger_label": "uplift-to-beta", "target_label": "branch:beta"},
+            {"source": "enterprise-main", "target": "enterprise-release", "trigger_label": "uplift-to-release", "target_label": "branch:release"},
+            {"source": "enterprise-beta", "target": "enterprise-release", "trigger_label": "uplift-to-release", "target_label": "branch:release"}
+          ]'
```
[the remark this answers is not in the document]
jonathanmendez: Updated the PR description.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,228 @@
...
+          
+          # Filter based on PR base ref and labels
+          BASE_REF='${{ github.event.pull_request.base.ref }}'
+          LABELS='${{ toJson(github.event.pull_request.labels.*.name) }}'
+
+          # Use jq to filter entries where source matches base.ref AND trigger_label is in labels
+          FILTERED=$(echo "$ALL_ENTRIES" | jq --arg base "$BASE_REF" --argjson labels "$LABELS" '[.[] | select(.source == $base and (.trigger_label | IN($labels[])))]')
+          
+          echo "matrix=$(echo $FILTERED | jq -c .)" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
jonathanmendez: This did actually seem to be correct, but unintuitive, from my testing, but `IN(...)`, which I confused with `in(...)`, seems to not be the most straightforward comparison, so I updated to a different filter.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,195 @@
...
+              if git diff --cached --quiet && git diff --quiet; then
+                echo "Empty cherry-pick for $commit → skipping"
+                git cherry-pick --skip
+                echo "- [$SHORT]($LINK) $MSG (already applied)" >> $SUMMARY_FILE
+                continue
+              fi
+
+              git add -A
+              git cherry-pick --continue
+              echo "- [$SHORT]($LINK) $MSG (⚠️ conflict)" >> $SUMMARY_FILE
```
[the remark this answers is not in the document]
jonathanmendez: `--continue` will not actually fail. I've addressed that this is intended behavior in a later comment.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,195 @@
...
+          # >2 fields → merge commit (2+ parents)
+          PARENTS=$(git rev-list --parents -n 1 $SHA)
+          COUNT=$(echo $PARENTS | wc -w)
+          if [ "$COUNT" -gt 2 ]; then
+            echo "Merge commit → multi-commit mode"
+            COMMITS=$(git rev-list --reverse --no-merges $SHA^1..$SHA)
+          else
+            echo "Single commit mode"
+            COMMITS=$SHA
+          fi
```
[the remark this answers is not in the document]
jonathanmendez: Duplicate comment. Discussion is above.

.github/workflows/uplift.yml
jonathanmendez: Took some figuring out but I like the matrix logic I ended up with.

.github/workflows/uplift.yml
jonathanmendez: Updated to use `gh` CLI

.github/workflows/uplift.yml
jonathanmendez: Added in the new matrix logic.

.github/workflows/uplift.yml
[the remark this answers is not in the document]
jonathanmendez: I haven't found a way to detect when the merge was done with squash vs rebase-and-merge. If we think it makes sense to remove rebase-and-merge as a PR merge strategy, then this workflow can work as-is. Otherwise I'll just have to leave a warning comment or something to go manually check if the original PR was merged via rebase-and-merge and tell the reviewers to abandon the created uplift PR and do it manually.

.github/workflows/uplift.yml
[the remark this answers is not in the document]
jonathanmendez: I further updated this to move the summary through an env var, but that required going back to an unquoted heredoc delimiter. Although the summary is created in this workflow and cannot have a line that is just 'EOF', I updated to a uuid delimiter to avoid future concerns.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+    steps:
+      - name: Set matrix
+        id: set-matrix
+        run: |
+          # Define all possible uplift paths
+          ALL_ENTRIES='[
+            {"source": "enterprise-main", "target": "enterprise-beta", "trigger_label": "uplift-to-beta", "target_label": "branch:beta"},
+            {"source": "enterprise-main", "target": "enterprise-release", "trigger_label": "uplift-to-release", "target_label": "branch:release"},
+            {"source": "enterprise-beta", "target": "enterprise-release", "trigger_label": "uplift-to-release", "target_label": "branch:release"}
+          ]'
```
lissyx, line 33: All in bash, that's lovely :p

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+jobs:
+  build-matrix:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    permissions: {}
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Set matrix
+        id: set-matrix
```
lissyx, line 26: Please use more descriptive name and ID

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+            {"source": "enterprise-main", "target": "enterprise-release", "trigger_label": "uplift-to-release", "target_label": "branch:release"},
+            {"source": "enterprise-beta", "target": "enterprise-release", "trigger_label": "uplift-to-release", "target_label": "branch:release"}
+          ]'
+
+          # Filter based on PR base ref and labels
+          BASE_REF='${{ github.event.pull_request.base.ref }}'
+          LABELS='${{ toJson(github.event.pull_request.labels.*.name) }}'
+
+          # Use jq to filter entries where source matches base.ref AND trigger_label is in labels
+          FILTERED=$(echo "$ALL_ENTRIES" | jq --arg base "$BASE_REF" --argjson labels "$LABELS" '[.[] | select(.source == $base and (.trigger_label as $tl | $labels | index($tl) != null))]')
```
lissyx, line 40: I dont get clearly what this is doing. We filter labels? You mention "entries" but the `ALL_ENTRIES` is just a structure we defined above, so it far from being clear.

Are we getting tricked by how `jq` works, and `$ALL_ENTRIES` being piped into `stdin` is just a way to pass the structure, but the real "input" is `$LABELS` ?

This needs to be better documented, especially since this is something running on CI so hard to debug locally when it fails. And the output is also not clear from there. What do we get, a list of `sha`?  Or just we find the entry in `ALL_ENTRIES` that we care about?

.github/workflows/uplift.yml
lissyx: > I intentionally avoided making a script because it is generally seen as a security vulnerability when using `on: pull_request_target` that executes with privilege. The idea of the security vulnerability is that while the existing/committed version of the workflow will be what runs (even if the PR contains changes to it), having the workflow call a script could in theory run a (maliciously) modified version of the script in the PR.

I fail to understand the rationale here. Script is checked out, right, so it would not change.

For review purpose, it would still be much easier, can you split it during review and move it back into the payload if it's really required, before we land this PR?

> 
> It may be enough mitigation that I am only running this code when the PR is closed and merged, but I guess I'll let the github admins weigh in on that.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+        include: ${{ fromJson(needs.build-matrix.outputs.matrix) }}
+    env:
+      MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
+      PR_NUMBER: ${{ github.event.pull_request.number }}
+      PR_TITLE: ${{ github.event.pull_request.title }}
+      REPO_URL: https://github.com/${{ github.repository }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
lissyx, line 63: Why not using tags? Or even branches? What does security github-admin thinks about this?

.github/workflows/uplift.yml
lissyx: Still not convinced.

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions"
+          git config user.email "github-actions@github.com"
```
lissyx, line 70: Shouldn't we have a valid user here, with properly limited tokens/credentials ?

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+
+          git checkout -b $BRANCH origin/$TARGET
+
+          touch $SUMMARY_FILE
+
+          # rev-list --parents outputs:
+          # <commit> <parent1> [parent2 ...]
+          # So:
+          # 2 fields → normal commit (1 parent)
+          # >2 fields → merge commit (2+ parents)
```
lissyx, line 111: That's much clear!

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+          fi
+
+          DELIM="GITHUB_OUTPUT_$(uuidgen)"
+          echo "summary<<$DELIM" >> $GITHUB_OUTPUT
+          cat $SUMMARY_FILE >> $GITHUB_OUTPUT
+          echo "$DELIM" >> $GITHUB_OUTPUT
+          EOF
+
+      - name: Uplift
+        id: uplift
```
lissyx, line 184: Please a more descriptive name and id
```suggestion
      - name: Use uplift script to generate uplift branch
        id: generate-uplift-branch
```

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+
+          PR_URL=$(gh pr create \
+            --base "${{ matrix.target }}" \
+            --head "$BRANCH" \
+            --title "[${{ matrix.target }} uplift] $PR_TITLE" \
+            --body-file pr_body.txt \
+            --label "${{ matrix.target_label }}" \
+            --label "uplift")
+
+          gh pr edit "$PR_URL" --add-reviewer "mozilla/enterprise-firefox-engineering"
```
lissyx, line 212: I really dislike this catch-all reviewer but I'm not sure we can have a better option, we would have to expose our self-generated "merge rotation duty" somewhere on github? Or can we tag explicitely people? Better: configured from an env var. Then the people who review can decide if others needs to verify as well or if a single reviewer is enough?

Otherwise we spam everybody ...

.github/workflows/uplift.yml
```diff
@@ -0,0 +1,231 @@
...
+            git fetch enterprise-firefox ${{ matrix.target }}
+            git checkout -b ${{ steps.uplift.outputs.branch }} enterprise-firefox/${{ matrix.target }}
+            git cherry-pick ${{ steps.uplift.outputs.commits }}
+            ```
+            EOF
+
+            gh pr comment "$PR_URL" --body-file comment_body.txt
+
+            # Add original author and mark PR as draft since manual resolution is required
+            gh pr edit "$PR_URL" --add-reviewer "${{ github.event.pull_request.user.login }}"
```
lissyx, line 229: Should we also set `assigned` field?

lissyx (approved):
LGTM

PR #794, opened 2026-04-27
Release merge 2026042700

jonathanmendez:
### Description 

Bugzilla: NO BUG
Daily merge from `upstream/release` to `enterprise-release`


lissyx (approved):
The cargo lint is again one from m-c I assume

jonathanmendez:
> The cargo lint is again one from m-c I assume

Yes it looks like it. I can historically find a m-c failure on `thin-vec` but not `rand` but either way our branch is no different than upstream for either one.

PR #796, opened 2026-04-27
Bug 2031652 - Enterprise: queue opening new window on remoting call of Felt

lissyx:
Interacting from Linux' dock or others over DBus to open a new window launches "firefox" without any parameter. This results in Felt being shown instead of opening a new window. Check in browser's command line handler section of Felt, and if there is no URL list for a **non initial launch** (i.e. a remoted call) then pass over queueing a new window.

### Description 

Bugzilla: Bug-2031652

---

### Testing 

* [x] Added tests

**Steps to verify changes:**
1. Download deb package from CI and install
1. OR: `mach build && mach package && .mach repackage deb --input obj-felt-dbg/dist/firefox-*.en-US.linux-x86_64.tar.bz2 --output obj-felt-dbg/dist/firefox-152.0a1.en-US.linux-x86_64.deb --arch x86_64 --build-number $(grep "BuildID" obj-felt-dbg/dist/bin/application.ini | cut -d'=' -f2 | tr -d '\n') --version $(grep "^Version=" obj-felt-dbg/dist/bin/application.ini | cut -d'=' -f2 | tr -d '\n') --templates ./browser/installer/linux/app/debian/ --release-product firefox --release-type nightly`
2. Launch, identify and get browser
3. Right click on the dock icon, "new window"

**Expected result:**
A new window is opened

### DIFF
--- a/browser/components/BrowserContentHandler.sys.mjs
+++ b/browser/components/BrowserContentHandler.sys.mjs
@@ -1669,6 +1669,8 @@ nsDefaultCommandLineHandler.prototype = {
     // window. Instead, forward any URLs to be opened in the real Firefox.
     if (Services.felt && Services.felt.isFeltUI()) {
       console.debug(`Felt: Found FeltUI in BrowserContentHandler.`);
+      const wasPreventDefault = cmdLine.preventDefault;
+
       cmdLine.preventDefault = true;
 
       // Consume Felt-specific flags that don't take parameters
@@ -1712,6 +1714,13 @@ nsDefaultCommandLineHandler.prototype = {
             });
           }
         }
+      } else if (
+        !wasPreventDefault &&
+        cmdLine.state != Ci.nsICommandLine.STATE_INITIAL_LAUNCH
+      ) {
+        queueFeltURL({
+          disposition: FELT_OPEN_WINDOW_DISPOSITION.NEW_WINDOW,
+        });
       }
       return;
     }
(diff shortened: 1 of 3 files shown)


lissyx:
Looks to be badly broken on macOS, I'll investigate

browser/extensions/felt/api.js
```diff
@@ -167,13 +167,14 @@ this.felt = class extends ExtensionAPI {
...
 
       try {
-        let args = null;
-        if (url) {
-          args = Cc["@mozilla.org/supports-string;1"].createInstance(
-            Ci.nsISupportsString
-          );
-          args.data = url;
+        if (!url) {
+          url = "about:blank";
```
gcp: This seems incorrect, as it would suppress the users' setting for "Homepage and new windows/New tabs"?

gcp: Buh, Copilot also caught this but didn't warn because it was "low confidence".

gcp (changes requested):

browser/extensions/felt/api.js
lissyx: let me check, I had to set that at some point to make sure there was no intermittence at the test level in how I was checking the opening

lissyx: `about:home` should be good I guess ?

lissyx: > Buh, Copilot also caught this but didn't warn because it was "low confidence".

Ah ok, I did not even saw it at all, I'll adapt.

gcp: >about:home should be good I guess ?

I'm not sure but probably not, because you **could** also have set `about:blank`. I think it has to really be empty?

lissyx: yes my reply was wrong, i've fixed it

gcp (approved):

PR #802, opened 2026-04-28
Bug 2035477 - Fix l10n issues in the signout and close dialog

jporter-dev:
### Description 

Splits warnOnSignout to use two separate FTL strings instead of a FTL variant. Using an FTL variant in this way is an antipattern and should be avoided.

Bugzilla: Bug-2035477

---

* https://github.com/mozilla-l10n/enterprise-firefox-l10n/pull/10

### DIFF
--- a/browser/locales/en-US/browser/enterprise/enterprise.ftl
+++ b/browser/locales/en-US/browser/enterprise/enterprise.ftl
@@ -19,7 +19,7 @@ enterprise-close-prompt-title = Close { -brand-short-name }?
 
 # Variables:
 #   $tabCount (Number): The number of tabs to be closed.
-enterprise-close-prompt-title-with-tabcount-warn =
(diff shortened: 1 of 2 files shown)


1rneh (approved):

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -19,34 +19,31 @@ enterprise-close-prompt-title = Close { -brand-short-name }?
 
 # Variables:
 #   $tabCount (Number): The number of tabs to be closed.
-#   $warnSignout (String): "true" when the sign-out warning is shown.
+enterprise-close-prompt-title-with-tabcount-warn =
```
1rneh: `nit:`
```suggestion
enterprise-close-prompt-title-with-tabcount-and-signout-warning
```

1rneh (approved):

PR #808, opened 2026-04-28
Bug 2036752: Use correct registry path for PreXULSkeletonUI in enterprise builds and allow skeleton UI when MOZ_BYPASS_FELT is set

victhorlopez:
### Description 

Bugzilla: Bug-2036752

In enterprise builds, the PreXULSkeletonUI startup check requires the -felt flag to be present, but mochitests don't pass this flag, so adding MOZ_BYPASS_FELT as an alternative allows the skeleton UI to run during tests and prevents browser_preXULSkeletonUIRegistry from failing on enterprise builds.

Additionally the MOZ_APP_BASENAME is used in C++ to compute the registry path, vs a hardcoded "firefox" on the test, which fails on enterprise

---

* Depends on: https://phabricator.services.mozilla.com/D298838 landing on enterprise

---

### Screenshots 

---

### Testing 

* [ ] Added tests
[See test pass with latest changes](https://treeherder.mozilla.org/logviewer?job_id=566227017&repo=enterprise-firefox-pr&task=RUDVu4nyTxqH7VeP9PXY_A.0&lineNumber=2623)
[See test fail without changes](https://treeherder.mozilla.org/logviewer?job_id=564334446&repo=enterprise-firefox-pr&task=RrHrCFPbS0Cdj04UeirgMA.0&lineNumber=2984)

### DIFF
--- a/mozglue/misc/PreXULSkeletonUI.cpp
+++ b/mozglue/misc/PreXULSkeletonUI.cpp
@@ -1656,13 +1656,22 @@ static Result<Ok, PreXULSkeletonUIError> ValidateCmdlineArguments(
     }
   }
 
+#if defined(MOZ_ENTERPRISE)
+  // If there is MOZ_BYPASS_FELT environment variable, then likely tests are
+  // executing and thus no `-felt <socket>` argument can be found to verify if
+  // this is running a browser. By essence of that variable it is known the
+  // browser is running and not Felt UI so explicitely use the PreSkeletonUI
+  // in this case.
+  if (EnvHasValue("MOZ_BYPASS_FELT")) {
+    return Ok();
+  }
+
   // When starting FELT, disable the PreXULSkeletonUI. Checking with
   // is_felt_browser() is not doable here because this symbol is defined in
   // libxul however the present code runs in firefox.exe. Hence checking the
   // existence of "-felt" as is_felt_browser() would be doing, so
   // PreXULSkeletonUI is kept for normal browser, just disabled for FELT
   // window.
-#if defined(MOZ_ENTERPRISE)
   if (!hasFeltFlag) {
     return Err(PreXULSkeletonUIError::Cmdline);
   }
--- a/testing/enterprise/test_felt_browser_skeleton_ui.py
+++ b/testing/enterprise/test_felt_browser_skeleton_ui.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from test_felt_browser_signout import BaseBrowserSignout
+
+
+class BrowserSkeletonUI(BaseBrowserSignout):
+    def test_skeleton_ui_not_shown_for_felt(self):
+        self.run_check_skeleton_ui(self._driver, expected=False)
+
+        super().run_felt_base()
+        self.connect_child_browser(
+            capabilities={
+                # Do not auto-handle prompts.
+                "unhandledPromptBehavior": "ignore"
+            }
+        )
+        # Skeleton UI is not shown for the child browser either, which is
+        # expected since we use FELT and don't need it.
+        self.run_check_skeleton_ui(self._child_driver, expected=False)
+        self._do_signout()
+
+        self._logger.info("Restarting felt out-of-process")
+        self.marionette.restart(in_app=False)
+        self._logger.info("Felt restarted")
+        self.run_check_skeleton_ui(self._driver, expected=False)
+        self._manually_closed_child = True
+
+    def run_check_skeleton_ui(self, driver, expected):
+        with driver.using_context(driver.CONTEXT_CHROME):
+            showed = driver.execute_script(
+                "return Services.startup.showedPreXULSkeletonUI;"
+            )
+        self._logger.info(f"showedPreXULSkeletonUI: {showed}")
+
+        assert showed == expected, (
+            f"Expected showedPreXULSkeletonUI to be {expected}, got {showed}"
+        )
--- a/testing/enterprise/manifest.toml
+++ b/testing/enterprise/manifest.toml
@@ -87,6 +87,11 @@ run-if = [
   "buildapp == 'browser'", # browser specific: enterprise panel
 ]
 
+["test_felt_browser_skeleton_ui.py"]
+run-if = [
+  "os == 'win'",
+]
+
 ["test_felt_browser_starts.py"]
 run-if = [
   "buildapp == 'browser'", # open new tab


browser/base/content/test/startup/browser_preXULSkeletonUIRegistry.js
```diff
@@ -2,6 +2,10 @@ ChromeUtils.defineESModuleGetters(this, {
   WindowsRegistry: "resource://gre/modules/WindowsRegistry.sys.mjs",
 });
 
+const kRegPath = AppConstants.MOZ_ENTERPRISE
+  ? "Software\\Mozilla\\FirefoxEnterprise\\PreXULSkeletonUISettings"
```
victhorlopez: in theory we could use here directly ```AppConstants.MOZ_APP_BASENAME```
but I followed the original schema? Using ```AppConstants.MOZ_APP_BASENAME``` would make it work, but can hide some issues as well? (i.e if somehow moz_app_name changes or has a non expected value?)

mozglue/misc/PreXULSkeletonUI.cpp
```diff
@@ -1663,7 +1663,7 @@ static Result<Ok, PreXULSkeletonUIError> ValidateCmdlineArguments(
   // PreXULSkeletonUI is kept for normal browser, just disabled for FELT
   // window.
 #if defined(MOZ_ENTERPRISE)
-  if (!hasFeltFlag) {
+  if (!hasFeltFlag && !EnvHasValue("MOZ_BYPASS_FELT")) {
```
lissyx: When we perform the bypass, we end up with a normal browser, so I am not sure what we are achieving here.

lissyx:
```
In enterprise builds, the PreXULSkeletonUI startup check requires the -felt flag to be present, but mochitests don't pass this flag, so adding MOZ_BYPASS_FELT as an alternative allows the skeleton UI to run during tests and prevents browser_preXULSkeletonUIRegistry from failing on enterprise builds.
```

At the minimum, the comment above should be part of what we comment in this code to explain why.

lissyx: `-felt` is passed when running the **browser** since it is the argument passing the socket to perform the IPC. When running mochitests, I see why we would have to verify `MOZ_BYPASS_FELT` since there is no `-felt` being passed, so it would behave as the Felt UI and not as a browser here.

lissyx: I think it would be much easier to follow by:
 - augmenting the comments above
 - special casing this specific handling to explicitely return Ok

```suggestion
  // If there is MOZ_BYPASS_FELT environment variable, then likely tests are executing and thus no `-felt <socket>` argument can be found to verify if this is running a browser. By essence of that variable it is known the browser is running and not Felt UI so explicitely use the PreSkeletonUI in this case.
  if (EnvHasValue("MOZ_BYPASS_FELT")) {
    return Ok();
  }
  
  if (!hasFeltFlag)) {
```

browser/base/content/test/startup/browser_preXULSkeletonUIRegistry.js
lissyx: I feel like hardcoding has more risk of hidden regressions: proof is we have to do this fix, because it was hardcoded. Having the keys used the variable, we would not have known.

mozglue/misc/PreXULSkeletonUI.cpp
victhorlopez: I will change then, I was just following the pattern in the function, if somebody would add code below (doing ```return Err```) our new change would skip that.

browser/base/content/test/startup/browser_preXULSkeletonUIRegistry.js
```diff
@@ -2,6 +2,8 @@ ChromeUtils.defineESModuleGetters(this, {
   WindowsRegistry: "resource://gre/modules/WindowsRegistry.sys.mjs",
 });
 
+const kRegPath = `Software\\Mozilla\\${AppConstants.MOZ_APP_BASENAME}\\PreXULSkeletonUISettings`;
```
lissyx, line 5: I think it's worth upstreaming this directly (and getting proper review from owners of that code).

Maybe @Mossop can help?

lissyx (approved):

browser/base/content/test/startup/browser_preXULSkeletonUIRegistry.js
Mossop: Sure, if you can submit it I can review it, easy r+!

victhorlopez: https://phabricator.services.mozilla.com/D298838

testing/enterprise/test_felt_browser_skeleton_ui.py
```diff
@@ -0,0 +1,35 @@
...
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from test_felt_browser_signout import BaseBrowserSignout
+
+
+class BrowserSkeletonUI(BaseBrowserSignout):
+    def test_skeleton_ui_not_shown_for_felt(self):
```
victhorlopez, line 15: @Mossop can you shed some light?

I wrote this test to verify that the skekelton UI is working, firefox enterprise actually is always using same binary.

That means we also share same registry keys cause they also have same MOZ_APP_BASENAME etc. 

[In this code](https://searchfox.org/firefox-main/rev/a358f48a49a4d72df7b214805c4b2d2049d34a5a/mozglue/misc/PreXULSkeletonUI.cpp#1844) we hold a lock on FELT (parent process).
```
[task 2026-05-07T13:58:38.991+00:00] 13:58:38     INFO -  PreXULSkeletonUI: Opening reg key
[task 2026-05-07T13:58:39.000+00:00] 13:58:39     INFO -  PreXULSkeletonUI: Opened/created reg key (disposition 2)
[task 2026-05-07T13:58:39.003+00:00] 13:58:39     INFO -  PreXULSkeletonUI: CreateAndStorePreXULSkeletonUI failed with error 9
```
Do we actually want to show the skeleton UI the first time we open the Felt browser (i.e child firefox process)?

Mossop: You definitely don't want to show the skeleton UI in the FELT process as it wouldn't look correct so you should be able to skip it in that case, which should resolve any locking issue with the child Firefox process.

Whether you want to show the skeleton UI at all is a product decision I guess, I'm not sure it is really relevant to the enterprise use.

mozglue/misc/PreXULSkeletonUI.cpp
```diff
@@ -1644,25 +1645,35 @@ static Result<Ok, PreXULSkeletonUIError> ValidateCmdlineArguments(
       if (!_stricmp(flag, approvedArg)) {
         approved = true;
 
-        if (i == profileArgIndex) {
+        if (j == profileArgIndex) {
```
victhorlopez: @Mossop can you check whether the following is a bug,
the original i here is an index on argc array,
https://searchfox.org/firefox-main/rev/0a95d145f781bc57ef4183545cbb88e35d27b66d/mozglue/misc/PreXULSkeletonUI.cpp#1609

profileArgIndex is based on approvedArguments indices => `approvedArguments.length() - 1`
 So I would assume we need to compare indices from approvedArguments vs approvedArguments?? 

Let me know if these makes sense to you


Also I added `sPreXULSkeletonUIShown` [cause it was not set](https://searchfox.org/firefox-main/search?q=sPreXULSkeletonUIShown&path=&case=false&regexp=false) and I wanted to used it for the test I wrote. I checked on searchfox and nobody was anyway calling this function.
 ```
if (sPreXULSkeletonUIWindow) {
    sPreXULSkeletonUIShown = true;
  }
```

The reason why I am mentioned this is because I wrote the marionette test to ensure we are not showing the skeletonUI, I removed the code we have to prevent the skeletonUI to understand if the test is working (i.e I wanted to see it failing) and then I found it was not working
with `i`
https://github.com/mozilla/enterprise-firefox/compare/enterprise-main...mozilla:enterprise-firefox-try:bug/PreXULSkeletonUI
https://firefox-ci-tc.services.mozilla.com/tasks/a4Z4LTUNTuqfvnmOAAbMUA/runs/0/logs/live/public/logs/live.log#L20517
with `j`
https://firefox-ci-tc.services.mozilla.com/tasks/VfAmMHaoR2Kz0TRE75a9Cw/runs/0/logs/public/logs/live.log#L20162
https://github.com/mozilla/enterprise-firefox/compare/mozilla:f3e619f...mozilla:ada0352#diff-17cf25c6be22cb5f667f6a28c1e92673b419f157202308c1a38531a4c9acfacbR1648

Comment (Mossop) [reply mozglue/misc/PreXULSkeletonUI.cpp]: > [@Mossop](https://github.com/Mossop) can you check whether the following is a bug, the original i here is an index on argc array, https://searchfox.org/firefox-main/rev/0a95d145f781bc57ef4183545cbb88e35d27b66d/mozglue/misc/PreXULSkeletonUI.cpp#1609
> 
> profileArgIndex is based on approvedArguments indices => `approvedArguments.length() - 1` So I would assume we need to compare indices from approvedArguments vs approvedArguments??
> 
> Let me know if these makes sense to you

Yeah this looks like a clear bug in the original implementation and probably caused us to show the skeleton UI in a few cases where we shouldn't have.

File: mozglue/misc/PreXULSkeletonUI.cpp [hunk 6]
```diff
@@ -2116,6 +2127,9 @@ HWND ConsumePreXULSkeletonUIHandle() {
   ::CloseHandle(sPreXULSKeletonUIAnimationThread);
   sPreXULSKeletonUIAnimationThread = nullptr;
   HWND result = sPreXULSkeletonUIWindow;
+  if (sPreXULSkeletonUIWindow) {
+    sPreXULSkeletonUIShown = true;
+  }
```
Mossop: Can we set this here instead: https://searchfox.org/firefox-main/rev/0a95d145f781bc57ef4183545cbb88e35d27b66d/mozglue/misc/PreXULSkeletonUI.cpp#2056

Mossop (changes requested):

mozglue/misc/PreXULSkeletonUI.cpp
victhorlopez: Should i create the change for m-c then?

Mossop: Yes please!

victhorlopez: https://phabricator.services.mozilla.com/D301489

Mossop (approved):

PR #809, opened 2026-04-28
Bug 2034845 - Fix broken Access Connector on policy modification

jporter-dev:
### Description 

This PR fixes an issue that occurs when the `AccessConnector` policy is modified in-place (not added/removed). This results in a broken access connector due to the `#inclusionList` not being updated when `MatchPatterns` changes, and the connection not being re-initialized with the new server when the serverlist changes (host/port in policy)

Bugzilla: Bug-2034845

- `IPProtectionServerlist.sys.mjs`: fix `PrefServerList` pref observer registration (was referencing instance instead of class), skip in maybeFetchList() when pref value is unchanged
- `IPPChannelFilter.sys.mjs`: register/unregister pref observer to refresh `#inclusionSet` when changed
- `IPPProxyManager.sys.mjs`: add `reconnect` method for reconnecting to the recommended server from the updated list
- `IPPAutoStart.sys.mjs`: add logic to `#handleListChanged` to reconnect on live changes

---

* Related: #788 

---

### Testing 

**Steps to verify changes:**

1. Enable AccessConnector policy
2. Launch browser and navigate to a matching URL
3. Observe active AccessConnector state
4. Remove or change matching URL from `MatchPatterns`
    - This updates the icon live without requiring new navigation or refresh 
6. Observe inactive AccessConnector state
7. Repeat above for host or port
    - Serverlist changes require new navigation to update
    - Previously new navigation used the old server
    - Changing to an invalid port will result in broken proxy, changing back should fix it

**Expected result:**

AccessConnector shows active/inactive properly when modifying policy

### DIFF
--- a/toolkit/components/ipprotection/IPPAutoStart.sys.mjs
+++ b/toolkit/components/ipprotection/IPPAutoStart.sys.mjs
@@ -54,16 +54,15 @@ class IPPAutoStartSingleton {
       false
     );
 
-    this.#handleListChanged = async () => {
+    this.#handleListChanged = () => {
       if (!lazy.IPProtectionServerlist.hasList) {
         return;
       }
-      if (lazy.IPPProxyManager.state === lazy.IPPProxyStates.ACTIVE) {
-        await lazy.IPPProxyManager.stop(false);
-        await lazy.IPPProxyManager.start(
-          false,
-          PrivateBrowsingUtils.permanentPrivateBrowsing
-        );
+      if (
+        this.autoStart &&
+        lazy.IPPProxyManager.state === lazy.IPPProxyStates.ACTIVE
+      ) {
+        lazy.IPPProxyManager.switch();
         return;
       }
       this.init();
--- a/toolkit/components/ipprotection/IPPAutoStart.sys.mjs
+++ b/toolkit/components/ipprotection/IPPAutoStart.sys.mjs
@@ -54,30 +54,30 @@ class IPPAutoStartSingleton {
       false
     );
 
-    this.#handleListChanged = async () => {
+    this.#handleListChanged = () => {
       if (!lazy.IPProtectionServerlist.hasList) {
         return;
       }
       if (this.autoStart) {
         const state = lazy.IPPProxyManager.state;
-        if (state === lazy.IPPProxyStates.ACTIVE) {
-          lazy.IPPProxyManager.switch();
-          return;
-        }
-        if (state === lazy.IPPProxyStates.ERROR) {
-          await lazy.IPPProxyManager.reset();
-          lazy.IPPProxyManager.start(
-            false,
-            PrivateBrowsingUtils.permanentPrivateBrowsing
-          );
-          return;
-        }
-        if (state === lazy.IPPProxyStates.READY) {
-          lazy.IPPProxyManager.start(
-            false,
-            PrivateBrowsingUtils.permanentPrivateBrowsing
-          );
-          return;
+        switch (state) {
+          case lazy.IPPProxyStates.ACTIVE:
+            lazy.IPPProxyManager.switch();
+            return;
+          case lazy.IPPProxyStates.ERROR:
+            lazy.IPPProxyManager.reset().then(() => {
+              lazy.IPPProxyManager.start(
+                false,
+                PrivateBrowsingUtils.permanentPrivateBrowsing
+              );
+            });
+            return;
+          case lazy.IPPProxyStates.READY:
+            lazy.IPPProxyManager.start(
+              false,
+              PrivateBrowsingUtils.permanentPrivateBrowsing
+            );
+            return;
         }
       }
       this.init();
--- a/toolkit/components/ipprotection/IPPAutoStart.sys.mjs
+++ b/toolkit/components/ipprotection/IPPAutoStart.sys.mjs
@@ -4,6 +4,7 @@
 
 import { PrivateBrowsingUtils } from "resource://gre/modules/PrivateBrowsingUtils.sys.mjs";
 import { XPCOMUtils } from "resource://gre/modules/XPCOMUtils.sys.mjs";
+import { IPPEarlyStartupFilter } from "moz-src:///toolkit/components/ipprotection/IPPEarlyStartupFilter.sys.mjs";
 
 const lazy = {};
 
@@ -11,8 +12,6 @@ ChromeUtils.defineESModuleGetters(lazy, {
   IPProtectionServerlist:
     "moz-src:///toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs",
   IPPProxyManager:
-    "moz-src:///toolkit/components/ipprotection/IPPProxyManager.sys.mjs",
-  IPPProxyStates:
     "moz-src:///toolkit/components/ipprotection/IPPProxyManager.sys.mjs",
   IPProtectionService:
     "moz-src:///toolkit/components/ipprotection/IPProtectionService.sys.mjs",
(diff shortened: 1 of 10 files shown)


lissyx:
@jporter-dev Now that we have #825 there is no reason not to rebase and add proper testing to ensure this does not regress

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
```diff
@@ -54,7 +54,34 @@ class IPPAutoStartSingleton {
       false
     );
 
-    this.#handleListChanged = () => this.init();
+    this.#handleListChanged = async () => {
```
lissyx: Do we really need the `async` ?

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
```diff
@@ -54,7 +54,34 @@ class IPPAutoStartSingleton {
...
+          return;
+        }
+        if (state === lazy.IPPProxyStates.READY) {
+          lazy.IPPProxyManager.start(
+            false,
+            PrivateBrowsingUtils.permanentPrivateBrowsing
+          );
+          return;
+        }
+      }
```
lissyx: Is this really the right place? Can we get feedback from people working on IPProtection for this?

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
```diff
@@ -54,7 +54,34 @@ class IPPAutoStartSingleton {
...
+          );
+          return;
+        }
+        if (state === lazy.IPPProxyStates.READY) {
+          lazy.IPPProxyManager.start(
+            false,
+            PrivateBrowsingUtils.permanentPrivateBrowsing
+          );
+          return;
+        }
```
lissyx: a `switch` would be easier to read

toolkit/components/ipprotection/IPPChannelFilter.sys.mjs
```diff
@@ -421,6 +421,10 @@ export class IPPChannelFilter {
       0 /* unsigned long aPosition */
     );
     this.#active = true;
+    this.#inclusionPrefObserver = () => {
+      this.#inclusionSet = IPPChannelFilter.getInclusionList();
+    };
+    Services.prefs.addObserver(INCLUSION_PREF, this.#inclusionPrefObserver);
```
lissyx: no risk of double `.start()` ? We would end up adding the observer twice.

lissyx: I'd feel safer if this was in the constructor for example

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -403,19 +404,21 @@ export class PrefServerList extends IPProtectionServerlistBase {
   }
 
   async initOnStartupCompleted() {
-    Services.prefs.addObserver(
-      IPProtectionServerlist.PREF_NAME,
-      this.#observer
-    );
+    Services.prefs.addObserver(PrefServerList.PREF_NAME, this.#observer);
```
lissyx, line 414: linter change?

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -403,19 +404,21 @@ export class PrefServerList extends IPProtectionServerlistBase {
...
-      this.#observer
-    );
+    Services.prefs.removeObserver(PrefServerList.PREF_NAME, this.#observer);
   }
+
   maybeFetchList(_forceUpdate = false) {
+    const newList = Services.prefs.getStringPref(PrefServerList.PREF_NAME, "");
+
+    // If the list hasn't changed, we don't need to fetch it again.
+    if (!_forceUpdate && newList === this.#previousList) {
```
lissyx: if we use `_forceUpdate` we should remove `_`
```suggestion
    if (!forceUpdate && newList === this.#previousList) {
```

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
```diff
@@ -403,19 +404,21 @@ export class PrefServerList extends IPProtectionServerlistBase {
...
 
   uninit() {
-    Services.prefs.removeObserver(
-      IPProtectionServerlist.PREF_NAME,
-      this.#observer
-    );
+    Services.prefs.removeObserver(PrefServerList.PREF_NAME, this.#observer);
   }
+
   maybeFetchList(_forceUpdate = false) {
```
lissyx:
```suggestion
  maybeFetchList(forceUpdate = false) {
```

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
lissyx: This is exactly the change that was making my life miserable the other day :)

lissyx (approved):
LGTM; please verify with a peer of IPProtection for the `#handleListChanged` implementation if that is the right place.

toolkit/components/ipprotection/IPProtectionServerlist.sys.mjs
jporter-dev: `IPProtectionServerlist.PREF_NAME` -> `PrefServerList.PREF_NAME` plus the linter compacting it to one line

This was referencing the wrong class so PREF_NAME was `undefined` so the observer was broken.

toolkit/components/ipprotection/IPPAutoStart.sys.mjs
jporter-dev: I've pinged @strseb for some feedback

strseb: The autostart component's job is to "connect once when ready, asap". 

So the way i'd read that code is, if autostart is enabled and something changes force-start no matter the current state. 
If that `always-on` is the desired behavior, i'd probably move this into it's own little helper component. And swap out this helper component with the `ipp-always-on` in  IPProtectionActivator.

Otherwise, given the bug-problem text, it sounds like just applying this while active should be sufficient. In that case ProxyManager should observe the serverlist and call `switch()` while in active.

jporter-dev: @strseb I've extracted the logic into a `enterprise/IPPAlwaysOn.sys.mjs` component that gets loaded in alongside the other helpers in a similar fashion to your `IPPEnterpriseAuthProvider` PR. 

Let me know what you think! If it's the right approach then I can prepare this for upstream

toolkit/components/ipprotection/IPProtectionActivator.sys.mjs
```diff
@@ -11,13 +12,23 @@ import { IPProtectionServerlist } from "moz-src:///toolkit/components/ipprotecti
...
+}
+
 const coreHelpers = [
   IPPStartupCache,
   IPProtectionServerlist,
   IPPProxyManager,
   IPPSessionPrefManager,
   IPPAutoRestoreHelper,
   ...IPPAutoStartHelpers,
+  ...(AppConstants.MOZ_ENTERPRISE ? lazy.IPPAlwaysOnHelpers : []),
```
strseb: Wondering whether you actually need the AutoRestore (on session restore starts vpn when it was on) or the IPPAutostartHelpers (for the start on boot pref) on enterprise. Otherwise less objects that can start/stop are preferable :)  
```
... AppConstants.MOZ_ENTERPRISE ? lazy.IPPAlwaysOnHelpers : [IPPAutoRestoreHelper,...IPPAutoStartHelpers ]
```

toolkit/components/ipprotection/tests/xpcshell/xpcshell.toml
```diff
@@ -31,4 +31,10 @@ prefs = [
 
 ["test_IPProtectionStates.js"]
 
+# On non-enterprise builds IPProtectionServerlist is RemoteSettingsServerlist and
+# ignores the pref-based server list set up in add_setup, so hasList stays false
+# and proxy activation tests time out.
+["test_IPPAlwaysOn.js"]
+run-if = ["enterprise"]
```
strseb, line 40: Thats a good point. Maybe we should have a head_enterpise.js for enterprise tests, so add_setup correctly creates the enviroment.

toolkit/components/ipprotection/enterprise/IPPAlwaysOn.sys.mjs
```diff
@@ -0,0 +1,311 @@
...
+ * Registers the channel filter at startup to prevent data leaks before the
+ * proxy connection is fully established in always-on mode. Mirrors
+ * IPPEarlyStartupFilter from IPPAutoStart but uses the AlwaysOn policy check.
+ */
+class IPPAlwaysOnEarlyStartupFilter {
+  #alwaysOnAtStartup = false;
+
+  constructor() {
+    this.handleEvent = this.#handleEvent.bind(this);
+  }
```
strseb: Wondering, this is pretty much the same as IPPEarlyStartupFilter - only with a diffrent condition. 

WDYT if we do 
```typescript
class IPPEarlyStartupFilter{
      constructor( shouldActivate:Function<bool>){...}

     init() {
        if(!this.shouldActivate()) return;
        ...
     }
}


export IPPAutostartFilter = new IPPEarlyStartupFilter(()=>IPPAutoStart.autoStart)
export IPPAlwaysOnEarlyStartupFilter = new IPPEarlyStartupFilter(()=>IPPAlwaysOn.alwaysOnEnabled))


```

strseb (approved):
Overall looks great! Just a few comments :)

toolkit/components/ipprotection/tests/xpcshell/xpcshell.toml
jporter-dev: Good call, I'll fix this up properly in the Phab PR.

toolkit/components/ipprotection/IPProtectionActivator.sys.mjs
jporter-dev: I may need to augment IPPAlwaysOn a bit more if removing those to handle the scenarios those handle, but it makes sense to rework it this way to avoid start/stop conflicts. I'll take a look at removing those and making sure IPPAlwaysOn has parity with them

jporter-dev:
As noted in Bugzilla, I've submitted this patch upstream so this PR is no longer needed.

PR #813, opened 2026-04-29
Bug 2028025: Fix xpcshell test_BackupService_crossProfileTypeRecovery.js on enterprise builds

victhorlopez:
### Description 

Bugzilla: Bug-2028025

Enterprise branding sets browser.profiles.enabled to false on the default pref branch (via firefox-enterprise.js). The test's createBackupAndRecover helper calls clearUserPref("browser.profiles.enabled") during cleanup, which removes the user pref and falls back to the default branch value. On mozilla-central the default is true, so this is transparent. On enterprise the default is false, which causes SelectableProfileService.isEnabled to return false unexpectedly. Setting the pref in the test's xpcshell.toml doesn't help because toml prefs are set on the user branch. clearUserPref removes them, falling back to the enterprise default of false. 

The fix overrides the default branch in add_setup so that clearUserPref falls back to true, matching mozilla-central's behavior.

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [ ] Added tests
[see test failing without the fix](https://treeherder.mozilla.org/logviewer?job_id=563487967&repo=enterprise-firefox-pr&task=TWRib8MSTd--3fRl7TS-Kg.0&lineNumber=10686)
[see test pass](https://treeherder.mozilla.org/logviewer?job_id=563373137&repo=enterprise-firefox-pr&task=Hhk9hTiXRnKkM4Q6ZhZATw.0&lineNumber=3350)

### DIFF
--- a/browser/components/backup/tests/xpcshell/test_BackupService_crossProfileTypeRecovery.js
+++ b/browser/components/backup/tests/xpcshell/test_BackupService_crossProfileTypeRecovery.js
@@ -7,6 +7,10 @@ const { SelectableProfileBackupResource } = ChromeUtils.importESModule(
   "resource:///modules/backup/SelectableProfileBackupResource.sys.mjs"
 );
 
+const { AppConstants } = ChromeUtils.importESModule(
+  "resource://gre/modules/AppConstants.sys.mjs"
+);
+
 const lazy = {};
 ChromeUtils.defineESModuleGetters(lazy, {
   SelectableProfileService:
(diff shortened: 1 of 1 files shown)


browser/components/backup/tests/xpcshell/test_BackupService_crossProfileTypeRecovery.js
```diff
@@ -15,6 +19,22 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
+    // user prefs, which clearUserPref() removes.
+    const defaultBranch = Services.prefs.getDefaultBranch("");
+    const originalDefault = defaultBranch.getBoolPref(
+      "browser.profiles.enabled"
+    );
+    defaultBranch.setBoolPref("browser.profiles.enabled", true);
+    registerCleanupFunction(() => {
+      defaultBranch.setBoolPref("browser.profiles.enabled", originalDefault);
+    });
+  }
```
lissyx, line 42: That should work, but we should also assert that the changed value that we do on Enterprise is properly set before changing the value here.

lissyx (approved):

browser/components/backup/tests/xpcshell/test_BackupService_crossProfileTypeRecovery.js
1rneh: @lissyx Is the purpose of the assertion to make sure it fails if we change the default in the enterprise build which would make this set up step obsolete?

1rneh (commented):
TIL that preferences in `firefox-enterprise.js` are set as default preferences.

browser/components/backup/tests/xpcshell/test_BackupService_crossProfileTypeRecovery.js
lissyx: Yes, that was the idea of asserting that the state we expect to be for Enterprise is the one we undo in this code, if we change the pref value later (or by mistake), then we would get some signal that something is wrong (and if it's on purpose, we know and we adjust)

1rneh (approved):

PR #816, opened 2026-04-29
Bug 2031399 - Restore focus to badge button when signout dialog is dismissed

jporter-dev:
### Description 

This PR fixes the a11y issue when using the Enterprise Badge button/panel to trigger a signout, and then dismissing or cancelling the dialog. The focus is given to `html:body` by default, the intended behavior is to return focus to the button which triggered the dialog. Since the actual signout button is in the panel, focus is instead returned to the Enterprise Badge button.

- [ ] TODO: Investigate the same focus issue for the "About Firefox Enterprise" dialog

Bugzilla: Bug-2031399

---

### Screenshots 

enterprise-badge-focus

_Enterprise Badge with focus after cancelling signout_

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. While using a screen reader, activate the "Firefox Enterprise" button using keyboard navigation
2. In the panel, activate the "Sign out…"
3. In the resulting Sign-Out Confirmation modal dialog, dismiss the dialog using either the "Close" buttons or the Esc key

**Expected result:**

For the Sign-Out Confirmation dialog, when the dialog closes, focus returns to the button that opened the dialog. Because the button that opened the dialog is in the "Company Menu" dropdown that closed, the focus returns to the "Firefox Enterprise" button that opened the "Company Menu" dropdown.

### DIFF
--- a/browser/components/enterprise/EnterpriseHandler.sys.mjs
+++ b/browser/components/enterprise/EnterpriseHandler.sys.mjs
@@ -123,10 +123,10 @@ export const EnterpriseHandler = {
   _tabCount: null,
 
   /**
-   * Reference to the element that opened the enterprise panel, used to
-   * restore focus if the signout dialog is cancelled. Cleared in onSignOut
-   * regardless of outcome. A weak reference is used so the element is not
-   * kept alive if the panel opener is destroyed before signout completes.
+   * Weak reference to the element that keyboard-activated the enterprise panel,
+   * used to restore focus if the signout dialog is cancelled. Only set for
+   * keyboard activations; null for mouse. Cleared in onSignOut regardless of
+   * outcome.
    *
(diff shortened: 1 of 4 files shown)


1rneh (commented):
Thanks for tackling this accessibility bug!

Rather than a code review, I've left a comment on Bug-2031399 - I was able to reproduce the same broken focus restoration on nightly and release on m-c with an unrelated dialog, which suggests this is a bug in m-c rather than something specific to our enterprise components and we should inherit a fix once it's addressed in m-c.

So I'd lean towards having consistency here with m-c rather than implementing an enterprise-badge-only workaround.

jporter-dev:
As noted by Janika, rather than solving this only for the Enterprise badge, this can be fixed for all instances of this scenario. Closing this in favor of Bug-2040031.

PR #822, opened 2026-04-30
Bug 2035477 - Split quit shortcut prompt title into separate FTL strings

jporter-dev:
### Description 

Splits `enterprise-quit-shortcut-prompt-title` into two separate FTL strings instead of using a FTL variant. Using an FTL variant in this way is an antipattern and should be avoided.

Bugzilla: Bug-2035477

---

* https://github.com/mozilla-l10n/enterprise-firefox-l10n/pull/10
* #802


1rneh (approved):
Thank you!

PR #823, opened 2026-04-30
Bug 2036114 - Enterprise: Add enterprise-repack-completion task

lissyx:
Add a new task kind that will wait on completion of the task repacks so that updates can be properly pulled from taskcluster.

Bugzilla: Bug-2036114

### DIFF
--- a/taskcluster/gecko_taskgraph/target_tasks.py
+++ b/taskcluster/gecko_taskgraph/target_tasks.py
@@ -497,6 +497,9 @@ def target_tasks_enterprise_firefox_with_tests(
     )
 
     def filter(task):
+        if "shippable" in task.label:
+            return True
+
         # Enabling tests suites triggers many jobs not required, limit execution
         # to enterprise builds for now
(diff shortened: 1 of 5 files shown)


taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -497,6 +497,9 @@ def target_tasks_enterprise_firefox_with_tests(
     )
 
     def filter(task):
+        if "shippable" in task.label:
+            return True
+
```
lissyx, line 502: This is only for testing, will remove when merging

ahal (changes requested):
This is something I've been meaning to add to Firefox anyway, so I took the liberty of re-implementing it in a more generic way here:
https://github.com/mozilla/enterprise-firefox/pull/824 (still pending testing)

This way we can trivially add new tasks for other cases where we might want a "barrier" without needing new kinds + transforms each time.

lissyx:
Droppping in favor of #824

PR #826, opened 2026-05-01
Bug 2036410 - Fix/skip IPProtection browser tests for enterprise builds

jporter-dev:
### Description 

IPProtection tests were failing in Enterprise builds for two reasons:
1. The shared `putServerInRemoteSettings()` helper always wrote to RemoteSettings, which Enterprise doesn't use (it uses a pref-based server list).
2. Many tests exercise UI or flows absent from the Enterprise panel (status card, upgrade/paused flow, unauthenticated state, FxA sign-out, location selector, etc.).

Bugzilla: Bug-2036410

---

#### Fully skipped in enterprise

> `skip-if = ["enterprise"]` in `browser.toml`

| File | Reason |
|------|--------|
| `browser_IPPProxyManager.js` | Default panel template not visible in Enterprise |
| `browser_IPPUsageHelper.js` | Default panel template not visible in Enterprise, no usage UI |
| `browser_IPProtectionService.js` | `IPPEnterpriseAuthProvider` is used, this test is not applicable |
| `browser_guardian_client.js` | `GuardianClient` is only used by `IPPFxaAuthProvider`; enterprise uses `IPPEnterpriseAuthProvider` which bypasses it entirely` |
| `browser_ipprotection_bandwidth_precision.js` | Default panel template not visible in Enterprise, no bandwidth UI |
| `browser_ipprotection_confirmation_hints_site_exclusion.js` | No site exclusion feature in Enterprise; confirmation hints on tab navigation never fire |
| `browser_ipprotection_content.js` | Default panel template not visible in Enterprise |
| `browser_ipprotection_content_site_exclusion.js` | Default panel template not visible in Enterprise, no site exclusion UI |
| `browser_ipprotection_content_unauthenticated.js` | Always authenticated, no unauthenticated panel UI |
| `browser_ipprotection_header.js` | Default panel template not visible in Enterprise |
| `browser_ipprotection_keyboard_navigation.js` | Default panel template not visible in Enterprise, no elements to navigate |
| `browser_ipprotection_location_selector_telemetry.js` | No toolbar button or panel containing location selector |
| `browser_ipprotection_locations_keyboard_navigation.js` | No toolbar button or panel containing location selector |
| `browser_ipprotection_message_bar.js` | Default panel template not visible in Enterprise, no message bar in UI |
| `browser_ipprotection_network_error.js` | Default panel template not visible in Enterprise, no network error UI |
| `browser_ipprotection_panel.js` | Default panel template not visible in Enterprise |
| `browser_ipprotection_status_box.js` | Default panel template not visible in Enterprise, no status box in UI |
| `browser_ipprotection_status_card.js` | Default panel template not visible in Enterprise, no status card in UI |
| `browser_ipprotection_telemetry.js` | No user-facing toggle, enrollment, or exclusion flow |
| `browser_ipprotection_toolbar.js` | No toolbar button or panel |

### DIFF
--- a/browser/components/ipprotection/tests/browser/browser.toml
+++ b/browser/components/ipprotection/tests/browser/browser.toml
@@ -1,4 +1,5 @@
 [DEFAULT]
+environment = ["MOZ_BYPASS_FELT=1"]
 support-files = [
   "head.js",
 ]
@@ -14,31 +15,38 @@ prefs = [
 ["browser_IPPChannelFilter.js"]
 
 ["browser_IPPProxyManager.js"]
+# Enterprise uses a different panel template without status card, paused, or upgrade UI.
+skip-if = ["enterprise"]
 
 ["browser_IPPUsageHelper.js"]
-# Enterprise has no status card or bandwidth meter; IPPUsageHelper has no UI to drive.
+# Enterprise does not show the IP Protection toolbar button or panel containing the usage.
 skip-if = ["enterprise"]
 
 ["browser_IPProtectionService.js"]
+# Enterprise uses IPPEnterpriseAuthProvider which has no enrollment, entitlement, or FxA sign-in
+# flow, and does not show the standard panel or toolbar button.
+skip-if = ["enterprise"]
 
 ["browser_guardian_client.js"]
+# Enterprise uses IPPEnterpriseAuthProvider which does not use the Guardian client.
+skip-if = ["enterprise"]
 
 ["browser_ipprotection_alert_manager.js"]
 
 ["browser_ipprotection_bandwidth_precision.js"]
-# Enterprise has no status card or bandwidth meter.
+# Enterprise does not show the IP Protection toolbar button or panel containing the bandwidth UI.
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_confirmation_hints_site_exclusion.js"]
-# Enterprise has no site exclusion feature; confirmation hints for exclusion toggles never fire.
+# Enterprise has no site exclusion feature; confirmation hints on tab navigation never fire.
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_content.js"]
-# Enterprise renders a minimal panel template without status card, upgrade, or paused UI.
+# Enterprise does not show the IP Protection toolbar button or panel.
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_content_site_exclusion.js"]
-# Enterprise uses a different panel template without the site exclusion toggle UI.
+# Enterprise does not show the IP Protection toolbar button or panel containing the site exclusion UI.
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_content_unauthenticated.js"]
@@ -46,7 +54,7 @@ skip-if = ["enterprise"]
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_header.js"]
-# Enterprise replaces the help-button header with an Access Connector status label.
+# Enterprise does not show the IP Protection toolbar button or panel containing the header.
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_infobar.js"]
@@ -56,36 +64,49 @@ skip-if = ["enterprise"]
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_location_selector_telemetry.js"]
-# Enterprise uses a different panel template without the status card or location selector UI.
+# Enterprise does not show the IP Protection toolbar button or panel containing the location selector.
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_locations_keyboard_navigation.js"]
-# Enterprise uses a different panel template without the status card or location selector UI.
+# Enterprise does not show the IP Protection toolbar button or panel.
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_locations_list.js"]
 
 ["browser_ipprotection_message_bar.js"]
+# Enterprise does not show the IP Protection toolbar button or panel containing the message bar.
+skip-if = ["enterprise"]
 
 ["browser_ipprotection_network_error.js"]
+# Enterprise does not show the IP Protection toolbar button or panel containing the network errors.
+skip-if = ["enterprise"]
 
 ["browser_ipprotection_optout.js"]
 
 ["browser_ipprotection_panel.js"]
+# Enterprise does not show the IP Protection toolbar button or panel.
+skip-if = ["enterprise"]
 
 ["browser_ipprotection_proxy_errors.js"]
 
 ["browser_ipprotection_status_box.js"]
+# Enterprise does not show the IP Protection toolbar button or panel containing the status box.
+skip-if = ["enterprise"]
 
 ["browser_ipprotection_status_card.js"]
-# Enterprise has no status card UI.
+# Enterprise does not show the IP Protection toolbar button or panel containing the status card.
 skip-if = ["enterprise"]
 
 ["browser_ipprotection_telemetry.js"]
+# Enterprise uses IPPEnterpriseAuthProvider with no user-facing toggle, enrollment, or exclusion
+# flow; nearly all telemetry events tested here cannot be triggered in Enterprise.
+skip-if = ["enterprise"]
 
 ["browser_ipprotection_toolbar.js"]
+# Enterprise does not show the IP Protection toolbar button or panel.
 skip-if = [
   "devedition",
+  "enterprise",
 ]
 
 ["browser_ipprotection_utils.js"]
--- a/browser/components/ipprotection/tests/browser/browser.toml
+++ b/browser/components/ipprotection/tests/browser/browser.toml
@@ -1,5 +1,4 @@
 [DEFAULT]
-environment = ["MOZ_BYPASS_FELT=1"]
 support-files = [
   "head.js",
 ]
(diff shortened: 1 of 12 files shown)


browser/components/ipprotection/tests/browser/browser.toml
```diff
@@ -1,4 +1,5 @@
 [DEFAULT]
+environment = ["MOZ_BYPASS_FELT=1"]
```
lissyx: this should not be here

victhorlopez:
I just checked cause I [saw some more broken tests](https://treeherder.mozilla.org/logviewer?job_id=563904005&repo=enterprise-firefox-pr&task=STExDsTwTcSYe_abE-BjtQ.0&lineNumber=7040), and I know I pushed you to fix this. They have a lot of tests that check UI elements, so probably some they don't make sense to have them enabled(like browser_ipprotection_network_error?). Here is some table with some summary, to be honest for the ones that they have a lot of UI usage, do we get any value(i.e are they testing some internal logic that can be useful) from having them with 5 guards?

| File | Total tasks | DOM tasks | Non-DOM tasks | Notes |                                                                                                                                                                                                                 
  |---|---|---|---|---|                                                                                                                                                                                                                                                      
  | `browser_ipprotection_bandwidth_precision.js` | 6 | **6/6** | 0 | 100% DOM — every task traverses `statusCardEl.statusBoxEl.shadowRoot` for bandwidth element. |                                                                                       
  | `browser_IPPProxyManager.js` | 8 | **6/8** | 2 | 75% DOM — 2 tasks test pure service logic (error events, channel filter cleanup) |                                                                            
  | `browser_IPPUsageHelper.js` | 2 | **2/2** | 0 | 100% DOM  |                                                                                                                                                                        
  | `browser_IPProtectionService.js` | 11 | **5/11** | 6 | 45% DOM — 6 tasks are pure service logic (eligibility, entitlement, Nimbus, telemetry). 5 tasks open panel for statusCard/actionButton/unauthenticatedEl. |                                   
  | `browser_ipprotection_confirmation_hints_site_exclusion.js` | 7 | **1/7** | 6 | 14% DOM — only `test_confirmation_hint_exclusions_toggle` opens panel for `siteExclusionToggleEl`. Other 6 test hint logic via spies/tabs. |                          
  | `browser_ipprotection_network_error.js` | 3 | **3/3** | 0 | 100% DOM — all 3 open panel and access statusCardEl/actionButtonEl/statusBoxEl. Skip entire file. |                                                                                                          
  | `browser_ipprotection_toolbar.js` | 8 | **2/8** | 6 | 25% DOM — `toolbar_icon_status` and `toolbar_icon_status_new_window` open the panel for statusCardEl. Other 6 test CustomizableUI/toolbar prefs. |                                             
  | `browser_IPPChannelFilter.js` | — | **0** | all | No panel interaction. Tests network channel filtering and proxy exclusions. |                                                                                                                                          
  | `browser_guardian_client.js` | — | **0** | all | No panel interaction. Tests Guardian client enrollment via HTTP mocking. |                                                                                                                                              
  | `browser_ipprotection_alert_manager.js` | — | **0** | all | No panel interaction. Tests alert dialogs via gDialogBox. |                                                                                                                                                  
  | `browser_ipprotection_infobar.js` | — | **0** | all | No panel interaction. Tests notification bar via gNotificationBox. |                                                                                                                                               
  | `browser_ipprotection_optout.js` | — | **0** | all | No panel interaction. Tests toolbar button visibility via getElementById. |                                                                                                                                         
  | `browser_ipprotection_proxy_errors.js` | — | **0** | all | No panel interaction. Tests network error observer events. |                                                                                                                                                  
  | `browser_ipprotection_usage.js` | — | **0** | all | No panel interaction. Tests `formatRemainingBandwidth` utility. |                                                                                                                                                    
  | `browser_ipprotection_utils.js` | — | **0** | all | No panel interaction. Tests formatting logic. |

browser/components/ipprotection/tests/browser/browser_IPPProxyManager.js
```diff
@@ -168,6 +164,10 @@ add_task(async function test_IPPProxyManager_bug_1999946() {
  * Tests that opening the panel when the IPPProxyManager state is PAUSED shows the paused view.
  */
 add_task(async function test_IPPProxyManager_paused_shown() {
+  if (AppConstants.MOZ_ENTERPRISE) {
+    // Enterprise renders a different panel template without paused/upgrade UI.
+    return;
+  }
```
lissyx: Without any reference to the failing test, it's hard to review and judge this. Can you link and include log content of test failures on the bug on bugzilla?

I worry we may be skipping tests completely here.

jporter-dev: Added a test example in bugzilla. Lots are timing out due to us hiding some UI elements in the panel, specifically for features we don't leverage (pause/stop/start, location switching, usage, support/settings buttons, etc).

Some of these are probably worth skipping completely, but some need a closer look.

browser/components/ipprotection/tests/browser/browser_IPPProxyManager.js
```diff
@@ -166,6 +162,10 @@ add_task(async function test_IPPProxyManager_bug_1999946() {
  * Tests that opening the panel when the IPPProxyManager state is PAUSED shows the paused view.
  */
 add_task(async function test_IPPProxyManager_paused_shown() {
+  if (AppConstants.MOZ_ENTERPRISE) {
```
victhorlopez: I just saw (in case we want to keep this approach), there is some notation to skip

https://searchfox.org/firefox-main/rev/1372522d3de2c9aeedc8020ec8d81326f1a9516e/browser/components/asrouter/tests/browser/browser_asrouter_newtab_multistage_messages.js#144

Not sure it would work also here

jporter-dev: Ah good find, this is a bit cleaner if it can work for our inline skips in case we dont want to skip the full file

browser/components/ipprotection/tests/browser/browser.toml
```diff
@@ -15,6 +15,7 @@ prefs = [
 ["browser_IPPProxyManager.js"]
 
 ["browser_IPPUsageHelper.js"]
+skip-if = ["enterprise"]
```
lissyx: For all those skips, please add a comment stating why we can allow skipping

browser/components/ipprotection/tests/browser/browser_IPPProxyManager.js
jporter-dev: I found there is some tradeoff with this approach. Using the `skip_if` seems like the right approach since it will skip the test properly rather than silently pass with the early returns (although I'm not sure how that looks in practice). However, adding the extra param to `add_task` will add a bit of extra noise to diffs due to the formatting (extra leading tab for entire function). 

What are your thoughts?

lissyx: First time I see `skip_if` there.

lissyx: Likely we sould `skip_if` at `add_task` level if we only need to skip one task ?

victhorlopez: it should be roughly same numbe of lines no? I would be in favour of this

```
 {
    skip_if: () => AppConstants.MOZ_ENTERPRISE,
  }
```

lissyx: > it should be roughly same numbe of lines no? I would be in favour of this
> 
> ```
>  {
>     skip_if: () => AppConstants.MOZ_ENTERPRISE,
>   }
> ```

Not just in term of number of lines, but it's semantically better I think and I suspect less prone to merge conflicts

jporter-dev:
@victhorlopez - our upstream changes simplified this a bit, I've updated the PR description with what tests are skipped. I've skipped all failing browser tests that are based around using the IPProtection UI (toolbarbutton/panel elements), since we override that in Enterprise. We don't modify the alert manager, and we want to keep the channel filter. I did not touch any tests that were passing. 

I will make note to revisit these in #683 since that PR will remove our button/panel overrides and may change what tests we want to keep.

victhorlopez:
> @victhorlopez - our upstream changes simplified this a bit, I've updated the PR description with what tests are skipped. I've skipped all failing browser tests that are based around using the IPProtection UI (toolbarbutton/panel elements), since we override that in Enterprise. We don't modify the alert manager, and we want to keep the channel filter. I did not touch any tests that were passing.
> 
> I will make note to revisit these in #683 since that PR will remove our button/panel overrides and may change what tests we want to keep.

looks good, the only thing that was suggested in another PR to me is that the comments in the .toml should be inline on the same "thing" you are skipping
skip-if = [
"enterprise" # reason to skip
]

browser/components/ipprotection/tests/browser/browser.toml
victhorlopez: Did you readd this? It is already in the CI

victhorlopez (approved):
lgtm

browser/components/ipprotection/tests/browser/browser.toml
jporter-dev: Yes, accident. Removed!

lissyx (approved):

PR #827, opened 2026-05-03
Enterprise main merge 2026050300

jonathanmendez:
### Description 

Bugzilla: NO BUG

Conflicts resolved in:
- browser/themes/shared/preferences/preferences.css
- taskcluster/gecko_taskgraph/transforms/test/other.py
- toolkit/components/ipprotection/fxa/GuardianClient.sys.mjs
- toolkit/components/ipprotection/tests/xpcshell/test_IPProtectionService.js


taskcluster/gecko_taskgraph/transforms/test/other.py
```diff
@@ -700,10 +700,10 @@ def handle_tier(config, tasks):
...
                 "macosx1470-64-devedition/opt",
-                "macosx1400-64-shippable-qr/opt",
-                "macosx1400-64-qr/debug",
-                "macosx1500-64-shippable/opt",
-                "macosx1500-64/debug",
+                "macosx1500-aarch64/opt",
+                "macosx1500-aarch64/debug",
+                "macosx1500-aarch64-shippable/opt",
+                "macosx1500-aarch64-devedition/opt",
                 "macosx1500-64-enterprise/opt",
```
jonathanmendez, line 707: @lissyx The above were changed as part of BUG-2034604. Do you think we should do the same for the enterprise mochitests?

lissyx: We only run on aarch64 already so I'm not sure, maybe we should fix/ update/ streamline our macos universal package?

lissyx (approved):

taskcluster/gecko_taskgraph/transforms/test/other.py
lissyx: What's confusing here is that `macosx1500-64-enterprise` is in fact `macosx1500-aarch64`, but on non enterprise builds, `macosx1500-64` is a universal package of amd64 + aarch64

PR #835, opened 2026-05-04
Bug 2036826: Allow update processing for non-Felt standalone firefox launches

victhorlopez:
### Description 

Bugzilla: Bug-2036826

When Firefox is launched standalone (not through Felt), the !is_felt_ui() check alone is too broad and makes test "marAppApplyUpdateStageSuccess.js" fail, using the flag "test-process-updates" solves the issue

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [ ] Added tests

[See test failing](https://treeherder.mozilla.org/logviewer?job_id=564255947&repo=enterprise-firefox-pr&task=NbTjC3A0Qe-UQq2ng4gHdw.0&lineNumber=23788)
[See test passing](https://treeherder.mozilla.org/logviewer?job_id=567140725&repo=enterprise-firefox-pr&task=WOc_6DnZTd6K0soR2yNwYw.0&lineNumber=6730)

### DIFF
--- a/toolkit/xre/nsAppRunner.cpp
+++ b/toolkit/xre/nsAppRunner.cpp
@@ -4801,8 +4801,16 @@ Maybe<ShouldNotProcessUpdatesReason> ShouldNotProcessUpdates(
   }
 
 #  if defined(MOZ_ENTERPRISE)
+
+  // The "test-process-updates" argument (bug 2036826) allows
+  // tests like marAppApplyUpdateSuccess.js and
+  // marAppApplyUpdateStageSuccess.js to bypass this check and force update
+  // processing.
+  if(CheckArgExists("test-process-updates")){
+    return Nothing();
+  }
   // Don't process updates when launching a Browser from FELT, only Felt should
-  // perform that step
+  // perform that step.
   if (!is_felt_ui()) {
     return Some(ShouldNotProcessUpdatesReason::FeltOnlyUpdates);
   }


toolkit/xre/nsAppRunner.cpp
```diff
@@ -4803,7 +4803,7 @@ Maybe<ShouldNotProcessUpdatesReason> ShouldNotProcessUpdates(
 #  if defined(MOZ_ENTERPRISE)
   // Don't process updates when launching a Browser from FELT, only Felt should
   // perform that step
-  if (!is_felt_ui()) {
+  if (!is_felt_ui() && is_felt_browser()) {
```
lissyx: Please explicitely document why we need to enforce here

lissyx (approved):

toolkit/xre/nsAppRunner.cpp
```diff
@@ -4802,8 +4802,11 @@ Maybe<ShouldNotProcessUpdatesReason> ShouldNotProcessUpdates(
 
 #  if defined(MOZ_ENTERPRISE)
   // Don't process updates when launching a Browser from FELT, only Felt should
-  // perform that step
-  if (!is_felt_ui()) {
+  // perform that step. We explicitly check is_felt_browser() because the
+  // browser can also run in standalone mode in tests, where both is_felt_ui() and
+  // is_felt_browser() are false, and standalone launches should still
+  // process updates.
```
lissyx: Technically it should not even be doable to perform a standalone launch, this is only scoped to tests.
Unfortunately I think it is too early in the startup process that preferences are available?

If we can, it would be better to scope this by a pref that is toggled by the tests that requires it.
If not , I think it is worth mentionning more in details which tests depends on that, with a searchfox stable link

victhorlopez: So originally I was going to use
``` CheckArg("test-process-updates") ```
would you be fine with this? The reason I did not use it, it is because it seems to me it is a "flag" misuse in this case

https://searchfox.org/firefox-main/rev/a249a43995d7634e6e4a56bfd6496039031648bc/toolkit/mozapps/update/tests/data/xpcshellUtilsAUS.js#4747

https://searchfox.org/firefox-main/rev/a249a43995d7634e6e4a56bfd6496039031648bc/toolkit/mozapps/update/tests/data/xpcshellUtilsAUS.js#4690

Comment (lissyx) [reply toolkit/xre/nsAppRunner.cpp]: it may be a better idea, whether it is a misuse I would reach to the owners of the tests on #desktop-integration. It seem safer from my point of view.

Comment (victhorlopez) [reply toolkit/xre/nsAppRunner.cpp]: Does it make sense to fix this one? or should we actually ensure it fails?

Comment (victhorlopez) [reply toolkit/xre/nsAppRunner.cpp]: @lissyx tried ```CheckArg("test-process-updates")``` and does not work for `toolkit/mozapps/update/tests/unit_update_binary/marAppApplyUpdateSuccess.js` , so my question above, should we actually skip this test? or we make it pass for the bypass case?

Comment (lissyx) [reply toolkit/xre/nsAppRunner.cpp]: > [@lissyx](https://github.com/lissyx) tried `CheckArg("test-process-updates")` and does not work for `toolkit/mozapps/update/tests/unit_update_binary/marAppApplyUpdateSuccess.js` , so my question above, should we actually skip this test? or we make it pass for the bypass case?

Why does it not work?

Comment (victhorlopez) [reply toolkit/xre/nsAppRunner.cpp]: okay, apparently it works, but [CheckArg by default removes the arg](https://searchfox.org/firefox-main/rev/8b3456fa29ad35f58da4cc0a1d182e039124ba5d/toolkit/xre/nsAppRunner.cpp#531), so I need to use CheckArgExists (otherwise second call fails)

File: toolkit/xre/nsAppRunner.cpp
Comment (lissyx) [inline toolkit/xre/nsAppRunner.cpp]: I would really prefer we have it as a distinct case to make it clear,
```suggestion
  if (CheckArgExists("test-process-updates") {
    return Ok();
  }
  
  if (!is_felt_ui()) {
```

victhorlopez: okay let me change that

victhorlopez: was your code at that position for a reason? Cause there are a bunch of checks after, so to be 100% correct, I would need to move it at the bottom if we do that.

edit: Ok, the test passes, so I guess for our use case this is okay

lissyx, line 4809: This was not linted?

victhorlopez: um, didnt all the checks pass except 2 tests? I just checked that

victhorlopez: okay looks like I have 1 warning on clang, but did not see it in the checks

victhorlopez: https://github.com/mozilla/enterprise-firefox/pull/923

PR #842, opened 2026-05-05
Bug 2037170 - Enterprise: properly handle UTF8 in browser path

lissyx:
The browser path is being forwarded from Rust's XPCOM so it needs to ensure it is passing proper UTF-8 at least, and even UTF-16 for Windows, back to Felt to properly handle the case of the installation being in a folder using such characters.

### Description 

Bugzilla: Bug-2037170

### Testing 

* [x] Added tests

### DIFF
--- a/browser/extensions/felt/rust/src/components.rs
+++ b/browser/extensions/felt/rust/src/components.rs
@@ -4,7 +4,7 @@
 use nserror::{
     nsresult, NS_ERROR_CONNECTION_REFUSED, NS_ERROR_FAILURE, NS_ERROR_NOT_CONNECTED, NS_OK,
 };
-use nsstring::{nsACString, nsCString};
+use nsstring::{nsACString, nsAString, nsCString, nsString};
 use std::cell::RefCell;
 use std::env;
 use std::ffi::{c_char, CStr, CString};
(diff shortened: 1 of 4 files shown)


testing/enterprise/test_felt_browser_starts_utf8.py
```diff
@@ -0,0 +1,54 @@
...
+        # But we want to copy whole .app directory
+        if sys.platform == "darwin":
+            existing = os.path.dirname(os.path.dirname(existing))
+            assert existing.endswith(".app"), "Existing path {existing} should be .app"
+
+        self._logger.info(f"Copying from {existing} to {self._root_dir}")
+        shutil.copytree(existing, self._root_dir, symlinks=False, dirs_exist_ok=True)
+
+        self._logger.info(f"Executing from {self._root_dir}")
+        self._binary = self._driver.instance.binary
```
jonathanmendez: nit: for clarity, consider renaming to `_original_binary` ?

testing/enterprise/test_felt_browser_starts_utf8.py
```diff
@@ -0,0 +1,54 @@
...
+            self._driver.instance.binary = os.path.join(self._root_dir, "firefox.exe")
+        else:
+            self._driver.instance.binary = os.path.join(self._root_dir, "firefox")
+        self._driver.start_session()
+
+    def teardown(self):
+        super().teardown()
+        self._driver.instance.binary = self._binary
+        if os.path.isdir(self._root_dir):
+            shutil.rmtree(self._root_dir)
```
jonathanmendez: Seeing a bare `rmtree` call here got concerned with "what if `_root_dir` somehow ends up being a bad directory (like the system root dir :) )". Instead of using `tempfile.mkdtemp` above, maybe consider `tempfile.TemporaryDirectory` instead to make this cleanup technically safer?

browser/extensions/felt/rust/src/components.rs
```diff
@@ -460,20 +460,29 @@ impl FeltXPCOM {
...
 
-    fn BinPath(&self, bin: *mut nsACString) -> nserror::nsresult {
+    fn BinPath(&self, bin: *mut nsAString) -> nserror::nsresult {
         match env::current_exe() {
-            Ok(exe_path) => match exe_path.to_str() {
-                Some(path) => {
-                    unsafe {
-                        (*bin).assign(path);
+            Ok(exe_path) => {
+                #[cfg(windows)]
```
jonathanmendez, line 468: Maybe a small comment saying that Windows encodes paths as wide strings directly, hence the split code path here?

jonathanmendez (approved):
A few considerations but none are critical

PR #843, opened 2026-05-05
Bug 2037196 - Enterprise: report launch failure to the user

lissyx:
Ensure that early launch failure of the browser gets reported to the user: Felt window shown again, error message with call to action

Bugzilla: Bug-2037196

---

### Screenshots 

screenshot_launch_failure

---

### Testing 

* [x] Added tests

### DIFF
--- a/browser/extensions/felt/api.js
+++ b/browser/extensions/felt/api.js
@@ -230,6 +230,7 @@ this.felt = class extends ExtensionAPI {
       );
       Services.ppmm.addMessageListener("FeltParent:FirefoxLogoutExit", this);
       Services.ppmm.addMessageListener("FeltParent:FirefoxAbnormalExit", this);
+      Services.ppmm.addMessageListener("FeltParent:FirefoxLaunchFailure", this);
       Services.ppmm.addMessageListener(
         "FeltParent:TransitionFeltToBackground",
         this
@@ -297,6 +298,12 @@ this.felt = class extends ExtensionAPI {
         const success = Services.felt.makeBackgroundProcess(false);
         lazy.log.debug(`FeltExtension: makeBackgroundProcess? ${success}`);
         this.showWindow("felt-browser-error-multiple-crashes");
+        break;
+      }
+
+      case "FeltParent:FirefoxLaunchFailure": {
+        Services.felt.makeBackgroundProcess(false);
+        this.showWindow("felt-browser-error-launch-failure");
         break;
       }
 
@@ -399,6 +406,10 @@ this.felt = class extends ExtensionAPI {
     }
 
     Services.ppmm.removeMessageListener("FeltChild:Loaded", this);
+    Services.ppmm.removeMessageListener(
+      "FeltParent:FirefoxLaunchFailure",
+      this
+    );
 
     if (this.chromeHandle) {
       this.chromeHandle.destruct();
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -456,8 +456,11 @@ export class FeltProcessParent extends JSProcessActorParent {
           });
       })
       .catch(err => {
+        lazy.log.error(
+          `Firefox launch failure (${err.result} / ${err.name}): ${err.message}`
+        );
         lazy.ConsoleClient.isSessionRefreshBlocked = false;
-        throw err;
+        Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure");
       });
   }
 
--- a/browser/locales/en-US/browser/enterprise/felt.ftl
+++ b/browser/locales/en-US/browser/enterprise/felt.ftl
@@ -35,6 +35,9 @@ felt-browser-error-sso-timeout2 =
     .message = Please try again, or contact your administrator if the problem persists.
 felt-browser-error-multiple-crashes2 =
     .heading = { -brand-short-name } crashed multiple times
+felt-browser-error-launch-failure =
+    .heading = { -brand-short-name } cannot start
+    .message = Please contact your administrator if the problem persists.
 
 ## Logout messages
 
--- a/browser/extensions/felt/content/felt.xhtml
+++ b/browser/extensions/felt/content/felt.xhtml
@@ -187,6 +187,12 @@
             dismissable="true"
           ></moz-message-bar>
           <moz-message-bar
+            class="felt-browser-error-launch-failure is-hidden"
+            data-l10n-id="felt-browser-error-launch-failure"
+            type="error"
+            dismissable="true"
+          ></moz-message-bar>
+          <moz-message-bar
             class="felt-browser-error-no-network is-hidden"
             data-l10n-id="felt-browser-error-no-network"
             type="error"
--- a/testing/enterprise/manifest.toml
+++ b/testing/enterprise/manifest.toml
@@ -56,6 +56,8 @@ run-if = [
 
 ["test_felt_browser_starts.py"]
 
+["test_felt_browser_starts_missing_bin.py"]
+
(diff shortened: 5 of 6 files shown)


browser/locales/en-US/browser/enterprise/felt.ftl
```diff
@@ -35,6 +35,8 @@ felt-browser-error-sso-timeout2 =
     .message = Please try again, or contact your administrator if the problem persists.
 felt-browser-error-multiple-crashes2 =
     .heading = { -brand-short-name } crashed multiple times
+felt-browser-error-launch-failure =
+    .heading = { -brand-short-name } cannot start, please contact your administrator if the problem persists.
```
1rneh: The call to action "Please contact your administrator if the problem persists." belongs in `.message`? This follows the heading/message pattern of the other error siblings.

browser/extensions/felt/api.js
```diff
@@ -300,6 +301,13 @@ this.felt = class extends ExtensionAPI {
         break;
       }
 
+      case "FeltParent:FirefoxLaunchFailure": {
+        const success = Services.felt.makeBackgroundProcess(false);
+        lazy.log.debug(`FeltExtension: makeBackgroundProcess? ${success}`);
```
1rneh: Do we need `success` and what for?

testing/enterprise/test_felt_browser_starts_missing.py
```diff
@@ -0,0 +1,58 @@
...
+
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_tests import FeltTests
+
+
+class FeltStartsBrowserMissing(FeltTests):
```
1rneh: Is it meant as in "Felt start the browser and but binary missing" or "Felt starts but browser missing"?
```suggestion
class FeltStartsBrowserMissingBin(FeltTests):
```

browser/locales/en-US/browser/enterprise/felt.ftl
1rneh: `felt-browser-error-launch-failure` could also replace `felt-browser-error-multiple-crashes2`?

It points to the same cause and the user doesn't need the detail of the browser having crashed multiple times, but that the browser couldn't be launched. What do you think?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -456,7 +456,13 @@ export class FeltProcessParent extends JSProcessActorParent {
           });
       })
       .catch(err => {
+        lazy.log.error(
+          `Firefox launch failure (${err.result} / ${err.name}): ${err.message}`
+        );
         lazy.ConsoleClient.isSessionRefreshBlocked = false;
+        Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure", {
+          performLogout: true,
```
1rneh: `performLogout` is dead code right now. We also only perform a logout on felt-side on`FeltParent:FirefoxNormalExit` and not when we return to felt after a failure.

testing/enterprise/test_felt_browser_starts_missing.py
1rneh: If you change it, can you also update the file name. please?

1rneh (changes requested):
Looking good so far!

browser/locales/en-US/browser/enterprise/felt.ftl
lissyx: > The call to action "Please contact your administrator if the problem persists." belongs in `.message`? This follows the heading/message pattern of the other error siblings.

maybe, i need to check this

lissyx: > `felt-browser-error-launch-failure` could also replace `felt-browser-error-multiple-crashes2`?
> 
> It points to the same cause and the user doesn't need the detail of the browser having crashed multiple times, but that the browser couldn't be launched. What do you think?

I do think we want to keep the distinction, at least to help error reporting from users, because it is not the same cause. When we crashed, we launched (maybe not super far), whereas here we are catching complete failure to launch, e.g. missing file on disk

browser/extensions/felt/api.js
lissyx: probably not

testing/enterprise/test_felt_browser_starts_missing.py
lissyx: > Is it meant as in "Felt start the browser and but binary missing" or "Felt starts but browser missing"?

Rather the first, "Felt attempts to starts the browser but the binary is missing" (that's one way to trigger it, you could imagine linker issues, disk failure that makes things disappear, broken AV breaking IOs etc)

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: Ok, should we file a follow up to remove this then?

browser/locales/en-US/browser/enterprise/felt.ftl
lissyx: oh and I asked matthew about the UX part, and he agreed (DM on slack, ~~maybe I should ask him to vet here?~~ I asked him his opinion as well)

DPX-designer:
Thank you both, I am happy with this!

lissyx:
note to self: do not forget to update/cherry-pick the latest version of this PR on #847 and #848

browser/extensions/felt/api.js
```diff
@@ -230,6 +230,7 @@ this.felt = class extends ExtensionAPI {
       );
       Services.ppmm.addMessageListener("FeltParent:FirefoxLogoutExit", this);
       Services.ppmm.addMessageListener("FeltParent:FirefoxAbnormalExit", this);
+      Services.ppmm.addMessageListener("FeltParent:FirefoxLaunchFailure", this);
```
1rneh, line 233: Do we need to remove message listeners? Currently we don't remove any...

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -456,7 +456,11 @@ export class FeltProcessParent extends JSProcessActorParent {
           });
       })
       .catch(err => {
+        lazy.log.error(
+          `Firefox launch failure (${err.result} / ${err.name}): ${err.message}`
+        );
         lazy.ConsoleClient.isSessionRefreshBlocked = false;
+        Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure");
         throw err;
```
1rneh: We shouldn't re-throw the error here, because `this.firefox` has no awaiter, so it becomes an unhandled rejection. And we actually handle the error here.

1rneh (changes requested):
Thanks for the update!

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: I'm never certain here, that's why I left it at first

browser/extensions/felt/api.js
lissyx: Given how the code is loaded, I dont think it's required, but maybe we should do it in a follow up? Or do you prefer we fix it right now?

1rneh: Yes, this can be a follow up bug. And can include removing all added message listeners, not only for `FeltParent:FirefoxLaunchFailure`

1rneh (approved):
LGTM!

browser/extensions/felt/api.js
lissyx: https://bugzilla.mozilla.org/show_bug.cgi?id=2038891

PR #867, opened 2026-05-08
Bug 2031249 - Set FullScreenPrimary at window creation to prevent crash in NSViewUpdateVibrancyForSubtree

victhorlopez:
### Description 

Bugzilla: Bug-2031249

We are preventing _implicitlyAllowsFullScreenPrimary from changing its value internally, which avoids creating extra views mid-enumeration. This is done by setting NSWindowCollectionBehaviorFullScreenPrimary explicitly at window creation time and via macnativefullscreen="true" in felt.xhtml, so AppKit's _effectiveCollectionBehavior never enters the implicit heuristic path.

On creating a window we have this scenario, we set the flag on step 1, and we make sure on 3 the flag does not change its value
1. Create window with some flags
2. Appkit calls displayIfNeeded
3. XUL Parsing, sets some attributes

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [ ] Added tests

500 runs, 0 crashes ( 3 test failures unrelated, unrelated testing/enterprise/test_felt_browser_crash_restart.py)
https://firefox-ci-tc.services.mozilla.com/tasks/groups/eaN8CSlEQpmU1ZTNQ6c1xg#marionette

500 runs, 0 crashes ( 5 tests failures unrelated, testing/enterprise/test_felt_browser_crash_restart.py)
https://firefox-ci-tc.services.mozilla.com/tasks/groups/daETusbpT12MtRorgcBVvA#marionette

### DIFF
--- a/browser/extensions/felt/content/felt.xhtml
+++ b/browser/extensions/felt/content/felt.xhtml
@@ -10,6 +10,7 @@
   class="system-font-size"
   id="felt-window"
   windowtype="navigator:browser"
+  macnativefullscreen="true"
 >
   <head>
     <meta charset="utf-8" />
--- a/widget/cocoa/nsCocoaWindow.mm
+++ b/widget/cocoa/nsCocoaWindow.mm
@@ -5071,11 +5071,18 @@ nsresult nsCocoaWindow::CreateNativeWindow(const NSRect& aRect,
     // particular space, override the "Assign To" space and display popups on
     // the active space. Does not work with multiple displays. See
     // NeedsRecreateToReshow() for multi-display with multi-space workaround.
-    mWindow.collectionBehavior = mWindow.collectionBehavior |
-                                 NSWindowCollectionBehaviorMoveToActiveSpace;
+    mWindow.collectionBehavior =
+        mWindow.collectionBehavior |
+        NSWindowCollectionBehaviorMoveToActiveSpace |
+        NSWindowCollectionBehaviorFullScreenAuxiliary |
+        NSWindowCollectionBehaviorFullScreenDisallowsTiling;
   } else {
     // Non-popup windows are always opaque.
     mWindow.opaque = YES;
+    mWindow.collectionBehavior =
+        mWindow.collectionBehavior |
+        NSWindowCollectionBehaviorFullScreenPrimary |
+        NSWindowCollectionBehaviorFullScreenAllowsTiling;
   }
 
   if (mAlwaysOnTop || mIsAlert) {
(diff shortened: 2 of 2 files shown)


victhorlopez:
Can you also review @mstange ?

browser/extensions/felt/content/felt.xhtml
```diff
@@ -10,6 +10,7 @@
   class="system-font-size"
   id="felt-window"
   windowtype="navigator:browser"
+  macnativefullscreen="true"
```
victhorlopez: this is just so [SetSupportsNativeFullscreen](https://searchfox.org/firefox-main/rev/1372522d3de2c9aeedc8020ec8d81326f1a9516e/widget/cocoa/nsCocoaWindow.mm#7175) is called with true(to match what we set)

lissyx: https://searchfox.org/enterprise-main/rev/5988d1d923d15f8e5691cf4957defa23f408eb13/browser/base/content/browser.xhtml#15 it's already here for `browser.xhtml` :)

widget/cocoa/nsCocoaWindow.mm
lissyx: `NSWindowStyleMaskResizable` is set by `macnativefullscreen` ?

lissyx: I'm wondering if we should limit this to felt itself? If it can happen on any window, we should directly upstream this

victhorlopez: NSWindowStyleMaskResizable is set by [resizable flag,](https://searchfox.org/enterprise-main/rev/f9048334eccd943cb2865aae56c2f2bffb5c285f/browser/extensions/felt/api.js#367-368)

macnativefullscreen calls [this](https://searchfox.org/firefox-main/rev/1372522d3de2c9aeedc8020ec8d81326f1a9516e/widget/cocoa/nsCocoaWindow.mm#7175) sets NSWindowCollectionBehaviorFullScreenPrimary.

I changed the PR description, but basically this is the idea
```
On creating a window we have this scenario, we set the flag on step 1, and we make sure on 3 the flag does not change its value

    1. Create window with some flags
    2. Appkit calls displayIfNeeded (here it crashes)
    3. XUL Parsing, sets some attributes
```

victhorlopez: TBD, I was not sure, I would wait until @mstange comments, 

I am not sure on all cases a window should have this flag set? i.e does resizable imply we can go full screen? Internally the appkit checks for different things

browser/extensions/felt/content/felt.xhtml
lissyx: If this is just for consistency, it should be a different bug

victhorlopez: I would leave the flag in, this calls the function  SetSupportsNativeFullscreen with a false if macnativefullscreen is not set, I am not sure later there is another weird scenario where this crashes?

lissyx: Either it's related to this issue and we should do our best to explicit why, because the bug is far from obvious, or it's not related and doing it in this bug is adding confusion and will be problematic later when we go back to that code.

victhorlopez: okay fine, I am removing this

edit: I readd, anyway I will make the change in m-c if it gets approved then we see when we add this flag

victhorlopez:
It is done on m-c

PR #868, opened 2026-05-11
Bug 2037899 - Enterprise: enable more tests

lissyx:
### Description 

Bugzilla: Bug-2037899

### DIFF
--- a/taskcluster/gecko_taskgraph/transforms/task.py
+++ b/taskcluster/gecko_taskgraph/transforms/task.py
@@ -2207,7 +2207,16 @@ def add_github_checks_route(config, tasks):
         if task.get("attributes", {}).get("code-review"):
             routes = task.setdefault("routes", [])
             tier = task.get("treeherder", {}).get("tier", 3)
-            if "checks" not in routes and tier == 1:
+            should_surface_checks = tier == 1 and "checks" not in routes
+
+            # Only surface test/mochitest for enterprise tasks
+            if (
+                config.kind in ["test", "mochitest"]
+                and not "enterprise" in task["label"]
+            ):
+                should_surface_checks = False
+
+            if should_surface_checks:
                 routes.append("checks")
 
         yield task
--- a/taskcluster/gecko_taskgraph/target_tasks.py
+++ b/taskcluster/gecko_taskgraph/target_tasks.py
@@ -497,9 +497,13 @@ def target_tasks_enterprise_firefox_with_tests(
     )
 
     def filter(task):
-        # Enabling tests suites triggers many jobs not required, limit execution
-        # to enterprise builds for now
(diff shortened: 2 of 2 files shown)


jonathanmendez (changes requested):
It looks like all of the non-enterprise tests are getting surface in the GitHub checks. Is that something we actually want?

ahal:
> It looks like all of the non-enterprise tests are getting surface in the GitHub checks. Is that something we actually want?

We need to ensure they don't get the `checks` route added to their task definitions. But along this line of thinking, do we even want these tests to run in the first place?

lissyx:
> > It looks like all of the non-enterprise tests are getting surface in the GitHub checks. Is that something we actually want?
> 
> We need to ensure they don't get the `checks` route added to their task definitions. But along this line of thinking, do we even want these tests to run in the first place?

I think having them run is the point of the PR ?

lissyx:
> It looks like all of the non-enterprise tests are getting surface in the GitHub checks. Is that something we actually want?

hopefully that is now fixed

ahal:
> I think having them run is the point of the PR ?

Oh sorry, I thought there were non-enterprise tests being run. I misunderstood.

ahal (approved):

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -2207,7 +2207,17 @@ def add_github_checks_route(config, tasks):
         if task.get("attributes", {}).get("code-review"):
             routes = task.setdefault("routes", [])
             tier = task.get("treeherder", {}).get("tier", 3)
-            if "checks" not in routes and tier == 1:
+            if tier != 1:
+                continue
```
jonathanmendez: It seems like this should yield the unmodified task rather than skipping it? Maybe
```suggestion
                yield task
                continue
```
although with needing to do the same below maybe it makes more sense to combine these checks for when to append the route and otherwise fall through without a `continue`

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -2207,7 +2207,17 @@ def add_github_checks_route(config, tasks):
...
+
+            # Only surface test/mochitest for enterprise tasks
+            if (
+                config.kind in ["test", "mochitest"]
+                and not "enterprise" in task["label"]
+            ):
+                continue
```
jonathanmendez: Same issue as above about needing to `yield task`

jonathanmendez (changes requested):

taskcluster/gecko_taskgraph/transforms/task.py
lissyx: yes that is what I did

jonathanmendez (approved):

PR #825, opened 2026-04-30
Bug 2036167 - Enterprise: Access Connector testing

lissyx:
Provide basic testing of access connector from policies pushed by the console, either being set enabled by default (from console) or disabled by default (from console). Coverage ensures prefs and set at expected values, and locked when applied and correctly unlocked and reset after.

Also try to load data with the policy applied and verify there is a proxy error because no proxy is really setup to handle the connections.

Bugzilla: Bug-2036167

### Testing 

* [x] Added tests
* [ ] Manual testing performed

PR #869, opened 2026-05-11
Bug 2038631 - Enterprise: make access connector test wait for the title

lissyx:
We have a network dependency that is hard to avoid since Access Connector's match pattern depends on having a hostname, hence testing wildcard is complicated with only localhost, no port can be added in the matching part. Thus testing is done against support.mozilla.org hence hitting the network which can be slow. Waiting loinger for the title to be read correctly to avoid intermittence.

### Description 

Bugzilla: Bug-2038631

---

* Depends on: #825

### DIFF
--- a/testing/enterprise/test_felt_browser_access_connector.py
+++ b/testing/enterprise/test_felt_browser_access_connector.py
@@ -185,6 +185,7 @@ class BrowserAccessConnector(FeltTests):
 
     def run_load_page_ok(self, url, expected_title):
         self.open_tab_child(url)
+        self._child_longwait.until(lambda d: len(d.title) > 0)
         found_title = self._child_driver.title
         assert found_title == expected_title, (
             f"No access connector used, expected '{expected_title}', found '{found_title}'"


victhorlopez (approved):
lgtm

testing/enterprise/test_felt_browser_access_connector.py
```diff
@@ -185,6 +185,7 @@ def run_change_access_connector_policy(self, new_value):
 
     def run_load_page_ok(self, url, expected_title):
         self.open_tab_child(url)
+        self._child_longwait.until(lambda d: len(d.title) > 0)
```
jporter-dev, line 188:
```
self._child_longwait.until(
            lambda d: d.title == expected_title,
            message=f"Expected title '{expected_title}', got '{self._child_driver.title}'",
        )
```

Not sure if it's worth waiting for the expected title or if waiting for _any_ title is fine.

jporter-dev (approved):
LGTM! Could wait for the specific title if it's still intermittent in case there's intermediate titles causing the flakiness.

testing/enterprise/test_felt_browser_access_connector.py
lissyx: Let's see and revisit if required

PR #871, opened 2026-05-11
Bug 2038609 -  Enable by default "browser.newtabpage.activity-stream.feeds.section.topstories" as it will be set disabled/enabled by policy

victhorlopez:
### Description 

Bugzilla: Bug-2038609

edit: Enabling again the original pref, as we support the feature and will just be disabled with policy

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [ ] Added tests
[See test pass](https://treeherder.mozilla.org/logviewer?job_id=565781665&repo=enterprise-firefox-pr&task=NbOnbR2ZTwevUTLrspA-TA.0&lineNumber=15503) filter for `browser/components/preferences/tests/home/`

### DIFF
--- a/browser/branding/enterprise/pref/firefox-enterprise.js
+++ b/browser/branding/enterprise/pref/firefox-enterprise.js
@@ -9,8 +9,6 @@ pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
 
 pref("enterprise.log_level", "Error");
 
-pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);
-
 // On Enterprise we want to enforce updates so we force it
 // Bug 2020768: Should those value be set/locked at runtime by FELT only
 //              or is it fine to apply it to any enterprise build?


browser/branding/enterprise/pref/firefox-enterprise.js
```diff
@@ -10,6 +10,7 @@ pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
 pref("enterprise.log_level", "Error");
 
 pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);
+pref("browser.newtabpage.activity-stream.discoverystream.endpointSpocsClear", "");
```
lissyx: From a product perspective, do we want Activity Stream at all?

victhorlopez: good question.. trying to clarify which features are we supposed to support.

It was already disabled with `pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);` but they have even more prefs. I guess your question was if we should disable tests, and ensure feature is disabled

victhorlopez: I had a look at two things, the first one is whether we should support this, 

As far as understood we need to support that features are toggled with policies , for these 3
Recommended Stories, Sponsored Content, Sponsored Tiles

```
Can be disabled with https://mozilla.github.io/policy-templates/#usermessaging
We should consider default disabling sponsored content in our builds
```



Second thing I did was checking documentation on prefs.
https://searchfox.org/firefox-main/rev/0a95d145f781bc57ef4183545cbb88e35d27b66d/browser/extensions/newtab/docs/v2-system-addon/preferences.md#131

After that then decided to move the pref to test only, the reason is we might need that endpoint if the user/policy sets that to true/false. thoughts?

lissyx: > good question.. trying to clarify which features are we supposed to support.
> 
> It was already disabled with `pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);` but they have even more prefs. I guess your question was if we should disable tests, and ensure feature is disabled

OK but that may have been disabled as a result of quickly boostrapping things

> I had a look at two things, the first one is whether we should support this,
> 
> As far as understood we need to support that features are toggled with policies , for these 3 Recommended Stories, Sponsored Content, Sponsored Tiles
> 
> ```
> Can be disabled with https://mozilla.github.io/policy-templates/#usermessaging
> We should consider default disabling sponsored content in our builds
> ```
> 
> Second thing I did was checking documentation on prefs. https://searchfox.org/firefox-main/rev/0a95d145f781bc57ef4183545cbb88e35d27b66d/browser/extensions/newtab/docs/v2-system-addon/preferences.md#131
> 
> After that then decided to move the pref to test only, the reason is we might need that endpoint if the user/policy sets that to true/false. thoughts?

So we have a subset we want to support ?

1rneh: Other sponsored content can also be disabled via the policy [FirefoxHome](https://firefox-admin-docs.mozilla.org/reference/policies/firefoxhome), this includes controling the preference `browser.newtabpage.activity-stream.feeds.section.topstories`. Hence we can remove [setting a default in our builds directly](https://searchfox.org/enterprise-main/rev/cee65aebe09d3814ffb5459068dff0de27d0b39b/browser/branding/enterprise/pref/firefox-enterprise.js#13).

1rneh (changes requested):
Left a comment stating that we can undo [setting a default in our builds directly](https://searchfox.org/enterprise-main/rev/cee65aebe09d3814ffb5459068dff0de27d0b39b/browser/branding/enterprise/pref/firefox-enterprise.js#13). because we can disable the activity stream preferences via the policy [FirefoxHome](https://firefox-admin-docs.mozilla.org/reference/policies/firefoxhome).

1rneh (approved):

lissyx (approved):

PR #872, opened 2026-05-11
Bug 2038707 - Enterprise: do not schedule complete kind for level < 3

lissyx:
The 'complete' kind aims at tracking shippable artifact status. It should not be scheduled on tasks of level < 3, otherwise it ends up pulling by the game of dependencies the shippable builds on PRs.

### Description 

Bugzilla: Bug-2038707

### DIFF
--- a/taskcluster/gecko_taskgraph/target_tasks.py
+++ b/taskcluster/gecko_taskgraph/target_tasks.py
@@ -519,6 +519,9 @@ def target_tasks_enterprise_firefox_with_tests(
         if ("shippable" in task.label or shippable) and level < 3:
             return False
 
+        if task.kind == "complete" and level < 3:
+            return False
+
         if not build_platform or not build_type:
             return True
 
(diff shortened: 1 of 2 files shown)


taskcluster/gecko_taskgraph/target_tasks.py
jonathanmendez, line 516: Maybe this is the different solution that @ahal is working on, but I would guess maybe the "complete" task should be considered "shippable" and thus should be filtered by the shippable check below instead of an additional specific filter for "complete". Is it enough to add attributes: shippable: true to the task?

jonathanmendez (approved):
This should work as-is, and my other comment/question is presumably something @ahal knows best and will resolve later. Looks good to me for now.

ahal (approved):

taskcluster/gecko_taskgraph/target_tasks.py
lissyx: I think it was on purpose that it was not set with that attribute and  only the index being shippable

PR #873, opened 2026-05-11
Bug 2038707 - Ensure 'complete' kind never causes dependencies to get…

ahal:
… pulled in


ahal:
Unfortunately I realized `if-dependencies` won't help us until https://github.com/taskcluster/taskgraph/issues/710 is fixed

PR #876, opened 2026-05-11
Bug 2038706 - Enterprise: add Android not implemented GetNetworkInterfaces

lissyx:
Implement the method but return an error since there is no Android build of Firefox Enterprise.

### Description 

Bugzilla: Bug-2038706


lissyx:
https://github.com/mozilla/enterprise-firefox-try/runs/75417056416

PR #880, opened 2026-05-12
Bug 2035781 - Register enterprise.prompt_on_signout pref default and add fresh-profile test

jporter-dev:
### Description 

This PR adds a default value for `enterprise.prompt_on_signout`. The functionality was working as expected in a fresh profile, but the pref did not have an explicit default.

Bugzilla: Bug-2035781

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. Start with a fresh profile
2. Open a new browser
3. Open `about:config` and see the pref `enterprise.prompt_on_signout` set to `true` by default
4. Trigger a quit of the browser
5. Observe the Enterprise signout dialog

**Expected result:**

User is presented with the Enterprise signout dialog when the browser exits.

### DIFF
--- a/testing/enterprise/test_felt_browser_close_warning.py
+++ b/testing/enterprise/test_felt_browser_close_warning.py
@@ -178,7 +178,7 @@ class BrowserCloseWarning(FeltTests):
 
     def test_prompt_on_signout_default_is_true(self):
         """enterprise.prompt_on_signout must have a registered default of true."""
-        super().run_felt_base()
+        self.run_felt_base()
         self.connect_child_browser()
 
         self._child_driver.set_context("chrome")
@@ -199,7 +199,7 @@ class BrowserCloseWarning(FeltTests):
 
     def test_browser_window_close_signout_warning_only(self):
         """Sign-out warn on, tabs warn off, single tab - enterprise signout dialog shows."""
-        super().run_felt_base()
+        self.run_felt_base()
         self.connect_child_browser()
         self.assert_user_signed_in(env=Environment.FIREFOX)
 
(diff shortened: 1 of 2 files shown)


lissyx (changes requested):
Can we have test coverage for that?
Ideally checking there is a default value, and it is `true`

testing/enterprise/test_felt_browser_close_warning.py
```diff
@@ -176,6 +176,27 @@ def is_closed(_):
 
         self._child_wait.until(is_closed)
 
+    def test_prompt_on_signout_default_is_true(self):
+        """enterprise.prompt_on_signout must have a registered default of true."""
+        super().run_felt_base()
```
lissyx: nit: 
```suggestion
        self.run_felt_base()
```

jporter-dev: Thanks, fixed the other references in here as well

testing/enterprise/test_felt_browser_close_warning.py
```diff
@@ -176,9 +176,30 @@ def is_closed(_):
...
+        )
+        default_value = self._child_driver.execute_script(
+            f"return Services.prefs.getDefaultBranch('').getBoolPref('{PREF_PROMPT_ON_SIGNOUT}', false);"
+        )
+        self._child_driver.set_context("content")
+
+        assert is_registered, (
+            f"{PREF_PROMPT_ON_SIGNOUT} has no registered default value"
+        )
+        assert default_value is True, (
```
lissyx:
```suggestion
        assert default_value, (
```
Should be enough?

lissyx (approved):

testing/enterprise/test_felt_browser_close_warning.py
jporter-dev: I suppose this can be simplified since the `is_registered` already checks for bools

PR #887, opened 2026-05-12
Bug 2039062 - macOS universal package

lissyx:
### Description 

Bugzilla: Bug-2039062

### DIFF
--- a/taskcluster/kinds/build/macosx.yml
+++ b/taskcluster/kinds/build/macosx.yml
@@ -240,23 +240,25 @@ macosx64/opt:
             - sysroot-wasm32-wasi
             - onnxruntime-x86_64-apple-darwin
 
-macosx64-enterprise-shippable/opt:
-    description: "MacOS X aarch64 Enterprise Shippable Cross-compile"
-    use-pgo: macosx64-aarch64-shippable/opt
-    index:
-        product: firefox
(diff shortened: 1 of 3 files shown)


lissyx:
try at: https://firefox-ci-tc.services.mozilla.com/tasks/groups/atkDYGmVSt--cvL_tPlpgA

taskcluster/kinds/build/macosx.yml
```diff
@@ -240,50 +240,234 @@ macosx64/opt:
...
+              dest: ../artifacts
+            - artifact: target.langpack.xpi
+              extract: false
+              dest: ../artifacts
+            - artifact: mozharness.zip
+              extract: false
+              dest: ../artifacts
+            - artifact: target.trainhop.tests.tar.zst
+              extract: false
+              dest: ../artifacts
```
ahal, line 428: This strikes me as something that should be handled via transform. You can inspect the dependency task's artifacts and translate them over in a loop to avoid listing them out.

ahal (approved):
I guess if you want to handle the comment in a follow-up that's fine

taskcluster/kinds/build/macosx.yml
lissyx: This likely need to be dealt with at m-c level though, I just took existing behavior :(

PR #891, opened 2026-05-13
Bug 2039183 - Enterprise: grant scope for fennec geoloc api key

lissyx:
Similar to desktop, grant scope for fetching the Fennec geolocation API key using gecko trust domain.

### Description 

Bugzilla: Bug-2039183


ahal:
This doesn't actually grant the scope, it'll need a corresponding PR in fxci-config for that.

ahal (approved):
I'll approve because the change itself is good, but we can't land until the secret is added in fxci-config

lissyx:
> I'll approve because the change itself is good, but we can't land until the secret is added in fxci-config

@ahal ping ?

PR #904, opened 2026-05-15
Bug 2040016, Bug 2040017 - Fix test race conditions waiting for felt window by specifically looking for a new window

jonathanmendez:
### Description 

Bugzilla: Bug-2040016, Bug-2040017

These two tests logically want to trigger a condition that returns the user to the felt auth window. The tests ran through the login workflow then immediately began a wait for the felt window to be present. However, in the race condition, the original felt window used for the login has not yet disappeared, and the test begins assertions before the actual condition-under-test has occurred. This change adds the ability to wait specifically for a new felt window to appear.

---

### Testing 

* [x] Manual testing performed (failed consistently locally without the change, succeeded after the change)

### DIFF
--- a/testing/enterprise/felt_tests.py
+++ b/testing/enterprise/felt_tests.py
@@ -15,7 +15,7 @@ import tempfile
 import time
 import urllib.parse
 import uuid
-from contextlib import closing
+from contextlib import closing, contextmanager
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from multiprocessing import Array, Process, Value
 
@@ -1074,3 +1074,27 @@ class FeltTests(FeltTestsBase):
 
     def await_felt_auth_window(self):
         self._wait.until(lambda mn: len(self._driver.chrome_window_handles) == 1)
+
+    @contextmanager
+    def expect_new_felt_auth_window(self):
+        """Context manager: assert FELT's auth window is replaced by a new one.
+
+        Captures the current FELT chrome window handle on entry. On normal
+        exit, waits for that window to be replaced by a new single chrome
+        window and switches to it via force_window(). Use to wrap actions
+        whose effect is that FELT closes its auth window and then re-opens
+        a new one (e.g. login completion that spawns a child Firefox which
+        then fails, causing FELT to re-open its auth window). Ensures the
+        test awaits the new FELT window and not the original one prior to
+        closing.
+        """
+        handles = self._driver.chrome_window_handles
+        previous_handle = handles[0] if len(handles) == 1 else None
+        yield
+
+        def replaced(_):
+            handles = self._driver.chrome_window_handles
+            return previous_handle not in handles and len(handles) == 1
+
+        self._wait.until(replaced)
+        self.force_window()
--- a/testing/enterprise/test_felt_browser_init_failures.py
+++ b/testing/enterprise/test_felt_browser_init_failures.py
@@ -15,10 +15,11 @@ from felt_tests import FeltTests
 class BrowserInitFailures(FeltTests):
     def test_browser_init_policy_fetch_fail(self):
         self.policies_fail_request.value = 1
-        super().run_felt_base()
-        self._manually_closed_child = True
-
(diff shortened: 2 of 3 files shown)


testing/enterprise/test_felt_browser_init_failures.py
```diff
@@ -15,10 +15,15 @@
...
+        initial_felt_handle = self.current_felt_chrome_window()
         super().run_felt_base()
         self._manually_closed_child = True
 
-        self.await_felt_auth_window()
+        # After SSO completion, FELT closes its auth window and
+        # spawns the child Firefox. The child fails its policy fetch and
+        # exits, which causes FELT to re-open its auth window. Wait for the
+        # new auth window to open.
+        self.await_new_felt_auth_window(initial_felt_handle)
```
lissyx: I am wondering if it would be easier to follow being a context manager?

lissyx (approved):

testing/enterprise/test_felt_browser_init_failures.py
jonathanmendez: Are you envisioning something like this?
```
with self.expect_new_felt_auth_window():
    super().run_felt_base()
    self._manually_closed_child = True
self.policies_fail_request.value = 0
self.force_window()
self.assert_user_signed_out(env=Environment.FELT)
```

lissyx: Something like that yes, and I assume we would wrap `self.force_window()` in it?
This would be handled on the exit of the context manager, if it works, I think it would be better?

testing/enterprise/test_felt_browser_init_failures.py
```diff
@@ -15,10 +15,11 @@
...
         self.policies_fail_request.value = 1
-        super().run_felt_base()
-        self._manually_closed_child = True
-
-        self.await_felt_auth_window()
+        # After SSO completion, FELT closes its auth window and spawns the
+        # child Firefox. The child fails its policy fetch and exits, which
+        # causes FELT to re-open a new auth window.
+        with self.expect_new_felt_auth_window():
+            super().run_felt_base()
```
lissyx: nit:
```suggestion
            self.run_felt_base()
```

testing/enterprise/test_felt_browser_starts_missing_bin.py
```diff
@@ -45,12 +45,15 @@ def test_felt_browser_start_missing_binary(self):
...
+        # to a non-existent path, so FELT re-opens a new auth window with the
+        # error element visible.
+        with self.expect_new_felt_auth_window():
+            self.run_felt_base()
         self.run_felt_check_error_message()
 
     def run_felt_check_error_message(self):
-        self.await_felt_auth_window()
-        self.force_window()
         self._driver.set_context("chrome")
```
lissyx, line 57: There are other uses, we should file a followup to convert them

lissyx (approved):

testing/enterprise/test_felt_browser_starts_missing_bin.py
jonathanmendez: Looking at https://searchfox.org/enterprise-main/search?q=await_felt_auth_window I think all other uses have stronger wait conditions (i.e. connecting to the child browser first) and don't need to specifically worry about finding a *new* felt auth window?

testing/enterprise/felt_tests.py
```diff
@@ -1074,3 +1074,27 @@ def run_felt_perform_sso_auth(self):
...
+        test awaits the new FELT window and not the original one prior to
+        closing.
+        """
+        handles = self._driver.chrome_window_handles
+        previous_handle = handles[0] if len(handles) == 1 else None
+        yield
+
+        def replaced(_):
+            handles = self._driver.chrome_window_handles
+            return previous_handle not in handles and len(handles) == 1
```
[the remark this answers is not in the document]
jonathanmendez: The intent is to find a new window. If there is no current window, then any window is new.

PR #911, opened 2026-05-18
Bug 2039213 - Fix  customizableui tests for enterprise build

victhorlopez:
### Description 

Bugzilla: Bug-2039213

fixes:

In general tests fail with `uncaught rejection: DocumentFragment.querySelector: '#' is not a valid selector`, this is because the we have two html link elements in the pannel, and the CustomizableUI uses id's programatically. I moved those ones in browser.xhtml as other items are doing.
- browser_884402_customize_from_overflow.js uses the FxA button from the "overflowable" menu and tries to customize it, because our badge is not customizable (i.e non removable) and we do not have FxA button, I used another button
- browser_923857_customize_mode_event_wrapping_during_reset.js same as above, I just used email button
- browser_editcontrols_update.js the test does different things, the problem in here is the expectation that we reach the "overflow" state, our "enterprise badge" has usually no size on this tests, as such I gave it a fake logo, to have a "organic size" (instead of just setting some random width)
- The test `browser_976792_insertNodeInWindow.js` is fixed upstream (see dependencies)

---

* Depends on: https://bugzilla.mozilla.org/show_bug.cgi?id=2041653
* https://phabricator.services.mozilla.com/D302990

---

### Screenshots 

---

### Testing 

* [ ] Added tests
[See test pass](https://treeherder.mozilla.org/logviewer?job_id=569225516&repo=enterprise-firefox-pr&task=XQJFeFQkTeSnfKAgAWpPzw.0&lineNumber=6026)

### DIFF
--- a/browser/components/customizableui/test/browser_884402_customize_from_overflow.js
+++ b/browser/components/customizableui/test/browser_884402_customize_from_overflow.js
@@ -18,15 +18,20 @@ registerCleanupFunction(function () {
 // show a context menu with options to move it.
 add_task(async function () {
   overflowPanel.setAttribute("animate", "false");
-  let fxaButton = document.getElementById("fxa-toolbar-menu-button");
-  if (BrowserTestUtils.isHidden(fxaButton)) {
-    // FxA button is likely hidden since the user is logged out.
+
+  const buttonId = AppConstants.MOZ_ENTERPRISE
+    ? "email-link-button"
+    : "fxa-toolbar-menu-button";
+  let button = document.getElementById(buttonId);
+
+  if (BrowserTestUtils.isHidden(button)) {
+    // Button is likely hidden since the user is logged out.
     let initialFxaStatus = document.documentElement.getAttribute("fxastatus");
     document.documentElement.setAttribute("fxastatus", "signed_in");
     registerCleanupFunction(() =>
       document.documentElement.setAttribute("fxastatus", initialFxaStatus)
     );
-    ok(BrowserTestUtils.isVisible(fxaButton), "FxA button is now visible");
+    ok(BrowserTestUtils.isVisible(button), "Button is now visible");
   }
 
   let navbar = document.getElementById(CustomizableUI.AREA_NAVBAR);
@@ -48,13 +53,12 @@ add_task(async function () {
     "customizationPanelItemContextMenu"
   );
   let shownContextPromise = popupShown(contextMenu);
-  ok(fxaButton, "fxa-toolbar-menu-button was found");
-  is(
-    fxaButton.getAttribute("overflowedItem"),
-    "true",
-    "FxA button is overflowing"
+  ok(button, buttonId + " was found");
+  await TestUtils.waitForCondition(
+    () => button.getAttribute("overflowedItem") == "true",
+    buttonId + " should be overflowing"
   );
-  EventUtils.synthesizeMouse(fxaButton, 2, 2, {
+  EventUtils.synthesizeMouse(button, 2, 2, {
     type: "contextmenu",
     button: 2,
   });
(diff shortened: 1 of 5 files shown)


browser/components/customizableui/test/browser_884402_customize_from_overflow.js
```diff
@@ -48,13 +53,12 @@ add_task(async function () {
     "customizationPanelItemContextMenu"
   );
   let shownContextPromise = popupShown(contextMenu);
-  ok(fxaButton, "fxa-toolbar-menu-button was found");
-  is(
-    fxaButton.getAttribute("overflowedItem"),
-    "true",
-    "FxA button is overflowing"
+  ok(customizableButton, buttonId + " was found");
+  await TestUtils.waitForCondition(
```
victhorlopez, line 57: the original conditional here, ok(customizableButton, buttonId + " was found"); was working because the first button that overflows is the fxa, the email button takes more time to overflow, and that shows the original check was racy...

browser/components/customizableui/test/browser_884402_customize_from_overflow.js
```diff
@@ -18,15 +18,20 @@ registerCleanupFunction(function () {
...
 add_task(async function () {
   overflowPanel.setAttribute("animate", "false");
-  let fxaButton = document.getElementById("fxa-toolbar-menu-button");
-  if (BrowserTestUtils.isHidden(fxaButton)) {
-    // FxA button is likely hidden since the user is logged out.
+
+  let buttonId = AppConstants.MOZ_ENTERPRISE
+    ? "email-link-button"
+    : "fxa-toolbar-menu-button";
+  let customizableButton = document.getElementById(buttonId);
```
1rneh: Let's use constants whenever possible

```suggestion
  const buttonId = AppConstants.MOZ_ENTERPRISE
    ? "email-link-button"
    : "fxa-toolbar-menu-button";
  const customizableButton = document.getElementById(buttonId);
```

1rneh: `nit`: Can `customizableButton` be renamed to `button` since the button being customizable is no characteristics that gets tested here.

browser/components/customizableui/test/browser_923857_customize_mode_event_wrapping_during_reset.js
```diff
@@ -8,7 +8,10 @@
 add_task(async function () {
   await startCustomizing();
   let devButton = document.getElementById("developer-button");
-  let fxaButton = document.getElementById("fxa-toolbar-menu-button");
+  let fxaButtonId = AppConstants.MOZ_ENTERPRISE
+    ? "enterprise-badge-toolbar-button"
+    : "fxa-toolbar-menu-button";
+  let fxaButton = document.getElementById(fxaButtonId);
```
1rneh: Can we give the variable a generic name as it's not the fxa button in enterprise builds.to be generic 
```suggestion
  const buttonId = AppConstants.MOZ_ENTERPRISE
    ? "enterprise-badge-toolbar-button"
    : "fxa-toolbar-menu-button";
  const button = document.getElementById(buttonId);
```

browser/components/customizableui/test/browser_editcontrols_update.js
```diff
@@ -3,6 +3,28 @@
 // It also verifies that the performance optimiation implemented by
 // updateEditUIVisibility in browser.js is applied.
 
+// The enterprise badge button sits in the navbar. Without a logo it can have
+// zero width, which throws off the kForceOverflowWidthPx resize threshold used
+// in test_panelui_customize_to_toolbar to force toolbar overflow.
+if (AppConstants.MOZ_ENTERPRISE) {
+  Services.prefs.setStringPref(
```
1rneh: We can use [`SpecialPowers.pushPrefEnv`](https://searchfox.org/enterprise-main/rev/017917c10c85768bfdf90714b660bb318053a879/testing/specialpowers/content/SpecialPowersParent.sys.mjs#460-483) to save the original user value of a preference, it automatically gets restored after test execution. 

Note: I've not mentioned this in other reviews because if I recall correctly you always modified the default value of a preference in those which the API doesn't cover.

1rneh (changes requested):
Thanks, some small nits and one blocker related to using [SpecialPowers.pushPrefEnv](https://searchfox.org/enterprise-main/rev/017917c10c85768bfdf90714b660bb318053a879/testing/specialpowers/content/SpecialPowersParent.sys.mjs#460-483) when instrumenting user preferences in tests.

PR #912, opened 2026-05-18
Bug 2040374 - Fix enterprise transparent primary password after nsIPK11TokenDB removal

beurdouche:
### Description 

Bugzilla: Bug-2040374

Upstream Bug 2037682 removed @mozilla.org/security/pk11tokendb;1 and the nsIPK11TokenDB interface in favour of obtaining the internal key token directly via @mozilla.org/security/internalkeytoken;1 / nsIPKCS11Token. The EnterpriseStorageEncryption.load idle task in BrowserGlue.sys.mjs still referenced the removed contract, so it threw immediately at the Cc[...].getService(...) lookup and the softoken stayed locked — which surfaced as the primary-password prompt on browsing sessions that should have been unlocked transparently.

This commit:

- Switches BrowserGlue to the new contract and interface.
- Extracts the task body into BrowserGlue.prototype._runEnterpriseStorageEncryptionLoad so the unlock can be unit-tested directly.
- Adds an xpcshell regression guard that fails if either the legacy contract reappears or the task body reintroduces the dead names.
- Adds an end-to-end smoke test that drives the task against a mocked console backend via a new EnterprisePolicyTesting.registerPrimarySecret helper.
- Fixes a pre-existing cleanup bug in EnterprisePolicyTesting that called the non-existent ConsoleClient.clearTokenData(); replaced with Services.felt.clearTokens(), which is the correct teardown API.
-->


browser/components/BrowserGlue.sys.mjs
```diff
@@ -1438,6 +1362,82 @@ BrowserGlue.prototype = {
...
+      );
+    } catch (e) {
+      console.error(
+        "EnterpriseStorageEncryption.load: Error getting PK11 token: " + e
+      );
+      return;
+    }
+
+    // Check if the PK11 token needs initialization
+    if (pk11token.needsUserInit) {
```
mozkeeler: The internal token should never need initialization: https://searchfox.org/firefox-main/rev/8b3456fa29ad35f58da4cc0a1d182e039124ba5d/security/certverifier/NSSCertDBTrustDomain.cpp#1599-1600

beurdouche: Nice catch, I'll remove that block. Thank you !

lissyx:
@beurdouche should we merge this?

beurdouche:
I would say yes. We need to test this IRL

browser/components/enterprisepolicies/tests/xpcshell/test_enterprise_storage_encryption_contracts.js
```diff
@@ -0,0 +1,92 @@
+/* Any copyright is dedicated to the Public Domain.
```
lissyx, line 1: xpchsell are not yet reliable, can we have some coverage using `marionette-enterprise` from under `testing/enterprise` if it makes any sense?

browser/components/enterprisepolicies/tests/xpcshell/xpcshell.toml
```diff
@@ -29,6 +29,11 @@ support-files = [
 
 ["test_empty_policy.js"]
 
+["test_enterprise_storage_encryption_contracts.js"]
```
lissyx, line 35: Why does this one does not have `run-if = ["enterprise"]` ?

lissyx (approved):
There's a lot of changes that may depend on @1rneh work on policies and tests, I think it warrants their review as well

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1438,6 +1362,80 @@ BrowserGlue.prototype = {
...
+      );
+      return;
+    }
+
+    // The internal key token is unconditionally PK11_InitPin'd (with an
+    // empty password) by NSS startup in the parent process — see
+    // security/certverifier/NSSCertDBTrustDomain.cpp:1594-1603. By the time
+    // this task runs (and in any case the createInstance above forces NSS
+    // init synchronously), pk11token.needsUserInit is guaranteed to be
+    // false, so we only need to handle the two remaining states.
```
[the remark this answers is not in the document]
beurdouche: That's the thing we just removed based on Dana's review comments

beurdouche:
I think this might have already be fixed upstream and merged by now...

1rneh:
> I think this might have already be fixed upstream and merged by now...

If that's the case, could you update the patch to only add test coverage?

browser/components/enterprisepolicies/tests/xpcshell/test_enterprise_storage_encryption_contracts.js
beurdouche: Ok, made an attempt to fix this.

browser/components/enterprisepolicies/tests/xpcshell/xpcshell.toml
beurdouche: Added a comment. This should always succeed

testing/enterprise/felt_browser_storage_encryption.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+
+from felt_tests import PRIMARY_SECRET, FeltTests
+
+
+class FeltBrowserStorageEncryption(FeltTests):
```
lissyx, line 10: IS this browser specific or toolkit level?

testing/enterprise/manifest.toml
```diff
@@ -60,6 +60,8 @@ run-if = [
 
 ["test_felt_browser_starts_utf8.py"]
 
+["test_felt_browser_storage_encryption.py"]
```
lissyx, line 63: if it's browser specific it should have adequat `run-if` like others

browser/components/enterprisepolicies/tests/xpcshell/xpcshell.toml
1rneh: I suppose this should probably land upstream if the behavior isn't specific to enterprise builds. 

**Edit:** Ignore my comment. It's outdated.

1rneh (commented):
As this fix was identified as critical during our meeting, I believe the actual fix should be split into a separate PR. The newly added tests may take longer to review and approve, and shouldn't block landing the fix.

fiji-flo:
I have the minimal fix in a stash. I can PR that.

beurdouche:
@fiji-flo Ok, please do. I am not sure if anyone is reviewing this PR...

bsmth:
See:

- [x] https://github.com/mozilla/enterprise-firefox/pull/1038

beurdouche:
This was superseeded and can be closed. https://github.com/mozilla/enterprise-firefox/pull/1059

PR #914, opened 2026-05-19
Bug 2033414 - Enterprise: add update channel in profile name

lissyx:
Add the update channel to the name of the profile being used both for Felt and for browser, to avoid running into "profile downgrade" errors.

### Description 

Bugzilla: Bug-2033414

Cherry-pick of #747 for beta


gcp (approved):
c

PR #924, opened 2026-05-19
Bug 2040630 - Enterprise: Add source of profile creation as 'felt'

lissyx:
Bug-2026972 introduced a "source" parameter when creating the profile, populating it with "felt".

### Description 

Bugzilla: Bug-2040630


gcp (changes requested):
`toolkit/profile/metrics.yaml` should perhaps be updated?

`felt-firstrun` or something similar? The Firefox one is `firstrun-created-default`

lissyx:
> `toolkit/profile/metrics.yaml` should perhaps be updated?
> 
> `felt-firstrun` or something similar? The Firefox one is `firstrun-created-default`

Yes but it's a `string` so I think we can put whatever we want

Mossop (approved):

gcp (approved):

PR #925, opened 2026-05-19
Bug 2037867 - Clear policy-set default preference values

1rneh:
### Description 

Bugzilla: Bug-2037867

- Clears a preference's default value on policy removal if none existed pre-policy
- Clears any modification to a preference's user value on policy removal if the preference was locked
- Introduces a `toolkit/components/enterprisepolicies/tests/browser/live` test suite and moves and enhances the tests covering restoring the preference state after a policy removal

---

### Testing 

* [X] Added tests


toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policies_restoring_preference_state.js
jonathanmendez: This doesn't seem to match the assertion.

jonathanmendez: The comment doesn't match what's being tested (that only the default is restored, when the preference was not locked).

jonathanmendez: And I guess also the name of the test function itself implies the intent is about the state only being cached once.

jonathanmendez: nit: extra 'only'
```suggestion
    "Verifying that the initial default value is restored, because preference state was only cached once"
```

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3967,14 +3993,16 @@ export var PoliciesUtils = {
...
    */
   setDefaultPref(prefName, prefValue, locked) {
-    let prefWasLocked = Services.prefs.prefIsLocked(prefName);
+    const prefWasLocked = Services.prefs.prefIsLocked(prefName);
     if (prefWasLocked) {
       Services.prefs.unlockPref(prefName);
     }
 
     this.savePreferenceState(prefName);
+    this._initialPrefState[prefName].isLocked =
```
jonathanmendez: I'm not sure why this logic is the way it is, combining "was locked" with "will be locked". If this is meant to capture "was ever locked" then you probably want to boolean or assign `|=` the `isLocked` property (or an equivalent that handles null for the first assignment).

If that's not the intent, then the `isLocked` state seems to just reflect some odd version of "recent history". For example, imagine a series of calls `setDefaultPref("foo", 1, true); setDefaultPref("foo", 2, false); setDefaultPref("foo", 3, false);` If you restore after 1 or 2, you end up restoring the original user pref, but if you restore after 3 you leave the current user pref.

The scenario may be implausible, but this logic for setting `isLocked` still seems weird to me.

jonathanmendez (changes requested):
The comment about setting null from copilot also seems concerning.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3851,93 +3851,132 @@ export var PoliciesUtils = {
...
     const prefState = this._initialPrefState[prefName];
 
+    delete this._initialPrefState[prefName];
+
     if (!prefState) {
       // Nothing to restore
       return;
     }
 
+    const type = Services.prefs.getPrefType(prefName);
```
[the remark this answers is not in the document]
1rneh: The reason I decided to get the current preference type and don't cache the previous one, is that for a non existing preference we'd cache the type `PREF_INVALID`. And I consider it very unlikely that the type of a preference is changed.

browser/components/enterprisepolicies/Policies.sys.mjs
1rneh: It is capturing _"Is the preference in a locked state while the policy is applied?"_, because that affects how we restore the preference.

But you pointing out something unintentional! The idea is to only capture the preference state once - the initial state - and restore that on policy removal. But here I'm overriding the `isLocked` property on each `setDefaultPref` call. When it comes to policy updates, in order to clear the previous state, my idea is to first remove the previous policy (calls `restorePreferenceState`) and then re-apply it (calls `savePreferenceState`) to end up with the correct settings. So we shouldn't run into subsequent calls to save the preference state.

But something the policy engine doesn't account for yet is multiple policies modifying the same preference causing those subsequent calls. Which might be a problem that I think doesn't need to be covered in this scope.

jonathanmendez (approved):

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policies_restoring_preference_state.js
jporter-dev, line 1: Should we have test coverage for `defaultValue=null, userValue!=null, isLocked=true`? This case is handled properly, but we're missing a test for it. 

For example, I can set `browser.ipProtection.override.serverlist` before enabling the policy, and my value does get restored when the policy is removed.

jporter-dev: nit: extra space at the end of the string

jporter-dev (approved):
Thank you for handling this! It seems to be working properly in the scenarios we came across for bug 2036744. Just noted one test gap that could be covered (though not sure how common this scenario is).

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policies_restoring_preference_state.js
1rneh: I added full test coverage for the preference state matrix. I also commented the matrix for a better overview of the expected post-restore expectations.

jporter-dev: Looks good! The matrix comment helps better understand all the possible combinations

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3851,93 +3851,138 @@ export var PoliciesUtils = {
...
+      const currentUserValue = Services.prefs.prefHasUserValue(prefName)
+        ? this._readPref(Services.prefs, prefName, type)
+        : null;
+      defaults.deleteBranch(prefName);
+      if (isClearUserValue) {
+        // Removing the preference also cleared the user value
+        return;
+      }
+      // Restoring the user value which recreates the preference but without the default value
+      this._writePref(Services.prefs, prefName, type, currentUserValue);
```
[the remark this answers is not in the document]
1rneh: Good catch! I added a non-nullish guard for setting the current user value. I also added an additional test covering that the user value of a locked preference is restored correctly when being it's being cleared during the policy lifecycle.

PR #927, opened 2026-05-19
Promote 151 release

lissyx:
### Description 

Release 151 promotion


jonathanmendez:
This PR is attempting to promote `enterprise-beta` to `enterprise-release`, but today is not the day to do that (it was done for version 151 in #897 and should happen for version 152 on June 10). For now we should be doing the daily merges from upstream/release to enterprise-release

PR #931, opened 2026-05-20
Enterprise main merge 2026052000

lissyx:
### Description 

Merge of the day, no conflict


lissyx:
There is no need to update l10n changeset since `mainè is still at the same value

1rneh (approved):

lissyx:
There's one failure due to the active work on IPProtection that, we are not going to block on it

lissyx:
Follow up handled by @jporter-dev: Bug-2040983

PR #944, opened 2026-05-21
Bug 2037872 - Enterprise: Fix access connector toolbar icon size regression

jporter-dev:
### Description 

Bugzilla: Bug-2037872

- Updates the Access Connector icon CSS rule to match upstream change to target only non-overflow panel icon

---

* Depends on: #885

### DIFF
--- a/browser/themes/shared/toolbarbutton-icons.css
+++ b/browser/themes/shared/toolbarbutton-icons.css
@@ -387,7 +387,7 @@
   }
   &:not(:is([overflowedItem="true"], [cui-areatype="panel"])) {
     > .toolbarbutton-icon {
-      width: calc(2 * var(--toolbarbutton-padding-inner) + var(--icon-width));
+      width: calc(2 * var(--toolbarbutton-padding-inner) + var(--icon-size-small));
     }
   }
   &.ipprotection-on {
--- a/browser/themes/shared/toolbarbutton-icons.css
+++ b/browser/themes/shared/toolbarbutton-icons.css
@@ -384,7 +384,11 @@
   }
   > .toolbarbutton-icon {
     list-style-image: url("chrome://browser/content/ipprotection/assets/ac-icon.svg");
-    width: calc(2 * var(--toolbarbutton-padding-inner) + 16px);
+  }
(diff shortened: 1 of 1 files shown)


browser/themes/shared/toolbarbutton-icons.css
```diff
@@ -384,7 +384,11 @@
   }
   > .toolbarbutton-icon {
     list-style-image: url("chrome://browser/content/ipprotection/assets/ac-icon.svg");
-    width: calc(2 * var(--toolbarbutton-padding-inner) + 16px);
+  }
+  &:not(:is([overflowedItem="true"], [cui-areatype="panel"])) {
+    > .toolbarbutton-icon {
+      width: calc(2 * var(--toolbarbutton-padding-inner) + var(--icon-width));
+    }
```
1rneh, line 391: This seems to be the same css rule as https://searchfox.org/enterprise-main/rev/cee65aebe09d3814ffb5459068dff0de27d0b39b/browser/themes/shared/toolbarbuttons.css#735-739, besides that `--icon-width` seems to be a hallucination as I don't see that token defined anywhere?

1rneh (changes requested):
I ran it and the size doesn't look correct (see screenshot)

Image

Could you include a screenshot with any future updates?

browser/themes/shared/toolbarbutton-icons.css
jporter-dev: It did [previously exist](https://searchfox.org/enterprise-main/rev/00689b9f9fd00c90e9066a83c6332d617e049e0e/browser/themes/shared/toolbarbuttons.css#737) in the other css file, I looked at the wrong line in the git diff when fixing this to use a token instead of hardcoded 16px and must have been out of sync when testing locally. Thanks for catching this!

I updated it to use the proper token for a 16px icon `--icon-size-small`

jporter-dev:
@1rneh sorry about the issue, should be good now! 

access-connector-icon-fixed

1rneh (approved):

1rneh:
Looks good now! Thank you!

PR #948, opened 2026-05-22
Bug 2031290 - Enterprise: PackageKitUpdateChecker

lissyx:
### Description 

Bugzilla: Bug-2031290

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [x] Added tests

### DIFF
--- a/toolkit/mozapps/update/AppUpdater.sys.mjs
+++ b/toolkit/mozapps/update/AppUpdater.sys.mjs
@@ -255,8 +255,10 @@ export class AppUpdater {
           "AppUpdater:check -" +
             "AppConstants.MOZ_UPDATER=" +
             AppConstants.MOZ_UPDATER +
-            "this.#updateDisabledByPackage: " +
-            this.#updateDisabledByPackage
+            " this.#updateDisabledByPackage: " +
+            this.#updateDisabledByPackage +
+            " this.aus: " +
+            this.aus
         );
         this.#setStatus(AppUpdater.STATUS.NO_UPDATER);
         return;
@@ -421,7 +423,12 @@ export class AppUpdater {
   // may not be, and we don't have a good way to tell the difference from here,
   // so we err to the side of less confusion for unmanaged users.
   get #updateDisabledByPackage() {
-    return Services.sysinfo.getProperty("isPackagedApp");
+    if (this.aus?.supportsUpdates) {
+      // There is an update service, and it is willing to handle updates now.
+      return false;
+    }
+    // Either there is no update service, or it is not willing to handle updates.
+    return true;
   }
 
   /**
@@ -778,24 +785,50 @@ export class AppUpdater {
   }
 }
 
-XPCOMUtils.defineLazyServiceGetter(
-  AppUpdater.prototype,
-  "aus",
-  "@mozilla.org/updates/update-service;1",
-  Ci.nsIApplicationUpdateService
-);
-XPCOMUtils.defineLazyServiceGetter(
-  AppUpdater.prototype,
-  "checker",
-  "@mozilla.org/updates/update-checker;1",
-  Ci.nsIUpdateChecker
-);
-XPCOMUtils.defineLazyServiceGetter(
-  AppUpdater.prototype,
-  "um",
-  "@mozilla.org/updates/update-manager;1",
-  Ci.nsIUpdateManager
-);
+if (Services.prefs.getBoolPref("app.update.use_package_kit", false)) {
+  LOG(
+    `UpdateService.sys.mjs: @mozilla.org/updates/packagekit-update-checker;1`
+  );
+  XPCOMUtils.defineLazyServiceGetter(
+    AppUpdater.prototype,
+    "aus",
+    "@mozilla.org/updates/packagekit-update-service;1",
+    Ci.nsIApplicationUpdateService
+  );
+  XPCOMUtils.defineLazyServiceGetter(
+    AppUpdater.prototype,
+    "checker",
+    "@mozilla.org/updates/packagekit-update-checker;1",
+    Ci.nsIUpdateChecker
+  );
+  XPCOMUtils.defineLazyServiceGetter(
+    AppUpdater.prototype,
+    "um",
+    // "@mozilla.org/updates/packagekit-update-manager;1",
+    "@mozilla.org/updates/update-manager;1",
+    Ci.nsIUpdateManager
+  );
+} else {
+  LOG(`UpdateService.sys.mjs: @mozilla.org/updates/update-checker;1`);
+  XPCOMUtils.defineLazyServiceGetter(
+    AppUpdater.prototype,
+    "aus",
+    "@mozilla.org/updates/update-service;1",
+    Ci.nsIApplicationUpdateService
+  );
+  XPCOMUtils.defineLazyServiceGetter(
+    AppUpdater.prototype,
+    "checker",
+    "@mozilla.org/updates/update-checker;1",
+    Ci.nsIUpdateChecker
+  );
+  XPCOMUtils.defineLazyServiceGetter(
+    AppUpdater.prototype,
+    "um",
+    "@mozilla.org/updates/update-manager;1",
+    Ci.nsIUpdateManager
+  );
+}
 
 AppUpdater.STATUS = {
   // Updates are allowed and there's no downloaded or staged update, but the
--- a/toolkit/mozapps/update/UpdateService.sys.mjs
+++ b/toolkit/mozapps/update/UpdateService.sys.mjs
@@ -26,24 +26,51 @@ ChromeUtils.defineESModuleGetters(lazy, {
   setTimeout: "resource://gre/modules/Timer.sys.mjs",
 });
 
-XPCOMUtils.defineLazyServiceGetter(
-  lazy,
-  "AUS",
-  "@mozilla.org/updates/update-service;1",
-  Ci.nsIApplicationUpdateService
-);
-XPCOMUtils.defineLazyServiceGetter(
-  lazy,
-  "UM",
-  "@mozilla.org/updates/update-manager;1",
-  Ci.nsIUpdateManager
-);
-XPCOMUtils.defineLazyServiceGetter(
-  lazy,
-  "CheckSvc",
-  "@mozilla.org/updates/update-checker;1",
-  Ci.nsIUpdateChecker
-);
+if (Services.prefs.getBoolPref("app.update.use_package_kit", false)) {
+  LOG(
+    `UpdateService.sys.mjs: @mozilla.org/updates/packagekit-update-checker;1`
+  );
+  XPCOMUtils.defineLazyServiceGetter(
+    lazy,
+    "CheckSvc",
+    "@mozilla.org/updates/packagekit-update-checker;1",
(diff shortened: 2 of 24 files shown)


toolkit/components/enterprise/modules/Updates.sys.mjs
cpdupuis, line 70: The delay here is OK for getting the first version checked in, but I recommend opening a bug to figure out how to control this more precisely.

One thing that might help with preventing the update from happening automatically is the pref `app.update.disabledForTesting`

cpdupuis, line 465: pending-elevate is a macos-specific state. pending-service is a Window-specific state. pending is a general-purpose state.

All 3 of these should probably be covered here.
See https://firefox-source-docs.mozilla.org/toolkit/mozapps/update/docs/Concepts.html#update-state

cpdupuis, line 464: Instead of putting a switch directly inside another switch, can you please move this inner switch statement to a separate function?

cpdupuis, line 493: This is another switch to extract to a separate function

toolkit/mozapps/update/packagekit/nsIPackageKitDBusProvider.idl
```diff
@@ -0,0 +1,137 @@
...
+   * Common values:
+   *   1 = unknown
+   *   6 = normal
+   *   7 = important
+   *   8 = security
+   *   9 = bugfix
+   *  10 = enhancement
+   *  11 = blocked
+   *
+   * See: https://www.freedesktop.org/software/PackageKit/gtk-doc/PackageKit-enumerations.html
```
cpdupuis: Your link returns a 404. Maybe their URL is case-sensitive? This one seems to work:

https://www.freedesktop.org/software/PackageKit/gtk-doc/PackageKit-Enumerations.html

toolkit/mozapps/update/packagekit/PackageKitDBusProvider.cpp
```diff
@@ -0,0 +1,1259 @@
...
+//                          RefPtr<PromiseType::Private> promise,
+//                          GCancellable* cancellable)
+// ---------------------------------------------------------------------------
+
+template <typename PromiseType, typename StartTransactionFn>
+static RefPtr<PromiseType> PackageKitPromise(
+    GCancellable* aCancellable, StartTransactionFn aStartTransaction) {
+  auto promise = MakeRefPtr<typename PromiseType::Private>(__func__);
+
+  CreateDBusProxyForBus(G_BUS_TYPE_SYSTEM,
```
cpdupuis, line 928: This is a lot of complexity to put into a static method.

How are you planning to unit test this?

toolkit/mozapps/update/packagekit/PackageKitDBusProvider.cpp
```diff
@@ -0,0 +1,1259 @@
...
+//   - an array of results (type differs)
+//   - a transaction proxy kept alive until Finished
+//   - Finished + ErrorCode subscription IDs (common to all transactions)
+//   - up to 2 operation-specific subscription IDs (mSignalId1/2)
+//
+// Using a template avoids duplicating the struct and UnsubscribeAll for each
+// operation, and lets HandleErrorCodeSignal / HandleFinishedSignal be shared.
+// ---------------------------------------------------------------------------
+
+template <typename PromisePrivateType, typename ResultType>
```
cpdupuis, line 194: This entire file should have unit testing at the function level.

toolkit/mozapps/update/AppUpdater.sys.mjs
```diff
@@ -421,6 +421,10 @@ export class AppUpdater {
   // may not be, and we don't have a good way to tell the difference from here,
   // so we err to the side of less confusion for unmanaged users.
   get #updateDisabledByPackage() {
+    if (Services.prefs.getBoolPref("app.update.use_package_kit", false)) {
```
cpdupuis: Make this function depend on the instantiated UpdateService implementation, if any:

```
get #updateDisabledByPackage() {
  if (lazy.AUS && lazy.AUS.supportsUpdates()) {
    // There is an update service, and it is willing to handle updates now.
    return false;
  } else {
    // Either there is no update service, or it is not willing to handle updates.
    return true;
  }
}
```

Then implement `supportsUpdates` in `PackageKitUpdateService.sys.mjs` to return true for the cases where you want it to be enabled (presumably where the app.update.use_package_kit pref is true), and the implementation for UpdateService.sys.mjs would be:

```
supportsUpdates() {
   return !Services.sysinfo.getProperty("isPackagedApp");
}
```

toolkit/mozapps/update/AppUpdater.sys.mjs
```diff
@@ -778,24 +782,50 @@ export class AppUpdater {
...
-  "@mozilla.org/updates/update-checker;1",
-  Ci.nsIUpdateChecker
-);
-XPCOMUtils.defineLazyServiceGetter(
-  AppUpdater.prototype,
-  "um",
-  "@mozilla.org/updates/update-manager;1",
-  Ci.nsIUpdateManager
-);
+if (Services.prefs.getBoolPref("app.update.use_package_kit", false)) {
```
cpdupuis, line 788: Is there a way to distinguish that this build is an enterprise build, or that this installation uses felt, or something that would be a clear differentiator from an consumer build?

I'd rather use something like that to control switching between these implementations.

toolkit/mozapps/update/UpdateService.sys.mjs
```diff
@@ -26,24 +26,51 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
-  "@mozilla.org/updates/update-manager;1",
-  Ci.nsIUpdateManager
-);
-XPCOMUtils.defineLazyServiceGetter(
-  lazy,
-  "CheckSvc",
-  "@mozilla.org/updates/update-checker;1",
-  Ci.nsIUpdateChecker
-);
+if (Services.prefs.getBoolPref("app.update.use_package_kit", false)) {
```
cpdupuis, line 29: Assuming we're aiming for the separation of implementations we discussed, this file should not have any checks for `app.update.use_package_kit`.

toolkit/mozapps/update/UpdateServiceStub.sys.mjs
```diff
@@ -116,14 +116,26 @@ export class UpdateServiceStub {
   }
 
   async #initUpdate() {
+    let updateCheckerContract = "@mozilla.org/updates/update-checker;1";
+    let updateServiceContract = "@mozilla.org/updates/update-service;1";
+    let updateManagerContract = "@mozilla.org/updates/update-manager;1";
+    if (Services.prefs.getBoolPref("app.update.use_package_kit", false)) {
```
cpdupuis, line 122: I think there should be a way to put all of these service lookups into one place and use that everywhere.

Something like:
```
class UpdateServiceFactory {
   defineLazyGetter(lazy, name, service_nickname, type) {
     const contract = #getContractByImpl(service_nickname);
      XPCOMUtils.defineLazyServiceGetter(lazy, name, contract, type);
  }
  #getContractByImpl(service_nickname) {
      const packageKitServices = {
        "update-service": "@mozilla.org/updates/packagekit-update-service;1",
       "update-checker": ...
      }
      const mozUpdaterServices = {
         "update-service": "@mozilla.org/updates/update-service;1",
        "update-checker": ...
      }
   if (!isPackagedApp) {
    return mozUpdaterServices[service_nickname];
  } else {
    return packageKitServices[service_nickname];
}
```

Something like that.

cpdupuis (changes requested):

toolkit/mozapps/update/packagekit/nsIPackageKitDBusProvider.idl
lissyx: GTK Doc seems to have that behavior, the link was present in my history with correct title :(

toolkit/mozapps/update/packagekit/PackageKitDBusProvider.cpp
lissyx: Yes, but `gtest` were not a thing back then on our builds so there was no way to do it at all ... I've enabled them and if the code moves from a PoC to something we land, I'll add coverage

toolkit/mozapps/update/AppUpdater.sys.mjs
lissyx: The limit here is that the enterprise build for non debian and debian package is the same (it's a repack, and we want to keep it this way), so we cannot rely only on `AppConstants.MOZ_ENTERPRISE` to switch implementation. We could check the pref only when `AppCoinstants.MOZ_ENTERPRISE` is true though.

toolkit/mozapps/update/packagekit/PackageKitDBusProvider.cpp
lissyx: Also on the topic of testing those interfaces, since `gtest` were not available, it's partly covered by:
 - `testing/enterprise/test_felt_browser_packagekit_no_firefox_update.py`
 - `testing/enterprise/test_felt_browser_packagekit_one_firefox_update.py`

It's not unit testing, but it allows to have a bit of confidence already.

toolkit/mozapps/update/AppUpdater.sys.mjs
lissyx: Thanks, done

lissyx: (except it's `this.aus` and not `lazy.AUS`)

lissyx: I'm debugging but when I update `UpdateServiceStub` code to be the same as your suggestion for `UpdateService`, I have a test failure:
```
 0:18.64 INFO patch:{'QueryInterface': {}, 'type': 'complete', 'URL': None, 'finalURL': None, 'size': 1048576, 'state': 'pending-service', 'errorCode': 0, 'selected': True, 'serialize': {}}
```

If I keep my `#updateDisabledByPackage()` in `UpdateServiceStub` the patch we get is correct:
```
 0:18.40 INFO patch:{'QueryInterface': {}, 'type': 'complete', 'URL': 'dbus://packagekit/firefoxenterprise;153.0.1;x86_64;updates/firefoxenterprise-153.0.1.x86_64.deb', 'finalURL': 'dbus://packagekit/firefoxenterprise;153.0.1;x86_64;updates/firefoxenterprise-153.0.1.x86_64.deb', 'size': 1048576, 'state': 'succeeded', 'errorCode': 0, 'selected': True, 'serialize': {}}
```

toolkit/mozapps/update/packagekit/PackageKitDBusProvider.cpp
lissyx: Likely not a good answer, but as much as I recall, there's no existing tooling that allows to properly test DBus-related things, I'm looking at how we can do something in that place. (we'd need a C++ level mock of PackageKit interface)

Keep in mind there's coverage with our python mock and tests. Not perfect but easier to do because the python dbus mock tooling is existing.

PR #950, opened 2026-05-22
Bug 2041830 - Fix syntax: Use ternary operator instead of incorrect use of object spread syntax to pass parameter

jonathanmendez:
### Description 

Bugzilla: Bug-2041830

The intent is to pass the single `extraHeaders` parameter as either `{}` (non-Enterprise) or populated with the bearer token (Enterprise). The `...` spread syntax as currently applied in the parameter list attempts to expand the object into the parameter list. It's clearer to instead use a ternary operator to determine what arg to pass.

---

### Dependencies / Related Issues 

* Regressed in: #770

PR #955, opened 2026-05-26
Bug 2042349 - Fix test_sync_auth_manager.js for enterprise builds

victhorlopez:
### Description 

Bugzilla: Bug-2042349

test_sync_auth_manager.js fails because the consoleClient throws when it attempts to call felt.getAccessTokenIfValid()

This patch adds an add_setup block that, only on enterprise builds:

- Sets a default identity.sync.tokenserver.uri pref so the token server client has a valid endpoint to work with. (i.e it just expect a url)
- Prevents the call to happen when it is not felt managed

---

* Depends on: https://github.com/mozilla/enterprise-firefox/pull/950

---

### Screenshots 

---

### Testing 

* [ ] Added tests

[See test failing](https://treeherder.mozilla.org/logviewer?job_id=568748340&repo=enterprise-firefox-pr&task=IPwSNZK_SMWqGDP4qIq2gQ.0&lineNumber=5535)
[See test pass](https://treeherder.mozilla.org/logviewer?job_id=568764996&repo=enterprise-firefox-pr&task=dhxXZJu6QyyF96_s-DZIjg.0&lineNumber=4742)

### DIFF
--- a/services/fxaccounts/FxAccountsClient.sys.mjs
+++ b/services/fxaccounts/FxAccountsClient.sys.mjs
@@ -797,7 +797,7 @@ FxAccountsClient.prototype = {
         method,
         credentials,
         jsonPayload,
-        AppConstants.MOZ_ENTERPRISE
+        AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltBrowser()
           ? {
               Authorization: `Bearer ${await lazy.ConsoleClient.getAccessToken()}`,
             }
--- a/services/sync/tests/unit/test_sync_auth_manager.js
+++ b/services/sync/tests/unit/test_sync_auth_manager.js
@@ -36,6 +36,28 @@ const { TokenServerClient, TokenServerClientServerError } =
 const { AccountState, ERROR_INVALID_ACCOUNT_STATE } =
   ChromeUtils.importESModule("resource://gre/modules/FxAccounts.sys.mjs");
 
+add_setup(function () {
+  const { AppConstants } = ChromeUtils.importESModule(
+    "resource://gre/modules/AppConstants.sys.mjs"
+  );
+  if (AppConstants.MOZ_ENTERPRISE) {
+    // Enterprise builds need a default tokenserver URI, otherwise
+    // TokenServerClient rejects the empty/missing URL.
+    const defaultBranch = Services.prefs.getDefaultBranch("");
+    const originalUri = defaultBranch.getStringPref(
+      "identity.sync.tokenserver.uri",
+      ""
+    );
+    defaultBranch.setStringPref(
+      "identity.sync.tokenserver.uri",
+      "https://token.services.mozilla.com/1.0/sync/1.5"
+    );
+    registerCleanupFunction(() => {
+      defaultBranch.setStringPref("identity.sync.tokenserver.uri", originalUri);
+    });
+  }
+});
+
 const SECOND_MS = 1000;
 const MINUTE_MS = SECOND_MS * 60;
 const HOUR_MS = MINUTE_MS * 60;


1rneh (commented):
I had a conversation with @lissyx about which kind of enterprise build we are running the m-c tests against and the conclusion was that setting `MOZ_BYPASS_FELT` runs a browser that should not be launched nor managed by felt - it should be "decoupled" from felt - hence the browser shouldn't call into felt.

Which call to the `ConsoleClient` is it that triggers the failing call to `Services.felt.getAccessTokenIfValid`?

victhorlopez:
> I had a conversation with @lissyx about which kind of enterprise build we are running the m-c tests against and the conclusion was that setting `MOZ_BYPASS_FELT` runs a browser that should not be launched nor managed by felt - it should be "decoupled" from felt - hence the browser shouldn't call into felt.
> 
> Which call to the `ConsoleClient` is it that triggers the failing call to `Services.felt.getAccessTokenIfValid`?

I see.... this is for the code that gets the token ( the one from here)

https://github.com/mozilla/enterprise-firefox/pull/950/changes , let me double check ( but if you add the !isFeltBrowser() ) or similar you will not detect regressions

Edit: yeah it's that call

lissyx:
> > I had a conversation with @lissyx about which kind of enterprise build we are running the m-c tests against and the conclusion was that setting `MOZ_BYPASS_FELT` runs a browser that should not be launched nor managed by felt - it should be "decoupled" from felt - hence the browser shouldn't call into felt.
> > Which call to the `ConsoleClient` is it that triggers the failing call to `Services.felt.getAccessTokenIfValid`?
> 
> I see.... this is for the code that gets the token ( the one from here)
> 
> https://github.com/mozilla/enterprise-firefox/pull/950/changes , let me double check ( but if you add the !isFeltBrowser() ) or similar you will not detect regressions
> 
> Edit: yeah it's that call

I guess it should be guarded by `Services.felt?.isFeltBrowser()` then?

lissyx:
> > > I had a conversation with @lissyx about which kind of enterprise build we are running the m-c tests against and the conclusion was that setting `MOZ_BYPASS_FELT` runs a browser that should not be launched nor managed by felt - it should be "decoupled" from felt - hence the browser shouldn't call into felt.
> > > Which call to the `ConsoleClient` is it that triggers the failing call to `Services.felt.getAccessTokenIfValid`?
> > 
> > 
> > I see.... this is for the code that gets the token ( the one from here)
> > https://github.com/mozilla/enterprise-firefox/pull/950/changes , let me double check ( but if you add the !isFeltBrowser() ) or similar you will not detect regressions
> > Edit: yeah it's that call
> 
> I guess it should be guarded by `Services.felt?.isFeltBrowser()` then?

Note that doing so will bypass our specificities, there may be a need for an enterprise-specific test here.

services/fxaccounts/FxAccountsClient.sys.mjs
```diff
@@ -797,7 +797,7 @@ FxAccountsClient.prototype = {
         method,
         credentials,
         jsonPayload,
-        AppConstants.MOZ_ENTERPRISE
+        AppConstants.MOZ_ENTERPRISE && Services.felt.isFeltBrowser()
```
lissyx:
```suggestion
        AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltBrowser()
```

victhorlopez:
> > > > I had a conversation with @lissyx about which kind of enterprise build we are running the m-c tests against and the conclusion was that setting `MOZ_BYPASS_FELT` runs a browser that should not be launched nor managed by felt - it should be "decoupled" from felt - hence the browser shouldn't call into felt.
> > > > Which call to the `ConsoleClient` is it that triggers the failing call to `Services.felt.getAccessTokenIfValid`?
> > > 
> > > 
> > > I see.... this is for the code that gets the token ( the one from here)
> > > https://github.com/mozilla/enterprise-firefox/pull/950/changes , let me double check ( but if you add the !isFeltBrowser() ) or similar you will not detect regressions
> > > Edit: yeah it's that call
> > 
> > 
> > I guess it should be guarded by `Services.felt?.isFeltBrowser()` then?
> 
> Note that doing so will bypass our specificities, there may be a need for an enterprise-specific test here.

Created bug https://bugzilla.mozilla.org/show_bug.cgi?id=2043312

lissyx (approved):

services/sync/tests/unit/test_sync_auth_manager.js
```diff
@@ -36,6 +36,28 @@ const { TokenServerClient, TokenServerClientServerError } =
...
+    // Enterprise builds need a default tokenserver URI, otherwise
+    // TokenServerClient rejects the empty/missing URL.
+    const defaultBranch = Services.prefs.getDefaultBranch("");
+    const originalUri = defaultBranch.getStringPref(
+      "identity.sync.tokenserver.uri",
+      ""
+    );
+    defaultBranch.setStringPref(
+      "identity.sync.tokenserver.uri",
+      "http://token-server"
```
1rneh: Can we set it to the url that m-c sets as default preference, that's `https://token.services.mozilla.com/1.0/sync/1.5`? Just as a preventive measure so that any additional checks in the future don't break us, e.g. a protocol check.

1rneh (approved):
Just one small `nit`.

PR #958, opened 2026-05-26
Ensure `./mach try ` displays enterprise specific tasks by default

ahal:
Currently `mach try` is using the default gecko-specific parameters to generate tasks locally. This means many enterprise specific tasks are being filtered out at the target phase. Instead make `mach try` on enterprise point to a parameter set pulled from a push to `enterprise-main`. This way all enterprise specific tasks will appear.

### DIFF
--- a/python/mach/mach/settings.py
+++ b/python/mach/mach/settings.py
@@ -121,6 +121,12 @@ class MachSettings:
                     "Remote name or url to push to.",
                     "git@github.com:mozilla/enterprise-firefox-try.git",
                 ),
+                (
+                    "try.params",
+                    "string",
+                    "Default set of parameters to generate tasks with.",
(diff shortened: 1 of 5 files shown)


python/mach/mach/settings.py
```diff
@@ -121,6 +121,12 @@ def try_config_settings():
                     "Remote name or url to push to.",
                     "git@github.com:mozilla/enterprise-firefox-try.git",
                 ),
+                (
+                    "try.params",
+                    "string",
+                    "Default set of parameters to generate tasks with.",
+                    "taskcluster/test/params/em-onpush.yml",
```
lissyx, line 128: So I assume this is updated by the `update.sh` script next to it? Who/when should it be ran?

lissyx (approved):

PR #960, opened 2026-05-27
Bug 2041210 - Move FELT UI out of being a system addon

lissyx:
### Description 

Bugzilla: Bug-2041210

Refactoring Felt out of being a system addon, and moving relevant parts to `toolkit/components` rather than living under `browser/components`.

### Scope

Non Firefox apps, e.g. Thunderbird, may not support system addons that FELT is relying on as of now. While Thunderbird may be getting it, it's still early. There is no real technical reason for felt being an addon, other than historical (early days, exploration showed we could package it as an addon that may be integration in any firefox build). The change was a bit invasive, and not required. Thunderbird support makes it relevant.

Given the move out of system addon, it requires also:
 - morphing into an XPCOM bound to the `app-startup` category

Given the targetting of non browser app, it requires also:
 - moving around things from `browser/` to `toolkit/` since they need to be available to any app

Target scope here is:
 - Ensuring Firefox does not regress
 - Ensuring basic Thunderbird can work

In light of this context, there may be things that were moved to `toolkit/` that needs to be revisited in the future, and others that are continuing to live under `browser/` that may need to move under `toolkit/`. However this depends on the feature scope for Thunderbird, which is not yet completely defined.

The amount of changes is not too big but obviously critical, so we want this to land as soon as possible to have enough time to exercize/debug. Any follow up required will be filed as blocking `firefox-client-mvp` immediately.

The PR is split in multiple commits that should ease reviewing, but there is no expectation that this split is correct in term of bisection. I expect to squash into one commit when ready.

### Remaining tasks

* [x] Fix `marionette-enterprise` execution on CI
* [x] Fix other tests execution on CI
* [x] Refactor so that moving to `toolkit/` is the last commit of the chain
* [x] Extract independant changes into separate bugs / PRs (to be listed below)
* [x] Sync with l10n to be ready for the move

### L10N Changes

Checked with L10n team, they need the PR to land to be able to perform the required changes on their side. There should be no requirement of renaming IDs.

### Status

* [x] Passes tests on Firefox
* [x] Passes tests on Thunderbird

CI is not yet up on Thunderbird side, although early PRs are running, e.g.: https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-pr&revision=130fe973b73f18d26cb4be73cf3e906d8fb698e2&searchStr=marionette-enterprise%2Cthunderbird
This needs to be read as: `MnEnt(comm, gecko)` where:
 - `comm` is Thunderbird-specific test set, very basic for now: one test that checks that FELT SSO flow works and launches Thunderbird, another that verifies Firefox Accounts can be used.
 - `gecko` is the `testing/enterprise/manifest.toml` from Firefox repo, executing tests that can run against Thunderbird

* Depends on:
 - #985 
 - #983 
 - #982 
 - #981 
 - #980 
 - #994 
 - #1015


toolkit/components/enterprise/moz.build
```diff
@@ -0,0 +1,20 @@
...
+    BUG_COMPONENT = ("Firefox Enterprise", "Client")
+
+EXTRA_JS_MODULES.enterprise += [
+    "EnterpriseCommon.sys.mjs",
+    "EnterpriseEndpoints.sys.mjs",
+    "EnterpriseHandler.sys.mjs",
+    "modules/ConsoleClient.sys.mjs",
+    "modules/EnterpriseAccountsStorage.sys.mjs",
+    "modules/EnterpriseOSInfo.sys.mjs",
+    "modules/Updates.sys.mjs",
```
1rneh, line 16: We have this `modules` subdirectory that contains some modules, but it hasn't been used consistently for storing them. But since we only have system modules at the moment, we could remove the subdirectory and place them directly in the parent folder instead.

toolkit/components/felt/content/felt.xhtml
```diff
@@ -15,11 +15,11 @@
     <meta charset="utf-8" />
     <meta http-equiv="Content-Security-Policy" content="default-src chrome:" />
     <link rel="stylesheet" href="chrome://felt/content/styles/felt.css" />
-    <link rel="localization" href="browser/enterprise/felt.ftl" />
+    <link rel="localization" href="toolkit/enterprise/felt.ftl" />
     <link rel="localization" href="branding/brand.ftl" />
     <link rel="localization" href="toolkit/global/textActions.ftl" />
-    <link rel="localization" href="browser/browser.ftl" />
-    <link rel="localization" href="browser/enterprise/enterprise.ftl" />
+    <!-- <link rel="localization" href="browser/browser.ftl" /> -->
```
1rneh: Do we even need strings from `browser/browser.ftl` here? Otherwise this comment can be removed.

browser/moz.configure
```diff
@@ -79,3 +52,16 @@ def requires_stub_installer(
 imply_option("MOZ_STUB_INSTALLER", True, when=requires_stub_installer)
 
 include("../toolkit/moz.configure")
+
+@depends("--enable-enterprise", "--enable-project")
+def firefox_enterprise(enterprise, project):
```
1rneh: `nit:` 
```suggestion
def is_firefox_enterprise(enterprise, project):
```

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -2,110 +2,65 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
1rneh, line 4: `"use strict";` can go back :)

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -250,14 +205,17 @@ this.felt = class extends ExtensionAPI {
     );
   }
 
-  async onStartup() {
+  async #init() {
+    if (this._ready) {
+      lazy.log.info(`Felt is already ready.`);
+      return;
+    }
```
1rneh: Why do we need this guard?

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -2,110 +2,65 @@
...
+  // QI — implement nsISupports + nsIObserver
+  QueryInterface = ChromeUtils.generateQI(["nsIObserver"]);
 
-    const manifestURI = Services.io.newURI(
-      "manifest.json",
-      null,
-      this.extension.rootURI
-    );
+  constructor() {
+    lazy.log.debug(`constructor()`);
```
1rneh: Left over debug message.

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -2,110 +2,65 @@
...
-      parent: {
-        esModuleURI: "chrome://felt/content/FeltProcessParent.sys.mjs",
-      },
-    });
+
+  // nsIObserver
+  observe(_subject, topic, _data) {
+    switch (topic) {
+      case "app-startup":
+        Services.obs.addObserver(this, "profile-after-change");
```
1rneh: Can't we add "profile-after-change" to the cateogries in `components.conf`?

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -2,110 +2,65 @@
...
+    "resource://gre/modules/enterprise/EnterpriseCommon.sys.mjs",
   WebAuthnPromptHelper:
     "moz-src:///toolkit/modules/WebAuthnPromptHelper.sys.mjs",
+
+  // These two should be it only in browser app, hitting elsewhere should fail,
+  // be diagnosed, and calling code guarded appropriately with isBuildAppBrowser()
+  // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
+  FELT_OPEN_WINDOW_DISPOSITION: "resource:///modules/FeltURLHandler.sys.mjs",
+  // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
+  BrowserWindowTracker: "resource:///modules/BrowserWindowTracker.sys.mjs",
```
1rneh: Can we also make sure the lazy imports are only added in the browser app?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -6,14 +6,24 @@ const lazy = {};
...
-  FeltStorage: "resource:///modules/FeltStorage.sys.mjs",
+  FeltStorage: "resource://gre/modules/FeltStorage.sys.mjs",
+
+  /* Call to those should be properly gated by isBuildAppBrowser() */
+  // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
+  gFeltPendingURLs: "resource:///modules/FeltURLHandler.sys.mjs",
+  // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
+  resetFeltFirefoxWindowReady: "resource:///modules/FeltURLHandler.sys.mjs",
+  // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
+  FELT_OPEN_WINDOW_DISPOSITION: "resource:///modules/FeltURLHandler.sys.mjs",
```
1rneh: Can we also make sure the lazy imports are only added in the browser app?

toolkit/components/felt/Felt.sys.mjs
1rneh: This is an ExtensionAPI lifecycle hook. As an XPCOM, nothing calls it.

1rneh: It's not clear to me why we have only run all the clean up logic below if the addon was uninstalled but not on app shutdown?

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -2,110 +2,65 @@
...
 
 ChromeUtils.defineLazyGetter(lazy, "log", () => {
   return lazy.createEnterpriseLogger("Felt");
 });
 
-this.felt = class extends ExtensionAPI {
-  FELT_PROCESS_ACTOR = "FeltProcess";
-  FELT_WINDOW_ACTOR = "FeltWindow";
-  FELT_ERROR_WINDOW_ACTOR = "FeltErrorWindow";
+const FeltCls = class {
```
1rneh:
```suggestion
export class Felt {
```

toolkit/components/felt/Felt.sys.mjs
1rneh, line 1: `Felt.sys.mjs` and `FeltProcessParent.sys.mjs` still log/name things as if it was still an add-on, e.g "FeltExtension:" log prefixes, Services.felt API names like `sendExtensionReady`and comments like "Notify that extension is ready". Can we leave out the extension naming bits?

browser/components/DesktopActorRegistry.sys.mjs
```diff
@@ -860,6 +846,26 @@ let JSWINDOWACTORS = {
   },
 };
 
+if (
+  !AppConstants.MOZ_ENTERPRISE ||
+  (AppConstants.MOZ_ENTERPRISE && !Services.felt?.isFeltUI())
```
1rneh: The `(AppConstants.MOZ_ENTERPRISE &&` part is redundant. 

When evaluating `(AppConstants.MOZ_ENTERPRISE && !Services.felt?.isFeltUI())` we already know that `AppConstants.MOZ_ENTERPRISE` is `true` otherwise the first statement `!AppConstants.MOZ_ENTERPRISE` would have been true. Same in `ActorManagerParent.sys.mjs`

toolkit/components/felt/moz.build
1rneh, line 6: Does this also need a `BUG_COMPONENT`?

```
with Files("**"):
    BUG_COMPONENT = ("Firefox Enterprise", "Client")
```

toolkit/modules/ActorManagerParent.sys.mjs
```diff
@@ -811,6 +816,76 @@ if (AppConstants.platform != "android") {
...
+          register();
+          isRegistered = true;
+        } else {
+          unregister();
+          isRegistered = false;
+        }
+      };
+
+      maybeRegister();
+    },
```
1rneh: `onAddActor` registers the actor asynchronously here, which is racy "in theory".

The `await ConsoleClient.ssoCallbackUriMatchPattern` defers the `register()` call and `_addActors` [doesn't await `onAddActor`](https://searchfox.org/enterprise-main/rev/71621ae4be192aa8fbdedcb0d1209b054874978d/toolkit/modules/ActorManagerParent.sys.mjs#846), and `RegisterWindowActor` [only wires up completely registered actors with existing content processes](https://searchfox.org/enterprise-main/rev/5fe866843302e864dd7b4fee3ee2d1acb902dd87/dom/ipc/jsactor/JSActorService.cpp#95-106). So if the sso callback window global is created before the deferred `register()` runs, it never gets a `FeltWindow` actor. It currently only works like this because both - actor registration and the sso navigation - block on the same preference and the navigation to the sso callback is "the longer path".

Another idea could be to register the actor synchronously instead, like the [CookieBanner actor](https://searchfox.org/enterprise-main/rev/71621ae4be192aa8fbdedcb0d1209b054874978d/toolkit/modules/ActorManagerParent.sys.mjs#315-347) does it. We'd observe the console address preference and build `matches` when it's changes. The callback uri isn't reached before the console address is set anyway.

toolkit/components/enterprise/EnterpriseHandler.sys.mjs
1rneh, line 1: `EnterpriseHandler.sys.mjs` has been moved to `toolkit/` but handles predominantly browser-specific content, it likely needs to be split up into toolkit logic and browser-specific logic?

toolkit/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
1rneh, line 1: I think `EnterpriseAccountsStorage.sys.mjs` doesn't belong into `toolkit/` nor `browser/` because it's limited to products that support FxA, hence maybe we move it to its consumer ` services/fxaccounts/`?

1rneh (changes requested):
Ok, I made my first pass! A lot of this already look good! I'm really happy about moving away form the system add-on architecture :)

toolkit/components/enterprise/moz.build
lissyx: I was thinking of the opposite when looking at that, making sure modules are under `modules`, wouldn't it be better?

toolkit/components/felt/content/felt.xhtml
lissyx: So yes we have at least one, for the webauthn icon that is not visible. That's why I'm adding it in some other changes

toolkit/components/felt/Felt.sys.mjs
lissyx: not 100% sure it's required, I prefered to be safe here, but maybe it's not strictly needed

toolkit/components/felt/Felt.sys.mjs
lissyx: I dont think we can add both, and `app-startup` was kinda important to ensure we make the enterLastWindow call from the constructor: the risk here is that there is no window yet started and it will shutdown. It does not happens on Firefox but it does on Thunderbird

lissyx: Maybe we could move all the work done in `profile-after-change` handling earlier? I'll explore this

toolkit/components/felt/Felt.sys.mjs
lissyx: I'm not sure how you want we do this, but if there's a way, I'm taking it

toolkit/components/felt/Felt.sys.mjs
lissyx: I'll try again but I think this was not working well at least on Thunderbird

toolkit/components/felt/Felt.sys.mjs
lissyx: We should definitively but that is the kind of thing I was preferring to deal with in a follow up, I dont think it's going to be complicated, but I worry it's going to add lot of changes to review here.

It's inconvenient, misleading, but it's not fundamentally blocking the required move here

toolkit/components/felt/moz.build
lissyx: Now with the rename it would be `Enterprise` but yes likely

toolkit/components/enterprise/EnterpriseHandler.sys.mjs
lissyx: Yes, this is why I mentionned there are likely things that will have to be improved in follow up. I had a look at some and it depends on the features we will have on Thunderbird, so I think it's better to do that more precisely when we know what we need

toolkit/components/enterprise/modules/EnterpriseAccountsStorage.sys.mjs
lissyx: Good point

toolkit/components/enterprise/moz.build
1rneh: I have no preference.

toolkit/components/enterprise/EnterpriseHandler.sys.mjs
lissyx: Bug-2047485

toolkit/components/felt/Felt.sys.mjs
lissyx: I've removed the `FeltExtension:` part of logging, and filed Bug-2047499 as a follow up for the more intrusive changes related to IDL and others.

toolkit/components/felt/Felt.sys.mjs
lissyx: OK, looks like we can.

toolkit/components/felt/Felt.sys.mjs
lissyx: It might be a leftover of a hacking mistake I made that triggered two constructions (`components.conf` referencing `Felt.sys.mjs` in two places), that was not committed but the file was still on disk. I cannot prove there's a problem now when removing the guard.

toolkit/components/felt/Felt.sys.mjs
lissyx: It's now properly bound to `xpcom-shutdown`

lissyx: I'm not really sure, but it is not relevant anymore, and from my current testing I could not expose a regression yet.

toolkit/components/felt/content/felt.xhtml
lissyx: So `browser/browser.ftl` was required for `urlbar-web-authn-anchor`. Making use of `felt-urlbar-web-authn-anchor` instead allows us to remove the dependency.

toolkit/components/felt/Felt.sys.mjs
lissyx: FTR linter removes it

toolkit/modules/ActorManagerParent.sys.mjs
lissyx: Looks like it's working on both firefox and thunderbird, so we can proceed

toolkit/components/felt/Felt.sys.mjs
lissyx: OK the patch may not be so difficult to bundle with this PR, but I'm keeping it a separate bug and commit if it ends up blocking this PR

services/fxaccounts/moz.build
```diff
@@ -31,6 +31,11 @@ EXTRA_JS_MODULES += [
     "FxAccountsWebChannel.sys.mjs",
 ]
 
+if CONFIG["MOZ_ENTERPRISE"]:
+    EXTRA_JS_MODULES += [
```
Mossop, line 34: Should this be in modules/enterprise?

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
```diff
@@ -500,6 +500,8 @@ export const EnterpriseHandler = {
       params.wrappedJSObject = params;
       Services.ww.openWindow(
         null,
+        // TODO: depends on Thunderbird features set?
```
Mossop: Can you elaborate on this comment?

toolkit/components/felt/moz.build
```diff
@@ -4,11 +4,19 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+with Files("**"):
+    BUG_COMPONENT = ("Enterprise Products", "Firefox")
+
 JAR_MANIFESTS += ["jar.mn"]
 
 EXTRA_JS_MODULES += [
```
Mossop: Should we also put these in the enterprise subdirectory while we're here?

Mossop (changes requested):
A couple of suggestions but just one comment I would like to understand further.

services/fxaccounts/moz.build
lissyx: So it was there, and @1rneh suggested this should be part of `services/fxaccounts` in fact, which does not looks like a bad idea to me either. Do you think we should not ?

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
lissyx: There's direct reference from `toolkit/` to `browser/` here, and it is not yet defined if this should be part of Thunderbird Enterprise features or not, because the feature set is not yet defined :). To the best of my knowledge, thunderbird does not have a warning close dialog at all for example.

But the close dialog logic has dependency on the browser itself, so moving it around is non trivial and I prefer to leave it like that for now.

toolkit/components/felt/moz.build
lissyx: Do you mean `toolkit/components/enterprise/felt` ?

Mossop: I mean use `EXTRA_JS_MODULES.enterprise` so they end up in `resource://gre/modules/enterprise/...`

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
Mossop: So should we wrap this in a test that we're running in the browser?

services/fxaccounts/moz.build
Mossop: Oh I didn't spot this is in `services/fxaccounts`. I think this is fine as-is.

toolkit/components/felt/moz.build
lissyx: Oh, we could, is it useful?

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
lissyx: yes this is something we can likely do we have `isBuildAppBrowser()`

toolkit/components/felt/moz.build
Mossop: Just helps make it a bit clearer where it came from rather than every module ending up in the same path. It isn't strictly necessary for anything though

services/fxaccounts/moz.build
```diff
@@ -31,6 +31,11 @@ EXTRA_JS_MODULES += [
...
+        "EnterpriseAccountsStorage.sys.mjs",
+    ]
```
1rneh, line 36: And on `else` when can build `FxAccountsStorage.sys.mjs`.

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
```diff
@@ -496,23 +498,30 @@ export const EnterpriseHandler = {
...
-        "chrome,centerscreen,modal,dialog",
-        params
-      );
-    } else {
-      if (window.gDialogBox.isOpen) {
-        window.gDialogBox.replaceDialogIfOpen();
+    // TODO: Thunderbird Enterprise may or may not need to support this.
+    // For now just check the build app is browser when requesting to show this
+    // so that chrome://browser/ makes sense.
+    if (lazy.isBuildAppBrowser()) {
```
1rneh: This only guards parts of the singout prompting process. Maybe it help to define what exactly you are intending to guard here?

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -426,12 +395,8 @@ this.felt = class extends ExtensionAPI {
...
   }
 
-  onShutdown(isAppShutdown) {
-    lazy.log.debug(`FeltExtension: onShutdown: ${isAppShutdown}`);
-
-    if (isAppShutdown) {
-      return;
-    }
+  #handleShutdown() {
+    lazy.log.debug(`handleShutdown()`);
```
1rneh, line 408: Previously we'd return early when isAppShutdown === true; now we continue executing. Was the previous behavior incorrect?

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -2,110 +2,87 @@
...
-      null,
-      this.extension.rootURI
-    );
-
-    this.chromeHandle = aomStartup.registerChrome(manifestURI, [
-      ["content", "felt", "content/"],
-    ]);
+/**
+ * Felt is an XPCOM component that manages the lifecycle of the
+ * Firefox Enterprise Login Tool (FELT) UI and its interaction with the
```
1rneh: Isn't it "Launcher"? 😄  
```suggestion
 * Firefox Enterprise Launcher Tool (FELT) UI and its interaction with the
```

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -451,10 +416,7 @@ this.felt = class extends ExtensionAPI {
     }
```
1rneh: Dead code

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -451,10 +416,7 @@ this.felt = class extends ExtensionAPI {
...
 
     if (Services.felt.isFeltUI()) {
-      ChromeUtils.unregisterWindowActor(this.FELT_WINDOW_ACTOR);
-      ChromeUtils.unregisterWindowActor(this.FELT_ERROR_WINDOW_ACTOR);
-      ChromeUtils.unregisterProcessActor(this.FELT_PROCESS_ACTOR);
       lazy.FeltStorage.uninit();
```
1rneh, line 422: Can this move into the branch above?

toolkit/modules/ActorManagerParent.sys.mjs
```diff
@@ -825,6 +827,62 @@ if (AppConstants.platform != "android") {
...
+if (AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI()) {
+  JSWINDOWACTORS.FeltWindow = {
+    parent: {
+      esModuleURI: "chrome://felt/content/FeltWindowParent.sys.mjs",
+    },
+    child: {
+      esModuleURI: "chrome://felt/content/FeltWindowChild.sys.mjs",
+      events: {
+        DOMContentLoaded: {},
+        load: {},
```
1rneh: The event `load` is not handled by `FeltWindowChild`. We can drop it.

toolkit/modules/ActorManagerParent.sys.mjs
```diff
@@ -825,6 +827,62 @@ if (AppConstants.platform != "android") {
...
+    },
+    allFrames: true,
+  };
+
+  const { ConsoleClient } = ChromeUtils.importESModule(
+    "resource://gre/modules/enterprise/ConsoleClient.sys.mjs"
+  );
+  ConsoleClient.ssoCallbackUriMatchPattern.then(m => {
+    JSWINDOWACTORS.FeltWindow.matches = [m];
+  });
```
1rneh: On my last review I've written [a comment](https://github.com/mozilla/enterprise-firefox/pull/960#discussion_r3404318568) on the challenge of getting the callback uri asynchronously while the actor registration happens synchronously. It wasn't a blocker, just emphasizing the theoretic raciness.

But your answer _"Looks like it's working on both firefox and thunderbird, so we can proceed"_ doesn't really address what I've pointed out.

In fact the updated actor registration in this PR conflicts with the synchronous actor registration in a different and this time "breaking" way, because the `.then()` that sets `matches` runs after the actor is already registered, hence the mutation is a no-op. The actor ends up registered **with no matches and `allFrames: true`**, in other words matching **every document**.

It still "works" because you have a _non-excluding_ match set, so the callback page is included and the tokens are still extracted. That's why a "looks like it's working" is not sufficient here.

Alternatively to my previous proposal, we can define our own registration callback `onAddActor` that registers the actor once the promise resolves:

```
    onAddActor(register) {
      ConsoleClient.ssoCallbackUriMatchPattern.then(
        uri => {
          this.matches = [uri];
          register();
        },
        e => console.error("Felt: FeltWindow actor registration failed", e)
      );
    },
```

I think I'd prefer this solution, but maybe @Mossop can propose how to best register the actor asynchronously.

1rneh (changes requested):
We're getting there :)

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
lissyx: This only guards the part that may be breaking for sure on Thunderbird. We may or may not want that on thunderbird, so we will have to revisit when this is known.

I can guard at the start of `async showSignoutPrompt(window)` function if you prefer ?

toolkit/components/felt/Felt.sys.mjs
lissyx: I'm not 100% sure (for the previous state), but I think what we do now is correct ?

toolkit/modules/ActorManagerParent.sys.mjs
lissyx: > On my last review I've written [a comment](https://github.com/mozilla/enterprise-firefox/pull/960#discussion_r3404318568) on the challenge of getting the callback uri asynchronously while the actor registration happens synchronously. It wasn't a blocker, just emphasizing the theoretic raciness.
> 
> But your answer _"Looks like it's working on both firefox and thunderbird, so we can proceed"_ doesn't really address what I've pointed out.
> 
> In fact the updated actor registration in this PR conflicts with the synchronous actor registration in a different and this time "breaking" way, because the `.then()` that sets `matches` runs after the actor is already registered, hence the mutation is a no-op. The actor ends up registered **with no matches and `allFrames: true`**, in other words matching **every document**.
> 
> It still "works" because you have a _non-excluding_ match set, so the callback page is included and the tokens are still extracted. That's why a "looks like it's working" is not sufficient here.

I had verified the behavior: removing `this.matches = ...` with the new implementation was breaking tests, so I'm not sure anymore ...

> 
> Alternatively to my previous proposal, we can define our own registration callback `onAddActor` that registers the actor once the promise resolves:
> 
> ```
>     onAddActor(register) {
>       ConsoleClient.ssoCallbackUriMatchPattern.then(
>         uri => {
>           this.matches = [uri];
>           register();
>         },
>         e => console.error("Felt: FeltWindow actor registration failed", e)
>       );
>     },
> ```
> 
> I think I'd prefer this solution, but maybe @Mossop can propose how to best register the actor asynchronously.

But then we go back to the first version I had ? What's the difference here, except that onAddActor always register (while other users of `onAddActor` tends to handle registraiton and unregistration) ?

services/fxaccounts/moz.build
lissyx: I can see a couple of places where we may be breaking tests doing so

toolkit/components/felt/Felt.sys.mjs
lissyx: Oops

toolkit/components/felt/Felt.sys.mjs
lissyx: Indeed

toolkit/components/felt/Felt.sys.mjs
lissyx: I got tricked because we removed this but yes, we can :)

toolkit/modules/ActorManagerParent.sys.mjs
lissyx: From what I can gather I  do not think we really have to care for `unregister` handling indeed.

toolkit/components/felt/Felt.sys.mjs
lissyx: It's possible we were missing some unregistration in the past indeed.

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
1rneh: Just checked and `showSignoutPrompt` is only consumed in `browser/*`:
https://searchfox.org/enterprise-main/search?q=showSignoutPrompt. 

Do we even need to guard in this case?

If I recall correctly you have also filed a follow-up bug that splits `EnterpriseHandler` into toolkit specific and browser specific changes?

lissyx: I did this to address https://github.com/mozilla/enterprise-firefox/pull/960#discussion_r3427328255

lissyx: ok, other call is https://searchfox.org/enterprise-main/source/browser/base/content/browser-sets.js#223

maybe being so defensive is not useful indeed

toolkit/modules/ActorManagerParent.sys.mjs
```diff
@@ -825,6 +827,70 @@ if (AppConstants.platform != "android") {
...
+      );
+    },
+  };
+
+  const { ConsoleClient } = ChromeUtils.importESModule(
+    "resource://gre/modules/enterprise/ConsoleClient.sys.mjs"
+  );
+  ConsoleClient.ssoCallbackUriMatchPattern.then(m => {
+    JSWINDOWACTORS.FeltWindow.matches = [m];
+  });
```
1rneh: You still need to remove the old registration path.
```suggestion
```

lissyx: i removed it, I think I forgot to git add

lissyx: thanks for spotting

services/fxaccounts/moz.build
lissyx: > I can see a couple of places where we may be breaking tests doing so

Should be better now

toolkit/modules/ActorManagerParent.sys.mjs
Mossop: I can't think of another actor that requires async work for the actor definition itself so we're in new territory here really. I think you have two options really, either do what you've done, so the async work in `onAddActor` and then call the `register` method, or you just ignore the normal actor registration path and call `ChromeUtils.registerWindowActor` yourself directly after you have built the actor definition. The latter is maybe a little clearer that something abnormal is going on, but on the whole I think the way you've done it is better as you keep everything in-line with all the other definitions.

I would suggest adding some comment pointing out that the registration happens asynchronously.

toolkit/modules/ActorManagerParent.sys.mjs
```diff
@@ -872,6 +874,68 @@ if (AppConstants.platform != "android") {
...
+      esModuleURI: "chrome://felt/content/FeltWindowParent.sys.mjs",
+    },
+    child: {
+      esModuleURI: "chrome://felt/content/FeltWindowChild.sys.mjs",
+      events: {
+        DOMContentLoaded: {},
+      },
+    },
+    allFrames: true,
+    onAddActor(register, _unregister) {
```
Mossop, line 892:
```suggestion
    matches: [],
    onAddActor(register, _unregister) {
```

This well (presumably) cause tests to fail should we fail to correctly set matches before registering.

Mossop (approved):

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -426,12 +395,8 @@ this.felt = class extends ExtensionAPI {
...
-    lazy.log.debug(`FeltExtension: onShutdown: ${isAppShutdown}`);
-
-    if (isAppShutdown) {
-      return;
-    }
+  #handleShutdown() {
+    lazy.log.debug(`handleShutdown()`);
 
     if (Services.felt.isFeltBrowser()) {
       Services.obs.removeObserver(this.urlObserver, "felt-open-url");
```
1rneh: An observer for `"felt-open-url"` also seems to be only relevant in the browser environment.

toolkit/components/felt/Felt.sys.mjs
1rneh: `lazy.FeltStorage.uninit();` was previously called if `Services.felt.isFeltUI()===true`, so either you can early return in the previous branch or guard this again.

1rneh (approved):
Thank you! 

Is there a meta-bug for all the follow-up bugs that you have filed through out the review iterations?

lissyx:
> Thank you!
> 
> Is there a meta-bug for all the follow-up bugs that you have filed through out the review iterations?

I made all blocking `firefox-client-mvp` so we can triage accordingly, but I can make a dedicated meta bug if you'd like

toolkit/components/felt/Felt.sys.mjs
lissyx: Indeed

lissyx: it also makes me thinks we should directly remove references to firefox / browser: https://bugzilla.mozilla.org/show_bug.cgi?id=2049322

toolkit/components/felt/Felt.sys.mjs
lissyx: We are still under `if (Services.felt.isFeltUI()) {`, no?

1rneh: But now it's also called when `Services.felt.isFeltBrowser()===true`

1rneh: Oh sorry for the confusion. The line `if (Services.felt.isFeltUI()) {` was collapsed in the GitHub UI and I didn't see that, so I thought `lazy.FeltStorage.uninit();` was called unconditionally.

jporter-dev (approved):
The changes here look great to me, I don't see anything that needs changing outside of what's already been mentioned. 

nit: one missed `browser/enterprise/enterprise.ftl` reference in a comment: https://searchfox.org/enterprise-main/source/tools/lint/fluent-lint/exclusions.yml#143

lissyx:
> The changes here look great to me, I don't see anything that needs changing outside of what's already been mentioned.
> 
> nit: one missed `browser/enterprise/enterprise.ftl` reference in a comment: https://searchfox.org/enterprise-main/source/tools/lint/fluent-lint/exclusions.yml#143

Thanks, fixed

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -10,7 +10,8 @@ ChromeUtils.defineESModuleGetters(lazy, {
   JsonSchemaValidator:
     "resource://gre/modules/components-utils/JsonSchemaValidator.sys.mjs",
   // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
```
jonathanmendez: nit: This eslint-disable can be removed now I think?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -19,7 +20,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   clearInterval: "resource://gre/modules/Timer.sys.mjs",
   setInterval: "resource://gre/modules/Timer.sys.mjs",
   // eslint-disable-next-line mozilla/no-browser-refs-in-toolkit
```
jonathanmendez: nit: This eslint-disable can be removed now I think?

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -268,17 +241,17 @@ this.felt = class extends ExtensionAPI {
       // In the real Firefox, register observer to handle URLs
       Services.obs.addObserver(this.urlObserver, "felt-open-url");
```
jonathanmendez: Shouldn't this be guarded, as in `#handleShutdown()`?
```suggestion
      if (lazy.isBuildAppBrowser()) {
        Services.obs.addObserver(this.urlObserver, "felt-open-url");
      }
```

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -2,110 +2,87 @@
...
+  static classDescription = "Felt";
+
+  // QI — implement nsISupports + nsIObserver
+  QueryInterface = ChromeUtils.generateQI(["nsIObserver"]);
+
+  constructor() {
+    if (Services.felt?.isFeltUI()) {
+      // Felt UI XPCOM is triggered by profile-after-change, make sure to block
+      // closing until startup has been executed.
+      Services.startup.enterLastWindowClosingSurvivalArea();
```
jonathanmendez, line 73: Is it the case that this is balanced by the existing `exitLastWindowClosingSurvivalArea()` when observing "domwindowopened"? And that same exit otherwise balances the enter that happens for "TransitionFeltToBackground"?

If yes, it may be worth a comment (maybe with the exit call) briefly explaining the lifecycle to help explain the seeming imbalance with one exit call but two enter calls in this file, i.e. that the exit call is handling both cases.

services/fxaccounts/FxAccounts.sys.mjs
```diff
@@ -953,7 +952,7 @@ FxAccountsInternal.prototype = {
   newAccountState(credentials) {
     let storage = AppConstants.MOZ_ENTERPRISE
       ? new lazy.EnterpriseStorageManager()
-      : new FxAccountsStorageManager();
+      : new lazy.FxAccountsStorageManager();
```
jonathanmendez: nit: The pattern you used in test_storage_manager.js initializes a single lazy storage manager value to either Enterprise or FxAccounts, while this file defines both in the lazy and switches on MOZ_ENTERPRISE at the call site. Should these use a consistent pattern?

jonathanmendez (approved):
Looks good! I found a few small suggestions.

toolkit/components/felt/Felt.sys.mjs
lissyx: yes

services/fxaccounts/FxAccounts.sys.mjs
lissyx: My version already has that change

lissyx: > nit: The pattern you used in test_storage_manager.js initializes a single lazy storage manager value to either Enterprise or FxAccounts, while this file defines both in the lazy and switches on MOZ_ENTERPRISE at the call site. Should these use a consistent pattern?

yes, let's use the same as in the test

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: yep

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx: done

toolkit/components/felt/Felt.sys.mjs
lissyx: yes, I'll update the comment

PR #964, opened 2026-05-27
Bug 2042370 - Uplift Bug-2028364 to release

lissyx:
### Description 

Bugzilla: Bug-2042370

Uplift of Bug-2028364 to address the situation where device sleep breaks forced update restart

---

### Testing 

* [x] Manual testing performed (QA)

### DIFF
--- a/browser/modules/UpdatePolicyEnforcer.sys.mjs
+++ b/browser/modules/UpdatePolicyEnforcer.sys.mjs
@@ -7,7 +7,7 @@ const PREF_APP_UPDATE_COMPULSORY_RESTART = "app.update.compulsory_restart";
 let deferredRestartTasks = null;
 
 ChromeUtils.defineESModuleGetters(lazy, {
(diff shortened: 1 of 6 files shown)


jonathanmendez (approved):
Checking my understanding:
The original fix was put in m-c, and through daily merges is present in `enterprise-main` and `enterprise-beta` already; the patch was not uplifted into Firefox's `release`, but we want it in `enterprise-release` because we are affected by the bug (as found by QA in our last release smoke testing). That is why we have an uplift to `enterprise-release` but not one for `enterprise-beta`.
If that's all true, looks good.

lissyx:
> Checking my understanding: The original fix was put in m-c, and through daily merges is present in `enterprise-main` and `enterprise-beta` already; the patch was not uplifted into Firefox's `release`, but we want it in `enterprise-release` because we are affected by the bug (as found by QA in our last release smoke testing). That is why we have an uplift to `enterprise-release` but not one for `enterprise-beta`. If that's all true, looks good.

Exactly

PR #965, opened 2026-05-28
Add policy for disabling crash reports

jonathanmendez:
### Description 

Bugzilla: Bug-2026983

Add policy to disable sending crash reports.

- For app crashes, uses the existing MOZ_CRASHREPORTER_NO_REPORT environment variable to block reports (adding an update method to re-read the environment variable since policies can be updated at runtime).
- For content crashes, add logic that checks whether the policy is disabled to block reports.

---

#### about:tabcrashed
| policy off | policy on |
| -- | -- |
| image | image |

#### about:crashes
| policy off | policy on |
| -- | -- |
| image | image |

---

### Testing 

* [x] Added tests

### DIFF
--- a/browser/base/content/aboutTabCrashed.js
+++ b/browser/base/content/aboutTabCrashed.js
@@ -149,7 +149,7 @@ var AboutTabCrashed = {
     let data = message.data;
 
     if (data.hasReport || data.policyAutoSubmit) {
-      this.hasReport = !data.policyAutoSubmit;
+      this.hasReport = data.hasReport;
       document.documentElement.classList.add("crashDumpAvailable");
 
       document.getElementById("sendReport").checked = data.sendReport;
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -1107,30 +1107,65 @@ export var Policies = {
 
   CrashReportsSubmit: {
     onBeforeAddons(_manager, param) {
-      if (param.ForceAutoSubmit) {
-        setAndLockPref("browser.crashReports.unsubmittedCheck.enabled", true);
-        setAndLockPref(
-          "browser.crashReports.unsubmittedCheck.autoSubmit2",
-          true
-        );
-        setAndLockPref("browser.tabs.crashReporting.sendReport", true);
-        setAndLockPref("browser.tabs.crashReporting.includeURL", true);
-        Services.env.set("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT", "1");
+      if ("Enabled" in param) {
+        if (param.Enabled) {
+          setAndLockPref(
+            "browser.crashReports.unsubmittedCheck.autoSubmit2",
+            true
+          );
+          setAndLockPref("browser.tabs.crashReporting.sendReport", true);
+          setAndLockPref("browser.crashReports.unsubmittedCheck.enabled", true);
+          setAndLockPref("browser.tabs.crashReporting.includeURL", true);
+          setEnvVar("MOZ_CRASHREPORTER_NO_REPORT", "");
+          setEnvVar("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT", "1");
+        } else {
+          setAndLockPref(
+            "browser.crashReports.unsubmittedCheck.autoSubmit2",
+            false
+          );
+          setAndLockPref("browser.tabs.crashReporting.sendReport", false);
+          setAndLockPref(
+            "browser.crashReports.unsubmittedCheck.enabled",
+            false
+          );
+          unsetAndUnlockPref("browser.tabs.crashReporting.includeURL");
+          setEnvVar("MOZ_CRASHREPORTER_NO_REPORT", "1");
+          setEnvVar("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT", "");
+        }
       } else {
-        // if ForceAutoSubmit is unset or is false
-        unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.enabled");
         unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.autoSubmit2");
         unsetAndUnlockPref("browser.tabs.crashReporting.sendReport");
+        unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.enabled");
         unsetAndUnlockPref("browser.tabs.crashReporting.includeURL");
-        Services.env.set("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT", "");
+        unsetEnvVar("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT");
+        unsetEnvVar("MOZ_CRASHREPORTER_NO_REPORT");
+      }
+      // The crash callback reads a cached, signal-safe atomic; tell it to
+      // re-read the env vars now that we've changed them.
+      try {
+        Services.appinfo
+          .QueryInterface(Ci.nsICrashReporter)
+          .updateShouldReport();
+      } catch (e) {
+        // nsICrashReporter is unavailable in builds without the crash reporter.
       }
     },
     onRemove(_manager, _oldParams) {
-      unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.enabled");
       unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.autoSubmit2");
       unsetAndUnlockPref("browser.tabs.crashReporting.sendReport");
+      unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.enabled");
       unsetAndUnlockPref("browser.tabs.crashReporting.includeURL");
-      Services.env.set("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT", "");
+      unsetEnvVar("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT");
+      unsetEnvVar("MOZ_CRASHREPORTER_NO_REPORT");
+      // The crash callback reads a cached, signal-safe atomic; tell it to
+      // re-read the env vars now that we've changed them.
+      try {
+        Services.appinfo
+          .QueryInterface(Ci.nsICrashReporter)
+          .updateShouldReport();
+      } catch (e) {
+        // nsICrashReporter is unavailable in builds without the crash reporter.
+      }
     },
   },
 
@@ -3920,6 +3955,20 @@ export function setAndLockPref(prefName, prefValue) {
 }
 
 /**
+ * setEnvVar
+ *
+ * Sets the value of an environment variable.
+ *
+ * @param {string} envVarName
+ *        The environment variable to be changed
+ * @param {string} envVarValue
+ *        The value to set the environment variable to
+ */
+export function setEnvVar(envVarName, envVarValue) {
+  PoliciesUtils.setEnvVar(envVarName, envVarValue);
+}
+
+/**
  * unsetAndUnlockPref
  *
  * Unsets the _default_ value of a pref, and unlocks it (meaning that
@@ -3930,6 +3979,19 @@ export function setAndLockPref(prefName, prefValue) {
  */
 export function unsetAndUnlockPref(prefName) {
   PoliciesUtils.unsetDefaultPref(prefName);
+}
+
+/**
+ * unsetEnvVar
+ *
+ * Unsets the value of an environment variable,
+ * restoring it to the value before any policy modification.
+ *
+ * @param {string} envVarName
+ *        The environment variable to be changed
+ */
+export function unsetEnvVar(envVarName) {
+  PoliciesUtils.unsetEnvVar(envVarName);
 }
 
 /**
@@ -3965,8 +4027,11 @@ export var PoliciesUtils = {
    * @property {boolean} isLocked - whether the preference is locked
    */
 
-  /** @type {PreferenceState} */
+  /** @type {{[key: string]: PreferenceState}} */
   _initialPrefState: {},
+
+  /** @type {{[key: string]: string}} */
+  _initialEnvVarState: {},
 
   /**
    * Reads a typed preference value from the given branch, returning null when
@@ -4043,6 +4108,22 @@ export var PoliciesUtils = {
   },
 
   /**
+   * Saves the current value of an env var before a policy changes it.
+   * No-op if the env var was already saved.
+   *
+   * @param {string} envVarName
+   */
+  saveEnvVarState(envVarName) {
+    if (envVarName in this._initialEnvVarState) {
+      return;
+    }
+
+    const envVarValue = Services.env.get(envVarName);
+
+    this._initialEnvVarState[envVarName] = envVarValue;
+  },
+
+  /**
    * Restores the preference to the state before any policy was applied. The user
    * value of a preference is only restored if the preference was locked by the policy
    * or was locked even before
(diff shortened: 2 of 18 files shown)


browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -1103,30 +1103,65 @@ export var Policies = {
...
+      }
+      // The crash callback reads a cached, signal-safe atomic; tell it to
+      // re-read the env vars now that we've changed them.
+      try {
+        Services.appinfo
+          .QueryInterface(Ci.nsICrashReporter)
+          .updateShouldReport();
+      } catch (e) {
+        // nsICrashReporter is unavailable in builds without the crash reporter.
       }
```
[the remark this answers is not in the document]
jonathanmendez: This is true and an inherent gap that we live with so that we can catch startup crashes.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -4001,6 +4090,25 @@ export var PoliciesUtils = {
...
+
+    if (!envVarState) {
+      // Nothing to restore
+      return;
+    }
+
+    Services.env.set(envVarName, envVarState.value);
+
+    delete this._initialEnvVarState[envVarName];
+  },
```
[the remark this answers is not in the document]
jonathanmendez: Will add comment.

toolkit/crashreporter/nsExceptionHandler.cpp
```diff
@@ -233,8 +233,11 @@ static MOZ_GLIBCXX_CONSTINIT xpstring memoryReportPath;
 // Where crash events should go.
 static MOZ_GLIBCXX_CONSTINIT xpstring eventsDirectory;
 
-// If this is false, we don't launch the crash reporter
-static bool doReport = true;
+// If this is false we don't launch the crash reporter client. Read from
+// the crash callback (which runs in a signal-handler-like context on POSIX),
+// so it must be an async-signal-safe lock-free atomic. Updated from
+// `SetExceptionHandler` at init and from `UpdateShouldReport()` afterwards.
+static mozilla::Atomic<bool> doReport(true);
```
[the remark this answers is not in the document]
jonathanmendez: This whole atomic<bool> was unnecessary, reverted.

browser/modules/ContentCrashHandlers.sys.mjs
```diff
@@ -584,6 +584,10 @@ export var TabCrashHandler = {
       return;
     }
 
+    if (Services.policies.getActivePolicies()?.CrashReportsSubmit?.Disabled) {
+      return;
+    }
```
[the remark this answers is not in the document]
jonathanmendez: Done.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3873,6 +3935,17 @@ export var PoliciesUtils = {
...
 
+  /**
+   * Cache for env var state before any policy changes the environment variables
+   *
+   * @typedef EnvVarState
+   * @type {object}
+   * @property {string} value - environment variable value
+   */
+
+  /** @type {EnvVarState} */
```
afranchuk: This type comment isn't accurate, I think. It should be describing an object with env var names as keys and values as `EnvVarState`.

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
```diff
@@ -9,4 +9,4 @@ policy-DownloadTelemetry = Enable and configure security logging/telemetry when
 policy-EnterpriseStorageEncryption = Enable enterprise-managed primary password for encrypted storage.
 policy-PrintPageTelemetry = Enable and configure security logging/telemetry when a page is printed.
 policy-Sync = Enable or disable sync and define which data to include.
-policy-CrashReportsSubmit = Configure crash report submission settings.
+policy-CrashReportsSubmit = Configure crash report submission settings. Use Disabled to prevent all crash report and crash ping submissions, or ForceAutoSubmit to enable automatic submission.
```
afranchuk: IIRC this needs to be a different fluent key for localization review to work (unless this has never been localized/released before?).

afranchuk (changes requested):
This looks good overall. The main thing that probably needs to change is the fluent key.

browser/components/enterprisepolicies/Policies.sys.mjs
jonathanmendez: I updated the type comment for the prefs cache and simplified the env var cache to just be <string, string>.

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
jonathanmendez: Good point, but I think this extra detail doesn't need to be here (and doesn't match other policy descriptions) so I just reverted to the original string.

xpcom/system/nsICrashReporter.idl
```diff
@@ -180,6 +180,15 @@ interface nsICrashReporter : nsISupports
...
   void UpdateCrashEventsDir();
 
+  /**
+   * Re-evaluate environment variables that gate whether the crash reporter
+   * client is launched after a crash (MOZ_CRASHREPORTER_NO_REPORT and
+   * MOZ_CRASHREPORTER_FULLDUMP). Call this after modifying those env vars
+   * so the signal-safe cached value used by the crash callback stays in
+   * sync.
+   */
+  void updateShouldReport();
```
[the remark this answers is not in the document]
jonathanmendez: UpdateCrashEventsDir is the odd one out in this interface. Everything else uses camelCase.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -4022,6 +4103,26 @@ export var PoliciesUtils = {
...
+
+    if (!envVarValue) {
+      // Nothing to restore
+      return;
+    }
+
+    Services.env.set(envVarName, envVarValue);
+
+    delete this._initialEnvVarState[envVarName];
+  },
```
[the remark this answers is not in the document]
jonathanmendez: Updated.

toolkit/crashreporter/nsExceptionHandler.cpp
```diff
@@ -2793,6 +2793,8 @@ nsresult SetSubmitReports(bool aSubmitReports) {
   return NS_OK;
 }
 
+void UpdateShouldReport() { doReport = ShouldReport(); }
```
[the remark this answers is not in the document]
jonathanmendez: Shouldn't be a practical concern but went ahead and updated it to be an `Atomic<bool>`

afranchuk (approved):

browser/components/enterprisepolicies/schemas/policies-schema.json
1rneh, line 871: When I read that `ForceAutoSubmit` and `Disabled` are mutually exclusive, I wondered if they can be combined in one key.

Two observations here:
1. The policy can only set two states: Either the crash report submissions are force disabled or force auto-submitted (which is effectively force enabled?).
2. From a glimpse through other policies, the general conversion seems to be to go with the key `Enabled` instead of `Disabled`.

Hence I would propose combining them to `Enabled=true|false`. That's is quite some change but better make that change now while nobody uses the policy yet? What do you think?

1rneh (commented):
I reviewed changes to `components/enterprisepolicies` and left a comment on the `CrashReportsSubmit` schema.

In addition, I have two questions:
1. Will this be a policy that can only be applied for enterprise builds? If so, we still need to make sure the added code is excluded in non-enterprise builds.
2. Can the policy tests that you've added move to `browser/components/enterprisepolicies/tests/browser` or do they depend on test harness in the respective modules that you've added them in?

browser/components/enterprisepolicies/schemas/policies-schema.json
jonathanmendez: The problem with a simple boolean is that semantically the two possible policy states are not opposites of each other:

1. You cannot send crash reports, even if you want to
2. Crash reports are automatically sent (immediately upon crash) without ability to prevent sending

The third state (if no policy is set) is "Crash reporting is enabled, and you can send crash reports if you choose to" (which is notably the actual opposite of policy state 1 above and arguably also the opposite of policy state 2 because you reports are _not_ automatically sent).

The history of this is that I originally understood the scope to only include the forced auto-submission behavior, but the question was raised what a `false` state would mean for that policy. At that time it was suggested that `false` could mean disabling crash reporting entirely and I raised the concern that the naming felt confusing to me for that setup. So with the current structure I've leaned into extra clarity with the names: you can make crash reports "Disabled" or you can "ForceAutoSubmit" of crash reports; the less clear part that needs the documentation is the mutual exclusivity.

I'm open to the idea that we instead lean into documenting the confusing semantics of what "true" and "false" mean for a single boolean property here, but I'd want more people to add their feedback.

afranchuk: If we anticipate these will never have some other combined behavior, maybe an enum would be clearer.

1rneh:
Regarding my second comment:
> 2. Can the policy tests that you've added move to browser/components/enterprisepolicies/tests/browser or do they depend on test harness in the respective modules that you've added them in?

I've found more tests living outside of `*/enterprisepolcies/tests/*`, so please ignore my request to move the test files.

browser/components/enterprisepolicies/schemas/policies-schema.json
1rneh: Actually, to me the two states do seem to be opposites of each other. The crash report submit is either force-enabled (automatically/always submitted without user choice) or force-disabled (never submitted without user choice)...

jonathanmendez: I've updated this to an enum with three values, "disabled", "auto-submit", and "force-auto-submit", with the following behaviors:

- disabled: You cannot send crash reports and are not prompted to do so
- auto-submit: The user preference "Automatically send crash reports" is locked and checked. Background crashes will automatically be submitted, and pending crashes will be sent on startup. You can decide to not submit any given report in the dialogs.
- force-auto-submit: Crash submission happens as soon as a crash happens. The usual dialog/page tells you that this happened rather than giving you the option to submit or not.

I added the "auto-submit" since it was easy and on my mind from this discussion, but if it's silly/useless it's easy to remove since it's just pref-based.

toolkit/components/crashes/CrashManager.in.sys.mjs
```diff
@@ -815,6 +828,9 @@ CrashManager.prototype = Object.freeze({
       // is passed the ping data (if applicable). This _doesn't_ mean the ping
       // was received by the telemetry server. We rely on the crashreporter
       // client's Glean instance to handle reliable delivery.
+      // When CrashReportsSubmit.Disabled is active we deliberately skip the
+      // send above but still mark as submitted so the ping is not retried
+      // forever; the admin chose not to deliver this data.
```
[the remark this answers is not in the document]
jonathanmendez: fixed

browser/components/enterprisepolicies/tests/xpcshell/test_simple_pref_policies.js
```diff
@@ -1374,9 +1374,73 @@ const POLICIES_TESTS = [
...
+      "browser.crashReports.unsubmittedCheck.autoSubmit2": false,
+    },
+    env: {
+      MOZ_CRASHREPORTER_NO_REPORT: "1",
+      MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT: "",
+    },
+  },
+
+  // POLICY: CrashReportsSubmit - auto-submit
+  // are set.
```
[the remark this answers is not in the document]
jonathanmendez: fixed

browser/modules/ContentCrashHandlers.sys.mjs
```diff
@@ -697,10 +714,14 @@ export var TabCrashHandler = {
...
+      };
+    }
+
     let dumpID = this.getDumpID(browser);
-    let policyAutoSubmit =
-      !!Services.policies.getActivePolicies()?.CrashReportsSubmit
-        ?.ForceAutoSubmit;
+    let policyAutoSubmit = reportSubmissionForcedByPolicy();
     // When ForceAutoSubmit is active, the crash report is submitted automatically
     // at crash time, so no dumpID is available. Allow the flow to continue.
```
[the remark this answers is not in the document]
jonathanmendez: fixed

browser/components/enterprisepolicies/schemas/policies-schema.json
jporter-dev: Should we keep the 3 modes, the enum approach looks good to me, but simplifying to an `Enabled` boolean would be much simpler/nicer. 

Whatever you go with, just make sure to update the admin reference docs with the key changes: https://github.com/mozilla/enterprise-admin-reference/blob/main/src/content/docs/reference/policies/CrashReportsSubmit.mdx

browser/base/content/aboutTabCrashed.js
```diff
@@ -149,7 +149,7 @@ var AboutTabCrashed = {
     let data = message.data;
 
     if (data.hasReport || data.policyAutoSubmit) {
-      this.hasReport = !data.policyAutoSubmit;
+      this.hasReport = data.hasReport;
```
jporter-dev, line 152: Thanks for remembering to fix this :)

browser/components/enterprisepolicies/schemas/policies-schema.json
1rneh: From an offline conversation:

m-c's user preference "Automatically send crash reports" doesn't actually automatically send all reports but just a limited subset of crash reports, e.g. background and pending crash reports. so we are not sure whether we should manifest this (confusing and maybe revampe-able) current behavior from m-c in a policy, which is what auto-submit would do.

@jonathanmendez will catch up with product folks to see what from an admin perspective that policy should look like and if two state policy `{ Enabled : true | false }` is sufficient.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -1107,30 +1107,74 @@ export var Policies = {
...
+            "browser.crashReports.unsubmittedCheck.autoSubmit2",
+            true
+          );
+          unsetAndUnlockPref("browser.tabs.crashReporting.sendReport");
+          unsetAndUnlockPref("browser.crashReports.unsubmittedCheck.enabled");
+          unsetAndUnlockPref("browser.tabs.crashReporting.includeURL");
+          setEnvVar("MOZ_CRASHREPORTER_NO_REPORT", "");
+          setEnvVar("MOZ_CRASHREPORTER_POLICY_AUTO_SUBMIT", "");
+          break;
+        default:
```
jporter-dev: Should this unset/unlock as well, for example when `Behavior` is removed? Previous there was an if/else for `params.ForceAutoSubmit`. Obviously not an issue if this moves from `Behavior` to `Enabled`.

jporter-dev: If this remains an enum, maybe `Behavior` should be a required attribute for the policy to avoid this case altogether

browser/components/enterprisepolicies/schemas/policies-schema.json
jonathanmendez: Ended up with an Enabled property for reflecting the two meaningful behaviors ("disabled" and "force-auto-submit"}

xpcom/system/nsICrashReporter.idl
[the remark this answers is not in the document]
jonathanmendez: as I understand it, this is N/A

browser/components/enterprisepolicies/Policies.sys.mjs
jonathanmendez: The new (boolean) code handles the case where "Enabled" isn't present (treats it the same as onRemove). And from discussion it seems that this will soon be unnecessary when live policies change their "there's a difference" behavior to remove first then apply.

1rneh (approved):

jporter-dev (approved):

PR #976, opened 2026-05-29
Bug 2043589 - fix browser_all_files_referenced.js for enterprise

victhorlopez:
### Description 

Bugzilla: Bug-2043589

- Allowlist programmatically-loaded URIs: crashreporter-enterprise.ftl and chrome://felt/content/felt.xhtml are loaded at runtime via code (not static markup), so browser_all_files_referenced.js can't detect them.
- Remove .inc.xhtml files from jar.mn: These fragment files are inlined at build time via #include.
- Add chrome:// override for enterprise.properties: nsDocShell.cpp loads chrome://global/locale/enterprise.properties (the toolkit version). The override directive redirects that URI to chrome://browser/locale/enterprise.properties so the browser-specific version is used instead. Without it, the browser's enterprise.properties would be packaged but never loaded, triggering the unreferenced file test.

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [ ] Added tests

### DIFF
--- a/browser/base/content/test/static/browser_all_files_referenced.js
+++ b/browser/base/content/test/static/browser_all_files_referenced.js
@@ -379,6 +379,17 @@ if (AppConstants.platform == "android") {
   });
 }
 
+if (AppConstants.MOZ_ENTERPRISE) {
+  allowlist.push(
+    // toolkit/crashreporter/client/app/src/lang/langpack.rs (string is built programmatically)
+    {
+      file: "resource://gre/localization/en-US/crashreporter/crashreporter-enterprise.ftl",
+    },
+    // Opened programmatically by the felt extension API
+    { file: "chrome://felt/content/felt.xhtml" }
+  );
+}
+
 if (AppConstants.MOZ_UPDATE_AGENT && !AppConstants.MOZ_BACKGROUNDTASKS) {
   // Task scheduling is only used for background updates right now.
   allowlist.push({
--- a/browser/components/enterprise/jar.mn
+++ b/browser/components/enterprise/jar.mn
@@ -3,13 +3,11 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 browser.jar:
+# Override the toolkit enterprise.properties with the browser version (loaded by nsDocShell.cpp).
+% override chrome://global/locale/enterprise.properties chrome://browser/locale/enterprise.properties
   content/browser/enterprise/enterprise-badge.css           (content/enterprise-badge.css)
-  content/browser/enterprise/enterprise-badge.inc.xhtml     (content/enterprise-badge.inc.xhtml)
   content/browser/enterprise/alert.svg                      (content/icons/alert.svg)
-  content/browser/enterprise/enterprise-panel.inc.xhtml     (content/enterprise-panel.inc.xhtml)
-  content/browser/enterprise/enterprise-urlbar-actions.inc.xhtml (content/enterprise-urlbar-actions.inc.xhtml)
   content/browser/enterprise/enterprise-urlbar-actions.css  (content/enterprise-urlbar-actions.css)
-  content/browser/enterprise/enterprise-urlbar-panels.inc.xhtml (content/enterprise-urlbar-panels.inc.xhtml)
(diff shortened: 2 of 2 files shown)


browser/extensions/felt/api.js
```diff
@@ -380,7 +380,7 @@ this.felt = class extends ExtensionAPI {
 
     this._win = Services.ww.openWindow(
       null,
-      `chrome://felt/content/felt.xhtml${queryString}`,
+      "chrome://felt/content/felt.xhtml" + queryString,
```
lissyx: why?

victhorlopez: cause the test cannot scan dynamic strings :/ and this dynamic string just adds ? , that is what I understood

lissyx: we can always add manually, I think you found it
(fun fact, this is one of the test I broke and got me backed out on https://bugzilla.mozilla.org/show_bug.cgi?id=2041209, so I was looking at this exactly this mornign)

victhorlopez: yeah, here I tested without this piece

https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&selectedTaskRun=ajHz3ntYQs2qTpboBKH6Qw.0

https://treeherder.mozilla.org/logviewer?job_id=569223730&repo=enterprise-firefox-try&task=ajHz3ntYQs2qTpboBKH6Qw.0&lineNumber=7025

Would you add 13(and any future one I guess)? maybe I am missing something

lissyx: I'd prefer to be safe and explicit rather than rely on magic.

victhorlopez: okay.. I changed it now.. I will check the CI if that works, I would expect this is ok?

browser/base/content/test/static/browser_all_files_referenced.js
```diff
@@ -385,6 +379,15 @@ if (AppConstants.platform == "android") {
   });
 }
 
+if (AppConstants.MOZ_ENTERPRISE) {
+  // toolkit/crashreporter/client/app/src/lang/langpack.rs (string is built programmatically)
+  allowlist.push({
+    file: "resource://gre/localization/en-US/crashreporter/crashreporter-enterprise.ftl",
+  });
+  // Opened programmatically by the felt extension API
+  allowlist.push({ file: "chrome://felt/content/felt.xhtml" });
```
lissyx: I think we can have a single push call?

browser/components/enterprise/jar.mn
```diff
@@ -3,13 +3,11 @@
...
+# Override the toolkit enterprise.properties with the browser version (loaded by nsDocShell.cpp).
+% override chrome://global/locale/enterprise.properties chrome://browser/locale/enterprise.properties
   content/browser/enterprise/enterprise-badge.css           (content/enterprise-badge.css)
-  content/browser/enterprise/enterprise-badge.inc.xhtml     (content/enterprise-badge.inc.xhtml)
   content/browser/enterprise/alert.svg                      (content/icons/alert.svg)
-  content/browser/enterprise/enterprise-panel.inc.xhtml     (content/enterprise-panel.inc.xhtml)
-  content/browser/enterprise/enterprise-urlbar-actions.inc.xhtml (content/enterprise-urlbar-actions.inc.xhtml)
   content/browser/enterprise/enterprise-urlbar-actions.css  (content/enterprise-urlbar-actions.css)
-  content/browser/enterprise/enterprise-urlbar-panels.inc.xhtml (content/enterprise-urlbar-panels.inc.xhtml)
   content/browser/enterprise/lockdown-mode.svg              (content/icons/lockdown-mode.svg)
```
lissyx, line 11: Those removal looks wrong? We would not be able to use anymore. They are `#include`'d but I'm not sure at what time this is processed. Can you ensure there is no regression here?

victhorlopez: I would expect if they are preprocessed it is before the jar, but okay, I will try will a shippable

Mossop: Yes included files do not need to be listed in the jar manifest as they are included as part of another entry.

Mossop (approved):

Mossop (approved):

lissyx (approved):

PR #980, opened 2026-06-01
Bug 2044005 - Distinguish FELT UI from Firefox in remote/Marionette

lissyx:
### Description 

Bugzilla: Bug-2044005

This is adding a way to expose the information that we are in the FeltUI case to Marionette, will be useful for non browser app.

### DIFF
--- a/remote/shared/AppInfo.sys.mjs
+++ b/remote/shared/AppInfo.sys.mjs
@@ -73,6 +73,10 @@ ChromeUtils.defineLazyGetter(AppInfo, "isEnterprise", () => {
   return Services.appinfo.ID == ID_FIREFOX && AppConstants.MOZ_ENTERPRISE;
 });
 
+ChromeUtils.defineLazyGetter(AppInfo, "isFeltUI", () => {
+  return !!(AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI());
+});
+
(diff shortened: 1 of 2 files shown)


whimboo:
This seems to be fine as long as the FeltUI implements all the necessary tabbrowser APIs.

lissyx:
> This seems to be fine as long as the FeltUI implements all the necessary tabbrowser APIs.

Thanks! Marionette is able to do its job so I think we are good :)

remote/shared/AppInfo.sys.mjs
```diff
@@ -73,6 +73,10 @@ ChromeUtils.defineLazyGetter(AppInfo, "isEnterprise", () => {
   return Services.appinfo.ID == ID_FIREFOX && AppConstants.MOZ_ENTERPRISE;
 });
 
+ChromeUtils.defineLazyGetter(AppInfo, "isFeltUI", () => {
+  return AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI();
```
jonathanmendez: It seems we wouldn't want to ever return undefined here?
```suggestion
  return !!(AppConstants.MOZ_ENTERPRISE && Services.felt?.isFeltUI());
```

jonathanmendez (changes requested):

jonathanmendez (approved):

PR #982, opened 2026-06-01
Bug 2044013 - Limit URL forwarding from Felt to browser app only

lissyx:
### Description 

Bugzilla: Bug-2044013

Forwarding handling of URLs only makes sense for browser app.

### DIFF
--- a/browser/extensions/felt/content/FeltProcessParent.sys.mjs
+++ b/browser/extensions/felt/content/FeltProcessParent.sys.mjs
@@ -8,6 +8,7 @@ ChromeUtils.defineESModuleGetters(lazy, {
   Subprocess: "resource://gre/modules/Subprocess.sys.mjs",
   ConsoleClient: "resource:///modules/enterprise/ConsoleClient.sys.mjs",
   CONSOLE_ADDRESS_PREF: "resource:///modules/enterprise/ConsoleClient.sys.mjs",
+  isBuildAppBrowser: "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
   isTesting: "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
   createEnterpriseLogger:
     "resource:///modules/enterprise/EnterpriseCommon.sys.mjs",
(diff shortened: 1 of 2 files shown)


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -447,7 +450,9 @@ export class FeltProcessParent extends JSProcessActorParent {
     this.exitReported = false;
     this.firefoxReady = false;
     this.extensionReady = false;
-    resetFeltFirefoxWindowReady();
+    if (lazy.isBuildAppBrowser()) {
```
jonathanmendez, line 453: It's not clear to me what this specific change is for, and I can't see how it's related to URL forwarding. Assuming that it's an intentional part of this PR, could you add a comment in the code explaining the intended effect?

jonathanmendez (changes requested):

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx: It's part of the code that deals with URL handling: https://searchfox.org/enterprise-main/rev/017917c10c85768bfdf90714b660bb318053a879/browser/components/FeltURLHandler.sys.mjs#140-142

lissyx: It's loaded from `"resource:///modules/FeltURLHandler.sys.mjs"` that will not be loadable in another app than `browser`

lissyx: here: https://github.com/mozilla/enterprise-firefox/pull/960/changes#diff-a7d1dbc1e60672a22916dcc2662534c4da1ce96a581cd4650a881270d90e5c98R19

So this code cannot be loaded (and does not make sense being) outside of browser app

jonathanmendez: Ok, if `FeltURLHandler.sys.mjs` will not be loadable, that makes me wonder about other usage in this file (e.g. in code reachable from `export function queueURL`), but at least the changes in here are good.

jonathanmendez (approved):

PR #985, opened 2026-06-01
Bug 2044008 - Use WebAuthnPromptHelper from toolkit in Felt

lissyx:
### Description 

Bugzilla: Bug-2044008

Let's make sure we load the WebAuthn helper when the app is not a browser (that loads it for us)

### Dependencies

#982 for `isBuildAppBrowser()`

### DIFF
--- a/browser/extensions/felt/content/felt.xhtml
+++ b/browser/extensions/felt/content/felt.xhtml
@@ -107,7 +107,7 @@
         id="webauthn-notification-icon"
         class="notification-anchor-icon"
         role="button"
-        data-l10n-id="urlbar-web-authn-anchor"
+        data-l10n-id="felt-urlbar-web-authn-anchor"
       />
     </xul:box>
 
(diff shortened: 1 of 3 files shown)


browser/extensions/felt/content/felt.xhtml
```diff
@@ -107,7 +107,7 @@
         id="webauthn-notification-icon"
         class="notification-anchor-icon"
         role="button"
-        data-l10n-id="urlbar-web-authn-anchor"
+        data-l10n-id="felt-urlbar-web-authn-anchor"
```
lissyx, line 110: Original string is part of `browser/` so we cannot use it outside of browser app: https://searchfox.org/enterprise-main/rev/017917c10c85768bfdf90714b660bb318053a879/browser/locales/en-US/browser/browser.ftl#41

It is required to be here anyway because of how the PopupNotification system works, but it's not visible to the user

lissyx:
We cannot merge this until https://phabricator.services.mozilla.com/D302705 makes it to `enterprise-main`

jonathanmendez (approved):

lissyx:
Current `enterprise-main` now has the dependency

PR #986, opened 2026-06-01
Revert "Bug 2040350 - Add missing strings for signout modal"

jporter-dev:
Reverts #972. Keeping these strings in `enterprise-main` is no longer needed since l10n [added support](https://github.com/mozilla-l10n/enterprise-firefox-l10n/pull/17) for `enterprise-beta` and `enterprise-release`.


lissyx:
@jporter-dev Do we still need this ?

lissyx (approved):

PR #987, opened 2026-06-01
Bug 2044080 - Fix test_backgroundupdate_exitcodes.js for enterprise builds

victhorlopez:
### Description 

Bugzilla: Bug-2044080

Same as `marAppApplyUpdateStageSuccess.js` Bug-2036826 , the problem is in this one the test expects to hit the code below. 

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [ ] Added tests
[see test pass](https://treeherder.mozilla.org/logviewer?job_id=569984899&repo=enterprise-firefox-pr&task=UishvNjvQ_CvU_wdH6fK0w.0&lineNumber=5953)

### DIFF
--- a/toolkit/xre/nsAppRunner.cpp
+++ b/toolkit/xre/nsAppRunner.cpp
@@ -4811,12 +4811,10 @@ Maybe<ShouldNotProcessUpdatesReason> ShouldNotProcessUpdates(
   // tests like marAppApplyUpdateSuccess.js and
   // marAppApplyUpdateStageSuccess.js to bypass this check and force update
   // processing.
-  if (CheckArgExists("test-process-updates")) {
-    return Nothing();
-  }
+  bool isTesting = CheckArgExists("test-process-updates") || CheckArgExists("test-should-not-process-updates");
   // Don't process updates when launching a Browser from FELT, only Felt should
   // perform that step.
-  if (!is_felt_ui()) {
+  if (!is_felt_ui() && !isTesting) {
     return Some(ShouldNotProcessUpdatesReason::FeltOnlyUpdates);
   }
 #  endif


toolkit/xre/nsAppRunner.cpp
```diff
@@ -4811,12 +4811,10 @@ Maybe<ShouldNotProcessUpdatesReason> ShouldNotProcessUpdates(
   // tests like marAppApplyUpdateSuccess.js and
   // marAppApplyUpdateStageSuccess.js to bypass this check and force update
   // processing.
-  if (CheckArgExists("test-process-updates")) {
-    return Nothing();
-  }
+  bool isTesting = CheckArgExists("test-process-updates") || CheckArgExists("test-should-not-process-updates");
```
lissyx, line 4814: Why not continueing to return `Nothing()` ? I really do not think we should mix the `FeltOnlyUpdate` value with testing-specific bits

victhorlopez: actually closing.. I got confused cause it was passing for me locally.... but it is actually a flaky on m-c... my bad, should have checked better, apologies for the noise and thanks for reviewing

https://bugzilla.mozilla.org/show_bug.cgi?id=1760099

victhorlopez: ok.. I will reopen, now I see what happened.... when I created the bug I created it for the wrong test... 

This is for this test
https://searchfox.org/firefox-main/rev/e28b34ab33dbf49364999070168cbb7e11e8e5bd/toolkit/components/backgroundtasks/tests/xpcshell/test_backgroundtask_shouldprocessupdates.js#20

Which is checking for the code after the felt.

Another option if you are not. happy with this, is to move your check on the very last part of the code? 

Instead of an early return, or are you using this reason somewhere? In the UI?

lissyx: > ok.. I will reopen, now I see what happened.... when I created the bug I created it for the wrong test...
> 
> This is for this test https://searchfox.org/firefox-main/rev/e28b34ab33dbf49364999070168cbb7e11e8e5bd/toolkit/components/backgroundtasks/tests/xpcshell/test_backgroundtask_shouldprocessupdates.js#20

So those tests should pass as expected when running as non felt-managed browser, and we should have our own test to verify that under felt we are skipping those?

> 
> Which is checking for the code after the felt.
> 
> Another option if you are not. happy with this, is to move your check on the very last part of the code?
> 
> Instead of an early return, or are you using this reason somewhere? In the UI?

victhorlopez: ok, then are y ou fine with this implementation https://github.com/mozilla/enterprise-firefox/pull/835#discussion_r3187933662 that was my first idea, and we add the test as you suggested?

```
if (!is_felt_ui() && is_felt_browser()) {
```

PR #993, opened 2026-06-03
Bug 2044567 - Enterprise: refactor Felt XPCOM to use xpcom_method macro

lissyx:
### Description 

Bugzilla: Bug-2044567

This is leveraging `xpcom_method!` to keep our rust implemenation cleaner.

### DIFF
--- a/browser/extensions/felt/rust/src/components.rs
+++ b/browser/extensions/felt/rust/src/components.rs
@@ -15,7 +15,7 @@ use xpcom::interfaces::{
     nsICategoryManager, nsIContentPolicy, nsICookie, nsILoadInfo, nsIObserver, nsIObserverService,
     nsISupports, nsIURI,
 };
-use xpcom::RefPtr;
+use xpcom::{xpcom_method, RefPtr};
 
 use log::{error, trace};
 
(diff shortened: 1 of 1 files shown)


browser/extensions/felt/rust/src/components.rs
```diff
@@ -72,45 +72,74 @@ impl FeltXPCOM {
...
             trace!("FeltXPCOM::SendCookies: oneCookie: {}", cookie.name);
-            rv = self.send(FeltMessage::Cookie(cookie));
+            let r = self.send(FeltMessage::Cookie(cookie));
+            if r.failed() {
+                rv = Err(r);
+            }
         });
 
         rv
     }
```
michaelvanstraten, line 102:
```suggestion
    xpcom_method!(send_cookies => SendCookies(cookies: *const ThinVec<Option<RefPtr<nsICookie>>>));
    fn send_cookies(
        &self,
        cookies: &ThinVec<Option<RefPtr<nsICookie>>>,
    ) -> Result<(), nserror::nsresult> {
        trace!("FeltXPCOM:SendCookies processing {}", cookies.len());
        if cookies.is_empty() {
            return Ok(());
        }

        cookies.iter().flatten().for_each(|x| {
            trace!("FeltXPCOM::SendCookies: oneCookie ....");
            let cookie = crate::utils::nsICookie_wrap(x);
            trace!("FeltXPCOM::SendCookies: oneCookie: {}", cookie.name);
            self.send(FeltMessage::Cookie(cookie)).to_result()?;
        });

        Ok(())
    }
```

browser/extensions/felt/rust/src/components.rs
```diff
@@ -642,45 +654,49 @@ impl FeltRestartForced {
...
-        NS_OK
+        Ok(self.is_restart_forced(a_content_location))
     }
 
-    fn ShouldProcess(
+    xpcom_method!(should_process => ShouldProcess(
+        a_content_location: *const nsIURI,
+        _a_load_info: *const nsILoadInfo
+    ) -> i16);
+
```
michaelvanstraten:
```suggestion
```

browser/extensions/felt/rust/src/components.rs
```diff
@@ -642,45 +654,49 @@ impl FeltRestartForced {
...
         let schemeStr = CString::new(scheme).unwrap();
         let mut isScheme = false;
         unsafe {
-            (*aContentLocation).SchemeIs(schemeStr.as_ptr(), &mut isScheme);
+            aContentLocation.SchemeIs(schemeStr.as_ptr(), &mut isScheme);
         }
         isScheme
     }
 
     fn is_restart_forced(&self, aContentLocation: *const nsIURI) -> i16 {
```
michaelvanstraten:
```suggestion
    fn is_restart_forced(&self, aContentLocation: &nsIURI) -> i16 {
```

We could pass a ref here and adjust the `should_load` and `should_process` `xpcom_macro`.

browser/extensions/felt/rust/src/components.rs
```diff
@@ -72,45 +72,74 @@ impl FeltXPCOM {
...
+    xpcom_method!(send_bool_preference => SendBoolPreference(
+        name: *const nsACString,
+        value: bool
+    ));
+    fn send_bool_preference(
+        &self,
+        name: &nsACString,
+        value: bool,
+    ) -> Result<(), nserror::nsresult> {
+        let name_s = name.to_string();
```
michaelvanstraten: Nit: I know this is not changed here but the rust norm is to shadow here, e.g. `let name = name.to_string();`.

michaelvanstraten (changes requested):

browser/extensions/felt/rust/src/components.rs
michaelvanstraten: Could not verify that this actually compiles, some macOS toolchain failure.

browser/extensions/felt/rust/src/components.rs
michaelvanstraten, line 634: What about this?

browser/extensions/felt/rust/src/components.rs
lissyx: As much as I verified, this was missing implementation on the xpcom_method side, I'll test again

browser/extensions/felt/rust/src/components.rs
lissyx: It's an observer, I do not think we can make much improvement, so left it out for now

browser/extensions/felt/rust/src/components.rs
lissyx: Looks like it builds at least

browser/extensions/felt/rust/src/components.rs
lissyx: `            self.send(FeltMessage::Cookie(cookie)).to_result()?;`

This would make us stop at the first failure, I think we want to continue anyway, at least that was the previous implementation.

lissyx: I must have made a mistake this morning because indeed it seems to work.

browser/extensions/felt/rust/src/components.rs
lissyx: Looks like observer still needs to be unsafe but we can harden a bit the pointer

browser/extensions/felt/rust/src/components.rs
```diff
@@ -626,59 +635,67 @@ impl FeltRestartForced {
...
+        a_content_location: *const nsIURI,
+        _a_load_info: *const nsILoadInfo
+    ) -> i16);
+    fn should_load(
         &self,
-        aContentLocation: *const nsIURI,
-        _aLoadInfo: *const nsILoadInfo,
-        retval: *mut i16,
-    ) -> ::nserror::nsresult {
+        a_content_location: *const nsIURI,
```
michaelvanstraten: This can't be:
```suggestion
        a_content_location: &nsIURI,
```
?

lissyx: right, forgot

browser/extensions/felt/rust/src/components.rs
```diff
@@ -591,28 +604,24 @@ impl FeltRestartForced {
...
-            (*contractID).assign("@mozilla-org/felt-restart-forced;1");
-            let mut retval = nsCString::new();
-            trace!(
-                "FeltRestartForced:new() register with nsICategoryManager: call in unsafe block"
-            );
-            let rv = catMan.AddCategoryEntry(
+        let category = nsCString::from("felt-restart-forced");
+        let contractID = nsCString::from("@mozilla-org/felt-restart-forced;1");
+        let mut retval = nsCString::new();
+        trace!("FeltRestartForced:new() register with nsICategoryManager: call in unsafe block");
```
michaelvanstraten: Nit: the comment is no longer accurate

browser/extensions/felt/rust/src/components.rs
```diff
@@ -626,59 +635,61 @@ impl FeltRestartForced {
...
         unsafe {
-            (*aContentLocation).SchemeIs(schemeStr.as_ptr(), &mut isScheme);
+            aContentLocation.SchemeIs(schemeStr.as_ptr(), &mut isScheme);
         }
         isScheme
     }
 
-    fn is_restart_forced(&self, aContentLocation: *const nsIURI) -> i16 {
+    fn is_restart_forced(&self, aContentLocation: &nsIURI) -> i16 {
+        // SAFETY: XPCOM guarantees `aContentLocation` is a valid, non-null pointer
```
michaelvanstraten: Forgotten comment

michaelvanstraten (approved):
Thank you!

browser/extensions/felt/rust/src/components.rs
```diff
@@ -72,45 +72,75 @@ impl FeltXPCOM {
...
+            return rv;
         }
 
         cookies.iter().flatten().for_each(|x| {
-            trace!("FeltXPCOM::SendCookies: oneCookie ....");
             let cookie = crate::utils::nsICookie_wrap(x);
             trace!("FeltXPCOM::SendCookies: oneCookie: {}", cookie.name);
-            rv = self.send(FeltMessage::Cookie(cookie));
+            let r = self.send(FeltMessage::Cookie(cookie.clone()));
+            if r.failed() {
```
michaelvanstraten, line 91: This is now also a behavioral change, just saying.

lissyx: Where do you see a behavior change? We're still updating `rv` (maybe it's not so good but)

michaelvanstraten: Before you where updating the rv in every iteration, now you are just updating on failure.

michaelvanstraten: Which is arguably better

lissyx: Ah you're right, I was not focused on it.

PR #994, opened 2026-06-03
Bug 2044585 - Refactor tests execution to account for non browser apps

lissyx:
### Description 

Bugzilla: Bug-2044585

Some tests have hard dependency on the app being a browser, mark them as such. For other that can be fixed, do it. Make sure naming is consistent as well.


remote/shared/AppInfo.sys.mjs
```diff
@@ -18,7 +18,8 @@ const ID_THUNDERBIRD = "{3550f703-e582-4d05-9a08-453d09bdfdc6}";
  * @property {boolean} isWindows - Whether the application runs on Windows.
  * @property {boolean} isFirefox - Whether the application is Firefox.
  * @property {boolean} isThunderbird - Whether the application is Thunderbird.
- * @property {boolean} isEnterprise - Whether the application is Firefox Enterprise.
+ * @property {boolean} isFirefoxEnterprise - Whether the application is Firefox Enterprise.
+ * @property {boolean} isEnterpriseEnterprise - Whether the application is Enterprise Enterprise.
```
jonathanmendez:
```suggestion
 * @property {boolean} isThunderbirdEnterprise - Whether the application is Thunderbird Enterprise.
```

remote/shared/AppInfo.sys.mjs
```diff
@@ -69,10 +70,14 @@ ChromeUtils.defineLazyGetter(AppInfo, "isThunderbird", () => {
   return Services.appinfo.ID == ID_THUNDERBIRD;
 });
 
-ChromeUtils.defineLazyGetter(AppInfo, "isEnterprise", () => {
+ChromeUtils.defineLazyGetter(AppInfo, "isFirefoxEnterprise", () => {
   return Services.appinfo.ID == ID_FIREFOX && AppConstants.MOZ_ENTERPRISE;
 });
 
+ChromeUtils.defineLazyGetter(AppInfo, "isEnterpriseEnterprise", () => {
```
jonathanmendez:
```suggestion
ChromeUtils.defineLazyGetter(AppInfo, "isThunderbirdEnterprise", () => {
```

testing/enterprise/manifest.toml
```diff
@@ -3,85 +3,151 @@
...
+]
 
 ["test_felt_browser_access_connector.py"]
+run-if = [
+  "buildapp == 'browser'", # open new tab
+]
 
 ["test_felt_browser_close_warning.py"]
+run-if = [
+  "buildapp == 'browser'", # open new tab
```
jonathanmendez, line 17: There are several tests being marked for browser here that look like they could be relevant/useful for Thunderbird, even if the fix is not trivial.

> For example, it looks like only some of the tests in the `test_felt_browser_close_warning.py` file require opening tabs, but the "close warning" feature may be or become relevant for Thunderbird?

It seems it could be good to have tracking bugs for eventually migrating the relevant tests; is that something you're already otherwise doing?

testing/enterprise/manifest.toml
```diff
@@ -3,85 +3,151 @@
...
-["test_felt_browser_updates_errors.py"]
+["test_felt_console_error.py"]
 
-["test_felt_browser_updates_forward.py"]
+["test_felt_context_menu.py"]
 
 ["test_felt_device_posture.py"]
+run-if = [
+  "buildapp == 'browser'", # hardcoded app name
+]
```
jonathanmendez: It looks like this test was actually adapted to run for Thunderbird? In which case remove the `run-if` for it.
```suggestion
["test_felt_device_posture.py"]
```

testing/enterprise/test_felt_device_posture.py
```diff
@@ -47,8 +47,18 @@ def run_device_posture_content(self):
...
+        expected_app_name = None
+        if app_name == "firefox":
+            expected_app_name = "FirefoxEnterprise"
+        elif app_name == "thunderbird":
+            expected_app_name = "ThunderbirdEnterprise"
+        else:
+            assert False, f"Unsupported app {app_name}"
+
+        assert device_posture["build"]["applicationName"] == expected_app_name, (
+            f"Device posture reports incrorrect applicationName: {device_posture['build']['applicationName']}"
```
jonathanmendez: nit: Most of the other assert messages in the file state the expectation (including the original message here), but maybe a more full message is better, like
```suggestion
            f"Expected device posture to report applicationName: '{expected_app_name}' but got '{device_posture['build']['applicationName']}'"
```
But even if that's not preferred, there's a typo that could be fixed:
```suggestion
            f"Device posture reports incorrect applicationName: {device_posture['build']['applicationName']}"
```

testing/enterprise/test_felt_starts_utf8.py
```diff
@@ -36,16 +36,17 @@ def setUp(self):
 
         self._logger.info(f"Executing from {self._root_dir.name}")
         self._original_binary = self._driver.instance.binary
+        executable = os.path.basename(self._original_binary)
         if sys.platform == "darwin":
             self._driver.instance.binary = os.path.join(
-                self._root_dir.name, "Contents", "MacOS", "firefox"
+                self._root_dir.name, "Contents", "MacOS", executable
             )
         elif sys.platform == "win":
```
jonathanmendez: I believe this clause can be removed.
1. On windows, `sys.platform` is "win32", so I don't believe this is currently being executed, but..
2. That's good because `os.path.basename("C:\\something\\firefox.exe")` returns "firefox.exe", so we don't need to (shouldn't) add the ".exe" extension again

jonathanmendez (changes requested):
A few small changes but also a question about tracking further migration of tests.

testing/enterprise/manifest.toml
lissyx: I think it's too early to be able to take a decision yet, that's why I did what was obvious/trivial. I have ported some of those that needs more changes into their equivalent but in the thunderbird repo for example

testing/enterprise/manifest.toml
lissyx: yep lost in the countless rebases I did, thanks

testing/enterprise/test_felt_device_posture.py
lissyx: taking the suggestion

testing/enterprise/manifest.toml
lissyx: like we have nothing at all yet on bugzilla, so a tracking bug would be hard to place somewhere ...

remote/shared/AppInfo.sys.mjs
```diff
@@ -18,7 +18,8 @@ const ID_THUNDERBIRD = "{3550f703-e582-4d05-9a08-453d09bdfdc6}";
  * @property {boolean} isWindows - Whether the application runs on Windows.
  * @property {boolean} isFirefox - Whether the application is Firefox.
  * @property {boolean} isThunderbird - Whether the application is Thunderbird.
- * @property {boolean} isEnterprise - Whether the application is Firefox Enterprise.
+ * @property {boolean} isFirefoxEnterprise - Whether the application is Firefox Enterprise.
+ * @property {boolean} isThunderbirdEnterprise - Whether the application is Enterprise Enterprise.
```
jonathanmendez: nit: leftover 'Enterprise Enterprise'
```suggestion
 * @property {boolean} isThunderbirdEnterprise - Whether the application is Thunderbird Enterprise.
```

testing/enterprise/manifest.toml
```diff
@@ -3,85 +3,148 @@
...
 
 ["test_felt_browser_signout.py"]
+run-if = [
+  "buildapp == 'browser'",
+]
 
 ["test_felt_browser_signout_on_exit.py"]
+run-if = [
+  "buildapp == 'browser'", # broken???
```
jonathanmendez: What does "broken???" imply here? Does this need a follow-up Bug?

testing/enterprise/manifest.toml
jonathanmendez: Ok, well at this stage I guess I'm fairly confident we will still revisit testing for TB more holistically, so this won't be lost. But if you're tracking "work to do to support TB Enterprise" then may as well add a line item to review these.

jonathanmendez (approved):

PR #1001, opened 2026-06-04
Bug 2045053 - Fix stylelint error for felt-logo background gradient token

jporter-dev:
### Description 

Bugzilla: Bug-2045053

The `use-design-tokens` lint rule validates `var` token names against design token categories when used with the `background` shorthand. Switch `background` to `background-image` for the felt logo gradient token, which does not validate as it does not expect a color token. This change is an established pattern for handling `linear-gradient` backgrounds, two examples below.

- https://searchfox.org/enterprise-main/source/browser/themes/shared/smartbar.css#14,221
- https://searchfox.org/enterprise-main/source/browser/components/aiwindow/ui/content/firstrun.css#30,66

---

### Screenshots

image

_FELT Login with `background-image`_

### DIFF
--- a/browser/extensions/felt/content/styles/felt.css
+++ b/browser/extensions/felt/content/styles/felt.css
@@ -29,7 +29,7 @@ body {
 
 .felt-logo {
   display: flex;
-  background: var(--felt-logo-background-gradient);
+  background-image: var(--felt-logo-background-gradient);
   overflow: hidden;
   flex-direction: column;
   justify-content: center;


lissyx (approved):
Can you just share screenshot to make sure there is no regression?

jporter-dev:
> Can you just share screenshot to make sure there is no regression?

Sure, added a screenshot to the description, with the CSS change shown in devtools.

PR #1002, opened 2026-06-04
Bug 2053429 - Adding Thunderbird Enterprise

lissyx:
### Description 

Bugzilla: Bug-2053429

Adding Thunderbird Enterprise building from `comm-central` on `enterprise-firefox` repo, via a second decision task.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -872,6 +872,8 @@ tasks:
                                         else: >
                                           cd /builds/worker/checkouts/gecko &&
                                           ln -s /builds/worker/artifacts artifacts &&
+                                          sed -i 's|^trust-domain: .*|trust-domain: enterprise|g' comm/taskcluster/config.yml &&
+                                          sed -i 's|^    scope-prefix: .*|    scope-prefix: project:enterprise:releng|g' comm/taskcluster/config.yml &&
                                           ./mach --log-no-times taskgraph decision \
                                            --root=comm/taskcluster \
                                            --pushlog-id='0' \
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -363,7 +363,7 @@ tasks:
                         eventAction: null
                       - $switch:
                             'tasks_for[:6] == "github"':
-                                ownTaskId: {$eval: as_slugid("decision_task")}
+                                ownTaskId: {$eval: as_slugid("gecko_decision_task")}
                                 eventType: '${tasks_for[7:]}'  # strip out 'github-'
                                 eventAction: '${event["action"]}'  # empty string if 'action' doesn't exist
                             'tasks_for in ["action", "cron", "pr-action"]':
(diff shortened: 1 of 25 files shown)


lissyx:
There are still a couple of commits on the branch to experiment things, so commits after (included) `WIP - FORCE TBIRD SHIPPABLE ON PR` should not be reviewed

browser/moz.configure
```diff
@@ -53,6 +53,28 @@ imply_option("MOZ_STUB_INSTALLER", True, when=requires_stub_installer)
...
+def enable_enterprise(value):
+    if value:
+        return True
+
+
+set_config("MOZ_ENTERPRISE", enable_enterprise)
+set_define("MOZ_ENTERPRISE", enable_enterprise)
+
+# Clear the default value to avoid mozilla.com url compiled into product
+imply_option("--with-crashreporter-url", "", when="--enable-enterprise")
```
lissyx, line 76: this will be duplicated in `browser/` and `mail/` until we land those bits on m-c. I plan on doing this soon after we have finished setting up CI for Thunderbird

python/mozbuild/mozbuild/artifacts.py
```diff
@@ -1303,13 +1320,15 @@ def __init__(
...
-        )
+        if substs.get("MOZ_BUILD_APP") == "comm/mail": # and substs.get("MOZ_ENTERPRISE") != "1":
+            job_configuration = ThunderbirdJobConfiguration
+        # elif substs.get("MOZ_BUILD_APP") == "comm/mail" and substs.get("MOZ_ENTERPRISE") == "1":
+        #     job_configuration = EnterpriseThunderbirdJobConfiguration
+        elif substs.get("MOZ_ENTERPRISE") == "1":
+            job_configuration = EnterpriseJobConfiguration
+        else:
+            job_configuration = None
+
```
lissyx: I am not sure if it's correct. The problem was that some of the thunderbird build step / task on CI hits some artifact build code path, and this was needed to move forward

taskcluster/docs/parameters.md
```diff
@@ -119,6 +119,10 @@ topic.
 
 : The GitHub pull request number, if any.
 
+`trigger_project`
+
+: The name of the GitHub repo, distinct from --project CLI parameter
+
```
lissyx: This was required to be able to run both from `enterprise-firefox` as a second decision task **and** from `enterprise-thunderbird`. This would still be valuable but if we are afraid we should be able to drop it

taskcluster/gecko_taskgraph/transforms/job/common.py
```diff
@@ -164,7 +164,12 @@ def support_vcs_checkout(config, job, taskdesc, repo_configs):
             "COMM_BASE_REPOSITORY": config.params["comm_base_repository"],
             "COMM_HEAD_REPOSITORY": config.params["comm_head_repository"],
             "COMM_HEAD_REV": config.params["comm_head_rev"],
+            "COMM_REPOSITORY_TYPE": "git",
```
lissyx: This and a few other bits may need adjustement to work on hg, but I think we have an uptodate mirror on GitHub now, so maybe we can be git-only here.

taskcluster/gecko_taskgraph/transforms/partner_repack.py
```diff
@@ -69,19 +69,18 @@ def populate_repack_manifests_url(config, tasks):
...
-                        "secrets:get:project/releng/gecko/build/level-{level}/partner-github-ssh".format(
-                            **config.params
-                        )
+                )
+            else:
+                task.setdefault("scopes", []).append(
+                    "secrets:get:project/releng/gecko/build/level-{level}/partner-github-ssh".format(
+                        **config.params
                     )
+                )
```
lissyx: this is just a temp hack to run PRs against `https` manifest repo until I start doing the refactor of repos on """official""" ones

lissyx:
It also contains https://github.com/mozilla/enterprise-firefox/pull/1228

.taskcluster.yml
```diff
@@ -657,3 +657,253 @@ tasks:
                                         then:
                                             cron: {$json: {$eval: 'cron'}}
                                       - tasks_for: '${tasks_for}'
+
+    # Second task: Comm (Thunderbird) decision task for git events.
+    - $switch:
+          'tasks_for[:6] == "github" || (tasks_for in ["action", "pr-action"] && parameters["repository_type"] == "git") || (tasks_for == "cron" && repository["type"] == "git")':
```
ahal: This is a `$switch` with a single arm. Would make more sense to use an `$if/then/else`

.taskcluster.yml
```diff
@@ -657,3 +657,253 @@ tasks:
...
+                                ref: 'refs/heads/${event.pull_request.head.ref}'
+                                baseRef: 'refs/heads/${event.pull_request.base.ref}'
+                                baseRev: '${event.pull_request.base.sha}'
+                                headRev: '${event.pull_request.head.sha}'
+                                prNumber: '${event.pull_request.number}'
+                                isPullRequest: true
+                            'tasks_for == "pr-action"':
+                                baseRepoUrl: '${repository.base_url}'
+                                eventType: action
+              in:
```
ahal, line 739: There's a lot of duplication between these two Git-based Decision tasks.. Might be worth considering some strategic yaml anchors to help make it a bit more maintainble

.taskcluster.yml
```diff
@@ -657,3 +657,253 @@ tasks:
...
+                                  env:
+                                      $merge:
+                                          - GECKO_BASE_REPOSITORY: '${baseRepoUrl}'
+                                            GECKO_BASE_REV: '${baseRev}'
+                                            GECKO_HEAD_REPOSITORY: '${repoUrl}'
+                                            GECKO_HEAD_REF: '${ref}'
+                                            GECKO_HEAD_REV: '${headRev}'
+                                            GECKO_REPOSITORY_TYPE: git
+                                            COMM_BASE_REPOSITORY: https://github.com/mozilla/enterprise-thunderbird
+                                            COMM_BASE_REF: '${baseRef}'
```
ahal: So we're not pinning to a specific Thunderbird revision here? This will lead to problems down the road. Maybe good enough for a first pass but I wouldn't recommend this as a long term solution.

.taskcluster.yml
```diff
@@ -657,3 +657,253 @@ tasks:
...
+                                            COMM_HEAD_REPOSITORY: https://github.com/lissyx/enterprise-thunderbird
```
ahal: Don't forget to change this. I'm not sure if COMM_HEAD_REPOSITORY is needed.. I don't think there is a HEAD_REPOSITORY as this isn't running in response to a PR on the thunderbird repo?

.taskcluster.yml
```diff
@@ -657,3 +657,253 @@ tasks:
...
+                                          - $switch:
+                                                'isPullRequest':
+                                                    GECKO_PULL_REQUEST_NUMBER: "${prNumber}"
+                                                'eventType == "action"':
+                                                    ACTION_TASK_GROUP_ID: '${action.taskGroupId}'
+                                                    ACTION_TASK_ID: {$json: {$eval: 'taskId'}}
+                                                    ACTION_INPUT: {$json: {$eval: 'input'}}
+                                                    ACTION_CALLBACK: '${action.cb_name}'
+                                  cache:
+                                      "${trustDomain}-level-${level}-enterprise-thunderbird-checkouts-git-shallow-v1": /builds/worker/checkouts
```
ahal: I think this can share a checkout cache with the other Decision task? It just means there might be two repos in the cache but that should be fine..

It's possible there's a reason to separate the caches, but I'm not immediately thinking of it.

.taskcluster.yml
```diff
@@ -657,3 +657,253 @@ tasks:
...
+
+                                  features:
+                                      taskclusterProxy: true
+                                      chainOfTrust: true
+
+                                  image: mozillareleases/taskgraph:run-task-latest
```
ahal: Probably want to pin to a specific version here

.taskcluster.yml
```diff
@@ -657,3 +657,253 @@ tasks:
...
+                                  maxRunTime: 1800
+                                  command:
+                                      - run-task
+                                      - '--gecko-checkout=/builds/worker/checkouts/gecko'
+                                      - '--gecko-shallow-clone'
+                                      - '--comm-checkout=/builds/worker/checkouts/gecko/comm'
```
ahal: You should also pass `--comm-shallow-clone`, or does that not work because we don't have a revision?

taskcluster/gecko_taskgraph/transforms/job/common.py
```diff
@@ -164,7 +164,12 @@ def support_vcs_checkout(config, job, taskdesc, repo_configs):
...
         })
+
+        repos = json.loads(taskdesc["worker"]["env"].get("REPOSITORIES", {}))
+        repos.update({"comm": "Mozilla Thunderbird"})
+        taskdesc["worker"]["env"]["REPOSITORIES"] = json.dumps(repos)
```
ahal: This should be defined in `taskcluster/config.yml` instead of code

taskcluster/docs/parameters.md
```diff
@@ -119,6 +119,10 @@ topic.
 
 : The GitHub pull request number, if any.
 
+`trigger_project`
+
+: The name of the GitHub repo, distinct from --project CLI parameter
```
ahal: Can you describe what it's needed for here?

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -281,6 +281,10 @@ def get_branch_rev(config):
     ]
 
 
+def get_root_branch_rev(config):
+    return config.params["head_rev"]
```
ahal: This helper seems a bit redundant

taskcluster/gecko_taskgraph/__init__.py
```diff
@@ -75,6 +77,14 @@ def _represent_ro_dict(dumper, data):
...
 del payload_builders["generic-worker"]
 
+orig_file_url = Parameters.file_url
+def local_file_url(self, path, pretty=False):
+    value = orig_file_url(self, path, pretty)
+    if self["trigger_project"] == self["project"]:
+        return value
+    return value.replace(self["comm_head_repository"], self["head_repository"])
+
+Parameters.file_url = local_file_url
```
ahal: This is kind of yucky.. Is there a nice way we could allow this in upstream Taskgraph? Though I agree I'm not sure I want to support the concept of `trigger_project` there.

ahal (changes requested):
Thanks, looks good overall! I mainly focused on the taskgraph bits and skimmed the other parts.

.taskcluster.yml
lissyx: There is, but I'm far from being at ease there to handle that kind of refactoring :(

.taskcluster.yml
lissyx: That's part of the things that will have to change, but yes, for now we decided it will pull from `comm-central` and not a pinned revision, at the risk of having broken thunderbird builds, because we have to balance between multiple things. We talked a lot about that with thunderbird people during the workweek and we'll see how things are working this way vs having to add an extra step of handling pinning

.taskcluster.yml
lissyx: As much as I remember, some of the code especially on `comm_taskgraph` depend on it

.taskcluster.yml
lissyx: I think I did it out of safety, but I'll check if we can avoid it

.taskcluster.yml
lissyx: We do not pin on gecko decision (but maybe we should?)

Which value should we pin to?

.taskcluster.yml
lissyx: I was not aware this was doable, and none of the `comm` decision code I saw was doing it (or I missed it at least), i'll see, it may help with the clone times I was having issues with ...

lissyx: OK it works, but it required a bit of a hack on how to execute `run-task` since we depend on `git ls-remote` to compute the sha1 here for `COMM_BASE_REV` and `COMM_HEAD_REV`.

On comm task there's https://searchfox.org/enterprise-main/source/taskcluster/docker/decision/comm-task-env but we do not have it here because we run `mozillareleases/taskgraph:run-task-latest` image while comm tasks by default runs `mozillareleases/gecko_decision`

.taskcluster.yml
lissyx: Also, technically, it's also a PR against my `enterprise-thunderbird` repo as of now

.taskcluster.yml
lissyx: That looks to be working and speeds things

taskcluster/gecko_taskgraph/transforms/task.py
lissyx: and it should not break execution on `comm-central`, so we need to scope by project

taskcluster/gecko_taskgraph/__init__.py
lissyx: This is required for CoT to be happy when checking things, I need to check if it's still required.

lissyx: Failure raises from https://github.com/mozilla-releng/scriptworker/blob/013a31a07303d7276eec534e0a201e3d19131627/src/scriptworker/cot/verify.py#L1955-L1960

I'm tracking it again but as much as I remember, it was because we run the decision task from gecko and so CoT checks incorrectly

lissyx: Check fails on https://firefox-ci-tc.services.mozilla.com/tasks/Rm3d-_2kSg6P_T_XdUq6MQ:
```
WARNING:scriptworker.cot.verify:verify_repo_matches_url: Paths don't match! /lissyx/enterprise-firefox /lissyx/enterprise-thunderbird/blob/bd0821d971c41e11aeb4891ba99f80ef4a6391fa/taskcluster/kinds/build
```

lissyx: @ahal 
my better fix for this was on `scritpworker` by giving a chance to check both:

```
(venv) alex@portable-alex:~/tmp/mozillasupport/scriptworker/scriptworker$ git diff 
diff --git a/src/scriptworker/cot/verify.py b/src/scriptworker/cot/verify.py
index 6cc77ec..adfbc10 100644
--- a/src/scriptworker/cot/verify.py
+++ b/src/scriptworker/cot/verify.py
@@ -1950,12 +1950,15 @@ def get_source_url(obj):
     source_env_prefix = obj.context.config["source_env_prefix"]
     task = obj.task
     log.debug("Getting source url for {} {}...".format(obj.name, obj.task_id))
+    comm_repo = get_repo(obj.task, source_env_prefix=source_env_prefix, repo_type="comm_head")
     repo = get_repo(obj.task, source_env_prefix=source_env_prefix)
     source = task["metadata"]["source"]
-    if repo and not verify_repo_matches_url(repo, source):
+    not_repo = repo and not verify_repo_matches_url(repo, source)
+    not_comm_repo = comm_repo and not verify_repo_matches_url(comm_repo, source)
+    if (not_repo) and (not_comm_repo):
         raise CoTError(
             "{name} {task_id}: {source_env_prefix} {repo} doesn't match source {source}!".format(
-                name=obj.name, task_id=obj.task_id, source_env_prefix=source_env_prefix, repo=repo, source=source
+                name=obj.name, task_id=obj.task_id, source_env_prefix=source_env_prefix, repo=[repo, comm_repo], source=source
             )
         )
     log.info("{} {}: found {}".format(obj.name, obj.task_id, source))
```

It should not break outside of enterprise, but I'm checking this. I also do not know how much this change is OK?

taskcluster/gecko_taskgraph/transforms/job/common.py
lissyx: I tried that for a long time, but I was hitting issues: https://treeherder.mozilla.org/logviewer?job_id=583757135&repo=enterprise-firefox-pr&task=D7wsX3GRQnSjYMTfH3f12g.0&lineNumber=42383

lissyx: (and then adding more fix was introducing more problems)

lissyx: So I get to:
```
diff --git a/taskcluster/config.yml b/taskcluster/config.yml
index 5c01992bde9..54fdbb19d79 100644
--- a/taskcluster/config.yml
+++ b/taskcluster/config.yml
@@ -171,8 +171,16 @@ taskgraph:
     repositories:
         gecko:
             name: 'Mozilla Firefox'
+            project-regex: enterprise-firefox
+            default-repository: https://github.com/mozilla/firefox
+            default-ref: main
+            type: git
         comm:
             name: 'Mozilla Thunderbird'
+            project-regex: enterprise-thunderbird
+            default-repository: https://github.com/mozilla/comm-central
+            default-ref: main
+            type: git

 workers:
     aliases:
```

From which things complicates because any `comm-checkout: true` is blocked by https://searchfox.org/enterprise-main/rev/36de309b784f868b38a9e60b293a35477203f2c7/taskcluster/gecko_taskgraph/transforms/job/run_task.py#78-79

lissyx: I can try bypassing this when running on a `enterprise*` project? let's see ...

lissyx: Once bypassed it's breaking:
```
[setup 2026-08-07T15:14:36.408+00:00] running as worker:worker
[vcs 2026-08-07T15:14:36.408+00:00] executing ['git', 'config', '--global', '--add', 'safe.directory', '/builds/worker/checkouts/gecko']
[vcs 2026-08-07T15:14:36.410+00:00] executing ['git', 'fetch', '--tags', '--force', '--depth=1', 'https://github.com/mozilla/firefox', 'main']
[vcs 2026-08-07T15:15:39.264+00:00] error: RPC failed; curl 28 Operation too slow. Less than 1024 bytes/sec transferred the last 60 seconds
[vcs 2026-08-07T15:15:39.264+00:00] fatal: protocol error: bad pack header
[vcs 2026-08-07T15:15:41.271+00:00] executing ['git', 'fetch', '--tags', '--force', '--depth=1', 'https://github.com/mozilla/firefox', 'main']
[vcs 2026-08-07T15:16:44.225+00:00] error: RPC failed; curl 28 Operation too slow. Less than 1024 bytes/sec transferred the last 60 seconds
[vcs 2026-08-07T15:16:44.225+00:00] fatal: protocol error: bad pack header
[vcs 2026-08-07T15:16:48.232+00:00] executing ['git', 'fetch', '--tags', '--force', '--depth=1', 'https://github.com/mozilla/firefox', 'main']
[vcs 2026-08-07T15:17:51.239+00:00] error: RPC failed; curl 28 Operation too slow. Less than 1024 bytes/sec transferred the last 60 seconds
[vcs 2026-08-07T15:17:51.239+00:00] fatal: protocol error: bad pack header
```

lissyx: I'll make sure the change is scoped only to enterprise project, and leave a comment

taskcluster/gecko_taskgraph/__init__.py
lissyx: https://github.com/mozilla-releng/scriptworker/pull/817

.taskcluster.yml
lissyx: OK, let's tackle this in a follow up for both decision tasks: https://bugzilla.mozilla.org/show_bug.cgi?id=2061750

.taskcluster.yml
lissyx: @ahal I'd prefer to tackle this in a follow up

taskcluster/gecko_taskgraph/__init__.py
lissyx: @ahal If the last fix you are preparing is not working, can we keep this repo-specific change (maybe improve it's scoping?) and fix correctly things on scriptworker side under less pressure in the following days?

browser/moz.configure
lissyx: https://bugzilla.mozilla.org/show_bug.cgi?id=2062894

python/mozbuild/mozbuild/artifacts.py
lissyx: @coreycb Can you verify this ?

taskcluster/docs/parameters.md
lissyx: This should now not be needed anymore

taskcluster/gecko_taskgraph/__init__.py
ahal: Yeah, let's keep it for now. I was mainly just wondering if there were alternatives. Hopefully we can solve that scriptworker issue, but let's not put it on the critical path.

lissyx: The last fix you deployed to dev was fine and now on production it's all fine as well :)

ahal (approved):
Thanks, this looks reasonable to me!

.taskcluster.yml
```diff
@@ -659,3 +659,274 @@ tasks:
...
+                    eventType: '${tasks_for}'
+                    eventAction: null
+                  - $switch:
+                        'tasks_for[:6] == "github"':
+                            geckoTaskId: {$eval: as_slugid("gecko_decision_task")}
+                            commTaskId: {$eval: as_slugid("comm_decision_task")}
+                            eventType: '${tasks_for[7:]}'
+                            eventAction: '${event["action"]}'
+                        'tasks_for in ["action", "cron", "pr-action"]':
+                            geckoTaskId: '${geckoTaskId}'
```
jcristau, line 680: This seems broken, see https://firefox-ci-tc.services.mozilla.com/tasks/S1IvhbENS_i7oNDm4IkaAg

lissyx: Nice finding

PR #1010, opened 2026-06-09
Bug 2028392 - Use brand strings for Access Connector feature name

jporter-dev:
### Description 

Bugzilla: Bug-2028392

Since Access Connector is the name of this enterprise feature, its name should not be included in the l10n strings for translation, and instead should be referenced from our brand.ftl.

---

### Screenshots 

image

_Access Connector button and panel using brand name_

image

_Access Connector policy description using brand name_

### DIFF
--- a/browser/branding/enterprise/locales/en-US/brand.ftl
+++ b/browser/branding/enterprise/locales/en-US/brand.ftl
@@ -25,9 +25,3 @@
 -brand-product-name = Firefox
 -vendor-short-name = Mozilla
 trademarkInfo = Firefox and the Firefox logos are trademarks of the Mozilla Foundation.
-
-## Enterprise Brand Features
-##
-## These feature names are enterprise product names and must not be translated.
-
--brand-feature-access-connector = Access Connector
--- a/browser/branding/enterprise/locales/en-US/brand.ftl
+++ b/browser/branding/enterprise/locales/en-US/brand.ftl
@@ -25,3 +25,9 @@
 -brand-product-name = Firefox
 -vendor-short-name = Mozilla
 trademarkInfo = Firefox and the Firefox logos are trademarks of the Mozilla Foundation.
+
+## Enterprise Brand Features
+##
+## These feature names are enterprise product names and must not be translated.
+
+-brand-feature-access-connector = Access Connector
--- a/browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
+++ b/browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-policy-AccessConnector = Configure an { -brand-feature-access-connector } for proxying web traffic.
+policy-AccessConnector2 = Configure an { -brand-feature-access-connector } for proxying web traffic.
 policy-AIChatbot = Configure available AI chatbot providers, default provider, and prompt features.
 policy-BlocklistDomainBrowsedTelemetry = Enable and configure security logging/telemetry when { -brand-short-name } blocks a visit to a blocklisted domain.
 policy-DownloadTelemetry = Enable and configure security logging/telemetry when a download is triggered.
--- a/browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
+++ b/browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-policy-AccessConnector = Configure an access connector for proxying web traffic.
+policy-AccessConnector = Configure an { -brand-feature-access-connector } for proxying web traffic.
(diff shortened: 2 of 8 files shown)


browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
```diff
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-policy-AccessConnector = Configure an access connector for proxying web traffic.
```
jporter-dev, line 5: @flodolo or @bcolsson - I assume I'll need to bump the identifiers for all of these since I'm replacing the previously translated "access connector" text with a brand string, despite the "meaning of the sentence" remaining the same. 

Just wanted to verify first, since changing the policy description in particular is a bit more involved due to how the l10n id is dynamic based on policy name.

bcolsson: Yes, you'll need to update the IDs as otherwise the old translations will still use the hardcoded translations.

browser/branding/enterprise/locales/en-US/brand.ftl
```diff
@@ -25,3 +25,9 @@
 -brand-product-name = Firefox
 -vendor-short-name = Mozilla
 trademarkInfo = Firefox and the Firefox logos are trademarks of the Mozilla Foundation.
+
+## Enterprise Brand Features
+##
+## These feature names are enterprise product names and must not be translated.
+
+-brand-feature-access-connector = Access Connector
```
lissyx: I'm surprised, I've often seen non brand-related items with that note to localizer to not translate. Given there is no explicit mentionning of the brand here, I'm a bit unsure we want to include it there.

flodolo: I don't think this string should live in `brand.ftl`.

This can live at the top of `enterprise.ftl`.

lissyx: I think it should stay within `browser/locales/en-US/browser/enterprise/enterprise.ftl` but we can extract it as a standalone string that we re-use ?

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -77,11 +77,11 @@ blocked-by-policy-title-enterprise = Access to this site is restricted
 neterror-blocked-by-policy-page-title-enterprise = Access to this site is restricted
 neterror-blocked-by-policy-contact-admin = If you believe this is an error or need access for work purposes, please contact your IT administrator.
 
-enterprise-access-connector-heading = Access Connector
+enterprise-access-connector-heading2 = { -brand-feature-access-connector }
```
lissyx:
```suggestion
enterprise-feature-access-connector = Access Connector
enterprise-access-connector-heading2 = { enterprise-feature-access-connector }
```

browser/locales/en-US/browser/enterprise/enterprise.ftl
```diff
@@ -77,11 +77,11 @@ blocked-by-policy-title-enterprise = Access to this site is restricted
...
 enterprise-access-connector-info-active = This site is being accessed through a secure company connection.
-enterprise-access-connector-button =
-  .label = Access Connector
-  .tooltiptext = Access Connector
+enterprise-access-connector-button2 =
+  .label = { -brand-feature-access-connector }
+  .tooltiptext = { -brand-feature-access-connector }
```
lissyx:
```suggestion
enterprise-access-connector-button2 =
  .label = { enterprise-feature-access-connector }
  .tooltiptext = { enterprise-feature-access-connector }
```

lissyx (changes requested):

browser/locales/en-US/browser/enterprise/enterprise.ftl
flodolo: Please keep it defined as a Fluent term, not a message. While it's not a brand, using a term is still the best approach.

jporter-dev: Thank you for the feedback! Moving it out of `brand.ftl` makes sense. I've moved it to the top of `enterprise.ftl` as a term and renamed as suggested. 

This also required loading `enterprise.ftl` in the `aboutPolicies.html` since it needs access to the term. It previously could access it since it was already loading `brand.ftl`.

lissyx (approved):
LGTM but if @flodolo can also confirm it's correct I'd feel safer

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
```diff
@@ -2,7 +2,7 @@
...
+policy-AccessConnector2 = Configure an { -enterprise-feature-access-connector } for proxying web traffic.
```
flodolo:
```suggestion
policy-AccessConnector2 = Configure { -enterprise-feature-access-connector } for proxying web traffic.
```

Would it work better without an article?

flodolo (approved):

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
jporter-dev: Thanks for pointing that out, it does sound a bit odd with the proper feature name. I spoke with @bsmth and we decided to update it to "Configure the Access Connector"

PR #1011, opened 2026-06-09
Bug 2046238 - Split live and non live test suite

1rneh:
### Description 

Bugzilla: Bug-2046238

- Restores `toolkit/components/enterprisepolicies/tests/browser/` from upstream/main and removes the live policy tests that had been mixed into the non-live suite.
- Adds back live policy tests to a dedicated live/ suite under toolkit/components/enterprisepolicies/tests/browser/.
- Introduces the preference browser.policies.remote.enabled to control remote policies support

Note for reviewer: Copilot review comments can be ignored on live policy test files because they will become obsolete due to further outstanding refactoring of those files.


toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -181,30 +196,42 @@ EnterprisePoliciesManager.prototype = {
...
+    if (this._isRemotePoliciesSupported()) {
+      remoteProvider = RemotePoliciesProvider.createInstance();
+      // Fetch first set of remote policies during the
+      // initialization of the policy engine
+      await remoteProvider.fetchPoliciesOnStartup();
+      if (remoteProvider.failed) {
+        // bug 2027006 will move the fetching of policies to felt
+        // and not shutdown will be needed then
+        lazy.EnterpriseHandler.initiateShutdown();
+      }
```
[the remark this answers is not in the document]
1rneh: I updated `_isRemotePoliciesSupported` so it can only ever return true on enterprise builds.

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policies_basic_tests.js
```diff
@@ -0,0 +1,139 @@
...
+  registerCleanupFunction(async () => {
+    await clearPolicyEngine();
+  });
+  await EnterprisePolicyTesting.ensureRemotePoliciesMockServer(
+    registerCleanupFunction
+  );
+});
+
+/**
+ * This is a copy of test_simple_policies from toolkit/components/enterprisepolicies/tests/browser/browser_policies_basic_tests.js
```
lissyx, line 16: Would be nice if we could just share the code, i.e., move the `add_task()`'s function into a JS that we load from both tests.

cf https://searchfox.org/enterprise-main/rev/416564dde2bc86317b8a43918223fc17bd74c6aa/security/sandbox/test/browser_content_sandbox_fs.js#12-16 and https://searchfox.org/enterprise-main/rev/416564dde2bc86317b8a43918223fc17bd74c6aa/security/sandbox/test/browser_content_sandbox_fs_tests.js as an example

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policies_basic_tests.js
```diff
@@ -0,0 +1,139 @@
...
+
+  Policies.simple_policy3 = {
+    onAllWindowsRestored(manager, param) {
+      is(param, false, "Param matches what was passed in config file");
+      policy3Ran = true;
+    },
+  };
+
+  // Ensuring that the endpoint /api/browser/policies was hiy
+  await EnterprisePolicyTesting.servePolicyWithJson(
```
lissyx, line 65: Ah so this is not an exact copy since we adapt here. Maybe it's not worth trying to re-use all this and it's better to just keep a separate copy

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policies_basic_tests.js
lissyx: (or not since we have differences in fact)

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -113,6 +115,18 @@ EnterprisePoliciesManager.prototype = {
...
 
+  /**
+   * Remote polling is enabled always when we are in felt launched browser
+   * or if the preference is set explicitly
+   *
+   * @returns {boolean} whether polling remote policies is enabled
+   */
+  _isRemotePoliciesSupported() {
+    return (
+      AppConstants.MOZ_ENTERPRISE && (Services.felt.isFeltBrowser() || Services.prefs.getBoolPref(PREF_REMOTE_POLICIES_ENABLED, false))
```
lissyx:
```suggestion
      AppConstants.MOZ_ENTERPRISE && (Services.felt?.isFeltBrowser() || Services.prefs.getBoolPref(PREF_REMOTE_POLICIES_ENABLED, false))
```

lissyx: I'm wondering if we need `AppConstants.MOZ_ENTERPRISE` really:
 - `Services.felt?.isFeltBrowser()` would gently fail on non enterprise builds,
 - the pref would be either true or false, but does not depend on it

lissyx (approved):
LGTM, maybe we want also an eye from @mkaply ?

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
1rneh: Without `AppConstants.MOZ_ENTERPRISE` one could enable fetching remote policies in a non-enterprise build by setting the preference `browser.policies.remote.enabled=true`.

lissyx: Right, my mind was still set on "this is on m-c and we support live policies not tied to the console". Which is not (yet?) true

1rneh:
> LGTM, maybe we want also an eye from @mkaply ?

Note that this PR and more PRs to come are working towards getting the changes from #321 in in smaller and more manageable pieces. So the majority of this PR is only moving files around and restoring the upstream test suite + introducing the live test suite - all of this with very little behavioral changes.

mkaply:
> > LGTM, maybe we want also an eye from @mkaply ?
> 
> Note that this PR and more PRs to come are working towards getting the changes from #321 in in smaller and more manageable pieces. So the majority of this PR is only moving files around and restoring the upstream test suite + introducing the live test suite - all of this with very little behavioral changes.

Yeah, I saw most of this already all looks good to me.

PR #1019, opened 2026-06-11
NO BUG - Update enterprise-main with release-only changes found during promotion

jonathanmendez:
### Description 

Bugzilla: NO BUG

Since we are new to using the "yolo merge" process for promoting a branch, I attempted to catch changes that might be unique to the release branch (if we committed them directly there instead of into enterprise-main and uplifts). To do this, I locally used `git merge -X theirs enterprise-firefox/enterprise-beta` from a local enterprise-release branch and compared to the results from the yolo merge.

The two items I found were:
- Commit 933b8a5dc65d had been made directly to enterprise-release and would be blapped by the yolo merge
- The enterprise taskcluster config.yml uses `{trust-domain}-t` as a provisioner instead of `gecko-t` but a new entry had been added (thus not encountering merge conflicts). This had been luckily caught and fixed in the enterprise-release merge but not in enterprise-main or enterprise-beta.

---

* Based on findings from: #1018 - the changes are re-added as follow-up commits in the promotion PR

### DIFF
--- a/browser/config/mozconfigs/linux64/release-enterprise
+++ b/browser/config/mozconfigs/linux64/release-enterprise
@@ -1,0 +1,5 @@
+. "$topsrcdir/browser/config/mozconfigs/linux64/common-opt"
+
+ac_add_options --enable-enterprise
+
+. "$topsrcdir/build/mozconfig.common.override"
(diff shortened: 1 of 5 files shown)


lissyx (approved):
LGTM but we should in fact remove the `--enable-instruments` they are only on nightly, not on beta mozconfig

jonathanmendez:
> LGTM but we should in fact remove the `--enable-instruments` they are only on nightly, not on beta mozconfig

I've updated to remove those and will uplift that change to `enterprise-release` as well then

PR #1020, opened 2026-06-11
Bug 2040418 - Posture POST, adding addons, machineids, EDR posture

gcp:
### Description

Extension of https://github.com/mozilla/enterprise-firefox/pull/526

https://bugzilla.mozilla.org/show_bug.cgi?id=2040418

Bugzilla: Bug-2040418

Change policy poll from GET to POST, include device posture in the poll, add reporting for addons, machine serial numbers, and add EDR detection and reporting. This hardcodes the list of EDR for now, we will restructure this in follow-ups.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -3776,11 +3776,13 @@ pref("services.common.log.logger.tokenserverclient", "Debug");
   pref("services.sync.lastversion", "firstrun");
   pref("services.sync.sendVersionInfo", true);
 
+#ifdef MOZ_ENTERPRISE
   // Enterprise: when enabled, the clients engine resets the FxA device
   // registration on sync startup if it detects the machine ID changed (e.g. a
   // profile cloned onto new hardware). Off by default; flip via enterprise
   // policy or pref to enable.
   pref("services.sync.client.machineId.detectChange", false);
+#endif
 
   pref("services.sync.scheduler.idleInterval", 3600);  // 1 hour
   pref("services.sync.scheduler.activeInterval", 600);   // 10 minutes
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -3776,6 +3776,12 @@ pref("services.common.log.logger.tokenserverclient", "Debug");
   pref("services.sync.lastversion", "firstrun");
   pref("services.sync.sendVersionInfo", true);
 
+  // Enterprise: when enabled, the clients engine resets the FxA device
+  // registration on sync startup if it detects the machine ID changed (e.g. a
+  // profile cloned onto new hardware). Off by default; flip via enterprise
+  // policy or pref to enable.
+  pref("services.sync.client.machineId.detectChange", false);
+
   pref("services.sync.scheduler.idleInterval", 3600);  // 1 hour
   pref("services.sync.scheduler.activeInterval", 600);   // 10 minutes
   pref("services.sync.scheduler.immediateInterval", 90);    // 1.5 minutes
--- a/services/sync/modules/engines/clients.sys.mjs
+++ b/services/sync/modules/engines/clients.sys.mjs
@@ -21,6 +21,7 @@
  */
 
 import { Async } from "resource://services-common/async.sys.mjs";
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
 
 import {
   DEVICE_TYPE_DESKTOP,
@@ -45,6 +46,10 @@ ChromeUtils.defineLazyGetter(lazy, "fxAccounts", () => {
   return ChromeUtils.importESModule(
     "resource://gre/modules/FxAccounts.sys.mjs"
   ).getFxAccountsSingleton();
+});
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  MachineId: "resource://gre/modules/MachineId.sys.mjs",
 });
 
 import { PREF_ACCOUNT_ROOT } from "resource://gre/modules/FxAccountsCommon.sys.mjs";
--- a/services/sync/modules/engines/clients.sys.mjs
+++ b/services/sync/modules/engines/clients.sys.mjs
@@ -50,7 +50,7 @@ ChromeUtils.defineLazyGetter(lazy, "fxAccounts", () => {
 
 if (AppConstants.MOZ_ENTERPRISE) {
   ChromeUtils.defineESModuleGetters(lazy, {
-    MachineId: "resource://gre/modules/MachineId.sys.mjs",
+    MachineId: "resource://gre/modules/enterprise/MachineId.sys.mjs",
   });
 }
 
--- a/services/sync/modules/engines/clients.sys.mjs
+++ b/services/sync/modules/engines/clients.sys.mjs
@@ -71,6 +71,10 @@ const COLLECTION_MODIFIED_REASON_FIRSTSYNC = "firstsync";
 const SUPPORTED_PROTOCOL_VERSIONS = [SYNC_API_VERSION];
 const LAST_MODIFIED_ON_PROCESS_COMMAND_PREF =
   "services.sync.clients.lastModifiedOnProcessCommands";
+const MACHINE_ID_PREF = "services.sync.client.machineId";
+const MACHINE_ID_DETECT_CHANGE_PREF =
+  "services.sync.client.machineId.detectChange";
+const MACHINE_ID_CHANGED_TOPIC = "sync:machine-id-changed";
 
 function hasDupeCommand(commands, action) {
   if (!commands) {
@@ -447,7 +451,76 @@ ClientEngine.prototype = {
     this._knownStaleFxADeviceIds = Utils.arraySub(localClients, fxaClients);
   },
 
+  async _checkMachineIdChanged() {
+    const storedMachineId = Services.prefs.getStringPref(MACHINE_ID_PREF, "");
+
+    let currentMachineId = null;
+    try {
+      currentMachineId = await lazy.MachineId.getHashedId();
+    } catch (error) {
+      this._log.warn("Could not get machine ID", error);
+      return false;
+    }
+
+    if (!currentMachineId) {
+      return false;
+    }
+
+    if (storedMachineId && storedMachineId !== currentMachineId) {
+      const oldLocalID = this.localID;
+
+      this._log.info(
+        "Machine ID changed - profile may have been copied to a different machine"
+      );
+
+      // Reset the FxA device registration first, then rotate the local Sync
+      // state. If any step fails (e.g. a transient network error), leave the
+      // stored machine ID untouched and bail rather than aborting the sync, so
+      // the change is detected and retried on the next sync instead of being
+      // silently lost. The individual resets are idempotent, so a retry after a
+      // partial run is safe.
+      try {
+        await this._resetFxADeviceRegistration();
+        Services.prefs.clearUserPref("services.sync.client.GUID");
+        await this.resetClient();
+        await this._tracker.removeChangedID(oldLocalID);
+      } catch (error) {
+        this._log.warn(
+          "Could not reset client state for machine ID change; will retry on next sync",
+          error
+        );
+        return false;
+      }
+
+      Services.obs.notifyObservers(null, MACHINE_ID_CHANGED_TOPIC);
+      Services.prefs.setStringPref(MACHINE_ID_PREF, currentMachineId);
+      return true;
+    }
+
+    if (!storedMachineId) {
+      Services.prefs.setStringPref(MACHINE_ID_PREF, currentMachineId);
+    }
+    return false;
+  },
+
+  async _resetFxADeviceRegistration() {
+    await this.fxAccounts._internal.updateUserAccountData({
+      device: null,
+      encryptedSendTabKeys: null,
+    });
+    await this.fxAccounts.device.getLocalId();
+  },
+
   async _syncStartup() {
+    if (
+      AppConstants.MOZ_ENTERPRISE &&
+      Services.prefs.getBoolPref(MACHINE_ID_DETECT_CHANGE_PREF, false) &&
+      (await this._checkMachineIdChanged())
+    ) {
+      this._log.info("Machine ID changed, regenerating client ID");
+      await this._tracker.addChangedID(this.localID);
+    }
+
     // Reupload new client record periodically.
     if (Date.now() / 1000 - this.lastRecordUpload > CLIENTS_TTL_REFRESH) {
       await this._tracker.addChangedID(this.localID);
--- a/services/sync/modules/engines/clients.sys.mjs
+++ b/services/sync/modules/engines/clients.sys.mjs
@@ -48,9 +48,11 @@ ChromeUtils.defineLazyGetter(lazy, "fxAccounts", () => {
   ).getFxAccountsSingleton();
 });
 
-ChromeUtils.defineESModuleGetters(lazy, {
-  MachineId: "resource://gre/modules/MachineId.sys.mjs",
-});
+if (AppConstants.MOZ_ENTERPRISE) {
+  ChromeUtils.defineESModuleGetters(lazy, {
+    MachineId: "resource://gre/modules/MachineId.sys.mjs",
+  });
+}
 
 import { PREF_ACCOUNT_ROOT } from "resource://gre/modules/FxAccountsCommon.sys.mjs";
 
--- a/services/sync/modules/engines/clients.sys.mjs
+++ b/services/sync/modules/engines/clients.sys.mjs
@@ -492,8 +492,10 @@ ClientEngine.prototype = {
         return false;
       }
 
-      Services.obs.notifyObservers(null, MACHINE_ID_CHANGED_TOPIC);
+      // Persist the new machine ID before notifying, so observers that read
+      // MACHINE_ID_PREF see the updated value rather than the old one.
       Services.prefs.setStringPref(MACHINE_ID_PREF, currentMachineId);
+      Services.obs.notifyObservers(null, MACHINE_ID_CHANGED_TOPIC);
       return true;
     }
 
--- a/services/sync/modules/engines/clients.sys.mjs
+++ b/services/sync/modules/engines/clients.sys.mjs
@@ -485,13 +485,25 @@ ClientEngine.prototype = {
         await this._resetFxADeviceRegistration();
         Services.prefs.clearUserPref("services.sync.client.GUID");
         await this.resetClient();
-        await this._tracker.removeChangedID(oldLocalID);
       } catch (error) {
         this._log.warn(
           "Could not reset client state for machine ID change; will retry on next sync",
           error
         );
         return false;
+      }
+
+      // Best-effort: the device and client rotation above already
+      // succeeded, so a failure here must not undo that progress by
+      // re-triggering it on the next sync. Worst case is a stale entry
+      // for an abandoned client ID left behind in the tracker.
+      try {
+        await this._tracker.removeChangedID(oldLocalID);
+      } catch (error) {
+        this._log.warn(
+          "Could not remove old client ID from tracker after machine ID change",
+          error
+        );
       }
 
       // Persist the new machine ID before notifying, so observers that read
--- a/services/sync/tests/unit/test_clients_engine.js
+++ b/services/sync/tests/unit/test_clients_engine.js
@@ -2119,6 +2119,44 @@ add_task(
   }
 );
 
+add_task(async function test_syncStartup_gates_machine_id_check_on_pref() {
+  // _syncStartup only consults the machine-ID change check when the
+  // detectChange pref is enabled (and on enterprise builds). Verify the pref
+  // actually gates it, rather than testing _checkMachineIdChanged() directly.
+  const DETECT_PREF = "services.sync.client.machineId.detectChange";
+  // Stub the superclass startup so we don't perform a real sync startup.
+  const superStartup = sinon
+    .stub(SyncEngine.prototype, "_syncStartup")
+    .resolves();
+  const checkStub = sinon
+    .stub(engine, "_checkMachineIdChanged")
+    .resolves(false);
+  const addChangedID = sinon.stub(engine._tracker, "addChangedID").resolves();
+
+  try {
+    Services.prefs.setBoolPref(DETECT_PREF, false);
+    await engine._syncStartup();
+    ok(
+      checkStub.notCalled,
+      "Should not run the machine ID check while detectChange is disabled"
+    );
+
+    checkStub.resetHistory();
+    Services.prefs.setBoolPref(DETECT_PREF, true);
+    await engine._syncStartup();
+    ok(
+      checkStub.calledOnce,
+      "Should run the machine ID check when detectChange is enabled"
+    );
+  } finally {
+    Services.prefs.clearUserPref(DETECT_PREF);
+    addChangedID.restore();
+    checkStub.restore();
+    superStartup.restore();
+    await cleanup();
+  }
+});
+
 add_task(async function test_machine_id_first_run_stores_without_reset() {
   const machineIdStub = sinon
     .stub(MachineId, "getHashedId")
--- a/services/sync/tests/unit/test_clients_engine.js
+++ b/services/sync/tests/unit/test_clients_engine.js
@@ -4,12 +4,18 @@
 const { ClientEngine, ClientsRec } = ChromeUtils.importESModule(
   "resource://services-sync/engines/clients.sys.mjs"
 );
-const { MachineId } = ChromeUtils.importESModule(
-  "resource://gre/modules/MachineId.sys.mjs"
-);
 const { AppConstants } = ChromeUtils.importESModule(
   "resource://gre/modules/AppConstants.sys.mjs"
 );
+// MachineId only ships on enterprise builds; import it behind the same gate so
+// this test loads in non-enterprise configurations (the machine-id tasks below
+// are skipped there).
+let MachineId;
+if (AppConstants.MOZ_ENTERPRISE) {
+  ({ MachineId } = ChromeUtils.importESModule(
+    "resource://gre/modules/MachineId.sys.mjs"
+  ));
+}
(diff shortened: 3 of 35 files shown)


browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -6,7 +6,23 @@ const lazy = {};
 
 const FELT_REFRESH_TIMEOUT = 60000;
 
+// Upper bound on a single console request. Without this, a server that accepts
```
gcp: Most of this comment is fluff.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -6,7 +6,23 @@ const lazy = {};
...
+// the connection but stalls the response is only bounded by necko's response
+// timeout (~5 min), during which the policy poller's in-flight guard would skip
+// every tick. Bounding here also activates the xhr.ontimeout handler below.
+const XHR_TIMEOUT_MS = 60000;
+
+// XXX: hardcoded for now. The console does not yet expose which EDR agents to
+// probe for, so we limit detection to the agents we currently care about. These
```
gcp: "These identifiers...unknown ones are silently ignored" this part is also fluff.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -206,12 +222,18 @@ export const ConsoleClient = {
   },
 
   /**
-   * Fetches remote enterprise policies.
+   * Fetches remote enterprise policies, including device posture in the body.
+   *
+   * The endpoint is POST-based; when posture is unavailable we still POST an
+   * empty object rather than falling back to GET, so the fetch keeps working
+   * against the POST endpoint (the fallback path is exactly when posture
```
gcp: fallback path comment is nonsensical, drop it

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -603,21 +628,48 @@ export const ConsoleClient = {
...
   /**
    * @typedef {object} DevicePosture
    * @property {object} os Telemetry-reported os information.
    * @property {object|undefined} security Telemetry-reported security software info (windows only)
    * @property {object} build Telemetry-reported build info info
    * @property {DeviceNetwork} network Network posture (placeholders for now).
+   * @property {DeviceAddon[]|null} extensions Installed browser addons, or null if not yet available.
+   * @property {DeviceMachineId|null} machineId Stable machine identifier, or null if unavailable.
+   * @property {boolean} secureBootEnabled Whether Secure Boot is enabled.
+   * @property {boolean} isDomainJoined Whether the machine is joined to a domain (Windows on-prem AD or Azure AD/Entra; false on other platforms for now).
```
gcp: "false on other platforms for now" is not useful information

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -628,6 +680,60 @@ export const ConsoleClient = {
...
+          enabled: addon.isActive,
+        }));
+      } catch {
+        return null;
+      }
+    };
+
+    // The console identifies a device by its raw machine identifier (e.g. the
+    // device serial); the source tier is reported alongside it so a change of
+    // source can be told apart from a genuine hardware change.
```
gcp: It's not like the admin can tell which case is true if the source changes. The comment is nonsensical.

browser/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -641,16 +747,41 @@ export const ConsoleClient = {
       ...(os_short_name != null && { os_short_name }),
     };
 
+    // Gather the EDR agents present on the client. The catalog of known agents
+    // lives in the EDR-checker component, not here; EdrDetection wraps the
```
gcp: "not here" => We see this captain obvious.

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+    Eset,                // ESET Endpoint Security
+    Cylance,             // BlackBerry Cylance
+    Symantec,            // Symantec Endpoint Security / Protection
+    TrendMicro,          // Trend Micro Apex One
+}
+
+impl EdrId {
+    // The complete catalog of known EDR agents. This is the single source of
+    // truth; callers (including the device posture payload) enumerate via the
+    // getPresentEdrs() XPCOM method rather than hardcoding identifiers.
```
gcp: "rather than hardcoding identifiers" is a lie - they have to hardcode strings. Remove the faulty comment.

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+            EdrId::Symantec => "symantec",
+            EdrId::TrendMicro => "trend-micro",
+        }
+    }
+
+    /// Maps a string identifier (as produced by `as_str`) back to an `EdrId`.
+    /// Returns `None` for unknown identifiers so callers can ignore them.
+    // Inherent method intentionally named to mirror `as_str`; it is not a
+    // `std::str::FromStr` implementation (the return type is `Option`, not
+    // `Result`).
```
gcp: It could have been a Result, eh.

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+    #[allow(clippy::should_implement_trait)]
```
gcp: Having to shut up clippy is a bad sign.

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+// Per-agent cache of (time detected, present?). Guarded by a Mutex because the
+// detection runs on a moz_task background thread.
+static CACHE: Mutex<Option<HashMap<EdrId, (Instant, bool)>>> = Mutex::new(None);
+
+// Upper bound on a single external probe (a service-status command, the
+// system-extension listing, etc.). A probe that exceeds this is treated as
+// "could not determine" so one wedged command cannot stall the whole sweep.
+// This is independent of, and shorter than, the JS-side EDR_DETECTION_TIMEOUT_MS
+// that bounds the caller.
+pub(crate) const PROBE_TIMEOUT: Duration = Duration::from_secs(2);
```
gcp: Seems very small.

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+
+// How often run_command_bounded re-checks a still-running child for exit.
+const PROBE_POLL_INTERVAL: Duration = Duration::from_millis(20);
```
gcp: Given that this is for hang detection(?), this also seems much too aggressive.

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+
+/// Captured result of a bounded external command. Each platform's probes read
+/// only the part they need -- Linux the exit status, macOS/Windows the stdout --
```
gcp: Don't they call this an enum in Rust?

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+
+/// Runs an external command, waiting up to `PROBE_TIMEOUT` for it to exit.
+///
+/// Unlike `Command::output()`, a command that overruns the timeout is **killed
+/// and reaped**, so a wedged probe cannot linger as a zombie or keep a thread
+/// blocked in a syscall across future sweeps. Returns `None` if the command
+/// could not be spawned, was killed for overrunning, or failed to wait.
+///
+/// stdout is drained only after the child exits. This assumes probe output is
+/// small (well under the OS pipe buffer); every command we run here
```
gcp: Sounds potentially brittle if the OS pipe buffer isn't well known or fixed?

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+static DETECTION_LOCK: Mutex<()> = Mutex::new(());
+
+/// Determines which of `requested` agents are present, returning their string
+/// identifiers. A single system snapshot is captured and reused across all
+/// agents that still need detection (i.e. are not already cached and fresh).
+///
+/// Runs on a background thread; must not touch main-thread-only state.
+fn detect_present_edrs(requested: &[EdrId]) -> Vec<&'static str> {
+    // Serialize sweeps. Only one detection runs at a time; a concurrent caller
+    // blocks here until the in-progress sweep has published its results, then
```
gcp: Isn't this explanation duplicated from above?

browser/extensions/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,570 @@
...
+    // reading a still-empty cache and reporting a false "none present". The
+    // lock is held across the probes, but each probe is itself bounded
+    // (PROBE_TIMEOUT), so the wait is bounded too. We recover a poisoned lock
+    // (a panicking probe) so detection is not wedged for the rest of the
+    // session.
+    let _sweep = DETECTION_LOCK.lock().unwrap_or_else(|e| e.into_inner());
+
+    let now = Instant::now();
+
+    // Figure out which agents are missing from the cache or have gone stale.
```
gcp: It's unlikely the console will switch which agents it wants to detect (a lot - it should be very rare), so perhaps this can be simplified a lot.

toolkit/components/felt/rust/src/edr_checker_win.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+    if ok == 0 {
+        return None;
+    }
+    Some(status.dwCurrentState == SERVICE_RUNNING)
+}
+
+/// Fallback service check via the `sc` command. Note this parses English
+/// output ("STATE" / "RUNNING") and is only used when the SCM API is
+/// unavailable.
```
gcp, line 151: I wonder if we want provenance data/telemetry on this at some point.

browser/extensions/felt/rust/nsIEdrChecker.idl
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * @param presentEdrs the identifiers of the requested EDR agents that were
+   *                    detected as present (e.g. "crowdstrike", "sentinelone").
+   */
+  void onComplete(in Array<ACString> presentEdrs);
+};
+
+/**
+ * Detects which well-known Endpoint Detection and Response (EDR) agents are
+ * present on the system. The catalog of known agents is defined internally;
```
gcp: The remark about the catalog is nonsensical and not useful.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -1064,16 +1065,40 @@ class RemotePoliciesProvider {
   }
 
   _performPolling() {
-    lazy.ConsoleClient.getRemotePolicies()
+    // Skip this tick if the previous poll has not finished. Posture collection
```
gcp: overly long comment

toolkit/modules/EdrDetection.sys.mjs
```diff
@@ -0,0 +1,69 @@
...
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  clearTimeout: "resource://gre/modules/Timer.sys.mjs",
+  setTimeout: "resource://gre/modules/Timer.sys.mjs",
+});
+
+// Default upper bound on detection, after which we report none present.
+const DEFAULT_TIMEOUT_MS = 10000;
```
gcp: Might be small if we really probed for everything?

xpcom/base/nsSystemInfo.cpp
```diff
@@ -1467,6 +1467,44 @@ bool GetSecureBootStatus_Windows() {
 
   return false;
 }
+
+#  include <lm.h>  // Net{Get,Free}* domain/AAD join APIs (netapi32)
+
+bool GetDomainJoinedStatus_Windows() {
+  // On-prem Active Directory: NetGetJoinInformation is the same Win32 API that
```
gcp: Nobody cares what Chromium does, only whether the code is correct or not.

toolkit/components/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,520 @@
...
+#[xpcom(implement(nsIEdrChecker), atomic)]
+pub struct EdrCheckerXPCOM {}
+
+#[allow(non_snake_case)]
+impl EdrCheckerXPCOM {
+    pub fn new() -> RefPtr<EdrCheckerXPCOM> {
+        EdrCheckerXPCOM::allocate(InitEdrCheckerXPCOM {})
+    }
+
+    fn GetPresentEdrs(
```
lissyx: There are some `unsafe` blocks we can avoid using `xpcom_method!` here

toolkit/components/felt/rust/src/edr_checker.rs
```diff
@@ -0,0 +1,520 @@
...
+        &self,
+        requested_ids: *const ThinVec<nsCString>,
+        callback: *const nsIEdrCheckerCallback,
+    ) -> nsresult {
```
lissyx: We should use `xpcom_method!` to avoid the `unsafe` blocks after.

argl:
This looks all fine to me. Difficult (impossible) to test all the different EDRs on all platforms, but at least for my work machine it correctly detected crowdstrike.

One note: We have [added POST support to the policies route](https://github.com/mozilla/enterprise-console/pull/305) in the console service to be prepared when this PR lands.

modules/libpref/init/all.js
```diff
@@ -3771,6 +3771,12 @@ pref("services.common.log.logger.tokenserverclient", "Debug");
   pref("services.sync.lastversion", "firstrun");
   pref("services.sync.sendVersionInfo", true);
 
+  // Enterprise: when enabled, the clients engine resets the FxA device
+  // registration on sync startup if it detects the machine ID changed (e.g. a
+  // profile cloned onto new hardware). Off by default; flip via enterprise
+  // policy or pref to enable.
+  pref("services.sync.client.machineId.detectChange", false);
```
lissyx, line 3784:
```suggestion
#ifdef MOZ_ENTERPRISE
  pref("services.sync.client.machineId.detectChange", false);
#endif
```

toolkit/modules/moz.build
```diff
@@ -162,6 +162,7 @@ EXTRA_JS_MODULES += [
     "CreditCard.sys.mjs",
     "DeferredTask.sys.mjs",
     "E10SUtils.sys.mjs",
+    "EdrDetection.sys.mjs",
```
lissyx: limit to `MOZ_ENTERPRISE` ?

toolkit/modules/moz.build
```diff
@@ -185,6 +186,7 @@ EXTRA_JS_MODULES += [
     "KeywordUtils.sys.mjs",
     "Log.sys.mjs",
     "LogManager.sys.mjs",
+    "MachineId.sys.mjs",
```
lissyx: limit to `MOZ_ENTERPRISE` ?

services/sync/modules/engines/clients.sys.mjs
```diff
@@ -47,6 +48,10 @@ ChromeUtils.defineLazyGetter(lazy, "fxAccounts", () => {
   ).getFxAccountsSingleton();
 });
 
+ChromeUtils.defineESModuleGetters(lazy, {
+  MachineId: "resource://gre/modules/MachineId.sys.mjs",
+});
+
```
lissyx: limit to `MOZ_ENTERPRISE` ?

services/sync/tests/unit/test_clients_engine.js
```diff
@@ -4,6 +4,12 @@
 const { ClientEngine, ClientsRec } = ChromeUtils.importESModule(
   "resource://services-sync/engines/clients.sys.mjs"
 );
+const { MachineId } = ChromeUtils.importESModule(
+  "resource://gre/modules/MachineId.sys.mjs"
+);
```
lissyx: limit to `MOZ_ENTERPRISE` ?

services/sync/tests/unit/test_clients_engine.js
```diff
@@ -2045,6 +2051,234 @@ add_task(async function update_known_stale_clients() {
   stubFetchFxADevices.restore();
 });
 
+add_task(
```
lissyx, line 2060: limit to `MOZ_ENTERPRISE` ?

services/sync/tests/unit/test_clients_engine.js
```diff
@@ -2045,6 +2051,234 @@ add_task(async function update_known_stale_clients() {
...
+      machineIdStub.restore();
+      resetClientStub.restore();
+      removeChangedIDStub.restore();
+      Services.prefs.clearUserPref("services.sync.client.machineId");
+      await cleanup();
+    }
+  }
+);
+
+add_task(async function test_syncStartup_gates_machine_id_check_on_pref() {
```
lissyx: limit to `MOZ_ENTERPRISE` ?

testing/enterprise/felt_tests.py
```diff
@@ -397,6 +400,11 @@ def do_GET(self):
             m = json.dumps(self.server.device_posture_payload)
             contentType = "application/json"
 
+        elif path == "/sso/get_device_posture_history":
```
lissyx, line 394: This is not an endpoint that exists on the real console right?
Can we do without? maybe via some shared variable of `self.server`

testing/enterprise/felt_tests.py
```diff
@@ -469,12 +477,31 @@ def do_POST(self):
...
+                self.rfile.read(int(self.headers.get("Content-Length")))
+            )
+            self.server.device_posture_payload = payload
+            if not hasattr(self.server, "device_posture_history"):
+                self.server.device_posture_history = []
+            self.server.device_posture_history.append(payload)
+            if self.server.policies_fail_request.value:
+                self.reply("", 500, "Internal Server Error", "application/json")
+                return
+            m = self.build_policies_response()
```
lissyx, line 494: There is already a handler on `ConsoleHttpHandler` class, are you redefining on purpose? Were you trying to just move it around ?

testing/enterprise/test_felt_device_posture.py
```diff
@@ -16,10 +16,88 @@
 
 
 class FeltDevicePosture(FeltTests):
+    # Identifiers of every EDR agent the EDR-checker knows about; must match
+    # EdrId::as_str() in toolkit/components/felt/rust/src/edr_checker.rs.
```
lissyx, line 21: We should have rather some api returning us the list so it does not risk to get stale ?

testing/enterprise/test_felt_device_posture.py
```diff
@@ -16,10 +16,88 @@
...
+        "eset",
+        "cylance",
+        "symantec",
+        "trend-micro",
+    }
+
     def test_felt_device_posture(self):
         super().run_felt_base()
         self.run_device_posture_content()
+        # The FELT login flow discards the parent's auth window, so EDR probing
```
lissyx: We have a pref you can use to keep the window open if it improve making the test closer to real life

testing/enterprise/test_felt_device_posture.py
```diff
@@ -184,6 +276,40 @@ def run_device_posture_content(self):
 
         assert found_one_ipv6, "Device posture reports network interfaces (IPv6)"
 
+        assert "extensions" in device_posture, "Device posture reports extensions"
```
lissyx, line 288: Do we want to assert the payload ?

testing/enterprise/test_felt_device_posture.py
```diff
@@ -184,6 +276,40 @@ def run_device_posture_content(self):
...
+
+        assert "machineId" in device_posture, "Device posture reports machineId"
```
lissyx: Same we may want to assert the payload

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -6,7 +6,17 @@ const lazy = {};
 
 const FELT_REFRESH_TIMEOUT = 60000;
 
+// Bound a single console request so a stalled server can't wedge the poller.
+const XHR_TIMEOUT_MS = 60000;
```
lissyx, line 10: nit: we have that kind of timeout with a similar value in different places, maybe we want to unify those

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -629,6 +673,57 @@ export const ConsoleClient = {
...
 
+    // AddonManager is only available in the full browser process, not in
+    // the FELT login window. Returns null in the FELT UI. When
+    // waitForAddons is true (periodic poll), blocks until AddonManager is
+    // ready so extensions are always reported. When false (startup),
+    // returns null if AddonManager isn't ready yet to avoid blocking.
+    const getExtensions = async () => {
+      try {
+        if (Services.felt.isFeltUI()) {
+          return null;
```
lissyx, line 698: I'll try to look into what we can do to get this from felt directly

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -642,16 +737,36 @@ export const ConsoleClient = {
       ...(os_short_name != null && { os_short_name }),
     };
 
+    const getPresentEDRs = async () =>
+      (await lazy.EdrDetection.getPresentEdrs(EDR_AGENTS_TO_PROBE)).map(
+        name => ({ name })
+      );
+
+    // These probes are independent, and some are slow (subprocess spawns, addon
```
lissyx, line 759: Maybe we can even do that eagerly during the startup phase so most/all is ready by the time we need it ?

toolkit/locales/en-US/toolkit/about/aboutSupport.ftl
```diff
@@ -33,6 +33,7 @@ security-software-name = Name
 security-software-antivirus = Antivirus
 security-software-antispyware = Antispyware
 security-software-firewall = Firewall
+security-software-edr = Endpoint Detection and Response
```
lissyx: If this string is specific to enterprise we should add it in enterprise specific fluent file

lissyx (changes requested):
There are places where we guard by `MOZ_ENTERPRISE` and others where we do not that feels like we should. There's also `xpcom_method!` we could leverage to limit `unsafe` blocks I think

toolkit/modules/moz.build
```diff
@@ -222,6 +220,12 @@ MOZ_SRC_FILES += [
     "InputPickerPanelCommon.sys.mjs",
 ]
 
+if CONFIG["MOZ_ENTERPRISE"]:
+    EXTRA_JS_MODULES += [
```
lissyx: nit: maybe we want `EXTRA_JS_MODULES.enterprise += ...` here?

lissyx (approved):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
```diff
@@ -1112,16 +1113,38 @@ class RemotePoliciesProvider {
   }
 
   _performPolling() {
-    lazy.ConsoleClient.getRemotePolicies()
+    // Skip this tick if the previous poll is still running; the interval is
+    // server-controlled and posture collection can be slow.
+    if (this._pollInFlight) {
+      return;
+    }
+    this._pollInFlight = true;
```
1rneh: This in-flight flag can be removed. I'm already adding one in #1069.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -303,10 +318,12 @@ export const ConsoleClient = {
   /**
    * Collect the device posture data and send them to the console.
    *
+   * @param {object} root0
+   * @param {boolean} root0.waitForAddons
```
1rneh: `nit`: 
```suggestion
   * @param {object} [options]
   * @param {boolean} [options.waitForAddons=false] - Whether to block until AddonManager is ready so extensions are always reported.
```
And same for the JSDoc of `collectDevicePosture`.

1rneh (changes requested):
_Note: Only reviewed on ConsoleClient.sys.mjs and EnterprisePoliciesParent.sys.mjs._

Two requests (first blocking; second a potential follow-up; inline comments are non-blocking nits):

1. `_performPolling` and `fetchPoliciesOnStartup` both repeat the same dance - collect posture, fall back to `null`, then call `getRemotePolicies(posture)`. Could `getRemotePolicies()` collect it? Building the request payload feels like `ConsoleClient`'s job, not the policy engine's. The callers only differ by startup vs. poll, so `getRemotePolicies({ isStartup })` - mapping internally to `waitForAddons = !isStartup` - collapses each to one line and keeps AddonManager details out of the provider entirely. 
Bonus: should sidestep the merge conflict with #1069 that whoever lands second would otherwise have to resolve. 🥲
2. The posture code has grown enough that I think it deserves its own `DevicePosture.sys.mjs` module, with `ConsoleClient` calling `DevicePosture.collect()` where the posture is needed. Fine as a follow-up although it looks like a straightforward change.

1rneh (approved):
From my earlier review:
> The callers only differ by startup vs. poll, so getRemotePolicies({ isStartup }) - mapping internally to waitForAddons = !isStartup - collapses each to one line and keeps AddonManager details out of the provider entirely.

Could you add this extra layer of abstraction? Let the policy engine pass `{ isStartup }` to `getRemotePolicies`, rather than device-posture internals like `waitForAddons`. That keeps the caller expressing intent (startup vs. poll) and leaves the waitForAddons mapping as an implementation detail inside the ConsoleClient or device posture collection.

r+ with that addressed.

PR #1021, opened 2026-06-11
Bug 2046862 - Disable noWindowAutoRestart on Enterprise

lissyx:
### Description 

Bugzilla: Bug-2046862

When running FELT after launching the browser, it is effectively a window-less instance of Firefox. On macOS it means there is a risk that background updates gets applied **and** that FELT restarts, without user action, creating troubles. Disable this feature.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -200,7 +200,14 @@ pref("app.update.langpack.enabled", true);
 #ifdef XP_MACOSX
   // If set to true, Firefox will automatically restart if it is left running
   // with no browser windows open.
-  pref("app.update.noWindowAutoRestart.enabled", true);
+  #if defined(MOZ_ENTERPRISE)
+    // Bug 2046862 : For Enterprise, FELT is running without a window, so it is
+    // required to disable this feature.
+    pref("app.update.noWindowAutoRestart.enabled", false);
+  #else
+    pref("app.update.noWindowAutoRestart.enabled", true);
+  #endif
+
   // How long to wait after all browser windows are closed before restarting,
   // in milliseconds. 5 min = 300000 ms
   pref("app.update.noWindowAutoRestart.delayMs", 300000);


jonathanmendez (approved):

browser/app/profile/firefox.js
```diff
@@ -200,7 +200,14 @@ pref("app.update.langpack.enabled", true);
 #ifdef XP_MACOSX
   // If set to true, Firefox will automatically restart if it is left running
   // with no browser windows open.
-  pref("app.update.noWindowAutoRestart.enabled", true);
+  #if defined(MOZ_ENTERPRISE)
```
1rneh, line 203: Or we have felt request from the browser whether it has any windows open in order to not loose this feature?

lissyx: > Or we have felt request from the browser whether it has any windows open in order to not loose this feature?

Which feature, the ability to perform full update cycle when no window is open on macOS ? Given FELT manages updates, the feature is not applicable here

lissyx: > Or we have felt request from the browser whether it has any windows open in order to not loose this feature?

Correct me if I'm wrong, but either felt is showing a window, or it's in background and the browser is running, we have no in-between where that feature is useful in the general case?

1rneh: How so? The feature in our setup would be for felt to perform full update cycle when the launched browser are windowless on macos.

lissyx: > How so? The feature in our setup would be for felt to perform full update cycle when the launched browser are windowless on macos.

Turns out you're right, I'm not used to that because I do not use macOS.

lissyx: Bug-2046895

browser/app/profile/firefox.js
```diff
@@ -200,7 +200,14 @@ pref("app.update.langpack.enabled", true);
...
+    // Bug 2046862 : For Enterprise, FELT is running without a window, so it is
+    // required to disable this feature.
+    pref("app.update.noWindowAutoRestart.enabled", false);
+  #else
+    pref("app.update.noWindowAutoRestart.enabled", true);
```
michaelvanstraten, line 208: I don't get this feature anyway because if I close all my windows I don't want you to just restart my instance. I have clear cookies on restart and sometimes I just close my windows with the intention to reuse my cookies.

michaelvanstraten (approved):

PR #1030, opened 2026-06-12
Enterprise main merge 2026061200

jonathanmendez:
### Description

Bugzilla: NO BUG  

No merge conflicts.


jporter-dev (approved):

taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -494,6 +494,10 @@ def filter(task):
         ):
             return False
 
+        # Skip source-test-python tasks which require significant greening effort for enterprise builds
+        if task.kind == "source-test" and task.label.startswith("source-test-python"):
+            return False
+
```
ahal: This will work, but I consider parsing labels in `target_tasks` to be a bit of an anti-pattern. Because you have the `target_tasks_default` filter at the start of this function, it means that `run-on-projects` should be honoured.

So instead of this, you could add:
```
run-on-projects: []
```

Above this line:
https://searchfox.org/firefox-main/source/taskcluster/kinds/source-test/python.yml#42

That will cause all `python-test` tasks to no run. But you might want to add it only to the specific tasks that are failing rather than the `task-defaults` section

PR #1033, opened 2026-06-12
Bug 2040381 - Show custom error page when Access Connector is unavailable

jporter-dev:
### Description 

Bugzilla: Bug-2040381

This is part 1 of 2 for this bug. When an enterprise's Access Connector routes a site through a secure connection service and that service is unavailable, Firefox currently shows a generic proxy failure error. This PR detects the Access Connector state (`isAccessConnectorActive()`) and on `proxyConnectFailure` / `proxyResolveFailure`, resolves to a new `accessConnectorFailure` error config instead of the standard proxy error.

Part 2 depends on the Access Connector UX PR landing, this piece is able to be completed without it.

---

* Depends on: #683 

---

### Screenshots 

access-connector-proxy-error

_Access Connector enabled with proxy error_

standard-proxy-error

_Access Connector disabled with proxy error_

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. Enable the `AccessConnector` policy with a proxy server that is unreachable (ex: change `Port` by 1)
2. Navigate to a site covered by the inclusion patterns
3. Observe `about:neterror` with the Access Connector error page

**Expected result:**

The Access Connector error page pictured in the screenshots renders and the generic proxy error no longer appears for this case. The proxy error page should still appear for non-Access Connector proxy failures.


toolkit/content/errors/error-lookup.mjs
```diff
@@ -37,12 +38,20 @@ export function resolveErrorID({
...
   if (noConnectivity) {
     return "NS_ERROR_OFFLINE";
   }
+  if (
+    (gErrorCode === "proxyConnectFailure" ||
+      gErrorCode === "proxyResolveFailure") &&
+    accessConnectorActive
+  ) {
+    return "accessConnectorFailure";
+  }
```
lissyx, line 54: I'm a bit sad that VPN handles it this way :(

toolkit/content/aboutNetErrorHelpers.mjs
```diff
@@ -43,6 +43,19 @@ export const VPN_ACTIVE = RPMGetBoolPref(
...
+ * Protection. AC sets mode to 3 (MODE_INCLUSION); VPN uses modes 0-2.
+ *
+ * @returns {boolean}
+ */
+export function isAccessConnectorActive() {
+  return (
+    RPMGetBoolPref("browser.ipProtection.enabled", false) &&
+    RPMGetIntPref("browser.ipProtection.mode", 0) === 3
+  );
+}
```
lissyx, line 58: So given VPN is not a feature we support, shouldn't we have that implementation wired into `isVpnActive`? This would simplify a bit code paths

lissyx (approved):
LGTM

testing/enterprise/test_felt_browser_access_connector.py
```diff
@@ -211,3 +220,56 @@ def run_load_page_without_access_connector(self):
...
+
+    def run_check_access_connector_neterror_card(self):
+        self._logger.info("Checking custom error title l10n id")
+        title_l10n_id = self._child_driver.execute_script(
+            """
+            const card = document.querySelector("net-error-card");
+            return card?.shadowRoot?.querySelector("#error-title")
+                ?.getAttribute("data-l10n-id");
+            """
+        )
```
lissyx: Why execute script? We should be able to directly find the elements using python level APIs ?

testing/enterprise/test_felt_browser_access_connector.py
```diff
@@ -211,3 +220,56 @@ def run_load_page_without_access_connector(self):
...
+        self._logger.info("Checking error page illustration image")
+        image_src = self._child_driver.execute_script(
+            """
+            const card = document.querySelector("net-error-card");
+            return card?.shadowRoot?.querySelector(".img-container img")
+                ?.getAttribute("src");
+            """
+        )
+        assert image_src == (
+            "chrome://global/skin/enterprise/access-connector-neterror.svg"
```
lissyx: nit: do we want to assert the whole path ? We could just assert `endswith("enterprise/access-connector-neterror.svg")` ?

toolkit/content/aboutNetErrorHelpers.mjs
jporter-dev: Yeah, I can simplify this. I wasn't sure if doing that would be better or worse for future changes and merge conflicts.

testing/enterprise/test_felt_browser_access_connector.py
jporter-dev: Simplified the image src check and updated these to use `find_element` on the `net-error-card` shadow root

1rneh (changes requested):
Looks good to me. 

The only request I have is to please gate enterprise specific changes so it doesn't leak into or break non-enterprise builds.

toolkit/actors/NetErrorChild.sys.mjs
```diff
@@ -204,6 +206,10 @@ export class NetErrorChild extends RemotePageChild {
     return lazy.AppInfo.isFirefox;
   }
 
+  RPMIsEnterprise() {
```
jporter-dev, line 209: @1rneh - Added this for enterprise gating since there's no direct access to `AppConstants` from the neterror page.

1rneh (approved):

PR #1065, opened 2026-06-20
Bug 2049134 -  Enable SQLite encryption using a Lockstore Password KEK

beurdouche:
### Description

Bugzilla: Bug-2049134

The objective of this change is to enable SQLite encryption by default in the Felt and Firefox profiles.
The change sets `security.storage.encryption.sqlite.enabled` to `true` and defines a Lockstore KEKs.
Since Felt starts without knowing the user, we use a Lockstore Local KEK to encrypt SQLite databases in Felt.
After user authentication via SSO, Felt fetches the PrimarySecret and passes it to the Firefox process through IPC
so that a Lockstore Password KEK is set for the Firefox Profile.

### Testing

* [x] Added tests as part of the SQLite Encryption patch in Bug-1996558

### DIFF
--- a/modules/libpref/init/StaticPrefList.yaml
+++ b/modules/libpref/init/StaticPrefList.yaml
@@ -19143,7 +19143,11 @@
 # ciphertext. Read at Connection::initialize, not hot-reloadable.
 - name: security.storage.encryption.sqlite.enabled
   type: RelaxedAtomicBool
+#if defined(MOZ_ENTERPRISE)
+  value: true
+#else
   value: false
+#endif
   mirror: always
 
 - name: security.tls13.aes_128_gcm_sha256
--- a/modules/libpref/init/StaticPrefList.yaml
+++ b/modules/libpref/init/StaticPrefList.yaml
@@ -19143,7 +19143,7 @@
 # ciphertext. Read at Connection::initialize, not hot-reloadable.
 - name: security.storage.encryption.sqlite.enabled
   type: RelaxedAtomicBool
-#if defined(MOZ_ENTERPRISE)
+#if defined(MOZ_ENTERPRISE) && !defined(MOZ_THUNDERBIRD)
   value: true
 #else
   value: false
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -4013,12 +4013,15 @@ pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", false
 pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", false, locked);
 #endif
 
-// Locked so end users cannot toggle SQLite encryption from about:config or
-// user.js. Enterprise policies (Preferences allowlist) and our keystore
-// auto-recovery use the unlock/setDefault/relock dance to override.
-// Defined in StaticPrefList.yaml; locked here because the YAML has no
-// "locked" attribute.
-pref("security.storage.encryption.sqlite.enabled", false, locked);
+#if defined(MOZ_ENTERPRISE)
+// SQLite at-rest encryption enabled and locked on (enterprise enforcement;
+// end users cannot toggle it from about:config or user.js). Per-DB DEKs are
+// wrapped under a Password KEK keyed by the enterprise primarySecret (D304778).
+// The StaticPrefList.yaml entry carries the enterprise-build default that the
+// startup launch gate reads before the pref files load; this `locked` line is
+// what actually prevents runtime toggling.
+pref("security.storage.encryption.sqlite.enabled", true, locked);
+#endif
 
 // Preferences for the form autofill toolkit component.
 // The truthy values of "extensions.formautofill.addresses.available"
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -4014,13 +4014,15 @@ pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", false
 #endif
 
 #if defined(MOZ_ENTERPRISE)
-// SQLite at-rest encryption enabled and locked on (enterprise enforcement;
-// end users cannot toggle it from about:config or user.js). Per-DB DEKs are
-// wrapped under a Password KEK keyed by the enterprise primarySecret (D304778).
-// The StaticPrefList.yaml entry carries the enterprise-build default that the
-// startup launch gate reads before the pref files load; this `locked` line is
-// what actually prevents runtime toggling.
-pref("security.storage.encryption.sqlite.enabled", true, locked);
+// SQLite at-rest encryption enabled by default on enterprise builds. Per-DB
+// DEKs are wrapped under a Password KEK keyed by the enterprise primarySecret.
+// Intentionally not `locked`: the authoritative profile-local signal that
+// gates obfsvfs's default-VFS registration is compatibility.ini's
+// EncryptedDatabases marker (set eagerly in CheckEncryptionCompatibility,
+// read back via IsProfileEncryptedDatabases()), and a declarative `locked`
+// here would silently mask the test-manifest `prefs = [..]` overrides that
+// the encryption-aware xpcshell tests rely on.
+pref("security.storage.encryption.sqlite.enabled", true);
 #endif
 
 // Preferences for the form autofill toolkit component.
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -4013,15 +4013,17 @@ pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", false
 pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", false, locked);
 #endif
 
-#if defined(MOZ_ENTERPRISE)
-// SQLite at-rest encryption enabled by default on enterprise builds. Per-DB
-// DEKs are wrapped under a Password KEK keyed by the enterprise primarySecret.
-// Intentionally not `locked`: the authoritative profile-local signal that
-// gates obfsvfs's default-VFS registration is compatibility.ini's
+#if defined(MOZ_ENTERPRISE) && !defined(MOZ_THUNDERBIRD)
+// SQLite at-rest encryption enabled by default on enterprise Firefox builds.
+// Per-DB DEKs are wrapped under a Password KEK keyed by the enterprise
+// primarySecret. Intentionally not `locked`: the authoritative profile-local
+// signal that gates obfsvfs's default-VFS registration is compatibility.ini's
 // EncryptedDatabases marker (set eagerly in CheckEncryptionCompatibility,
 // read back via IsProfileEncryptedDatabases()), and a declarative `locked`
 // here would silently mask the test-manifest `prefs = [..]` overrides that
-// the encryption-aware xpcshell tests rely on.
+// the encryption-aware xpcshell tests rely on. Gated on !MOZ_THUNDERBIRD so
+// Thunderbird (which can in principle be built with --enable-enterprise too)
+// stays plaintext for now.
 pref("security.storage.encryption.sqlite.enabled", true);
 #endif
 
--- a/security/lockstore/gtest/TestLockstoreKeystore.cpp
+++ b/security/lockstore/gtest/TestLockstoreKeystore.cpp
@@ -725,3 +725,66 @@ TEST_F(LockstoreKeystoreTest, SwitchKekMissingOldWrapping) {
   rv = keystore_switch_kek(mKeystore, &coll, &otherLocal, &mLocalKekRef);
   ASSERT_EQ(rv, NS_ERROR_NOT_AVAILABLE);
 }
+
+// "Refresh Firefox" copies an encrypted database into a NEW profile whose own
+// per-profile LocalKey cannot unwrap the source profile's DEK. SQLiteEncryption
+// recovers it by reading the raw DEK from the source keystore and re-importing
+// it into the destination keystore under the destination's own KEK -- a re-wrap,
+// never a keystore-file copy. This verifies that round-trip across two distinct
+// keystores preserves the DEK bytes, so the copied ciphertext stays readable.
+TEST_F(LockstoreKeystoreTest, CrossKeystoreDekRewrap) {
+  // Source keystore (this test's profile dir) + its own LocalKey + a DEK.
+  nsresult rv = keystore_open(&mProfilePath, &mKeystore);
+  ASSERT_NS_SUCCEEDED(rv);
+  MintLocalKek();
+  const nsCString coll("places.sqlite"_ns);
+  rv = keystore_create_dek(mKeystore, &coll, &mLocalKekRef,
+                           /* extractable */ true, 32);
+  ASSERT_NS_SUCCEEDED(rv);
+  nsTArray<uint8_t> srcDek;
+  rv = keystore_get_dek(mKeystore, &coll, &mLocalKekRef, &srcDek);
+  ASSERT_NS_SUCCEEDED(rv);
+  ASSERT_EQ(srcDek.Length(), 32u);
+
+  // A separate destination profile keystore with its OWN (distinct) LocalKey.
+  nsCOMPtr<nsIFile> destDir;
+  rv = NS_GetSpecialDirectory(NS_OS_TEMP_DIR, getter_AddRefs(destDir));
+  ASSERT_NS_SUCCEEDED(rv);
+  rv = destDir->AppendNative("lockstore_ks_dest_test"_ns);
+  ASSERT_NS_SUCCEEDED(rv);
+  rv = destDir->CreateUnique(nsIFile::DIRECTORY_TYPE, 0700);
+  ASSERT_NS_SUCCEEDED(rv);
+  nsAutoString destPathWide;
+  rv = destDir->GetPath(destPathWide);
+  ASSERT_NS_SUCCEEDED(rv);
+  NS_ConvertUTF16toUTF8 destPath(destPathWide);
+
+  KeystoreHandle* destKeystore = nullptr;
+  rv = keystore_open(&destPath, &destKeystore);
+  ASSERT_NS_SUCCEEDED(rv);
+
+  const nsCString kekType("local"_ns);
+  const nsCString empty;
+  nsCString destKekRef;
+  rv = keystore_create_kek(destKeystore, &kekType, &empty, &empty,
+                           /* cache_timeout_ms */ 0, &destKekRef);
+  ASSERT_NS_SUCCEEDED(rv);
+  // Distinct keystores have independent random LocalKeys.
+  EXPECT_FALSE(destKekRef.Equals(mLocalKekRef));
+
+  // Re-wrap: import the source DEK bytes under the destination's own KEK.
+  rv = keystore_import_dek(destKeystore, &coll, &destKekRef, srcDek.Elements(),
+                           srcDek.Length(), /* extractable */ true);
+  ASSERT_NS_SUCCEEDED(rv);
+
+  // The destination unwraps to the SAME DEK bytes, so a database encrypted with
+  // that DEK in the source profile remains readable after the copy.
+  nsTArray<uint8_t> destDek;
+  rv = keystore_get_dek(destKeystore, &coll, &destKekRef, &destDek);
+  ASSERT_NS_SUCCEEDED(rv);
+  ASSERT_EQ(destDek.Length(), srcDek.Length());
+  EXPECT_EQ(memcmp(destDek.Elements(), srcDek.Elements(), srcDek.Length()), 0);
+
+  EXPECT_NS_SUCCEEDED(keystore_close(destKeystore));
+  destDir->Remove(true);
+}
--- a/security/lockstore/gtest/TestLockstoreKeystore.cpp
+++ b/security/lockstore/gtest/TestLockstoreKeystore.cpp
@@ -729,9 +729,10 @@ TEST_F(LockstoreKeystoreTest, SwitchKekMissingOldWrapping) {
 // "Refresh Firefox" copies an encrypted database into a NEW profile whose own
 // per-profile LocalKey cannot unwrap the source profile's DEK. SQLiteEncryption
 // recovers it by reading the raw DEK from the source keystore and re-importing
-// it into the destination keystore under the destination's own KEK -- a re-wrap,
-// never a keystore-file copy. This verifies that round-trip across two distinct
-// keystores preserves the DEK bytes, so the copied ciphertext stays readable.
+// it into the destination keystore under the destination's own KEK -- a
+// re-wrap, never a keystore-file copy. This verifies that round-trip across two
+// distinct keystores preserves the DEK bytes, so the copied ciphertext stays
+// readable.
 TEST_F(LockstoreKeystoreTest, CrossKeystoreDekRewrap) {
   // Source keystore (this test's profile dir) + its own LocalKey + a DEK.
   nsresult rv = keystore_open(&mProfilePath, &mKeystore);
--- a/storage/SQLiteEncryption.cpp
+++ b/storage/SQLiteEncryption.cpp
@@ -11,7 +11,9 @@
 #include "mozilla/StaticMutex.h"
 #include "mozilla/StaticPtr.h"
 #include "mozilla/StaticPrefs_security.h"
+#include "mozilla/SpinEventLoopUntil.h"
 #include "mozilla/SyncRunnable.h"
+#include "mozilla/TimeStamp.h"
 #include "mozilla/dom/quota/IPCStreamCipherStrategy.h"
 #include "mozilla/security/lockstore/lockstore_ffi_generated.h"
 #include "ScopedNSSTypes.h"
@@ -20,10 +22,12 @@
 #include "nsCOMPtr.h"
 #include "nsDirectoryServiceDefs.h"
 #include "nsDirectoryServiceUtils.h"
+#include "nsIFelt.h"
 #include "nsIFile.h"
 #include "nsIObserver.h"
 #include "nsIObserverService.h"
 #include "nsLocalFile.h"
+#include "nsServiceManagerUtils.h"
 #include "nsString.h"
 #include "nsTArray.h"
 #include "nsThreadUtils.h"
@@ -37,12 +41,33 @@ mozilla::LogModule* GetSQLiteEncryptionLog() {
 
 namespace {
 
+using mozilla::security::lockstore::keystore_add_kek;
 using mozilla::security::lockstore::keystore_close;
 using mozilla::security::lockstore::keystore_create_dek;
 using mozilla::security::lockstore::keystore_create_kek;
+using mozilla::security::lockstore::keystore_delete_kek;
 using mozilla::security::lockstore::keystore_get_dek;
+using mozilla::security::lockstore::keystore_list_deks;
+using mozilla::security::lockstore::keystore_list_keks;
 using mozilla::security::lockstore::keystore_open;
+using mozilla::security::lockstore::keystore_remove_kek;
+using mozilla::security::lockstore::keystore_switch_kek;
+using mozilla::security::lockstore::keystore_unlock_kek;
 using mozilla::security::lockstore::KeystoreHandle;
+
+// Deterministic kek_ref literals (the random-suffix in
+// `lockstore::kek::<type>:<id>`) for the two KEKs the storage layer
+// owns. Defined as string literals; consumers wrap them in a local
+// `nsCString` before passing &-of to the FFI (which expects
+// `const nsACString*`, not the more-derived `nsLiteralCString*`).
+#define KEK_REF_LOCAL_SQLITE "lockstore::kek::local:sqlite"
+#define KEK_REF_PASSWORD_SQLITE "lockstore::kek::password:sqlite"
+
+// Practical max for cache_timeout_ms; "session-unlimited". Lockstore's
+// 0 means "don't cache" (zero-and-discard), not "infinity".
+// u32::MAX ms == ~49.7 days; if a session ever runs that long, the
+// re-unlock path in GetEncryptionKey transparently re-derives.
+constexpr uint32_t kKekCacheTimeoutMs = UINT32_MAX;
 
 constexpr size_t kDekBytes = 32;
 
@@ -64,9 +89,22 @@ static_assert(kDekBytes ==
 mozilla::StaticMutex sStateMutex;
 KeystoreHandle* sHandle MOZ_GUARDED_BY(sStateMutex) = nullptr;
 nsString sCachedProfilePath MOZ_GUARDED_BY(sStateMutex);
-// Deterministic kek_ref (lockstore::kek::local:sqlite) under which every
-// SQLite DEK is wrapped; resolved lazily via create_kek's get-or-create.
+// Deterministic kek_ref under which every SQLite DEK is wrapped.
+// `lockstore::kek::password:sqlite` (a Password KEK with
+// primarySecret as the password); the legacy
+// `lockstore::kek::local:sqlite` value is migrated away at first
+// launch. Resolved lazily via the unlock-or-create flow in
+// GetEncryptionKey.
 nsCString sKekRef MOZ_GUARDED_BY(sStateMutex);
+// Console-supplied primarySecret (64-char hex), cached for the
+// lifetime of the session. Delivered by the Felt IPC bridge during early
+// profile startup; EnsurePrimarySecretCached waits for it (it can run from
+// either profile-do-change or profile-after-change, whichever opens the first
+// encrypted database). Consumed by the Password KEK unlock/create call and by
+// the transparent re-unlock path after cache expiry. Held in the same
+// StaticMutex lifecycle as sHandle so it survives across background-thread DB
+// opens. Cleared on xpcom-will-shutdown.
+nsCString sPrimarySecret MOZ_GUARDED_BY(sStateMutex);
 // Set once xpcom-will-shutdown has torn the keystore down, so later calls
 // don't re-open it or mint key material that would never be destroyed.
 bool sShuttingDown MOZ_GUARDED_BY(sStateMutex) = false;
@@ -87,6 +125,133 @@ class ProfileObserver final : public nsIObserver {
 mozilla::StaticRefPtr<ProfileObserver> sObserver;
 
 NS_IMPL_ISUPPORTS(ProfileObserver, nsIObserver)
+
+// The Password KEK is keyed by the Felt-delivered primarySecret, which only
+// exists in enterprise (Felt) builds, so compile this waiter out elsewhere -- it
+// would otherwise be an unused function (its sole caller is already gated on
+// MOZ_ENTERPRISE).
+#if defined(MOZ_ENTERPRISE)
+// Wait for the console-supplied primarySecret to be delivered by the Felt IPC
+// bridge. The Felt parent process fetches primarySecret before spawning the
+// browsing Firefox and sends it as its first IPC message; the spawned child's
+// FeltClientThread pushes it straight into sPrimarySecret via the
+// mozStorageSetSqlitePrimarySecret entry point (below) -- it is never stored on
+// the Felt side. This call blocks until that delivery happens.
+//
+// The browser cannot decrypt its profile without primarySecret, so wait
+// generously and then fail closed (no plaintext fallback). The Felt parent
+// already aborts the launch when its own getPrimarySecret() fetch fails, so a
+// non-arrival here is not expected in practice.
+//
+// Returns NS_OK once delivered, NS_ERROR_NOT_AVAILABLE on timeout or shutdown.
+nsresult EnsurePrimarySecretCached() {
+  MOZ_ASSERT(NS_IsMainThread());
+  constexpr uint32_t kTimeoutMs = 30000;
+  const TimeStamp deadline =
+      TimeStamp::Now() + TimeDuration::FromMilliseconds(kTimeoutMs);
+  // Spin the event loop rather than blocking the main thread with a sleep, the
+  // same way the Felt startup barrier waits in nsAppRunner: keep IPC and other
+  // main-thread work running while we wait for Felt to deliver the secret. Stop
+  // on delivery, shutdown, or the timeout.
+  SpinEventLoopUntil("storage::EnsurePrimarySecretCached"_ns, [&]() -> bool {
+    {
+      StaticMutexAutoLock lock(sStateMutex);
+      if (!sPrimarySecret.IsEmpty() || sShuttingDown) {
+        return true;
+      }
+    }
+    return TimeStamp::Now() >= deadline;
+  });
+  StaticMutexAutoLock lock(sStateMutex);
+  if (!sPrimarySecret.IsEmpty()) {
+    return NS_OK;
+  }
+  // Timed out or shutting down: fail closed, no plaintext fallback.
+  MOZ_LOG(GetSQLiteEncryptionLog(), LogLevel::Error,
+          ("primarySecret not delivered by Felt (timeout or shutdown); failing "
+           "closed"));
+  return NS_ERROR_NOT_AVAILABLE;
+}
+#endif  // MOZ_ENTERPRISE
+
+// One-shot migration. Called at most once per profile, from the
+// create branch of GetEncryptionKey. Rotates every DEK currently
+// wrapped under `lockstore::kek::local:sqlite` to instead be wrapped
+// under `lockstore::kek::password:sqlite`, then deletes the now-orphan
+// LocalKey record. Idempotent: on a true-fresh profile there's no
+// local:sqlite to rotate and this is a no-op.
+//
+// Caller must hold sStateMutex.
+nsresult MigrateLocalToPasswordKek() {
+  sStateMutex.AssertCurrentThreadOwns();
+  MOZ_LOG(GetSQLiteEncryptionLog(), LogLevel::Info,
+          ("Migrating local:sqlite -> password:sqlite"));
+  const nsCString localKek(KEK_REF_LOCAL_SQLITE ""_ns);
+  const nsCString passwordKek(KEK_REF_PASSWORD_SQLITE ""_ns);
+  nsTArray<nsCString> collections;
+  nsresult rv = keystore_list_deks(sHandle, &collections);
+  if (NS_FAILED(rv)) {
+    return rv;
+  }
+  uint32_t rotated = 0;
+  for (const auto& coll : collections) {
+    nsTArray<nsCString> keks;
+    if (NS_FAILED(keystore_list_keks(sHandle, &coll, &keks))) {
+      continue;
+    }
+    bool hasLocal = false;
+    for (const auto& k : keks) {
+      if (k.Equals(localKek)) {
+        hasLocal = true;
+        break;
+      }
+    }
+    if (!hasLocal) {
+      continue;
+    }
+    // Add the password wrapping, switch primary to it, drop the
+    // local wrapping. Each step independently fallible; log and
+    // move on so a single bad collection doesn't strand the rest.
+    nsresult ar = keystore_add_kek(sHandle, &coll, &localKek, &passwordKek);
+    if (NS_FAILED(ar)) {
+      MOZ_LOG(GetSQLiteEncryptionLog(), LogLevel::Warning,
+              ("migrate: add_kek failed for %s: 0x%" PRIx32, coll.get(),
+               static_cast<uint32_t>(ar)));
+      continue;
+    }
+    nsresult sr =
+        keystore_switch_kek(sHandle, &coll, &localKek, &passwordKek);
+    if (NS_FAILED(sr)) {
+      MOZ_LOG(GetSQLiteEncryptionLog(), LogLevel::Warning,
+              ("migrate: switch_kek failed for %s: 0x%" PRIx32, coll.get(),
+               static_cast<uint32_t>(sr)));
+      continue;
+    }
+    nsresult rr = keystore_remove_kek(sHandle, &coll, &localKek);
+    if (NS_FAILED(rr)) {
+      MOZ_LOG(GetSQLiteEncryptionLog(), LogLevel::Warning,
+              ("migrate: remove_kek failed for %s: 0x%" PRIx32, coll.get(),
+               static_cast<uint32_t>(rr)));
+      continue;
+    }
+    rotated++;
+  }
+  // After all collections are rotated, drop the LocalKey record
+  // entirely. Lockstore's in-use guard rejects this if any
+  // collection slipped through above; in that case the LocalKey
+  // record survives and we'll retry on the next launch.
+  nsresult dr = keystore_delete_kek(sHandle, &localKek);
+  if (NS_FAILED(dr)) {
+    MOZ_LOG(GetSQLiteEncryptionLog(), LogLevel::Warning,
+            ("migrate: delete_kek(local:sqlite) failed: 0x%" PRIx32
+             " (will retry on next launch)",
+             static_cast<uint32_t>(dr)));
+  } else {
+    MOZ_LOG(GetSQLiteEncryptionLog(), LogLevel::Info,
+            ("Migrated %u collections; local:sqlite deleted", rotated));
+  }
+  return NS_OK;
+}
 
 // Resolve the profile directory and cache it. MAIN-THREAD ONLY:
 // NS_GetSpecialDirectory -> nsDirectoryService::Get asserts NS_IsMainThread()
@@ -225,6 +390,23 @@ NS_IMETHODIMP ProfileObserver::Observe(nsISupports*, const char* aTopic,
     // blocked main thread.
     EnsureProfilePathCached();
     EnsureNSSInitializedForEncryptionIfReady();
+    // Also pull the Felt-supplied primarySecret into our cache here,
+    // but only in the spawned-Browser-Firefox case. The Felt UI
+    // process itself never receives primarySecret over IPC (it's the
+    // entity that fetches it), so polling there would burn the full
+    // 5s timeout for nothing -- and GetEncryptionKey routes the UI
+    // process to the LocalKey path regardless.
+    if (StaticPrefs::security_storage_encryption_sqlite_enabled()) {
+      nsCOMPtr<nsIFelt> felt =
+          do_GetService("@mozilla.org/toolkit/library/felt;1");
+      bool isFeltSpawnedBrowser = false;
+      if (felt) {
+        (void)felt->IsFeltBrowser(&isFeltSpawnedBrowser);
+      }
+      if (isFeltSpawnedBrowser) {
+        (void)EnsurePrimarySecretCached();
+      }
+    }
   } else if (!strcmp(aTopic, "profile-after-change")) {
     EnsureProfilePathCached();
     MarkProfileEncryptedIfNeeded();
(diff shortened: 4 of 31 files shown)


browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +476,34 @@ export class FeltProcessParent extends JSProcessActorParent {
       gObserversRegistered = true;
     }
 
+    // Pre-fetch primarySecret from the console BEFORE spawning Firefox
+    // so the child's storage encryption layer (mozStorage / obfsvfs)
+    // can unlock its `lockstore::kek::password:sqlite` Password KEK at
+    // profile-do-change without racing the first cookies.sqlite open.
+    // Best-effort: if the call fails (no console reachable, expired
+    // token), we still spawn Firefox; the child times out waiting for
```
lissyx: I am not sure this is a good thing, we better want to directly fail, no?
Otherwise this behavior means the user will have no Felt UI and no browser, timing out in weird ways?

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +476,34 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    // PrimarySecret and the encryption gate refuses launch via the
+    // existing CheckEncryptionCompatibility path.
+    try {
+      const payload = await lazy.ConsoleClient.getPrimarySecret();
+      if (payload?.data) {
+        Services.felt.setPrimarySecret(payload.data);
```
lissyx: What is the usecase here? Does felt really needs to store it?

browser/extensions/felt/rust/nsIFelt.idl
```diff
@@ -69,6 +69,43 @@ interface nsIFelt : nsISupports {
   [must_use]
   void clearTokens();
 
+  /**
+   * Store the console-supplied primarySecret on the Felt UI process
+   * after it's fetched via ConsoleClient.getPrimarySecret(). Held in
+   * memory only; never logged.
+   */
+  [must_use]
+  void setPrimarySecret(in ACString hex);
```
lissyx: I am not sure why we want to store it ?

modules/libpref/init/all.js
```diff
@@ -4003,12 +4003,11 @@ pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", false
...
-// user.js. Enterprise policies (Preferences allowlist) and our keystore
-// auto-recovery use the unlock/setDefault/relock dance to override.
-// Defined in StaticPrefList.yaml; locked here because the YAML has no
-// "locked" attribute.
-pref("security.storage.encryption.sqlite.enabled", false, locked);
+// SQLite at-rest encryption enabled and locked on (enterprise enforcement;
+// end users cannot toggle it from about:config or user.js). Per-DB DEKs are
+// wrapped under a Password KEK keyed by the enterprise primarySecret (D304778),
+// with a LocalKey fallback when no primarySecret is available.
+pref("security.storage.encryption.sqlite.enabled", true, locked);
```
lissyx:
```suggestion
#if defined(MOZ_ENTERPRISE)
// SQLite at-rest encryption enabled and locked on (enterprise enforcement;
// end users cannot toggle it from about:config or user.js). Per-DB DEKs are
// wrapped under a Password KEK keyed by the enterprise primarySecret (D304778),
// with a LocalKey fallback when no primarySecret is available.
pref("security.storage.encryption.sqlite.enabled", true, locked);
#endif
```

modules/libpref/init/StaticPrefList.yaml
```diff
@@ -18900,7 +18900,7 @@
 # ciphertext. Read at Connection::initialize, not hot-reloadable.
 - name: security.storage.encryption.sqlite.enabled
   type: RelaxedAtomicBool
-  value: false
+  value: true
```
lissyx, line 19147:
```suggestion
#if defined(MOZ_ENTERPRISE)
  value: true
#endif
```

modules/libpref/init/all.js
lissyx: Since it's set in `StaticPrefList.yaml` why do we need it in `all.js` ?

testing/enterprise/felt_tests.py
```diff
@@ -229,6 +229,18 @@ def do_GET(self):
...
+        elif path == "/api/browser/key":
+            # The primarySecret for SQLite at-rest encryption. The Felt UI
+            # process fetches it (ConsoleClient.getPrimarySecret ->
+            # /api/browser/key) before spawning Firefox and forwards it as the
+            # Password KEK password; without it the spawned browser cannot open
+            # its encrypted profile databases.
+            m = json.dumps({
+                "data": "00112233445566778899aabbccddeeff"
+                "00112233445566778899aabbccddeeff"
+            })
```
lissyx: nit: do we want something more generated than hard-coded? May help ensure we avoid issues with stale caches anywhere.

browser/extensions/felt/content/FeltProcessParent.sys.mjs
beurdouche: There is something to do here when the profile is already encrypted and has the ability to unlock using another type of key (eg. a PKCS11 or FIDO2 token) but you are right that we should start by the simple thing and not start if we can't get the secret for now. I will possibly do a separate PR for when the console is not reachable (eg. we still want to be able to open Thunderbird offline)

browser/extensions/felt/content/FeltProcessParent.sys.mjs
beurdouche: We don't need to anymore in the current patch

browser/extensions/felt/rust/nsIFelt.idl
beurdouche: I agree, I did it differently in the past, but we don't technically have to anymore, so I removed that part.

modules/libpref/init/all.js
beurdouche: The locking mechanism for prefs happens via `all.js`

modules/libpref/init/StaticPrefList.yaml
beurdouche: Done

testing/enterprise/felt_tests.py
beurdouche: Done

lissyx:
There's a marionette test failure related to key fetch do we care about that? https://treeherder.mozilla.org/logviewer?job_id=574144215&repo=enterprise-firefox-pr&task=UdV-CzniQJWozSzLVZC-FA.0&lineNumber=1289

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +477,54 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    // retry. The child's storage encryption layer (mozStorage / obfsvfs)
+    // blocks at profile-do-change waiting for it to unlock the
+    // `lockstore::kek::password:sqlite` Password KEK, so the browser cannot
+    // function without it. If the fetch ultimately fails, abort the launch
+    // rather than spawn a browser that hangs waiting for a secret that never
+    // arrives. The value is held only in this local until it is relayed to the
+    // spawned browser below; Felt never stores it.
+    let primarySecret;
+    const kMaxAttempts = 3;
+    for (let attempt = 1; attempt <= kMaxAttempts; attempt++) {
```
lissyx: I'm not sure it's really required to try three times like that.

And I guess if we want to do so, it should be part of `getPrimarySecret()` and not handled here.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +477,54 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      }
+    }
+    if (!primarySecret) {
+      lazy.log.error(
+        "startFirefox: primarySecret unavailable after retries; aborting browser launch"
+      );
+      throw new Error(
+        "Felt: refusing to launch Firefox without the storage-encryption primarySecret"
+      );
+    }
```
lissyx, line 495: We should:
 - make sure this properly pops a dedicated error to the user
 - add test on https://searchfox.org/enterprise-main/rev/b451268c510f1f01858e04a3afb42c712f815f30/testing/enterprise/test_felt_console_error.py to cover for that error

lissyx: In term of UX (error message, user feedback on failure path), depending on your constraints, either we validate with matthew or we land something decent right now with a follow up blocking `firefox-client-mvp` so UX can be addressed properly

browser/extensions/felt/rust/src/components.rs
```diff
@@ -240,6 +240,25 @@ impl FeltXPCOM {
         self.set_tokens_impl("".to_string(), "".to_string(), 0)
     }
 
+    // Send the console-supplied primarySecret to the spawned Firefox over the
+    // existing Felt IPC channel. Called from the Felt UI process right after
+    // spawning Firefox; the secret is never stored on either side -- the UI
+    // fetches it and relays it here, and the browser hands it straight to the
+    // storage encryption layer on receipt (see client.rs). No-op if invoked
+    // from the browser side. Mirrors SendAccessToken.
+    fn SendPrimarySecret(&self, hex: *const nsACString) -> nserror::nsresult {
```
lissyx: Use `xpcom_method!` macro like elsewhere

browser/extensions/felt/rust/src/components.rs
```diff
@@ -240,6 +240,25 @@ impl FeltXPCOM {
...
+        if self.is_felt_browser {
+            trace!("FeltXPCOM::SendPrimarySecret: called from the browser, no-op");
+            return NS_OK;
+        }
+        let hex_s = unsafe { (*hex).to_string() };
```
lissyx: This `unsafe` can be removed once you use `xpcom_method!`

toolkit/components/felt/rust/src/message.rs
```diff
@@ -55,6 +55,14 @@ pub enum FeltMessage {
...
     StartupReady,
     AccessToken((String, i64)),
+    /// The console-supplied primarySecret (64-char hex string), shipped
+    /// from the Felt UI process to the spawned browsing Firefox.
+    /// Storage encryption uses it as the password to unlock the
+    /// `lockstore::kek::password:sqlite` Password KEK that wraps every
+    /// per-DB DEK. Must arrive BEFORE the storage layer initializes
+    /// (i.e. before `profile-do-change`); sent unconditionally on the
+    /// first IPC handshake, ahead of `AccessToken` / `Ready`.
+    PrimarySecret(String),
```
lissyx, line 65: Update version number

browser/extensions/felt/rust/src/message.rs
lissyx:
```suggestion
pub const FELT_IPC_VERSION: u32 = 11;
```

browser/extensions/felt/rust/src/components.rs
lissyx:
```suggestion

    xpcom_method!(send_primary_secret => SendPrimarySecret(
        hex: *const nsACString,
    ));
    fn send_primary_secret(
        &self,
        hex: &nsACString,
    ) -> Result<(), nserror::nsresult> {
```

browser/extensions/felt/rust/src/components.rs
lissyx:
```suggestion
        let hex = hex.to_string();
```

browser/extensions/felt/rust/src/components.rs
```diff
@@ -240,6 +240,25 @@ impl FeltXPCOM {
...
+    // Send the console-supplied primarySecret to the spawned Firefox over the
+    // existing Felt IPC channel. Called from the Felt UI process right after
+    // spawning Firefox; the secret is never stored on either side -- the UI
+    // fetches it and relays it here, and the browser hands it straight to the
+    // storage encryption layer on receipt (see client.rs). No-op if invoked
+    // from the browser side. Mirrors SendAccessToken.
+    fn SendPrimarySecret(&self, hex: *const nsACString) -> nserror::nsresult {
+        if self.is_felt_browser {
+            trace!("FeltXPCOM::SendPrimarySecret: called from the browser, no-op");
+            return NS_OK;
```
lissyx:
```suggestion
            return Ok(());
```

lissyx: Also wondering if it should not be a hard panic/assert, this should likely never happen

browser/extensions/felt/rust/src/components.rs
```diff
@@ -240,6 +240,25 @@ impl FeltXPCOM {
...
+        }
+        let hex_s = unsafe { (*hex).to_string() };
+        if hex_s.is_empty() {
+            trace!("FeltXPCOM::SendPrimarySecret: empty primarySecret, not sending");
+            return NS_ERROR_FAILURE;
```
lissyx:
```suggestion
            return Err(NS_ERROR_FAILURE);
```

browser/extensions/felt/rust/src/client.rs
```diff
@@ -391,6 +397,15 @@ impl FeltClientThread {
...
                                     }
                                 }
+                                Ok(FeltMessage::PrimarySecret(hex)) => {
+                                    // Hand the console-supplied primarySecret straight to
+                                    // storage/SQLiteEncryption.cpp; Felt keeps no copy.
+                                    // Do NOT trace the value.
+                                    let c: nsCString = hex.as_str().into();
+                                    unsafe { mozStorageSetSqlitePrimarySecret(&*c) };
+                                    trace!("FeltClientThread::felt_client::ipc_loop(): PrimarySecret delivered to storage");
+                                    utils::notify_observers("felt-firefox-primary-secret-received".to_string());
```
lissyx: This observer is not used.

browser/extensions/felt/rust/src/client.rs
```diff
@@ -15,6 +15,12 @@ use log::trace;
 use crate::message::{nsICookieWrapper, FeltMessage, FELT_IPC_VERSION};
 use crate::utils::{self, Tokens, TOKENS};
 
+extern "C" {
+    // Defined in storage/SQLiteEncryption.cpp. Hands the console-supplied
+    // primarySecret to the storage encryption layer; Felt keeps no copy.
+    fn mozStorageSetSqlitePrimarySecret(hex: *const nsACString);
```
lissyx: nit: move to `utils.rs` and wrap as e.g. `moz_storage_set_sqlite_primary_secret(hex: String)` so the wrapper handles the `unsafe {}` block at the same time

testing/enterprise/base_test.py
```diff
@@ -196,7 +197,11 @@ def _child_longwait(self):
         return self._longwaiter(self._child_driver)
 
     def _waiter(self, driver):
-        return Wait(driver, 10)
+        # Debug builds start the Felt browser appreciably slower (notably the
+        # at-rest-encryption keystore is brought up during startup), so the
+        # short waits need more headroom there to avoid spurious timeouts.
+        timeout = 15 if mozinfo.info.get("debug") else 10
+        return Wait(driver, timeout)
```
lissyx: Do we have any waiter call depending on process startup ?
If so, maybe we want to have a dedicated waiter, rather than augment the timeout for all, I'm worried of side effects here

testing/enterprise/felt_tests.py
```diff
@@ -229,6 +237,15 @@ def do_GET(self):
                 "extra_prefs": [["marionette.port", 0]],
             })
 
+        elif path == "/api/browser/key":
```
lissyx, line 264:
```suggestion
        elif path == "/api/browser/key":
            if not self.check_auth():
                return
```

modules/libpref/init/StaticPrefList.yaml
```diff
@@ -18900,7 +18900,11 @@
 # ciphertext. Read at Connection::initialize, not hot-reloadable.
 - name: security.storage.encryption.sqlite.enabled
   type: RelaxedAtomicBool
+#if defined(MOZ_ENTERPRISE)
```
lissyx: Do we want to support that on Thunderbird as well?

modules/libpref/init/all.js
```diff
@@ -4003,12 +4003,15 @@ pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", false
...
-// "locked" attribute.
-pref("security.storage.encryption.sqlite.enabled", false, locked);
+#if defined(MOZ_ENTERPRISE)
+// SQLite at-rest encryption enabled and locked on (enterprise enforcement;
+// end users cannot toggle it from about:config or user.js). Per-DB DEKs are
+// wrapped under a Password KEK keyed by the enterprise primarySecret (D304778).
+// The StaticPrefList.yaml entry carries the enterprise-build default that the
+// startup launch gate reads before the pref files load; this `locked` line is
+// what actually prevents runtime toggling.
+pref("security.storage.encryption.sqlite.enabled", true, locked);
```
lissyx: This will enable for any app, including Thunderbird, is it OK ? I am not sure, we probably want to scope to browser builds only?

toolkit/components/felt/rust/src/message.rs
beurdouche: Done

modules/libpref/init/all.js
beurdouche: I think this is ok for Thunderbird Enterprise but worth thinking about, what do you suggest to scope to Firefox Enterprise only ?

modules/libpref/init/StaticPrefList.yaml
beurdouche: I would say yes, but let's discuss

browser/extensions/felt/rust/src/components.rs
beurdouche: Done

browser/extensions/felt/rust/src/client.rs
beurdouche: Done — moved the mozStorageSetSqlitePrimarySecret extern into utils.rs and wrapped it as moz_storage_set_sqlite_primary_secret(hex: String)

browser/extensions/felt/rust/src/client.rs
beurdouche: Done — removed the notify_observers("felt-firefox-primary-secret-received") call

browser/extensions/felt/rust/src/components.rs
beurdouche: Done

browser/extensions/felt/rust/src/components.rs
beurdouche: Done

browser/extensions/felt/rust/src/components.rs
beurdouche: Done

browser/extensions/felt/rust/src/message.rs
beurdouche: Done

browser/extensions/felt/rust/src/components.rs
beurdouche: Done

browser/extensions/felt/rust/src/components.rs
beurdouche: Done

modules/libpref/init/StaticPrefList.yaml
lissyx: I dont think Thunderbird is yet in a state where we can safely enable this

modules/libpref/init/all.js
lissyx: `MOZ_BUILD_APP_IS_BROWSER` maybe? https://searchfox.org/enterprise-main/search?q=MOZ_BUILD_APP_IS_BROWSER&path=&case=false&regexp=false

browser/extensions/felt/content/FeltProcessParent.sys.mjs
beurdouche: Agreed — dropped the retry loop. ConsoleClient.getPrimarySecret() → _fetch already refreshes the session and retries once

testing/enterprise/test_felt_init_failures.py
```diff
@@ -23,3 +23,23 @@ def test_browser_init_policy_fetch_fail(self):
...
+        self._manually_closed_child = True
+        self.key_fail_request.value = 0
+
+        self.await_felt_auth_window()
+        self.force_window()
+        self._driver.set_context("chrome")
+        error_msg = self.get_elem(".felt-browser-error-launch-failure")
+        assert error_msg.is_displayed(), (
+            "launch-failure error shown when primarySecret is unavailable"
+        )
```
lissyx, line 56: Can you run locally with `UX_SCREENSHOT=1` and share it on the bug?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
beurdouche: It now routes through the existing FeltParent:FirefoxLaunchFailure → showWindow("felt-browser-error-launch-failure"), so the user gets the already-localized launch-failure window.

testing/enterprise/base_test.py
beurdouche: Reverted _waiter to the 10s default and moved the debug headroom into _child_wait only (the spawned browser carries the at-rest-encryption keystore startup cost; the Felt UI waiter is unaffected)

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +476,45 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    if (!primarySecret) {
+      // The spawned browser cannot open its encrypted profile databases
+      // without the primarySecret, so do not launch it. Surface the existing
+      // launch-failure error to the user rather than leaving Felt backgrounded
+      // with no browser. A dedicated, primarySecret-specific error message is
+      // tracked as a follow-up.
+      lazy.log.error(
+        "startFirefox: primarySecret unavailable; aborting browser launch"
+      );
+      Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure");
```
lissyx: I'd still prefer if we could augment this message so we can show more customs error messages. It is only handling launch failures that are related to fork/exec right now mostly, this is different enough that I think we want to surface to end user (wording to be validated by Matthew, but it's easier to improve once we have it done here) to ease diagnosing for support.

lissyx: So maybe we could pass data in the message that would carry the type of error, and update the call site at https://searchfox.org/enterprise-main/rev/b451268c510f1f01858e04a3afb42c712f815f30/browser/extensions/felt/content/FeltProcessParent.sys.mjs#527 as well

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +476,45 @@ export class FeltProcessParent extends JSProcessActorParent {
...
     this.firefox = this.startFirefoxProcess();
     this.firefox
       .then(async () => {
+        // Send primarySecret FIRST, before any other state, so the
+        // child's storage encryption KEK is unlocked before any
+        // mozStorage consumer opens a database. This bypasses the
+        // `firefoxReady=true` gate that sendAccessToken / sendReady
+        // wait for. The browser hands it straight to the storage layer
+        // on receipt; neither side stores it.
+        Services.felt.sendPrimarySecret(primarySecret);
```
lissyx: it's not done correctly on other calls, so maybe worth a followup for those, but I believe we should ensure `sendPrimarySecret` returned `NS_OK` here?

lissyx: (and fail with the error message we would have added earlier)

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +476,45 @@ export class FeltProcessParent extends JSProcessActorParent {
...
       .then(async () => {
+        // Send primarySecret FIRST, before any other state, so the
+        // child's storage encryption KEK is unlocked before any
+        // mozStorage consumer opens a database. This bypasses the
+        // `firefoxReady=true` gate that sendAccessToken / sendReady
+        // wait for. The browser hands it straight to the storage layer
+        // on receipt; neither side stores it.
+        Services.felt.sendPrimarySecret(primarySecret);
         await this.sendPrefsToFirefox();
         Services.felt.sendAccessToken();
```
lissyx, line 520: needs a followup to check return value

browser/extensions/felt/content/FeltProcessParent.sys.mjs
lissyx, line 525: needs a followup to check return value

storage/SQLiteEncryption.cpp
```diff
@@ -225,6 +566,23 @@ NS_IMETHODIMP ProfileObserver::Observe(nsISupports*, const char* aTopic,
...
     EnsureNSSInitializedForEncryptionIfReady();
+    // Also pull the Felt-supplied primarySecret into our cache here,
+    // but only in the spawned-Browser-Firefox case. The Felt UI
+    // process itself never receives primarySecret over IPC (it's the
+    // entity that fetches it), so polling there would burn the full
+    // 5s timeout for nothing -- and GetEncryptionKey routes the UI
+    // process to the LocalKey path regardless.
+    if (StaticPrefs::security_storage_encryption_sqlite_enabled()) {
+      nsCOMPtr<nsIFelt> felt =
+          do_GetService("@mozilla.org/toolkit/library/felt;1");
```
lissyx: This is only relevant when you need it, but on non enterprise build it should not run at all

storage/SQLiteEncryption.cpp
```diff
@@ -225,6 +566,23 @@ NS_IMETHODIMP ProfileObserver::Observe(nsISupports*, const char* aTopic,
     // blocked main thread.
     EnsureProfilePathCached();
     EnsureNSSInitializedForEncryptionIfReady();
+    // Also pull the Felt-supplied primarySecret into our cache here,
+    // but only in the spawned-Browser-Firefox case. The Felt UI
+    // process itself never receives primarySecret over IPC (it's the
+    // entity that fetches it), so polling there would burn the full
+    // 5s timeout for nothing -- and GetEncryptionKey routes the UI
+    // process to the LocalKey path regardless.
+    if (StaticPrefs::security_storage_encryption_sqlite_enabled()) {
```
lissyx:
```suggestion
   #if defined(MOZ_ENTERPRISE)
   #  include "mozilla/browser/extensions/felt/felt.h"
   #endif
   
   [...]
   
    if (StaticPrefs::security_storage_encryption_sqlite_enabled() && is_felt_browser()) {
```

storage/SQLiteEncryption.cpp
```diff
@@ -225,6 +566,23 @@ NS_IMETHODIMP ProfileObserver::Observe(nsISupports*, const char* aTopic,
...
+      nsCOMPtr<nsIFelt> felt =
+          do_GetService("@mozilla.org/toolkit/library/felt;1");
+      bool isFeltSpawnedBrowser = false;
+      if (felt) {
+        (void)felt->IsFeltBrowser(&isFeltSpawnedBrowser);
+      }
```
lissyx: Would not be required anymore

storage/SQLiteEncryption.cpp
```diff
@@ -213,6 +371,189 @@ void EnsureNSSInitializedForEncryptionIfReady() {
...
+    bool isFeltSpawnedBrowser = false;
+    {
+      nsCOMPtr<nsIFelt> felt =
+          do_GetService("@mozilla.org/toolkit/library/felt;1");
+      if (felt) {
+        (void)felt->IsFeltBrowser(&isFeltSpawnedBrowser);
+      }
+    }
+
+    if (isFeltSpawnedBrowser) {
```
lissyx:
```suggestion
    if (is_felt_browser()) {
```

storage/SQLiteEncryption.cpp
```diff
@@ -509,6 +862,26 @@ nsresult GetEncryptionKey(const nsACString& aDatabasePath, OpenIntent aIntent,
...
+// PrimarySecret message arrives. Stores it for the Password KEK and wakes the
+// EnsurePrimarySecretCached waiter. C linkage so the Felt Rust crate can call
+// it directly; thread-safe (any thread) via sStateMutex. The secret is held
+// only here, never on the Felt side.
+extern "C" void mozStorageSetSqlitePrimarySecret(const nsACString* aHex) {
+  if (!aHex || aHex->IsEmpty()) {
+    return;
+  }
+  StaticMutexAutoLock lock(sStateMutex);
+  if (sShuttingDown) {
```
lissyx, line 999: it looks weird to take the lock before checking for shutdown

lissyx: I think I'd move `sShuttingDown` to the top directly even before checking `aHex`

testing/enterprise/base_test.py
```diff
@@ -189,7 +190,13 @@ def _longwait(self):
 
     @property
     def _child_wait(self):
-        return self._waiter(self._child_driver)
+        # The spawned browser's startup carries the at-rest-encryption keystore
+        # init (the Password KEK is unlocked at profile-do-change), which is
+        # appreciably slower on debug builds. Give only this child-process
+        # waiter extra headroom there; the Felt UI waiter (_wait) is unaffected,
```
lissyx: but it's still going to impact **all** waiting of elements and others on child side. Please make a dedicated waiter

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +476,45 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      primarySecret = payload?.data;
+    } catch (e) {
+      lazy.log.error(`startFirefox: getPrimarySecret() failed: ${e}`);
+    }
+    if (!primarySecret) {
+      // The spawned browser cannot open its encrypted profile databases
+      // without the primarySecret, so do not launch it. Surface the existing
+      // launch-failure error to the user rather than leaving Felt backgrounded
+      // with no browser. A dedicated, primarySecret-specific error message is
+      // tracked as a follow-up.
```
lissyx: Please add bug reference

browser/extensions/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -476,9 +476,45 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+    try {
+      const payload = await lazy.ConsoleClient.getPrimarySecret();
+      primarySecret = payload?.data;
+    } catch (e) {
+      lazy.log.error(`startFirefox: getPrimarySecret() failed: ${e}`);
+    }
+    if (!primarySecret) {
+      // The spawned browser cannot open its encrypted profile databases
+      // without the primarySecret, so do not launch it. Surface the existing
+      // launch-failure error to the user rather than leaving Felt backgrounded
```
lissyx: How is it going to behave if the received key is wrong?

toolkit/components/felt/rust/src/components.rs
```diff
@@ -240,6 +240,28 @@ impl FeltXPCOM {
...
+    // fetches it and relays it here, and the browser hands it straight to the
+    // storage encryption layer on receipt (see client.rs). No-op if invoked
+    // from the browser side. Mirrors SendAccessToken.
+    xpcom_method!(send_primary_secret => SendPrimarySecret(
+        hex: *const nsACString
+    ));
+    fn send_primary_secret(&self, hex: &nsACString) -> Result<(), nserror::nsresult> {
+        if self.is_felt_browser {
+            trace!("FeltXPCOM::SendPrimarySecret: called from the browser, no-op");
+            return Ok(());
```
lissyx: I'm not sure we should not return an error here, it's not supposed to happen we should catch it

lissyx:
@beurdouche  How much is migration expected to work?
image

security/lockstore/gtest/TestLockstoreKeystore.cpp
```diff
@@ -725,3 +725,66 @@ TEST_F(LockstoreKeystoreTest, SwitchKekMissingOldWrapping) {
...
   ASSERT_EQ(rv, NS_ERROR_NOT_AVAILABLE);
 }
+
+// "Refresh Firefox" copies an encrypted database into a NEW profile whose own
+// per-profile LocalKey cannot unwrap the source profile's DEK. SQLiteEncryption
+// recovers it by reading the raw DEK from the source keystore and re-importing
+// it into the destination keystore under the destination's own KEK -- a re-wrap,
+// never a keystore-file copy. This verifies that round-trip across two distinct
+// keystores preserves the DEK bytes, so the copied ciphertext stays readable.
+TEST_F(LockstoreKeystoreTest, CrossKeystoreDekRewrap) {
```
lissyx, line 736: We do not run `gtest` against `enterprise` builds

lissyx: Bug-2050540

storage/SQLiteEncryption.cpp
```diff
@@ -64,9 +95,28 @@ static_assert(kDekBytes ==
...
+// Deterministic kek_ref under which every SQLite DEK is wrapped.
+// `lockstore::kek::password:sqlite` (a Password KEK with
+// primarySecret as the password); the legacy
+// `lockstore::kek::local:sqlite` value is migrated away at first
+// launch. Resolved lazily via the unlock-or-create flow in
+// GetEncryptionKey.
 nsCString sKekRef MOZ_GUARDED_BY(sStateMutex);
+// Console-supplied primarySecret (64-char hex), cached for the
+// lifetime of the session. Set once via EnsurePrimarySecretCached at
+// profile-do-change; consumed by the Password KEK unlock/create call
```
lissyx: From pernosco, I have seen it being triggered at `profile-after-change` also.

storage/SQLiteEncryption.cpp
```diff
@@ -88,6 +138,145 @@ mozilla::StaticRefPtr<ProfileObserver> sObserver;
...
+nsresult EnsurePrimarySecretCached() {
+  MOZ_ASSERT(NS_IsMainThread());
+  constexpr uint32_t kTimeoutMs = 30000;
+  constexpr uint32_t kStepMs = 50;
+  uint32_t waited = 0;
+  for (;;) {
+    {
+      StaticMutexAutoLock lock(sStateMutex);
+      if (!sPrimarySecret.IsEmpty()) {
+        return NS_OK;
```
lissyx: Above comment on `sPrimarySecret` mentions that `EnsurePrimarySecretCached` should only be called by `profile-do-change` but running under pernosco, I have seen it being also triggered by `profile-after-change`.

Either we should update the comment, or we should assert here instead of `return NS_OK;` ?

storage/SQLiteEncryption.cpp
```diff
@@ -88,6 +138,145 @@ mozilla::StaticRefPtr<ProfileObserver> sObserver;
...
+      if (sShuttingDown) {
+        return NS_ERROR_NOT_AVAILABLE;
+      }
+    }
+    if (waited >= kTimeoutMs) {
+      break;
+    }
+    PR_Sleep(PR_MillisecondsToInterval(kStepMs));
+    waited += kStepMs;
+  }
```
lissyx: I assume this code runs under the same constraints as felt ipc, i.e. blocking startup until we receive the information. We could use `SpinEventLoopUntil` maybe? https://searchfox.org/enterprise-main/rev/47dc8d1ebc6f5d741820c82c339758847b713a7a/toolkit/xre/nsAppRunner.cpp#6163-6164

storage/SQLiteEncryption.cpp
```diff
@@ -64,9 +95,28 @@ static_assert(kDekBytes ==
...
+// launch. Resolved lazily via the unlock-or-create flow in
+// GetEncryptionKey.
 nsCString sKekRef MOZ_GUARDED_BY(sStateMutex);
+// Console-supplied primarySecret (64-char hex), cached for the
+// lifetime of the session. Set once via EnsurePrimarySecretCached at
+// profile-do-change; consumed by the Password KEK unlock/create call
+// and by the transparent re-unlock path after cache expiry. Held in
+// the same StaticMutex lifecycle as sHandle so it survives across
+// background-thread DB opens. Cleared on xpcom-will-shutdown.
+nsCString sPrimarySecret MOZ_GUARDED_BY(sStateMutex);
```
lissyx, line 122: If it's enterprise specific, `#if defined(MOZ_ENTERPRISE)`

storage/SQLiteEncryption.cpp
```diff
@@ -88,6 +138,145 @@ mozilla::StaticRefPtr<ProfileObserver> sObserver;
...
+// mozStorageSetSqlitePrimarySecret entry point (below) -- it is never stored on
+// the Felt side. This call blocks until that delivery happens.
+//
+// The browser cannot decrypt its profile without primarySecret, so wait
+// generously and then fail closed (no plaintext fallback). The Felt parent
+// already aborts the launch when its own getPrimarySecret() fetch fails, so a
+// non-arrival here is not expected in practice.
+//
+// Returns NS_OK once delivered, NS_ERROR_NOT_AVAILABLE on timeout or shutdown.
+nsresult EnsurePrimarySecretCached() {
```
lissyx, line 162: If it's enterprise specific, `#if defined(MOZ_ENTERPRISE)`

toolkit/locales/en-US/toolkit/enterprise/felt.ftl
```diff
@@ -48,6 +48,9 @@ felt-browser-error-multiple-crashes2 =
 felt-browser-error-launch-failure =
     .heading = { -brand-short-name } cannot start
     .message = Please contact your administrator if the problem persists.
+felt-browser-error-primary-secret =
+    .heading = { -brand-short-name } cannot start securely
+    .message = Your secure profile key could not be retrieved. Please try again, or contact your administrator if the problem persists.
```
lissyx: Are we going to support encryption on Thunderbird ?

toolkit/locales/en-US/toolkit/enterprise/felt.ftl
```diff
@@ -48,6 +48,9 @@ felt-browser-error-multiple-crashes2 =
 felt-browser-error-launch-failure =
     .heading = { -brand-short-name } cannot start
     .message = Please contact your administrator if the problem persists.
+felt-browser-error-primary-secret =
```
lissyx:
```suggestion
felt-error-primary-secret =
```

testing/enterprise/felt_tests.py
```diff
@@ -1078,7 +1106,13 @@ def run_felt_perform_sso_auth(self):
...
 
     def await_felt_auth_window(self):
-        self._wait.until(lambda mn: len(self._driver.chrome_window_handles) == 1)
+        # Use the long waiter: the Felt UI auth window appears only after the
+        # FELT process has finished starting up, and with at-rest encryption
+        # enabled that startup (keystore + NSS init) is appreciably slower on
+        # both opt and debug builds -- the short 10s wait times out. This is a
+        # dedicated headroom for the startup-dependent window wait, not a bump
+        # of the general element waits.
+        self._longwait.until(lambda mn: len(self._driver.chrome_window_handles) == 1)
```
lissyx: We should not need this, as investigated on pernosco and hopefully fixed; please remove

testing/enterprise/felt_tests.py
```diff
@@ -1101,5 +1135,7 @@ def replaced(_):
             handles = self._driver.chrome_window_handles
             return previous_handle not in handles and len(handles) == 1
 
-        self._wait.until(replaced)
+        # Long waiter: re-opening the Felt auth window goes through the slow
+        # encryption-enabled FELT startup (see await_felt_auth_window).
+        self._longwait.until(replaced)
```
lissyx: We should not need this, as investigated on pernosco and hopefully fixed; please remove

testing/enterprise/test_felt_init_failures.py
```diff
@@ -23,3 +23,44 @@ def test_browser_init_policy_fetch_fail(self):
...
+        self._driver.set_context("chrome")
+
+        def launch_failure_shown(_):
+            try:
+                handles = self._driver.chrome_window_handles
+                if len(handles) != 1:
+                    return False
+                self._driver.switch_to_window(handles[0])
+                self._driver.set_context("chrome")
+                error = self.get_elem(".felt-browser-error-primary-secret")
```
lissyx:
```suggestion
                error = self.get_elem(".felt-error-primary-secret")
```

toolkit/components/felt/content/felt.xhtml
```diff
@@ -200,6 +200,12 @@
             type="error"
             dismissable="true"
           ></moz-message-bar>
+          <moz-message-bar
+            class="felt-browser-error-primary-secret is-hidden"
+            data-l10n-id="felt-browser-error-primary-secret"
```
lissyx:
```suggestion
            class="felt-error-primary-secret is-hidden"
            data-l10n-id="felt-error-primary-secret"
```

taskcluster/kinds/test/marionette.yml
```diff
@@ -135,15 +135,16 @@ marionette-enterprise:
...
     tier: default
+    # At-rest encryption makes every test's FELT/browser startup appreciably
+    # slower, so the whole enterprise suite needs more wall-clock. (The
+    # `.*-debug.*` key does not match the enterprise platform string
+    # `*-enterprise/debug`, so every enterprise job uses `default`.)
     max-run-time:
         by-test-platform:
             .*-debug.*: 3600
-            default: 2700
+            default: 3600
```
lissyx: We should not need this, as investigated on pernosco and hopefully fixed; please remove

taskcluster/kinds/test/marionette.yml
```diff
@@ -135,15 +135,16 @@ marionette-enterprise:
...
     instance-size: default
-    chunks:
-        by-test-platform:
-            .*-debug.*: 2
-            default: 1
+    chunks: 1
```
lissyx: same.

lissyx: If anything, we should rather fix chunking, in a follow up?

toolkit/components/felt/rust/src/message.rs
lissyx: you're a magician, it disappeared

toolkit/components/felt/rust/src/message.rs
lissyx:
```suggestion
pub const FELT_IPC_VERSION: u32 = 12;
```

it got lost somehow?

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -303,7 +303,11 @@ export class Felt {
 
       case "FeltParent:FirefoxLaunchFailure": {
         Services.felt.makeBackgroundProcess(false);
-        this.showWindow("felt-browser-error-launch-failure");
+        const errorClass =
+          message.data?.errorType === "primarySecret"
+            ? "felt-browser-error-primary-secret"
+            : "felt-browser-error-launch-failure";
```
lissyx:
```suggestion
            ? "felt-error-primary-secret"
            : "felt-browser-error-launch-failure";
```

storage/mozStorageService.cpp
```diff
@@ -356,21 +357,26 @@ nsresult Service::initialize() {
...
+  // first launch of a fresh enterprise profile still gets encrypted.
+  // mozStorage callers that explicitly name `obfsvfs::GetVFSName()` keep
+  // using it by name in either case; callers that name `basevfs` /
+  // `quotavfs` / the read-only-no-lock VFS keep their named choice
+  // (sqlite3_open_v2 with an explicit zVfs argument bypasses the default).
   rc = mObfuscatingSqliteVFS.Init(
       obfsvfs::ConstructVFS(quotavfs::GetVFSName()),
       /* aMakeDefault = */
-      StaticPrefs::security_storage_encryption_sqlite_enabled());
+      mozilla::IsProfileEncryptedDatabases());
```
lissyx: nit: I think that kind of change would be valuable upstreamed in a follow up?

storage/SQLiteEncryption.cpp
lissyx: this one was addressed

storage/SQLiteEncryption.cpp
beurdouche: This + !(MOZ_THUNDERBIRD)

storage/SQLiteEncryption.cpp
```diff
@@ -213,6 +412,203 @@ void EnsureNSSInitializedForEncryptionIfReady() {
...
+          NS_NewLocalFileInputStream(getter_AddRefs(markerStream), marker))) {
+    return NS_ERROR_NOT_AVAILABLE;
+  }
+  nsAutoCString sourcePath;
+  nsresult rv = NS_ReadInputStreamToString(markerStream, sourcePath, -1);
+  markerStream->Close();
+  if (NS_FAILED(rv)) {
+    return NS_ERROR_NOT_AVAILABLE;
+  }
+  sourcePath.Trim(" \t\r\n");
```
lissyx, line 569: Looks top weird, at least warrants a comment on why

lissyx:
For reference, this is what is shown to the user starting on an existing profile, as of current state:
image

toolkit/components/felt/content/felt.xhtml
```diff
@@ -200,6 +200,12 @@
             type="error"
             dismissable="true"
           ></moz-message-bar>
+          <moz-message-bar
+            class="felt-error-primary-secret is-hidden"
+            data-l10n-id="felt-error-primary-secret"
+            type="error"
+            dismissable="true"
+          ></moz-message-bar>
```
lissyx: note to self: make sure to run the marionette tests that exercize this and get screenshots

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -303,7 +303,11 @@ export class Felt {
 
       case "FeltParent:FirefoxLaunchFailure": {
         Services.felt.makeBackgroundProcess(false);
-        this.showWindow("felt-browser-error-launch-failure");
+        const errorClass =
+          message.data?.errorType === "primarySecret"
+            ? "felt-error-primary-secret"
+            : "felt-browser-error-launch-failure";
+        this.showWindow(errorClass);
         break;
```
lissyx, line 311:
```suggestion
        this.showWindow(message.data?.errorType || "felt-browser-error-launch-failure");
break;
```

toolkit/locales/en-US/chrome/mozapps/profile/profileSelection.properties
```diff
@@ -48,8 +48,12 @@ profileDeletionFailedTitle=Deletion Failed
...
 profileEncryptedButPrefOff=Your %S profile has encrypted databases, but this build of %S is configured to open them unencrypted. To avoid corrupting the profile, %S will not start. Enable database encryption in your settings or policy and try again.
 profileEncryptedButPrefOffTitle=Encrypted Profile
-profileNotEncryptedButPrefOn=Your %S profile contains unencrypted databases, but this build of %S is configured to require encryption. %S cannot migrate existing data automatically. Disable the encryption requirement or use a fresh profile to continue.
+profileNotEncryptedButPrefOn=%S now requires encrypted storage and your existing profile is unprotected. A fresh encrypted profile will be created and Sync will restore most of your data.\n\nWe strongly recommend Deleting the old profile as Archiving leaves it unprotected on disk.
 profileNotEncryptedButPrefOnTitle=Profile Migration Required
+# LOCALIZATION NOTE (profileMismatchDeleteAndContinue, profileMismatchKeepAndContinue, profileMismatchQuit): Button labels for the profile encryption migration dialog.
+profileMismatchDeleteAndContinue=Delete (Recommended)
+profileMismatchKeepAndContinue=Archive
+profileMismatchQuit=Quit
```
lissyx, line 56: Since it's enterprise specific it should live in enterprise-specific file.

lissyx: (not doing so will make troubles for l10n process)

toolkit/xre/nsAppRunner.cpp
```diff
@@ -2915,6 +2919,228 @@ static nsresult ProfileEncryptionMismatchDialog(const char* aMsgKey,
...
+      nsCOMPtr<nsIStringBundleService> sbs =
+          mozilla::components::StringBundle::Service();
+      NS_ENSURE_TRUE(sbs, NS_ERROR_FAILURE);
+
+      nsCOMPtr<nsIStringBundle> sb;
+      sbs->CreateBundle(kProfileProperties, getter_AddRefs(sb));
+      NS_ENSURE_TRUE_LOG(sb, NS_ERROR_FAILURE);
+
+      NS_ConvertUTF8toUTF16 appName(gAppData->name);
+      AutoTArray<nsString, 1> params = {appName};
```
lissyx, line 3072: I suspect we want to use correct branding here

toolkit/locales/en-US/chrome/mozapps/profile/profileSelection.properties
```diff
@@ -48,8 +48,12 @@ profileDeletionFailedTitle=Deletion Failed
 # LOCALIZATION NOTE (profileEncryptedButPrefOff, profileNotEncryptedButPrefOn): %S is the application name.
 profileEncryptedButPrefOff=Your %S profile has encrypted databases, but this build of %S is configured to open them unencrypted. To avoid corrupting the profile, %S will not start. Enable database encryption in your settings or policy and try again.
 profileEncryptedButPrefOffTitle=Encrypted Profile
-profileNotEncryptedButPrefOn=Your %S profile contains unencrypted databases, but this build of %S is configured to require encryption. %S cannot migrate existing data automatically. Disable the encryption requirement or use a fresh profile to continue.
+profileNotEncryptedButPrefOn=%S now requires encrypted storage and your existing profile is unprotected. A fresh encrypted profile will be created and Sync will restore most of your data.\n\nWe strongly recommend Deleting the old profile as Archiving leaves it unprotected on disk.
```
lissyx, line 51: `%S` should have a localization note that it is "brandShortName" like in other places where we use the brand

toolkit/xre/nsAppRunner.cpp
lissyx: We would need to create a `brandBundle` from `"chrome://branding/locale/brand.properties"` so we can source `brandShortName` and use it as a parameter.

toolkit/xre/nsAppRunner.cpp
```diff
@@ -2915,6 +2919,228 @@ static nsresult ProfileEncryptionMismatchDialog(const char* aMsgKey,
...
+  Choice choice = Choice::Quit;
+
+  // Test-automation shortcut: when this env is set, skip the modal.
+  const char* autoConfirm = PR_GetEnv("MOZ_TEST_AUTO_CONFIRM_PROFILE_RESET");
+  if (autoConfirm && *autoConfirm) {
+    nsDependentCString val(autoConfirm);
+    if (val.EqualsLiteral("delete")) {
+      choice = Choice::Delete;
+    } else if (val.EqualsLiteral("keep")) {
+      choice = Choice::Keep;
```
lissyx, line 3017: We need to make sure this is properly tested by QA, my local test with "keep" prompted me back to Felt login instead of starting browser, but the logic somehow performed the archive/new profile since on next start it was fine.

toolkit/xre/nsAppRunner.cpp
```diff
@@ -5809,6 +6076,43 @@ int XREMain::XRE_mainStartup(bool* aExitFlag,
     }
   }
 
+#if defined(MOZ_ENTERPRISE)
+  // SQLite at-rest encryption + Profile Refresh: copy the source profile's NSS
```
lissyx, line 6137: also something we want to be covered by QA I think

toolkit/profile/nsToolkitProfileService.cpp
```diff
@@ -2163,6 +2163,57 @@ nsresult nsToolkitProfileService::ApplyResetProfile(
...
+  // to launch with -P.
+  nsCOMPtr<nsIToolkitProfile> previousDefault;
+  (void)GetDefaultProfile(getter_AddRefs(previousDefault));
+
+  // Rename old profile aside so its slot in profiles.ini does not collide
+  // with the new (encrypted) profile that takes over the original name.
+  nsCString renamedName(originalName);
+  renamedName.AppendLiteral("-plaintext-");
+  renamedName.AppendInt(static_cast<int64_t>(PR_Now() / PR_USEC_PER_SEC));
+  rv = aOldProfile->SetName(renamedName);
```
lissyx, line 2193: we should mention this does not rename the directory itself, thats where I got tricked

testing/enterprise/test_browsing_child_encryption_mismatch.py
lissyx, line 1: this is not added to `manifest.toml` ?

testing/enterprise/test_felt_ui_scratch_reset.py
lissyx, line 1: This is not added to `manifest.toml` ?

testing/enterprise/test_browsing_child_encryption_mismatch.py
```diff
@@ -0,0 +1,231 @@
...
+    "relaunched headless Firefox has no Marionette socket and nothing tells "
+    "it to exit. The 'quit' branch returns NS_OK and exits cleanly, so only "
+    "that case is observable today. Possible rewrites: drop --tryExec for "
+    "the relaunch (so subprocess.run sees the parent exit immediately and "
+    "the child becomes a detached process to be terminated by the test), "
+    "OR add an instrumentation hook (write a marker file at the new profile "
+    "before LaunchChild) that the test can poll for. The current test is "
+    "preserved as a structural starting point."
+)
+class BrowsingChildEncryptionMismatch(MarionetteTestCase):
```
lissyx, line 46: This is completely bypassing the infrastructure we have in place

testing/enterprise/test_browsing_child_encryption_mismatch.py
```diff
@@ -0,0 +1,231 @@
...
+        self._logger = self.logger
+
+    def tearDown(self):
+        shutil.rmtree(self._workdir, ignore_errors=True)
+        super().tearDown()
+
+    def _env(self, extra=None):
+        env = os.environ.copy()
+        env.update({
+            "MOZ_HEADLESS": "1",
```
lissyx, line 65:
```suggestion
            "MOZ_HEADLESS": "1",
            "MOZ_AUTOMATION": "1",
```

testing/enterprise/test_browsing_child_encryption_mismatch.py
```diff
@@ -0,0 +1,231 @@
...
+            check=False,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+        )
+
+    def _create_profile(self, profiles_dir, name):
+        path = os.path.join(profiles_dir, name)
+        os.makedirs(path, exist_ok=True)
+        result = self._run(
+            ["--CreateProfile", f"{name} {path}", "-no-remote"],
```
lissyx, line 89: This may be broken badly, nothing happens except it starts Felt

lissyx:
Error handling added:
screenshot_test_browser_init_primary_secret_fetch_fail

testing/enterprise/test_felt_init_failures.py
```diff
@@ -23,3 +23,44 @@ def test_browser_init_policy_fetch_fail(self):
...
+                if len(handles) != 1:
+                    return False
+                self._driver.switch_to_window(handles[0])
+                self._driver.set_context("chrome")
+                error = self.get_elem(".felt-error-primary-secret")
+                return error if error.is_displayed() else False
+            except Exception:
+                return False
+
+        error_msg = self._longwait.until(launch_failure_shown)
```
lissyx:
```suggestion
        self.await_felt_auth_window()
        self.force_window()

        self._driver.set_context("chrome")
        error_msg = self.get_elem(".felt-error-primary-secret")
```

testing/enterprise/test_felt_init_failures.py
lissyx: I did it

testing/enterprise/test_felt_ui_scratch_reset.py
```diff
@@ -0,0 +1,126 @@
...
+    "in SelectStartupProfile (nsAppRunner.cpp ~3296) and SKIPS the "
+    "${TMPDIR}/felt-${MOZ_UPDATE_CHANNEL} scratch-profile branch the test "
+    "needs to exercise. The planted plaintext file therefore lives in a "
+    "directory the launched Firefox never inspects, and "
+    "ResetFeltUIScratchProfile is never reached. Two viable rewrites: "
+    "(a) drive Firefox without Marionette (so no --profile is passed) and "
+    "scrape stdout/stderr for the silent-wipe log line, or "
+    "(b) plant the plaintext file inside the Marionette-supplied profile dir "
+    "and assert the gate's eager-mark path wipes it (different code path "
+    "than ResetFeltUIScratchProfile). The current test is preserved as a "
```
lissyx, line 41: does it means it would not exercise the same code path? `self._driver.profile` should be the profile you want

toolkit/xre/nsAppRunner.cpp
```diff
@@ -2915,6 +2919,228 @@ static nsresult ProfileEncryptionMismatchDialog(const char* aMsgKey,
...
+
+  nsCOMPtr<nsIFile> newLocal;
+  rv = newProfile->GetLocalDir(getter_AddRefs(newLocal));
+  NS_ENSURE_SUCCESS(rv, rv);
+
+  SaveFileToEnv("XRE_PROFILE_PATH", newRoot);
+  SaveFileToEnv("XRE_PROFILE_LOCAL_PATH", newLocal);
+
+  *aExitFlag = true;
+  return LaunchChild(/*aBlankCommandLine*/ false, /*aTryExec*/ true);
```
lissyx: This may not be good at all, Felt is tracking the launched process.

storage/mozStorageService.cpp
beurdouche: Ack. I'll handle a round of upstreaming after this lands

testing/enterprise/test_felt_browser_starts.py
```diff
@@ -16,3 +18,30 @@ def test_felt_browser_start_from_cli(self):
         super().run_felt_base()
         self.run_felt_browser_started()
         self.run_ensure_firefox_config_set_in_browser()
+        # self.run_ensure_profile_encryption(self._driver.profile)
```
lissyx:
```suggestion
        self.run_ensure_profile_encryption(self._driver.profile)
```

lissyx: fixed in my PR

storage/SQLiteEncryption.cpp
```diff
@@ -88,6 +140,153 @@ mozilla::StaticRefPtr<ProfileObserver> sObserver;
...
+#endif  // MOZ_ENTERPRISE
+
+// One-shot migration. Called at most once per profile, from the
+// create branch of GetEncryptionKey. Rotates every DEK currently
+// wrapped under `lockstore::kek::local:sqlite` to instead be wrapped
+// under `lockstore::kek::password:sqlite`, then deletes the now-orphan
+// LocalKey record. Idempotent: on a true-fresh profile there's no
+// local:sqlite to rotate and this is a no-op.
+//
+// Caller must hold sStateMutex.
```
lissyx, line 199: Don't we have an annotation system for this ?

storage/SQLiteEncryption.cpp
```diff
@@ -213,6 +412,203 @@ void EnsureNSSInitializedForEncryptionIfReady() {
...
+    }
+    MOZ_LOG(GetSQLiteEncryptionLog(), LogLevel::Info, ("Lockstore opened"));
+  }
+
+  if (sKekRef.IsEmpty()) {
+#if defined(MOZ_ENTERPRISE)
+    const bool isFeltSpawnedBrowser = is_felt_browser();
+#else
+    const bool isFeltSpawnedBrowser = false;
+#endif
```
lissyx, line 457: nit:
```suggestion
  const bool isFeltSpawnedBrowser = 
#if defined(MOZ_ENTERPRISE)
  is_felt_browser()
#else
  false
#endif
  ;
```

storage/SQLiteEncryption.cpp
```diff
@@ -213,6 +412,203 @@ void EnsureNSSInitializedForEncryptionIfReady() {
...
+// this profile's own KEK, and read it. The source profile is still on disk at
+// this point (the reset's file deletion is a deferred background task). Never
+// copies the source keystore (its LocalKey is foreign). Caller holds
+// sStateMutex and has resolved sHandle + sKekRef. Returns NS_OK with aOutDek
+// filled on success, or NS_ERROR_NOT_AVAILABLE (no marker / no source DEK /
+// etc.) so the caller falls through to its normal missing-DEK handling.
+nsresult TransferRefreshedDekLocked(nsIFile* aProfileDir,
+                                    const nsACString& aCollection,
+                                    nsTArray<uint8_t>& aOutDek) {
+  sStateMutex.AssertCurrentThreadOwns();
```
lissyx, line 546: annotation missing?

storage/SQLiteEncryption.cpp
```diff
@@ -333,20 +741,49 @@ nsresult GetDatabaseEncryptionStatus(const nsACString& aDatabasePath,
...
+  if (normalizedDbPath.Length() > profileLen &&
+      normalizedDbPath.CharAt(profileLen) == kDirSeparator) {
+#ifdef XP_WIN
+    // Windows filesystems are case-insensitive; match nsLocalFileWin::Contains.
+    isUnder = _wcsnicmp(char16ptr_t(normalizedProfilePath.get()),
+                        char16ptr_t(normalizedDbPath.get()), profileLen) == 0;
+#else
+    isUnder = memcmp(normalizedProfilePath.get(), normalizedDbPath.get(),
+                     profileLen * sizeof(char16_t)) == 0;
+#endif
```
lissyx, line 803: `nsAutoString` we cannot make use of `==` ?

lissyx (approved):
I'll fix the remainder in my own branch. Waiting on QA to confirm first.

storage/SQLiteEncryption.cpp
lissyx: no we cannot easily

storage/SQLiteEncryption.cpp
lissyx: OK I can see the annotations

storage/SQLiteEncryption.cpp
lissyx: moved to just before `sPrimarySecret

testing/enterprise/test_browsing_child_encryption_mismatch.py
lissyx: (test is disabled, I had a look it's likely too much pain for low gain)

testing/enterprise/test_browsing_child_encryption_mismatch.py
lissyx: unrelevant on a disabled test, and likely not enough

testing/enterprise/test_felt_init_failures.py
lissyx: Fixed

testing/enterprise/test_felt_ui_scratch_reset.py
lissyx: likely not relevant as a test anyway

toolkit/components/felt/content/felt.xhtml
lissyx: Image

lissyx:
screenshot_test_browser_init_primary_secret_fetch_fail

toolkit/components/felt/Felt.sys.mjs
lissyx: my suggestion was wrong

toolkit/locales/en-US/chrome/mozapps/profile/profileSelection.properties
lissyx: the note is above so it's OK

toolkit/locales/en-US/chrome/mozapps/profile/profileSelection.properties
lissyx: after checking it should be fine

toolkit/xre/nsAppRunner.cpp
```diff
@@ -6118,6 +6155,51 @@ nsresult XREMain::XRE_mainRun() {
...
+                marker->Append(u".sqlite-refresh-source"_ns);
+                nsCOMPtr<nsIOutputStream> out;
+                nsresult mrv =
+                    NS_NewLocalFileOutputStream(getter_AddRefs(out), marker);
+                if (NS_SUCCEEDED(mrv)) {
+                  uint32_t written = 0;
+                  mrv = out->Write(aProfilePath.get(), aProfilePath.Length(),
+                                   &written);
+                  out->Close();
+                }
```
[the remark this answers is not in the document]
lissyx: This was fixed

storage/SQLiteEncryption.cpp
lissyx: OK it's just a trimming security

lissyx:
QA reported tests successfull on all three platforms.

lissyx:
The remainder of tests failures are known test intermittents

PR #1070, opened 2026-06-22
Bug 2049389 - Remove :shutdown to close console/SSO server and use terminate()

lissyx:
### Description 

Bugzilla: Bug-2049389

On Windows it looks like there can be a ReadTimeout exception raised when we `POST` to `/:shutdown`. Handling this over an http request is prone to risks, just `terminate()`.

This also tackle some issues that arised with UTF-8 path handling

### DIFF
--- a/testing/enterprise/felt_tests.py
+++ b/testing/enterprise/felt_tests.py
@@ -93,18 +93,6 @@ class LocalHttpRequestHandler(BaseHTTPRequestHandler):
         self.end_headers()
         self.wfile.write(bytes(payload, "utf8"))
 
-    def do_POST(self):
-        print("POST", self.path)
-
-        if self.path == "/:shutdown":
-            print("Shutting down as requested")
-            self.reply("OK")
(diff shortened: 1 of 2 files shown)


testing/enterprise/felt_tests.py
```diff
@@ -810,13 +798,23 @@ def teardown(self):
...
+            )
+            self.console_httpd.kill()
+            self.console_httpd.join(timeout=2)
         self._logger.info("Shutting down SSO")
-        requests.post(f"http://localhost:{self.sso_port}/:shutdown", timeout=2)
-        self._logger.info("Stopping process console")
-        self.console_httpd.join()
-        self._logger.info("Stopping process SSO")
-        self.sso_httpd.join()
+        self.sso_httpd.terminate()
```
jporter-dev, line 810: nit: this duplicates the console shutdown, could create a helper method for it if desired

jporter-dev (approved):
LGTM

PR #1074, opened 2026-06-23
NO BUG - Version bump for Enterprise a=release

lissyx:
Version was not updated in previous merges


1rneh (approved):
Thank you for correcting this!

PR #1076, opened 2026-06-23
Bug 2048644 - Resolve the enterprise console address from AutoConfig, falling back to distribution.ini

mkaply:
### Description

Bugzilla: Bug-2048644

The enterprise console address (enterprise.console.address) is migrating from distribution.ini to AutoConfig (firefox.cfg). Teach both the browser and the crash reporter client to prefer AutoConfig while keeping distribution.ini working, so existing deployments are unaffected during the migration.

Browser (nsAppRunner): once AutoConfig has been evaluated, if enterprise.console.address is locked (i.e. set via lockPref in firefox.cfg), derive the crash submission and update URLs from it and register the crash ServerURL, overriding any value read earlier from distribution.ini. The early distribution.ini read is left intact as the fallback, and the lock check means non-AutoConfig (distribution.ini) deployments keep their existing behavior.

Crash reporter client: resolve the console base URL in priority order:
  1. the ServerURL crash annotation recorded by the browser,
  2. the AutoConfig file (firefox.cfg, byte-shift decoded), then
  3. distribution.ini (legacy, retained for migration). This feeds both the crash submission URL and the Glean telemetry endpoint. The distribution.ini reader is preserved as the final fallback rather than removed.

This is just the enabling code. Actually putting the firefox.cfg in the build will be a separate patch.


mkaply:
### Background
The enterprise console address (`enterprise.console.address`) is moving from `distribution.ini` to AutoConfig (`firefox.cfg`). Today the browser reads it from `distribution.ini` very early in startup (`XRE_mainInit`, before prefs/JS exist) to build the crash-submission and update URLs, and the crash reporter client reads `distribution.ini` directly for its Glean telemetry endpoint. AutoConfig isn't evaluated until much later (`XRE_mainRun` → `FinishInitializingUserPrefs`), so it can't feed that early code path.

This patch makes both the browser and the crash reporter client **prefer AutoConfig while keeping `distribution.ini` fully working**, so deployments can migrate incrementally without breaking.

### Browser (`toolkit/xre/nsAppRunner.cpp`)
After AutoConfig has been evaluated, if `enterprise.console.address` is **locked** — i.e. set via `lockPref` in `firefox.cfg` — derive the crash-submission and update URLs from it via `XRE_ParseEnterpriseServerURL` and register the crash `ServerURL`, overriding anything read earlier from `distribution.ini`.

- The lock state is the discriminator for "this came from AutoConfig": `lockPref` locks the pref, whereas `distribution.ini`-applied prefs and test prefs do not. This is what lets AutoConfig take precedence without disturbing `distribution.ini` deployments.
- The existing early `distribution.ini` read in `XRE_mainInit` is **left untouched** as the fallback.
- The phase-3 block replaces the previous test-only (`enterprise.felt_tests.read_update_url_from_prefs`) path. Because the new gate is `IsLocked(...) || readUpdateUrlFromPref`, and the pref is never locked in the selenium harness, the existing update-URL tests (`test_felt_updates_basic`, `test_felt_updates_distIni`) keep their current behavior.

### Crash reporter client (`toolkit/crashreporter/client/app`)
New `autoconfig` module (`#[cfg(feature = "enterprise")]`) resolves the console base URL in priority order:
1. the `ServerURL` crash annotation already recorded by the browser,
2. the AutoConfig file (`firefox.cfg`) — bytes are un-shifted by `general.config.obscure_value` (tries the default 13, then plaintext) and scanned for the `lockPref`/`defaultPref`/`pref` setting,
3. `distribution.ini` (the original reader, relocated here and kept as the legacy fallback).

This single resolver now feeds both consumers:
- **Submit URL** (`config.rs`): previously only the `ServerURL` annotation; now, when that annotation is absent or a domain-less placeholder (`/submit?...`, which is what the empty built-in crash URL produces for enterprise), it falls back through the chain above.
- **Telemetry URL** (`glean.rs`): previously read `distribution.ini` directly; now goes through the same resolver.

Supporting changes: a bytes-returning `fs::read` helper; generalized `prefs_parser` so it can match `lockPref`/`defaultPref`/`pref` (not just `user_pref`). These two are generic and intentionally not feature-gated.

### What is explicitly preserved
`distribution.ini` reading is **retained, not removed**, in both the browser (early read, still the fallback) and the crash client (final step of the resolver). Non-AutoConfig deployments behave exactly as before.

### Testing
- New Rust unit tests cover all three resolution paths (annotation, AutoConfig-encoded file, `distribution.ini`) plus the `lockPref` parser case.
- `rustfmt` and `clang-format` clean.
- [Fill in] try run / `mach rusttests` for the crash client, and the felt update selenium tests.

toolkit/xre/nsAppRunner.cpp
```diff
@@ -6173,18 +6173,25 @@ nsresult XREMain::XRE_mainRun() {
...
+      // the same path without AutoConfig.
       bool readUpdateUrlFromPref = false;
-      rv = Preferences::GetBool(
-          "enterprise.felt_tests.read_update_url_from_prefs",
-          &readUpdateUrlFromPref);
-      if (NS_SUCCEEDED(rv) && readUpdateUrlFromPref) {
-        NS_WARNING("Setting appUpdateURL from pref value");
+      Preferences::GetBool("enterprise.felt_tests.read_update_url_from_prefs",
+                           &readUpdateUrlFromPref);
+      if (Preferences::IsLocked("enterprise.console.address") ||
```
jonathanmendez: With more recent changes, I don't think this block needs to be guarded at all anymore.

Previously, we had the "enterprise.console.address" pref defined with a default value in a `browser/branding/enterprise/pref/firefox-enterprise.js`. This code was added to enable re-setting the update URL to the expected value for tests (by setting the console address pref there), but it needed to be guarded in production so as to not blap the distribution.ini value with the default pref value.

Now, we have [removed the pref's definition/default value](https://github.com/mozilla/enterprise-firefox/pull/679), so we don't need to be worried about that blapping failure case anymore, and with the autoconfig changes the normal case will be to expect the pref to be present. @lissyx may weigh in but I think it would be safe to remove the guard (including the new `IsLocked` part).

toolkit/crashreporter/client/app/src/autoconfig.rs
```diff
@@ -0,0 +1,163 @@
...
+//! crash annotation, which the browser records once AutoConfig has run. A crash
+//! earlier than that (e.g. during startup) produces no usable annotation, so we
+//! recover the address ourselves, preferring the AutoConfig (`.cfg`) file and
+//! falling back to `distribution.ini` (retained during the migration away from
+//! it).
+
+use crate::config::installation_resource_path;
+use crate::prefs_parser::find_string_pref_call;
+
+/// The AutoConfig file name (the default value of `general.config.filename`).
```
jonathanmendez: nit: I can't seem to find an actual default value for `general.config.filename`? The public documentation for setting up AutoConfig does use "firefox.cfg". But presumably we will be in complete control of this so maybe reword or rename to specify "the enterprise AutoConfig file"/"ENTERPRISE_AUTOCONFIG_FILENAME"

toolkit/crashreporter/client/app/src/autoconfig.rs
```diff
@@ -0,0 +1,163 @@
...
+const PREF_FUNCTIONS: &[&str] = &["lockPref", "defaultPref", "pref"];
+
+/// Resolve the enterprise console base URL (without a trailing slash).
+///
+/// Resolution order:
+/// 1. `server_url` (the `ServerURL` crash annotation, i.e. the submission
+///    endpoint) by stripping the submission path back to the base;
+/// 2. the AutoConfig file;
+/// 3. `distribution.ini` (legacy, retained during migration).
+pub fn console_base(server_url: Option<&str>) -> anyhow::Result<String> {
```
jonathanmendez: I suggest renaming this to something like `enterprise_prefs.rs` and adding:
- a new `console_report_url()` fn that constructs the report url (moved from `config.rs`)
- A new `GLEAN_SUBMIT_PATH` const and a new `console_glean_url()` fn that constructs the glean url (moved from `construct_enterprise_console_endpoint` in `glean.rs`

This concentrates all of the console hard-coded paths and logic in this one place.

I'm guessing with the file rename, it may make sense to rename `read_string_pref` to something like `read_autoconfig_string_pref`

toolkit/crashreporter/client/app/src/enterprise_prefs.rs
```diff
@@ -0,0 +1,163 @@
...
+/// The AutoConfig file name (the default value of `general.config.filename`).
+const AUTOCONFIG_FILENAME: &str = "firefox.cfg";
+
+/// The pref holding the enterprise console address.
+pub const CONSOLE_ADDRESS_PREF: &str = "enterprise.console.address";
+
+/// Path appended to the console address to form the crash submission endpoint.
+pub const CRASH_SUBMIT_PATH: &str = "api/browser/crash-reports/submit";
+
+/// Default byte shift applied to AutoConfig files (`general.config.obscure_value`).
```
jonathanmendez, line 33: Should we lock this value for enterprise, so that a user doesn't mess this up? And then reword/rename here as above to indicate we have controlled the value?

toolkit/crashreporter/client/app/src/config.rs
```diff
@@ -269,6 +269,31 @@ impl Config {
...
+        // crash) the annotation is absent or a domainless placeholder
+        // (`/submit?...`, from the empty built-in crash URL), so recover the
+        // console address from AutoConfig directly.
+        #[cfg(feature = "enterprise")]
+        {
+            let usable = self
+                .report_url
+                .as_ref()
+                .and_then(|u| u.to_str())
+                .is_some_and(|u| u.starts_with("http"));
```
jonathanmendez: Consider `url::Url::parse` here?

jonathanmendez (changes requested):
Looks good, I just have a few cleanup requests.

toolkit/crashreporter/client/app/src/autoconfig.rs
mkaply: Yeah, in the followup where I enable this, it's going to be hardcoded in firefox-branding.js.

It is not set by default.

toolkit/crashreporter/client/app/src/enterprise_prefs.rs
mkaply: Which value specifically? The console address is locked.

jonathanmendez: Sorry, I expected the comment to limit to the one line: The "general.config.obscure_value" pref's value.

toolkit/xre/nsAppRunner.cpp
mkaply: Which guard? 

The enterprise.felt_tests.read_update_url_from_prefs part?

mkaply: I get it now.

mkaply:
@jonathanmendez Finally getting back to this.

So this is an update patch that removes the distrbution.ini code completely.

I decided to go ahead and add the autoconfig files in the repack step now.

Once I verify that the files land correctly in a build (unused), I can merge this patch.

Removing the distribution.ini code completely made things much simpler.

I did have to make some test changes that messed with the server.

lissyx:
> @jonathanmendez Finally getting back to this.
> 
> So this is an update patch that removes the distrbution.ini code completely.
> 
> I decided to go ahead and add the autoconfig files in the repack step now.
> 
> Once I verify that the files land correctly in a build (unused), I can merge this patch.
> 
> Removing the distribution.ini code completely made things much simpler.
> 
> I did have to make some test changes that messed with the server.

@mkaply Do you still plan to land this ? I can take over if required

mkaply:
Yes. I didn't realize that GitHub didn't re-ask for review.. Jonathan is
taking a look..

browser/branding/enterprise/distribution.ini
jonathanmendez: I don't know if we can perhaps remove this file from local builds entirely now? At least at the time we added it, it was only needed for the pref that's been moved.

However, since then, upstream [added a default distribution.ini](https://phabricator.services.mozilla.com/D281651) and maybe there could now be things that expect a distribution.ini to be present.

testing/enterprise/base_test.py
```diff
@@ -117,61 +133,65 @@ def tearDown(self):
...
+
+            self._logger.info(f"Writing console pref in {self.autoconfig_path}")
+            # The first line of an AutoConfig file is always skipped by the
+            # parser, so the comment is required rather than decorative.
+            #
+            # defaultPref, not lockPref as the shipped firefox.cfg uses: tests
+            # such as test_felt_console_error.py override the console address
+            # with using_prefs() to point at error-inducing hosts, and a locked
+            # pref cannot be overridden. This matches the default-branch
+            # semantics the [Preferences] section of distribution.ini had.
```
jonathanmendez: No need to include the historical context here.
```suggestion
            # pref cannot be overridden.
```

toolkit/crashreporter/client/app/src/config.rs
```diff
@@ -269,6 +269,26 @@ impl Config {
...
+        // from application.ini (`/submit?...`), so fall back to reading the
+        // console address from AutoConfig when the annotation isn't usable.
+        #[cfg(feature = "enterprise")]
+        {
+            let current = self
+                .report_url
+                .as_ref()
+                .and_then(|u| u.to_str())
+                .map(str::to_owned);
+            match crate::enterprise_prefs::console_report_url(current.as_deref()) {
```
jonathanmendez: I think this can be simplified to:

```suggestion
            let current = self.report_url.as_deref().and_then(OsStr::to_str);
            match crate::enterprise_prefs::console_report_url(current) {
```

jonathanmendez (approved):
A few minor suggestions

browser/branding/enterprise/distribution.ini
mkaply: We should be able to. Actually makes the Linux story better because we're not clashing with theirs.

PR #1095, opened 2026-06-30
Enterprise main merge 2026063000

jporter-dev:
### Description

Bugzilla: NO BUG  

Resolved conflicts:

*   browser/base/content/navigator-toolbox.js
*   browser/components/aboutlogins/tests/browser/browser_osAuthDialog.js
*   browser/extensions/newtab/lib/FrecencyBoostProvider/FrecencyBoostProvider.mjs
*   browser/extensions/newtab/lib/Wallpapers/WallpaperFeed.sys.mjs
*   toolkit/components/passwordmgr/LoginHelper.sys.mjs

Note: bug-2043699 removed fallback URLs for remote settings attachments, `FrecencyBoostProvider.mjs` and `WallpaperFeed.sys.mjs` were detected as merge conflicts, `MLEngineParent.sys.mjs` did not show a conflict but had a hanging `AppConstants` import that I removed.


toolkit/components/passwordmgr/LoginHelper.sys.mjs
```diff
@@ -1718,6 +1718,8 @@ export const LoginHelper = {
         telemetryEvent,
       };
     }
+    // We'll attempt to re-auth via Primary Password, so log out.
+    await token.logout();
```
jonathanmendez: I think we maybe don't want to add this. See the comment on line 1755, which implies we don't want the forced logout for enterprise (the forced logout moved to 1759 for non-enterprise).

At least this is how I had interpreted and resolved conflicts here before, so if this doesn't match what you think we should discuss.

jonathanmendez (changes requested):

toolkit/components/passwordmgr/LoginHelper.sys.mjs
jporter-dev: I think you're right, I'll remove that one

jonathanmendez (approved):

PR #1109, opened 2026-07-04
Bug 2052638 - Handle isDomainJoined property throwing r?gcp!

lissyx:
### Description 

Bugzilla: Bug-2052638

Default isDomainJoined to `false` and handle throwing from `getPropertyAsBool`


gcp (changes requested):
I think this is cleaner to express with `??` but also what you describe is not possible from the code. I want to understand what is going on.

lissyx:
Closing as I cannot reproduce anymore.

PR #1110, opened 2026-07-06
Bug 2039838 - add watermark policy

fiji-flo:
### Description 

Bugzilla: Bug-2039838

- New `Watermark` policy (`Match`, `Copy`, `Color`, `FontSize`, `Angle`, `SecondaryCopy`, `Size`) overlays a tiled, diagonal watermark on configured pages, added to `policies-schema.json`.
- `WatermarkPolicy.sys.mjs`: validates `Match`/`Color`/`FontSize`/`Angle`/`Size`, falls back `Copy` to the logged-in user's email (via `ConsoleClient.getLoggedInUserInfo()`) or a generic default, publishes the config through `sharedData`, and dynamically (re-)registers a `WatermarkPolicy` `JSWindowActor` scoped to the configured `Pages` so it's never created outside them.
- `WatermarkChild.sys.mjs`: draws the on-screen watermark as native anonymous content (tamper-resistant against page script), sized to the full document height; separately inserts a real DOM node around `beforeprint`/`afterprint` for print output, since anonymous content isn't captured by the print document clone. `Copy`/`SecondaryCopy` support `%t` (timestamp) / `%e` (email) template placeholders.
- Tests: `test_watermark.js` (xpcshell) for config publishing, defaults, clamping/sanitization, and email fallback; `browser_policy_watermark.js` for on-screen/print display and live policy apply/remove.

---

### Screenshots 
image

(policy value: 
```
{
  "Copy": "MOZILLA CONFIDENTIAL",
  "SecondaryCopy": "%t - %e",
  "Angle": -45,
  "Color": "rgba(200, 0, 0, 0.5)",
  "Match": [
    "https://*.mozilla.org/*",
    "https://example.com/*",
    "https://*.corp.example.com/*"
  ],
  "FontSize": 20,
  "Size": 250
}
```

---

### Testing 

* [x] Added tests

**Steps to verify changes:**
1. Add your user to the `Watermarking` org unit on the stage console
2. Visit https://example.com or https://www.mozilla.org
3. Confirm the watermark is displayed (also when taking a screenshot and printing)


browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
```diff
@@ -0,0 +1,343 @@
...
+const PREF_LOGLEVEL = "browser.policies.loglevel";
+
+const lazy = {};
+
+ChromeUtils.defineLazyGetter(lazy, "log", () => {
+  let { ConsoleAPI } = ChromeUtils.importESModule(
+    "resource://gre/modules/Console.sys.mjs"
+  );
+  return new ConsoleAPI({
+    prefix: "Watermark Policy",
```
lissyx:
```suggestion
    prefix: "WatermarkPolicy",
```

As much as I remember, it's used in `MOZ_LOG` and space there are making things complicated

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
```diff
@@ -0,0 +1,343 @@
...
+
+    let win = this.contentWindow;
+    if (!win) {
+      return;
+    }
+
+    // Captured once per watermark application.
+    let watermarkConfig = {
+      ...config,
+      timestamp: new Date().toLocaleDateString(),
```
lissyx: Why use `toLocaleDateString()` here and `toISOString()` in the print version?

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,195 @@
...
+  BrowserWindowTracker: "resource:///modules/BrowserWindowTracker.sys.mjs",
+  ConsoleClient: "resource://gre/modules/enterprise/ConsoleClient.sys.mjs",
+});
+
+ChromeUtils.defineLazyGetter(lazy, "log", () => {
+  let { ConsoleAPI } = ChromeUtils.importESModule(
+    "resource://gre/modules/Console.sys.mjs"
+  );
+  return new ConsoleAPI({
+    prefix: "Watermark Policy",
```
lissyx: same

```suggestion
    prefix: "WatermarkPolicy",
```

browser/locales/en-US/browser/policies/policies-descriptions.ftl
```diff
@@ -263,6 +263,8 @@ policy-UseSystemPrintDialog = Print using the system print dialog.
 
 policy-VisualSearchEnabled = Enable or disable visual search.
 
+policy-Watermark = Display a tiled, diagonal watermark over a list of websites.
```
lissyx: Please mention it also applies to printing

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -2849,6 +2849,56 @@
       "description": "Enable or disable visual search.",
       "examples": [true, false]
     },
+    "Watermark": {
+      "type": "object",
+      "x-category": "Miscellaneous",
+      "description": "Display a tiled, diagonal watermark over a list of websites.",
```
lissyx, line 2987: Mention printing as well

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -2849,6 +2849,56 @@
...
+      "properties": {
+        "Pages": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "Copy": {
+          "type": "string",
+          "description": "Text to display in the watermark. Falls back to the logged-in user's email, or a generic default, if not set. %t is replaced with the current timestamp, and %e with the logged-in user's email. Defaults to \"%e\""
```
lissyx:
```suggestion
          "description": "Text to use in the watermark for display or printing. Falls back to the logged-in user's email, or a generic default, if not set. %t is replaced with the current timestamp, and %e with the logged-in user's email. Defaults to \"%e\""
```

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,195 @@
...
+
+const PREF_LOGLEVEL = "browser.policies.loglevel";
+
+// Key used to publish the watermark configuration to content processes through
+// `sharedData`. Must be kept in sync with WatermarkChild.sys.mjs.
+export const WATERMARK_SHARED_DATA_KEY = "EnterprisePolicies:Watermark";
+
+const WATERMARK_ACTOR_NAME = "WatermarkPolicy";
+
+const LIST_LENGTH_LIMIT = 1000;
```
lissyx: Why this arbitrary limit ?

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,195 @@
...
+  });
+});
+
+export let WatermarkPolicy = {
+  async init(pages, copy, color, fontSize, angle, secondaryCopy, size) {
+    let validPages = [];
+
+    for (let i = 0; i < pages.length && i < LIST_LENGTH_LIMIT; i++) {
+      try {
+        new MatchPattern(pages[i]);
```
lissyx: You need to cleanup any port reference from `pages[i]` otherwise the `MatchPattern` may not work

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
```diff
@@ -0,0 +1,343 @@
...
+      return false;
+    }
+
+    let uri = this.document?.documentURI;
+    if (!uri) {
+      return false;
+    }
+
+    try {
+      return new MatchPatternSet(config.pages).matches(uri);
```
lissyx: `config.pages` should contains URLs that have no port

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,195 @@
...
+          ? fontSize
+          : DEFAULT_FONT_SIZE,
+        MAX_FONT_SIZE
+      ),
+      angle: (typeof angle === "number" ? angle : DEFAULT_ANGLE) % 360,
+      secondaryCopy,
+      email: email || "",
+    };
+
+    this._registerActor(validPages);
```
lissyx: So here we register the actor to match against a list of URLs
 - I am not sure how much the size of `matches` is expected to behave correctly
 - there's a hardcoded limit of 1000 pages, was this derived from testing?
I'm worried this may negatively impact perfs.

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
```diff
@@ -0,0 +1,343 @@
...
+    } catch (e) {
+      lazy.log.error(`Failed to match ${uri}: ${e}`);
+      return false;
+    }
+  }
```
lissyx: It sounds to me like we are doing matching twice ? There is matching applied when registering the actor and then when trying to apply the watermark we match again

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
```diff
@@ -0,0 +1,343 @@
...
+
+    // Captured once per watermark application.
+    let watermarkConfig = {
+      ...config,
+      timestamp: new Date().toLocaleDateString(),
+    };
+
+    try {
+      let { HighlighterEnvironment, CanvasFrameAnonymousContentHelper } =
+        lazy.DevTools;
```
lissyx: Is it OK to load devtools things like that in the actor? Would welcome a feedback from a devtools peer

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,195 @@
...
+    ChromeUtils.unregisterWindowActor(WATERMARK_ACTOR_NAME);
+    ChromeUtils.registerWindowActor(WATERMARK_ACTOR_NAME, {
+      child: {
+        esModuleURI: "resource:///modules/policies/WatermarkChild.sys.mjs",
+        events: {
+          DOMContentLoaded: {},
+          pageshow: {},
+        },
+      },
+      matches,
```
lissyx: we need to make sure it's a valid array, it might be empty

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
fiji-flo: Because I changed my mind and forgot that I need to change it twice.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
fiji-flo: I'm doing (trying at least) the same we do in `WebsiteFilter`. The match pattern also do need at least a "/" as path, but we do not sanitize them. I'd keep it the same. We should really improve our JSON schema to enforce this.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
fiji-flo: Not arbitrary, same as WebsiteFilter.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
lissyx: I cannot speak about `WebsiteFilter`, it may have different constraints on where the value gets from?

> The match pattern also do need at least a "/" as path, but we do not sanitize them. I'd keep it the same. We should really improve our JSON schema to enforce this.

Are you OK to land something that may not work on purpose and without warning/error feedback possible?
It's just a matter of making a `URI` and getting everything except the port.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
lissyx: It warrants at least a comment to explain it then, but `WebsiteFilter` provides no rationale either.

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
fiji-flo: Yes, I can reach out. Took it from: https://firefox-source-docs.mozilla.org/devtools/tools/highlighters.html

browser/components/enterprisepolicies/tests/browser/browser.toml
```diff
@@ -159,6 +159,11 @@ support-files = [
 
 ["browser_policy_usermessaging.js"]
 
+["browser_policy_watermark.js"]
+support-files = ["policy_watermark.html"]
+# Only run in MOZ_ENTERPRISE builds
+run-if = ["enterprise"]
```
lissyx, line 168: Reminder that for now we are not green on those suites, so it's hard to keep track of the regressions here.

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
fiji-flo: We need that for
- removing watermarks if the policy changed
- doing the match for printing

I can look into this more if you have big doubts.

lissyx: > We need that for
> 
>     * removing watermarks if the policy changed
> 

If policy changes, we remove and re-install the actor, so the list we pass to `matches` should already be correct, no?

>     * doing the match for printing
> 

I'm not sure of the specifics here, that may be a good reason.

> 
> I can look into this more if you have big doubts.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,195 @@
...
+  },
+
+  _registerActor(matches) {
+    ChromeUtils.unregisterWindowActor(WATERMARK_ACTOR_NAME);
+    ChromeUtils.registerWindowActor(WATERMARK_ACTOR_NAME, {
+      child: {
+        esModuleURI: "resource:///modules/policies/WatermarkChild.sys.mjs",
+        events: {
+          DOMContentLoaded: {},
+          pageshow: {},
```
lissyx, line 172: I'm not sure you want `pageshow` here, it's triggered multiple times during the loading phase?

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
lissyx: Good, I just wonder what happens if e.g. we have a policy that disables devtools :)

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
```diff
@@ -0,0 +1,324 @@
...
+
+  #applyPrintWatermark(
+    config = Services.cpmm.sharedData.get(WATERMARK_SHARED_DATA_KEY)
+  ) {
+    this.#removePrintWatermark();
+
+    let doc = this.document;
+    if (!doc?.documentElement) {
+      return;
+    }
```
lissyx, line 459: It's applied at the content document level? Would it get exposed that devtools allows to remove it?
If so we should add an event listener so any removal done not by us would force apply again?

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,195 @@
...
+      messageManagerGroups: ["browsers"],
+      safeForUntrustedWebProcess: true,
+    });
+  },
+
+  _unregisterActor() {
+    ChromeUtils.unregisterWindowActor(WATERMARK_ACTOR_NAME);
+  },
+
+  _refreshOpenTabs(config) {
```
lissyx, line 195: nit:

```suggestion
  _refreshExistingTabs(config) {
```

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,195 @@
...
+      // No browser windows available (e.g. during early startup or in
+      // non-browser contexts); nothing to refresh.
+      return;
+    }
+    for (let win of windows) {
+      for (let browser of win.gBrowser?.browsers ?? []) {
+        try {
+          browser.sendMessageToActor(
+            "Watermark:Refresh",
+            { config },
```
lissyx, line 209: We already have the `Services.ppmm.sharedData` to share info?

lissyx: All the call sites of `_refreshOpenTabs()` currently are following immediately calls to `Services.ppmm.sharedData.set(WATERMARK_SHARED_DATA_KEY);` / `Services.ppmm.sharedData.delete(WATERMARK_SHARED_DATA_KEY);`, so is there any reason we want to pass here in the message when the information is already shared ?

browser/components/enterprisepolicies/tests/browser/browser_policy_watermark.js
```diff
@@ -0,0 +1,142 @@
...
+        "Watermark should appear on the already-open tab after the policy is applied"
+      );
+      ok(true, "Watermark appears on an already-open tab (live apply)");
+
+      await setupPolicyEngineWithJson("");
+      await BrowserTestUtils.waitForCondition(
+        async () => !(await isShowingWatermark(browser)),
+        "Watermark should be removed from the already-open tab after the policy is removed"
+      );
+      ok(true, "Watermark removed from an already-open tab (live remove)");
```
lissyx: I'm not certain this would really reflect a live change. 
It might be better to move this test to `toolkit/components/enterprisepolicies/tests/browser/live/` and follow e.g. `toolkit/components/enterprisepolicies/tests/browser/live/browser_live_*`

browser/components/enterprisepolicies/tests/browser/browser_policy_watermark.js
lissyx, line 1: At least until we have coverage correct on all test suites, I think it's critical we get some basic testing of that feature on `testing/enterprise` using a marionette-enterprise test.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
fiji-flo: If a user enters a specific port and it's applied everywhere it also not great. At least this is consistent behavior and I think WebsiteFilter is very similar. We maybe should address this more broadly?

lissyx: If there's a port defined in the string used to build `MatchPattern` the matcher will just not work, it will match nothing at all.

lissyx: https://searchfox.org/enterprise-main/rev/e0b7d51f9d292d6b77c4759ec657bd180b51f7fa/toolkit/components/enterprise/modules/ConsoleClient.sys.mjs#200-206

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
fiji-flo: I'd look for a better solution down the way, more similar to the anonymous content that's used for non printing. As far as I can see you cannot inspect the DOM that's printed in dev tools, only with the browser toolbox, but that's game over anyway. I you have access to this code you can inject an eventListener and invalidate the watermark, but I'd say that's pretty sophisticated.

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
fiji-flo: Talked with emilio and using insertAnonymousContent directly now.

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
lissyx: Fair enough, should we have a follow up there? While for an end user it may be pretty sophisticated, I was wondering about malware trying to exfiltrate data

fiji-flo: I would let this be midterm. Even if we manage to get this hardened on the browser side, removing the layer from the PDF is still simple. I think we'll revisit this as a feature and would have a print as image policy. But even there I'd say that removing a watermark is a noop in the AI days. I think this is 99% for the "remind folks not to share it" use case.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
fiji-flo: https://searchfox.org/enterprise-main/search?q=opentabs&path=&case=false&regexp=false -> Number of results: 1653

https://searchfox.org/enterprise-main/search?q=existingtabs&path=&case=false&regexp=false -> Number of results: 16

Voting for opentabs

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
fiji-flo: From reading the code `sharedData` seems very async and sending the config explicitly avoids races.

browser/components/enterprisepolicies/schemas/policies-schema.json
fiji-flo: @bsmth That's documentation. I think printing is somewhat implied but at the same time not that important to mention it.

bsmth: Indeed, we'll cover in docs, I think the current string is sufficient.

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
lissyx: Mid term or not, we should have a follow up to make sure it's tracked at some point.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
lissyx: I still get mislead by the name of the method, but if you think it's good enough ...

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
lissyx: `sharedData` is an `EventTarget`: https://searchfox.org/enterprise-main/rev/e0b7d51f9d292d6b77c4759ec657bd180b51f7fa/dom/chrome-webidl/MozSharedMap.webidl#18-26

So you can use it and augment your existing `handleEvent` to handle the changes

lissyx: Doing so you could even completely get rid of `refreshOpenTabs`, since it should propagate the event to all listeners you installed in each actor

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,256 @@
...
+   * Removes the watermark from all open tabs and unregisters the actor.
+   */
+  cleanup() {
+    this._unregisterActor();
+
+    Services.ppmm.sharedData.delete(WATERMARK_SHARED_DATA_KEY);
+    Services.ppmm.sharedData.flush();
+
+    this._refreshOpenTabs(null);
+  },
```
lissyx, line 154: Re-reading this, we have `_refreshOpenTabs` that will send message to all actors, but we have just unregistered ?

I guess you want to first delete the key, let it propagate and refresh tabs (I think with the event handler this should be better) and then only unregister the actor?

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
lissyx: Ok there is no real rationale here: https://bugzilla.mozilla.org/show_bug.cgi?id=1429178#c7

lissyx: I'm going to ask around to make sure the value is sound

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,256 @@
...
+      try {
+        new MatchPattern(match[i]);
+        validMatches.push(match[i]);
+      } catch (e) {
+        lazy.log.error(`Invalid pattern on Watermark. Pages: ${match[i]}`);
+      }
+    }
+
+    if (!validMatches.length) {
+      lazy.log.error("Watermark policy has no valid pages; not applying.");
```
jporter-dev: There's some leftover references to `Pages` in the JSDoc and validation errors that may be confusing to trace since the policy got refactored to use `Match`

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -2849,6 +2849,56 @@
...
+      "properties": {
+        "Match": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "Copy": {
+          "type": "string",
+          "description": "Text to display in the watermark. Falls back to the logged-in user's email, or a generic default, if not set. %t is replaced with the current timestamp, and %e with the logged-in user's email. Defaults to \"%e\""
```
jporter-dev:
```suggestion
          "description": "Text to display in the watermark. Falls back to the logged-in user's email, or a generic default, if unavailable. %t is replaced with the current timestamp, and %e with the logged-in user's email. Defaults to \"%e\""
```

The wording here confused me a bit wrt the default. The JSDoc for this mentions it falls back to `CONFIDENTIAL` if email is unavailable, which is a bit more accurate than saying when it's "not set."

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,256 @@
...
+      } catch (e) {
+        lazy.log.error(`Failed to get logged in user info for Watermark: ${e}`);
+      }
+    }
+
+    if (!copy) {
+      copy = email || "CONFIDENTIAL";
+    }
+
+    let config = {
```
jporter-dev: Could we move the config validation/normalization to a private helper function? For example, `buildConfig` that validates and normalizes params, returning this JSON payload. 

Passing several positional args to `init` and normalizing `Match` at the call site spreads things out a bit. `init` is pretty heavy already. 

This could instead become `WatermarkPolicy.init(param)`, and `init` can call `buildConfig()` which will normalize `param.Match || []`. Added benefit is the param normalization can be tested directly.

jporter-dev (changes requested):
Overall this is looking good!

I do have a question on the application of the watermark - when in "Reader Mode" on pages that support it, the watermark is not shown since this loads `about:reader?url=https%3A%2F%2Fwww.mozilla.org%2Fen-US%2F` which does not match our patterns. This means a user can effectively bypass the watermark policy by entering reader mode and can screenshot or print from there, so that may be a path we want to handle.

fiji-flo:
> Overall this is looking good!
> 
> I do have a question on the application of the watermark - when in "Reader Mode" on pages that support it, the watermark is not shown since this loads `about:reader?url=https%3A%2F%2Fwww.mozilla.org%2Fen-US%2F` which does not match our patterns. This means a user can effectively bypass the watermark policy by entering reader mode and can screenshot or print from there, so that may be a path we want to handle.

Good catch and a horrible rabbit hole. There's also view:source and then the worst. Printing from reader mode....

Got them all I hope. @jporter-dev

fiji-flo:
@lissyx can you take another look. I wanna keep the limit and behavior like WebsiteFilter for consistency. I removed ppmm after talking to folks and everyone's reaction was "this still exists, don't use it". Talked with @bsmth about the copy (see his comment).

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
```diff
@@ -0,0 +1,485 @@
...
+   * @param {WatermarkConfig} config Watermark configuration.
+   * @returns {Element}
+   */
+  #buildNode(config) {
+    let doc = this.document;
+    let container = doc.createElementNS("http://www.w3.org/1999/xhtml", "div");
+    container.setAttribute("id", "enterprise-watermark");
+    container.setAttribute(
+      "style",
+      watermarkScreenStyle(config, documentHeight(doc))
```
jporter-dev:
```suggestion
      watermarkScreenStyle(config, documentWidth(doc), documentHeight(doc))
```

missing a param, i think the observer corrects this immediately anyway

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
lissyx: OK there are other cases that were suggested

jporter-dev (commented):
This looks good to me wrt to reader mode! Not to send you down more rabbitholes, but I did come across a couple other edge cases/issues 😅 some of which may be okay as a followup... 

Mostly related to source code, so may be okay as followups...

- JSON loads in a `resource://devtools` in the JSON previewer with no watermark due to the `resource://devtools` url pattern. Disabling this restores the watermark (`devtools.jsonview.enabled=false`)
    - [Sample URL](https://product-details.mozilla.org/1.0/all.json)
- Some source code pages don't show watermark, not 100% sure why.
    - [Sample URL](https://hg-edge.mozilla.org/mozilla-central/raw-file/tip/modules/libpref/init/all.js)
    - [Another example](https://hg-edge.mozilla.org/mozilla-central/raw-file/tip/layout/style/res/html.css)
    - [This one works?](https://developer.mozilla.org/static/client/527.27b4368313f61142.js) - much larger file, slow to load, maybe timing/race?

This one may be worthwhile to fix now since it affects printing PDFs which are a common watermark usecase.

- Sizing/styling issue when printing PDFs from the pdf.js viewer. Reader mode reminded me to check that pdf.js showed the watermark since it loads in a previewer, which it does, but printing is broken. Unlike json preview, disabling with `pdfjs.disabled=true` just downloads it so watermarking is not applied.

broken pdf print

browser/components/enterprisepolicies/helpers/moz.build
```diff
@@ -9,6 +9,9 @@ EXTRA_JS_MODULES.policies += [
     "AIChatbotPolicies.sys.mjs",
     "BookmarksPolicies.sys.mjs",
     "ProxyPolicies.sys.mjs",
+    "WatermarkChild.sys.mjs",
+    "WatermarkParent.sys.mjs",
+    "WatermarkPolicy.sys.mjs",
```
lissyx: It should be under `if CONFIG["MOZ_ENTERPRISE"]:`

browser/components/enterprisepolicies/helpers/moz.build
lissyx, line 21:
```suggestion
EXTRA_JS_MODULES.policies += [
"AIChatbotPolicies.sys.mjs",
"BookmarksPolicies.sys.mjs",
"ProxyPolicies.sys.mjs",
"WebsiteFilter.sys.mjs",
]

if CONFIG["MOZ_ENTERPRISE"]:
EXTRA_JS_MODULES.policies += [
"SyncPolicy.sys.mjs",
"WatermarkChild.sys.mjs",
"WatermarkParent.sys.mjs",
"WatermarkPolicy.sys.mjs",
]
```

browser/components/enterprisepolicies/helpers/WatermarkChild.sys.mjs
```diff
@@ -0,0 +1,485 @@
...
+    }
+    if (!this.#matchesDocument(config)) {
+      return;
+    }
+
+    // Reflects when the document was printed, rather than when it was loaded.
+    let watermarkConfig = {
+      ...config,
+      timestamp: Temporal.Now.instant().toString({ smallestUnit: "second" }),
+    };
```
lissyx: nit:
```suggestion
    let watermarkConfig = {
      ...config,
      // Timestamp here reflects when the document was printed, rather than when it was loaded.
      timestamp: Temporal.Now.instant().toString({ smallestUnit: "second" }),
    };
```

browser/components/enterprisepolicies/tests/browser/browser.toml
```diff
@@ -157,8 +157,17 @@ support-files = [
...
+# Bypass the Felt UI launcher process so the harness's profile/marionette
+# setup reaches the actual browser process being tested.
+environment = ["MOZ_BYPASS_FELT=1"]
```
lissyx: This is done on CI already, we should not pollute here.

browser/components/enterprisepolicies/tests/browser/browser.toml
```diff
@@ -157,8 +157,17 @@ support-files = [
...
+
 ["browser_policy_websitefilter.js"]
 
 ["browser_policy_websitefilter_telemetry.js"]
 # Only run in MOZ_ENTERPRISE builds
 run-if = ["enterprise"]
+environment = ["MOZ_BYPASS_FELT=1"]
```
lissyx: same, remove this

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
lissyx: I have not been able to get answers :(

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
```diff
@@ -0,0 +1,306 @@
...
+  size,
+  fontSize,
+  angle,
+  secondaryCopy,
+} = {}) {
+  const validMatches = [];
+
+  for (let i = 0; i < match.length && i < LIST_LENGTH_LIMIT; i++) {
+    try {
+      new MatchPattern(match[i]);
```
lissyx, line 260: Please make sure we remove any port here, otherwise your matches will not work if the user submits with a port.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
lissyx: This precise code moved so the coment is outdated, but I made a new one on the new code, the requirement to drop the port is still here.

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
lissyx: https://searchfox.org/enterprise-main/rev/e0b7d51f9d292d6b77c4759ec657bd180b51f7fa/toolkit/components/enterprise/modules/ConsoleClient.sys.mjs#200-206

lissyx:
> @lissyx can you take another look. I wanna keep the limit and behavior like WebsiteFilter for consistency. I removed ppmm after talking to folks and everyone's reaction was "this still exists, don't use it". Talked with @bsmth about the copy (see his comment).

This looks simpler with the removal of `ppmm` but there are still things to be addressed, the more imporant are IMHO:
 - basic `marionette-enterprise` testing to ensure we can discover regressions
 - making sure the submitted `Match` does not include a port
 - `pageshow` is triggered multiple times by the system when rendering a page, do we want to react to it ?

browser/components/enterprisepolicies/helpers/WatermarkPolicy.sys.mjs
fiji-flo: That's intended. It's just a check if the watermark is applied on the happy path.

fiji-flo:
@jporter-dev thanks for all the edge cases. Everything except for PDF is working now, but that's out of scope.
@lissyx I would not remove the ports. I would very much expect this to behave like WebsiteFilter and there we don't strip the port.

lissyx:
> @jporter-dev thanks for all the edge cases. Everything except for PDF is working now, but that's out of scope. @lissyx I would not remove the ports. I would very much expect this to behave like WebsiteFilter and there we don't strip the port.

But it's not really a choice you have here or not. WebExtensions matches that are used under the hood **requires** not to use a port, that's documented, and that is enforced on console client when dealing with actors for doing the SSO callback match. Keeping the port will make the match never work

lissyx:
@fiji-flo https://bugzilla.mozilla.org/show_bug.cgi?id=1468162 and https://bugzilla.mozilla.org/show_bug.cgi?id=1362809

jporter-dev (approved):
Changes look good, thanks!

fiji-flo:
@lissyx I do see that using `"https://*.mozilla.org:443/*"` will never work. What I care about is that the same is true for the `WebsiteFilter` Policy, if `"https://*.mozilla.org:443/*"` is in `"Block`" it is silently ignored. I value consistency in this case. What we do in SSO callback was relevant due to development on localhost where this is common. I would rather tackle these 8/9 year old bugs than diverting policy behavior.

Maybe @mkaply has an opinion. But if that's the blocker and you insist I can live with silently dropping the port.

mkaply:
> @lissyx I do see that using `"https://*.mozilla.org:443/*"` will never work. What I care about is that the same is true for the `WebsiteFilter` Policy, if `"https://*.mozilla.org:443/*"` is in `"Block`" it is silently ignored. I value consistency in this case. What we do in SSO callback was relevant due to development on localhost where this is common. I would rather tackle these 8/9 year old bugs than diverting policy behavior.
> 
> Maybe @mkaply has an opinion. But if that's the blocker and you insist I can live with silently dropping the port.

I agree. We shouldn't special case things in policy for broken Firefox features. If we want to fix match patterns, we should fix match patterns in Firefox and it will work everywhere.

I'm not convinced ports are that important anyway for things like this. We have never received a complaint about ports not working properly in WebsiteFilter.

lissyx:
> @lissyx I do see that using `"https://*.mozilla.org:443/*"` will never work. What I care about is that the same is true for the `WebsiteFilter` Policy, if `"https://*.mozilla.org:443/*"` is in `"Block`" it is silently ignored. I value consistency in this case. What we do in SSO callback was relevant due to development on localhost where this is common. I would rather tackle these 8/9 year old bugs than diverting policy behavior.

Then when someone needs to filter a site with a specific port, they get something not working at all with the current implementation of `matches` ; while if we drop the port, since you have code enforcing manual matching later, then it should work.

Current situation means it can fail to work in hard to discover ways for users, except if they dig into the implementation. The fact that `WebsiteFilter` policy may be hit by the same thing is IMHO not relevant  since it's not the current code under review. If it would warrant a follow up to be fixed there as well, we should do it.

> 
> Maybe @mkaply has an opinion. But if that's the blocker and you insist I can live with silently dropping the port.

If you consider not dropping the port is a good idea, do as you wish. You still have not addressed others of my comments, especially around tests. We have landed in the past features without test coverage, and it proved ot be a bad idea.

> I agree. We shouldn't special case things in policy for broken Firefox features. If we want to fix match patterns, we should fix match patterns in Firefox and it will work everywhere.

The problem, I insist, is that we can have people insert legitimately a port and it would silently not work in hard to debug ways. Maybe nobody complained about `WebsiteFilter` policy but that does not mean people may not have been hit by the bug.

lissyx (approved):
r=me with some basic `marionette-enterprise` test coverage to make sure we have something we seriously monitor that would catch some regression of the feature

Given I'll be on PTO next week, and I trust you can fix this, I do not want to block more from landing once your have those tests in place

testing/enterprise/test_felt_browser_watermark.py
fiji-flo, line 1: @lissyx I did add these as basic tests.

lissyx: Oh, my bad, I had reloaded the PR but I was not seeing them

testing/enterprise/manifest.toml
```diff
@@ -113,6 +113,11 @@ run-if = [
   "buildapp == 'browser'", # browser specific: notification
 ]
 
+["test_felt_browser_watermark.py"]
+run-if = [
+  "buildapp == 'browser'", # browser specific: watermark drawn on a tab
```
lissyx, line 118: While current implementation is browser only, it's useful we keep it in mind because it may well make a lot of sense on Thunderbird as well

lissyx (approved):

1rneh (approved):
r+ for the test `testing/enterprise/test_felt_browser_watermark.py`.

Note: The fluent string `policy-Watermark` needs to be added in [enterprise-policies-descriptions.ftl](https://searchfox.org/enterprise-main/source/browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl).

PR #1115, opened 2026-07-06
Bug 2052919 - Collect MAC address on network interfaces reporting as Ethernet

lissyx:
### Description 

Bugzilla: Bug-2052919

Bug 2013652 removed the requirement for MAC address when receiving NetLink messages because WWAN interface do not have one. However a WiFI network interface have, and because NetLink sends multiple messages, it may be missing at first and it is thus required to continue receiving messages to get the MAC address.

### Testing 

* [x] Added tests

### DIFF
--- a/netwerk/system/netlink/NetlinkService.cpp
+++ b/netwerk/system/netlink/NetlinkService.cpp
@@ -275,6 +275,9 @@ class NetlinkLink {
     len = aNlh->nlmsg_len - NLMSG_LENGTH(sizeof(*iface));
 
     bool hasName = false;
+#if defined(MOZ_ENTERPRISE)
+    bool hasMAC = false;
+#endif
     for (attr = IFLA_RTA(iface); RTA_OK(attr, len);
          attr = RTA_NEXT(attr, len)) {
       if (attr->rta_type == IFLA_IFNAME) {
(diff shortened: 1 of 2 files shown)


netwerk/system/netlink/NetlinkService.cpp
jonathanmendez: In the earlier version of this code, this would return false if `!hasName || !hasMAC`; should this incoporate `hasMAC` again?
```suggestion
    if (!hasName
#if defined(MOZ_ENTERPRISE)
		      || (IsTypeEther() && !hasMAC)
#endif
      ) {
```
(side note: I copied the formatting from above which includes tabs, which I didn't think I'd seen in cpp files before)

lissyx: Yes I may have forgot this one

netwerk/system/netlink/NetlinkService.cpp
```diff
@@ -284,15 +287,24 @@ class NetlinkLink {
...
+#endif
+      ) {
         break;
       }
     }
 
-    if (!hasName) {
+    if (!hasName
+#if defined(MOZ_ENTERPRISE)
+        && (IsTypeEther() && !hasMAC)
```
jonathanmendez: This should be `||` to match previous behavior.

I'm interpreting this block to say "if the required properties are not found, do not finish the init (do not copy the data into `mIface`)". As I understand, this change is trying to say "if the network interface is an Ethernet, it should have a MAC address"; and therefore for an Ethernet interface, if the MAC address is not found, we should not finish the init.
```suggestion
        || (IsTypeEther() && !hasMAC)
```

lissyx: You're right that was the implementation back then :(

jonathanmendez (approved):

PR #1116, opened 2026-07-07
Bug 2039258 - Send security event in case of unsafe site visit

mz-darnold:
### Description

Bugzilla: Bug-2039258

Send telemetry when a user visits a URL identified as malicious by Google Safe Browsing. Also applies to unsafe downloads.

---

### Testing

* [x] Added tests

---

**Steps to verify changes:**
1. Visit https://testsafebrowsing.appspot.com/
2. Click links
3. Confirm unsafe browsing telemetry in console


browser/actors/BlockedSiteChild.sys.mjs
```diff
@@ -2,12 +2,23 @@
...
+
 const lazy = {};
 
 ChromeUtils.defineESModuleGetters(lazy, {
   SafeBrowsing: "resource://gre/modules/SafeBrowsing.sys.mjs",
 });
 
+// Safe Browsing threat types that represent an unsafe site visit. Note that
+// "addon" (addonBlocked) is intentionally excluded as it is not a site threat.
+const UNSAFE_SITE_THREAT_TYPES = new Set([
```
jonathanmendez: Since the bug is requesting a security event "in case of unsafe site visit", it's not clear to me why we would only care about a subset of unsafe site threats instead of reporting them all. I don't think we want to filter the threat type at all, unless that has been requested at the Product level (i.e. by @bsmth )

browser/actors/BlockedSiteChild.sys.mjs
```diff
@@ -45,6 +56,21 @@ export class BlockedSiteChild extends JSWindowActorChild {
     let blockedInfo = getSiteBlockedErrorDetails(this.docShell);
     let provider = blockedInfo.provider || "";
 
+    // In enterprise builds, report a security event when a user lands on an
+    // unsafe site so that administrators can monitor Safe Browsing hits.
+    if (
```
jonathanmendez: This code is happening in `onAboutBlockedLoaded` which is the browser's user-facing reaction to a decision to block the page. Our other security events are instead happening at the code that makes the decision. It's worth investigating whether this can be moved to the decision point instead of here at the reaction point (and if the decision point happens in the main process, then you can avoid the need for the extra IPC being done here, although that's not a huge deal).

browser/actors/BlockedSiteParent.sys.mjs
```diff
@@ -120,6 +121,90 @@ export class BlockedSiteParent extends EscapablePageParent {
...
+   * @param {string} data.list - The Safe Browsing list that matched the site.
+   */
+  _recordUnsafeSiteVisit({ url, threatType, provider, list }) {
+    if (!AppConstants.MOZ_ENTERPRISE) {
+      return;
+    }
+    if (
+      !Services.prefs.getBoolPref(
+        "browser.safebrowsing.enterprise.telemetry.unsafeSiteVisit.enabled",
+        true
```
jonathanmendez: This matches what we have elsewhere, but I think we should use `false` for the default value, i.e. only send these events if the policy to do so has been set (and we should fix the other uses). Probably worth double-checking with @gcp in case we want to default to `true` for some reason.

browser/actors/metrics.yaml
```diff
@@ -28,3 +28,46 @@ fullscreen:
...
+
+#ifdef MOZ_ENTERPRISE
+security:
+  unsafe_site_visit:
+    type: event
+    data_sensitivity:
+      - highly_sensitive
+    description: >
+      Recorded when a user navigates to a top-level site blocked by Safe
+      Browsing (malware, phishing, unwanted software, or harmful application).
```
jonathanmendez: Noting to also remove this parenthetical listing the specific types of blocks that are being filtered.

browser/actors/metrics.yaml
```diff
@@ -28,3 +28,46 @@ fullscreen:
...
+      url:
+        description: >
+          The URL of the unsafe site that was blocked. Collection level is
+          configurable via the
+          browser.safebrowsing.enterprise.telemetry.unsafeSiteVisit.urlLogging
+          pref: "full" (default), "domain", or "none".
+        type: string
+      threat_type:
+        description: >
+          The type of threat detected: "malware", "phishing", "unwanted", or
```
jonathanmendez: Noting to remove this list of possible threats if we don't filter them.

browser/components/safebrowsing/content/test/browser_enterprise_unsafe_site_telemetry.js
```diff
@@ -0,0 +1,165 @@
...
+        event.extra.url,
+        url,
+        "Telemetry should include the blocked URL"
+      );
+      Assert.equal(
+        event.extra.threat_type,
+        threatType,
+        "Telemetry should include the threat type"
+      );
+      Assert.equal(
```
jonathanmendez, line 107: I don't know if any aspect of these visits are mocked or otherwise controlled in the test harness, but if we do have firm knowledge of what to expect for 'provider' and 'list' for the sites being tested, we should check the specific values being returned instead of just "it's a string".

jonathanmendez (changes requested):
Looks structurally good overall, but there are a few things to follow up on.

browser/actors/BlockedSiteParent.sys.mjs
gcp: I don't know what the default is for the others but it should at least be consistent.

browser/actors/BlockedSiteParent.sys.mjs
```diff
@@ -120,6 +121,90 @@ export class BlockedSiteParent extends EscapablePageParent {
...
+   * @returns {string|null} The processed URL, or null when logging is disabled
+   *   or the URL cannot be parsed.
+   */
+  _processTelemetryUrl(sourceUrl) {
+    if (!sourceUrl) {
+      return null;
+    }
+
+    const policy = Services.prefs.getCharPref(
+      "browser.safebrowsing.enterprise.telemetry.unsafeSiteVisit.urlLogging",
```
gcp: @jonathanmendez I guess the idea is that we then make this a part of https://bugzilla.mozilla.org/show_bug.cgi?id=2016141 ?

jonathanmendez: Yes, that's where it should end up. I'll make a note there and connect the bugs.

browser/actors/BlockedSiteChild.sys.mjs
mz-darnold: I think this should be resolved for both unsafe site visits and downloads now?

gcp:
Combined human/Claude review:

```
Overall: Prefs, Glean event metadata, defaults, and tests match the existing enterprise security telemetry well. C++ placement is the right call. The findings below are the actionable ones after discussion.

1. (High) No enterprise Policy wiring

Skip that for now, do this in https://bugzilla.mozilla.org/show_bug.cgi?id=2016141

2. (High) Ping volume at the classifier layer

RecordUnsafeSiteVisit fires from nsChannelClassifier::OnClassifyComplete for every hit — top-level, subframe, and subresource — and calls Enterprise.Submit() each time. The existing features never hit this because they record one layer up (WebsiteFilter at the document channel, about:blocked at the error-page layer), so they're naturally bounded to navigations. Per-record submission itself matches precedent; the concern is purely the higher-frequency layer. Since we're keeping subresource coverage, coalesce ping submission rather than submit-per-hit. (unsafe_download is naturally low-frequency — no change needed there.)

3. (Medium) Factor the duplicated C++ recording logic

The enabled-check → urlLogging (full/domain/none) → GetHost/GetSpec → record → gated Enterprise.Submit() sequence is duplicated near-verbatim in nsChannelClassifier.cpp and ApplicationReputation.cpp. Pull it into one shared helper (pref prefix + extra struct in; redaction + submit handled once) so the policy can't drift between the two events. This also resolves the inconsistency where the download side ignores GetHost's nsresult while the site-visit side checks NS_SUCCEEDED.

4. (Medium) Rename the Glean category security → safebrowsing

security is a generic top-level category shared across all of Firefox — collision-prone and it reads like a global metric. Both events are Safe Browsing hits and the prefs live under browser.safebrowsing.enterprise.telemetry.*, so safebrowsing is the specific, coherent choice, still shared across both metrics.yaml files. (No need to encode "enterprise" — the enterprise ping already carries that.)

5. (Low) Add a referrer extra key

unsafe_site_visit currently logs only the blocked subresource an admin can't see which page embedded it. WebsiteFilter already logs referrer (originalReferrer.spec); the channel/query has it available on both sides. Add a referrer extra key subject to the same urlLogging redaction.
```

toolkit/components/enterprise/EnterpriseTelemetry.cpp
```diff
@@ -0,0 +1,98 @@
...
+static TimeStamp sLastEnterprisePingTime;
+static uint64_t sLastEnterpriseBrowserId = 0;
+
+bool EventReportingEnabled(const nsACString& aPrefPrefix) {
+  nsAutoCString pref(aPrefPrefix);
+  pref.AppendLiteral(".enabled");
+  return Preferences::GetBool(pref.get(), true);
+}
+
+void RedactUrl(const nsACString& aPrefPrefix, nsIURI* aURI,
```
jonathanmendez: nit: Since this does not always redact, consider renaming (e.g. "MaybeRedactUrl" or "RedactUrlAsNecessary").

toolkit/components/enterprise/EnterpriseTelemetry.cpp
```diff
@@ -0,0 +1,98 @@
...
+               nsACString& aResult) {
```
jonathanmendez: nit: Consider renaming the return value to a more specific `aProcessedUrl` or similar

toolkit/components/enterprise/EnterpriseTelemetry.cpp
```diff
@@ -0,0 +1,98 @@
...
+    nsAutoCString host;
+    if (NS_SUCCEEDED(aURI->GetHost(host))) {
+      aResult = host;
+    }
+  } else {
+    NS_GetSanitizedURIStringFromURI(aURI, aResult);
+  }
+}
+
+EnterprisePingAction ThrottleEnterprisePing(uint64_t aBrowserId) {
```
jonathanmendez: I think there are several issues with the throttling approach here:

1. I don't think this belongs in a shared "Enterprise Telemetry" space. The throttling we have implemented is for very specific issues with a couple of events, and any throttling should stay specific to those events rather than "any security event that happens in the same browserId within a window". In addition the pref that is checked for the RecordOnly case is safebrowsing specific, further indicating this belongs somewhere else.
2. Is this trying to solve the same use case as blockedUrl? In that situation, we are working around a Firefox-specific implementation issue where we detect a blocked URL twice for the same user action (once for a speculative load and once for a real load). I don't actually love the cache solution I built for that, and would like to replace it with a better approach that can distinguish between the two loads. This implementation seems even less fit for that purpose because it does not distinguish e.g. different URLs loaded in the same tab, which are not the same user action.
3. However, the comment indicates that this may be trying to solve a different problem? Supposedly a burst could happen in the same tab from various subresources. Have you tested whether that actually happens or is that LLM speculation about a theoretical event that doesn't practically happen? I think understanding the real scenarios may help clarify what approach to take here (for example, I don't know all the things that might count as a 'subresource' in this context, but as an admin I would want a page with 3 separate blocked iframes to report all 3 and not just the first one that happened to be detected).
4. Related to the above issues, if the browserId continues to be used as a cache distinguisher, I think that `0` should be special cased and never dropped, since things like downloads "being from no tab" don't fit either reason for dropping events.

toolkit/components/enterprise/EnterpriseTelemetry.cpp
```diff
@@ -0,0 +1,98 @@
...
+  if (withinCooldown && aBrowserId == sLastEnterpriseBrowserId) {
+    return EnterprisePingAction::Drop;
+  }
+
+  sLastEnterprisePingTime = now;
+  sLastEnterpriseBrowserId = aBrowserId;
+  return EnterprisePingAction::RecordAndSubmit;
+}
+
+void SubmitEnterprisePing() {
```
jonathanmendez: I don't think this currently is a useful helper method. However, if the various events that want to disable submission for testing purposes consolidated to use the same single pref (e.g. "browser.enterprise.securityEvents.disableSubmitForTesting"), then it could be useful to put that check here and have the other events call the helper (so the check doesn't have to be repeated at all call sites).

toolkit/components/enterprise/EnterpriseTelemetry.cpp
```diff
@@ -0,0 +1,98 @@
...
+#  include "nsThreadUtils.h"
+
+namespace mozilla::enterprise {
+
+// Time and originating tab of the last submitted enterprise ping, used to
+// throttle submissions.
+static TimeStamp sLastEnterprisePingTime;
+static uint64_t sLastEnterpriseBrowserId = 0;
+
+bool EventReportingEnabled(const nsACString& aPrefPrefix) {
```
jonathanmendez: The `EventReportingEnabled` and `RedactUrl` functions (and a potentially modified `SubmitEnterprisePing` function) are genuinely useful helpers to share among the security events; however this PR does not update any of the other events to use them.
1. Please either update the PR to make the other events use the shared helpers, or file a follow-up Bug to do so.
2. To make these helpers useful to all call sites, some of which are C++ and some of which are js, it will be necessary to make an XPIDL component

If you do tackle this now, consider consulting with @1rneh in case she has any thoughts in context of Bug-2016141

browser/components/safebrowsing/content/test/browser_enterprise_unsafe_site_telemetry.js
```diff
@@ -0,0 +1,398 @@
...
+      BrowserTestUtils.removeTab(tab);
+      Services.fog.testResetFOG();
+    }
+  }
+});
+
+add_task(async function test_records_for_subframe_load() {
+  // A safe top-level page that embeds an unsafe iframe. The old about:blocked
+  // hook only reported top-level blocks, so this produced no telemetry; the
+  // classifier-level hook records it regardless of frame level.
```
jonathanmendez: Remove this claude-records-the-history-of-your-local-changes comment (and any others throughout the PR).

netwerk/url-classifier/nsChannelClassifier.cpp
```diff
@@ -381,6 +390,98 @@ nsresult nsChannelClassifier::SendThreatHitReport(nsIChannel* aChannel,
   return NS_OK;
 }
 
+#ifdef MOZ_ENTERPRISE
+// Records an enterprise security event for every Safe Browsing hit (top-level,
+// subframe, or subresource) so administrators can monitor unsafe-site access.
+// The enabled-check, URL redaction and (throttled) ping submission use the
+// shared mozilla::enterprise helpers so the policy stays in sync with other
+// enterprise security events.
```
jonathanmendez: Remove this part of the comment that isn't relevant to understanding `RecordUnsafeSiteVisit` (and is currently false since the other events aren't using the helpers).
```suggestion
```

toolkit/components/reputationservice/ApplicationReputation.cpp
```diff
@@ -1485,6 +1489,71 @@ nsresult PendingLookup::DoLookupInternal() {
   return LookupNext();
 }
 
+#ifdef MOZ_ENTERPRISE
+// Records an enterprise security event whenever download protection flags a
+// download as unsafe so administrators can monitor unsafe downloads. Recording
+// is independent of the block prefs so a detection is reported even when the
+// corresponding block pref is disabled. The enabled-check, URL redaction and
+// (throttled) ping submission use the shared mozilla::enterprise helpers so the
+// policy stays in sync with other enterprise security events.
```
jonathanmendez:
```suggestion
// corresponding block pref is disabled.
```

toolkit/components/reputationservice/ApplicationReputation.cpp
```diff
@@ -1485,6 +1489,71 @@ nsresult PendingLookup::DoLookupInternal() {
...
+    default:
+      // VERDICT_SAFE or unknown; nothing to record.
+      return;
+  }
+
+  // Downloads have no originating tab, so pass the 0 sentinel: they share the
+  // enterprise cooldown window but are never dropped on account of another
+  // tab's Safe Browsing hits.
+  const mozilla::enterprise::EnterprisePingAction action =
+      mozilla::enterprise::ThrottleEnterprisePing(0);
```
jonathanmendez: As mentioned in my comment about the throttle implementation, I would guess that this event should not throttle at all.

jonathanmendez (changes requested):

toolkit/components/enterprise/EnterpriseTelemetry.cpp
mz-darnold: Removed it.

toolkit/components/enterprise/EnterpriseTelemetry.h
```diff
@@ -0,0 +1,70 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef mozilla_EnterpriseTelemetry_h
+#define mozilla_EnterpriseTelemetry_h
+
+#ifdef MOZ_ENTERPRISE
```
1rneh: Checking the `MOZ_ENTERPRISE` build flag inside of `toolkit/components/enterprise` is redundant since the entire `enterprise`component is behind the `MOZ_ENTERPRISE` build flag, see https://searchfox.org/enterprise-main/rev/bd43ca06db2530f304cfaf030a847e814611078e/toolkit/components/moz.build#160-161

toolkit/components/enterprise/EnterpriseTelemetry.cpp
1rneh: OK, I'm jumping into this with very little involvement in our existing security events and the related policy controls. I've just had a look at Bug-2016141 (note this bug was intended to ensure documentation coverage only) and I've filed Bug-2058036 (blocks Bug-2016141) to track adding the general `SecurityLogging` policy.

With the context I have now, I'd suggest this PR to leave out any bigger security event refactors and simply read a preference that controls whether to send the event, just like the other security event related policies (see [`BlocklistDomainBrowsedTelemetry`](https://searchfox.org/enterprise-main/rev/bd43ca06db2530f304cfaf030a847e814611078e/browser/components/downloads/DownloadsTelemetry.enterprise.sys.mjs#51-61)).

As we'll be combining all security events policy into one in Bug-2058036, it makes more sense to me to revisit implementing security event helper functions and migrate our existing security event implementations as part of this bug..

mz-darnold: Leaving out making other events use the helpers. It can go in its own bug.

toolkit/components/url-classifier/metrics.yaml
1rneh: Same here.

```suggestion
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=TBD
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=TBD
```

toolkit/components/reputationservice/metrics.yaml
1rneh: So far for all enterprise metrics, we've put TBD. This is updated once we upstream and actually receive a data review.

```suggestion
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=TBD
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=TBD
```

1rneh (changes requested):
Thanks for the update! 

`EventReportingEnabled` and `MaybeRedactUrl` have moved but are still written in a generic way. Can we remove the generality  and pref building and tie each copy to its specific event. It could look like this:

Unsafe Site Visit in `nsChannelClassifier.cpp`:
- `browser.safebrowsing.enterprise.telemetry.unsafeSiteVisit.enabled` read in `ShouldRecordUnsafeSiteVisit` or just inline
- `browser.safebrowsing.enterprise.telemetry.unsafeSiteVisit.urlLogging` read in `RedactUnsafeSiteVisitUrl` or just inline

Unsafe Download in `ApplicationReputation.cpp`:
- `browser.safebrowsing.enterprise.telemetry.unsafeDownload.enabled` read in e.g. `ShouldRecordUnsafeDownload` or just inline
- `browser.safebrowsing.enterprise.telemetry.unsafeDownload.urlLogging` read in `RedactUnsafeDownloadUrl` or just inline

1rneh (approved):
Looks good to me now, thanks!

jonathanmendez (approved):

jonathanmendez:
BTW @mz-darnold make sure to use "Squash and merge" for completing this PR, not "Create a merge commit"

PR #1048, opened 2026-06-16
Bug 2040381 - Show error state in Access Connector urlbar button and panel when proxy is unavailable

jporter-dev:
### Description 

Bugzilla: Bug-2040381

This is part 2 of 3 for this bug. When the Access Connector proxy is unavailable and Firefox shows the `accessConnectorFailure` error page (implemented in #1033), the urlbar button now reflects the error state. The icon swaps to an "unavailable" variant displayed in a `critical` color, the tooltip and label update, and the Access Connector panel displays dedicated error content.

---

### Dependencies / Related Issues 

* [x] Depends on: #683
* [x] Part 2 of: #1033

---

### Screenshots 

Screenshot 2026-06-16 at 1 04 48 PM

_Access Connector in active state_

access-connector-unavailable

_Access Connector in error state_

---

### Testing 

* [x] Added tests
* [x] Manual testing performed

**Steps to verify changes:**
1. Enable the `AccessConnector` policy with a proxy server that is unreachable (ex: change `Port` by 1)
2. Navigate to a site covered by the inclusion patterns
3. Observe the Access Connector urlbar button and panel in an error state

**Expected result:**

The Access Connector error button renders with an error icon, and the panel title and message show error text.

PR #1121, opened 2026-07-07
Bug 2040381 - Register security event for Access Connector connection error

jporter-dev:
### Description 

Bugzilla: Bug-2040381

Part 3 of 3. Adds a security event that records when the Access Connector enters an error state because a site failed to load due to a proxy connection error. 

- Registers a new `access_connector.proxy_error` event metric with the domain as additional metadata, sent in the `enterprise` ping.
- Since Access Connector is an Enterprise-only feature, the new metric is added in `enterprise/metrics.yaml` using the `Enterprise Products :: Firefox` tag.
- Metric is uses `data_sensitivity: highly_sensitive` since it sends the browsed domain as an extra key.
- Records the event once per error transition. Adds a guard to prevent re-firing on subsequent `update` calls for the same domain while the page remains in the error state.

> Note: this is stacked on top of #1048. Relevant commit is `Bug 2040381 - Register telemetry metric for Access Connector proxy error`.

---

* Depends on: #1048 
* Part 3 of: #1033 

---

### Testing 

**Steps to verify changes:**

Follow steps outlined in #1048 

**Expected result:**

`access_connector.proxy_error` event is submitted to console. Verifiable in either the Browser Toolbox network tab or in the console itself.

### DIFF
--- a/browser/components/enterprise/AccessConnectorButton.sys.mjs
+++ b/browser/components/enterprise/AccessConnectorButton.sys.mjs
@@ -31,6 +31,7 @@ export class AccessConnectorButton {
   #window = null;
   #progressListener = null;
   #onClick = null;
+  #lastErrorDomain = null;
 
   /**
    * @param {Window} window - The chrome window that owns the button.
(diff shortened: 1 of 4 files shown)


browser/components/enterprise/content/enterprise-urlbar-actions.css
```diff
@@ -2,12 +2,22 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+:root {
+  --lockdown-mode-icon: url("chrome://browser/content/enterprise/site-restrictions.svg");
+  --access-connector-icon: url("chrome://browser/content/enterprise/access-connector.svg");
+  --access-connector-error-icon: url("chrome://browser/content/enterprise/access-connector-unavailable.svg");
+}
```
lissyx: we have `browser/branding/enterprise/content/felt-tokens.css`, maybe we could have another one for browser enterprise things?
e.g., `browser/branding/enterprise/content/browser-tokens.css`

jporter-dev: Good idea, I can add this for these and any future tokens for non-FELT UX

jporter-dev: I'll add this to #1048 since that's the UX component

1rneh (approved):
Lgtm!

PR #1126, opened 2026-07-08
Bug 2050550 - Use leaf name without path for remoting.

mkaply:
### Description

Bugzilla: Bug-2050550 

Change window remoting name to not use the path (since it can change).


toolkit/xre/nsAppRunner.cpp
lissyx: I think it may help to make it more obvious to people reading this code later that this is under the assumptions that:
 - the felt profile does not really contain anything important, that is why it lives under a temp directory
 - the directory name itself is hardcoded and includes the update channel in its name, limiting the risks of collisions between different branches running in parallel

lissyx (approved):
r=me

PR #1129, opened 2026-07-09
Bug 2037079 - Add support for captive portal

mkaply:
### Description

Bugzilla: Bug-2037079

This patch updates our update/login flow to support captive portals

Screenshots (I have not received a design for this. I have requested it)

Screenshot 2026-07-10 070358

Screenshot 2026-07-10 070410

### DIFF
--- a/toolkit/components/enterprise/modules/CaptivePortal.sys.mjs
+++ b/toolkit/components/enterprise/modules/CaptivePortal.sys.mjs
@@ -1,0 +1,252 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  FeltCommon: "chrome://felt/content/FeltCommon.sys.mjs",
+  FeltErrorReport: "resource://gre/modules/enterprise/FeltErrorReport.sys.mjs",
+});
+
+/**
+ * Pre-authentication captive-portal sign-in for FELT. Shows a
+ * banner rather than auto-opening anything, and hosts the portal in an
+ * embedded browser rather than a second top-level window (which would disturb
+ * FELT's window-lifecycle bookkeeping). Dismissal follows CaptivePortalService's
+ * connectivity signal, not the portal's navigation.
+ *
+ * Owns the captive-portal parts of the FELT login DOM (the banner and the
+ * portal browser) and drives them by selector off the document passed to
+ * init(). It does not touch the sign-in flow itself: the FELT window injects
+ * callbacks (resetLoginUi, onConnectivityRestored, suspend/resumeUpdates) so
+ * this module never calls into window.js's sign-in logic directly.
+ */
+export const CaptivePortal = {
+  _cps: Cc["@mozilla.org/network/captive-portal-service;1"].getService(
+    Ci.nsICaptivePortalService
+  ),
+
+  // Document of the FELT window, provided at init().
+  _doc: null,
+
+  // Injected by init():
+  //   onConnectivityRestored() - network is back; FELT may resume sign-in.
+  //   resetLoginUi()          - flip the UI back to the login form (aborting any
+  //                             in-flight SSO), keeping a pending sign-in.
+  //   suspendUpdates()        - stop the startup update check behind a portal.
+  //   resumeUpdates()         - re-run the update check once connectivity is back.
+  _onConnectivityRestored: null,
+  _resetLoginUi: null,
+  _suspendUpdates: null,
+  _resumeUpdates: null,
+
+  // True while we've suspended the update check for a portal, so we only resume
+  // a check we actually stopped (and never disturb an in-progress one).
+  _updatesSuspended: false,
+
+  get canonicalURL() {
+    return Services.prefs.getCharPref("captivedetect.canonicalURL");
+  },
+
+  /**
+   * @param {Document} doc - The FELT window document.
+   * @param {object} [hooks]
+   * @param {Function} [hooks.onConnectivityRestored]
+   * @param {Function} [hooks.resetLoginUi]
+   * @param {Function} [hooks.suspendUpdates]
+   * @param {Function} [hooks.resumeUpdates]
+   */
+  init(
+    doc,
+    { onConnectivityRestored, resetLoginUi, suspendUpdates, resumeUpdates } = {}
+  ) {
+    this._doc = doc;
+    this._onConnectivityRestored = onConnectivityRestored;
+    this._resetLoginUi = resetLoginUi;
+    this._suspendUpdates = suspendUpdates;
+    this._resumeUpdates = resumeUpdates;
+
+    Services.obs.addObserver(this, "captive-portal-login");
+    Services.obs.addObserver(this, "captive-portal-login-success");
+    Services.obs.addObserver(this, "captive-portal-login-abort");
+    Services.obs.addObserver(this, "network:captive-portal-connectivity");
+    this._doc.defaultView.addEventListener("unload", () => this.uninit(), {
+      once: true,
+    });
+
+    this._doc
+      .getElementById("felt-open-network-login")
+      .addEventListener("click", () => this.showPortalBrowser());
+
+    // Already behind a portal at startup: stop the update check and surface the
+    // banner. Otherwise let the update check (already running) proceed; if a
+    // portal turns up mid-check our observer interrupts it.
+    if (this._cps.state == this._cps.LOCKED_PORTAL) {
+      this.showBanner();
+    } else if (this._cps.state == this._cps.UNKNOWN) {
+      // Probe so a portal surfaces promptly rather than as an update failure.
+      this._cps.recheckCaptivePortal();
+    }
+  },
+
+  // Track that we suspended, so we only ever resume a check we stopped.
+  _suspendUpdatesForPortal() {
+    if (!this._updatesSuspended) {
+      this._updatesSuspended = true;
+      this._suspendUpdates?.();
+    }
+  },
+
+  // Resume the suspended check once connectivity is confirmed and no sign-in is
+  // holding it off. Only clears the flag if the check actually ran, so a resume
+  // declined for a pending sign-in is retried when that sign-in ends rather than
+  // lost. _cps.state is safe to read here (not the login observer's racy path).
+  maybeResumeUpdates() {
+    const state = this._cps.state;
+    const connected =
+      state == this._cps.NOT_CAPTIVE || state == this._cps.UNLOCKED_PORTAL;
+    if (this._updatesSuspended && connected && this._resumeUpdates?.()) {
+      this._updatesSuspended = false;
+    }
+  },
+
+  uninit() {
+    Services.obs.removeObserver(this, "captive-portal-login");
+    Services.obs.removeObserver(this, "captive-portal-login-success");
+    Services.obs.removeObserver(this, "captive-portal-login-abort");
+    Services.obs.removeObserver(this, "network:captive-portal-connectivity");
+    this._onConnectivityRestored = null;
+    this._resetLoginUi = null;
+    this._suspendUpdates = null;
+    this._resumeUpdates = null;
+    this._doc = null;
+  },
+
+  // Re-probe so a real portal surfaces as the banner, not a dead-end error.
+  recheck() {
+    this._cps.recheckCaptivePortal();
+  },
+
+  observe(subject, topic) {
+    switch (topic) {
+      case "captive-portal-login":
+        this.showBanner();
+        break;
+      case "captive-portal-login-success":
+        // DNS lookups made while captive may be cached stale; flush so console
+        // requests resolve freshly now that we're connected.
+        Services.dns.clearCache(true);
+        this.hidePortalBrowser();
+        this.hideBanner(true);
+        this.maybeResumeUpdates();
+        break;
+      case "captive-portal-login-abort":
+        // Keep updates suspended until connectivity is confirmed. Abort from
+        // CaptivePortalService::Stop() self-heals: its Start() re-probe fires
+        // network:captive-portal-connectivity, which resumes us.
+        this.hidePortalBrowser();
+        this.hideBanner(false);
+        break;
+      case "network:captive-portal-connectivity":
+        this.maybeResumeUpdates();
+        break;
+    }
+  },
+
+  showBanner() {
+    // Suspend the update check so it doesn't error behind the portal; this also
+    // restores the login pane for the banner to sit on.
+    this._suspendUpdatesForPortal();
+
+    // Clear any stray connection-error bar from a failed posture/SSO attempt.
+    lazy.FeltErrorReport.reset();
+
+    // Stop any in-flight SSO load and flip back to the login form. resetLoginUi()
+    // is the shared reset (also clears the back button/status panel) and keeps
+    // the pending sign-in so it resumes once the portal clears.
+    this._doc.getElementById("browser").stop();
+    this._resetLoginUi?.();
+
+    this._doc
+      .querySelector(".felt-browser-error-captive-portal")
+      .classList.remove("is-hidden");
+  },
+
+  hideBanner(restored) {
+    this._doc
+      .querySelector(".felt-browser-error-captive-portal")
+      .classList.add("is-hidden");
+
+    // Network is back: let the FELT window decide whether to resume a sign-in.
+    if (restored) {
+      this._onConnectivityRestored?.();
+    }
+  },
+
+  showPortalBrowser() {
+    const url = this.canonicalURL;
+    // Exempt the plaintext probe from HTTPS upgrade. HTTPS-First is on by default
+    // in private browsing and this browser is private, so the http probe would be
+    // upgraded and a portal couldn't intercept it. pbId must be 1 to match. Same
+    // as desktop's browser-captivePortal.js.
+    const uri = Services.io.newURI(url);
+    const principal = Services.scriptSecurityManager.createContentPrincipal(
+      uri,
+      { privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID }
+    );
+    Services.perms.addFromPrincipal(
+      principal,
+      "https-only-load-insecure",
+      Ci.nsIPermissionManager.ALLOW_ACTION,
+      Ci.nsIPermissionManager.EXPIRE_SESSION
+    );
+
+    // Hide the banner: it sits outside .felt-login, so the portal overlay
+    // won't cover it.
+    this._doc
+      .querySelector(".felt-browser-error-captive-portal")
+      .classList.add("is-hidden");
+
+    this._doc
+      .querySelector(".felt-login__portal")
+      .classList.remove("is-hidden");
+
+    const browser = this._doc.getElementById("portal-browser");
+    // Browser starts in-process (no remote="true" in markup, saving a startup
+    // content process); DocumentLoadListener switches it to a content process
+    // from the URL (gated on the markup's maychangeremoteness), so untrusted
+    // portal content stays isolated.
+
+    // Untrusted third-party content: null principal carrying the private
+    // browsing id so its origin attributes match the docshell.
+    browser.fixupAndLoadURIString(url, {
+      triggeringPrincipal: Services.scriptSecurityManager.createNullPrincipal({
+        privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID,
+      }),
+    });
+    browser.focus();
+  },
+
+  hideOverlay() {
+    this._doc.querySelector(".felt-login__portal").classList.add("is-hidden");
+  },
+
+  hidePortalBrowser() {
+    const browser = this._doc.getElementById("portal-browser");
+    this.hideOverlay();
+    // The portal shares the private session (pbId=1) with SSO, and FELT collects
+    // every cookie in it, so drop the portal's cookies on teardown. Clearing the
+    // whole jar is safe: an SSO the portal interrupted was aborted and restarts
+    // clean, so there are no console cookies to lose.
+    Services.cookies.removeCookiesWithOriginAttributes(
+      JSON.stringify({ privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID })
+    );
+    // Discard the portal page; dismissal follows the connectivity signal.
+    browser.fixupAndLoadURIString("about:blank", {
+      triggeringPrincipal: Services.scriptSecurityManager.createNullPrincipal({
+        privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID,
+      }),
+    });
+  },
+};
--- a/toolkit/components/felt/content/felt.xhtml
+++ b/toolkit/components/felt/content/felt.xhtml
@@ -190,6 +190,19 @@
                 maychangeremoteness="true"
               ></browser>
             </div>
+            <div class="felt-login__portal is-hidden">
+              <!-- No remote="true": it would spin up a content process at every
+                   startup. Loading the portal URL switches it to one then
+                   (URL-driven, gated on maychangeremoteness here). -->
+              <browser
+                xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+                id="portal-browser"
+                type="content"
+                disablehistory="true"
+                disableglobalhistory="true"
+                maychangeremoteness="true"
+              ></browser>
+            </div>
           </div>
           <div id="felt-statuspanel" class="is-hidden">
             <span
@@ -242,6 +255,18 @@
             dismissable="true"
           >
             <span slot="message" class="felt-browser-error-details"></span>
+          </moz-message-bar>
+          <moz-message-bar
+            class="felt-browser-error-captive-portal is-hidden"
+            data-l10n-id="felt-browser-error-captive-portal"
+            type="info"
+          >
+            <moz-button
+              slot="actions"
+              id="felt-open-network-login"
+              data-l10n-id="felt-captive-portal-open-login"
+              size="small"
+            ></moz-button>
           </moz-message-bar>
           <moz-message-bar
             class="felt-browser-error-connection is-hidden"
--- a/toolkit/components/felt/content/window.js
+++ b/toolkit/components/felt/content/window.js
@@ -8,6 +8,7 @@ const lazy = {};
 
 ChromeUtils.defineESModuleGetters(lazy, {
   AppConstants: "resource://gre/modules/AppConstants.sys.mjs",
+  CaptivePortal: "resource://gre/modules/enterprise/CaptivePortal.sys.mjs",
   ConsoleClient: "resource://gre/modules/enterprise/ConsoleClient.sys.mjs",
   FeltCommon: "chrome://felt/content/FeltCommon.sys.mjs",
   FeltErrorReport: "resource://gre/modules/enterprise/FeltErrorReport.sys.mjs",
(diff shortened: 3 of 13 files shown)


toolkit/components/felt/content/window.js
```diff
@@ -440,6 +644,7 @@ window.addEventListener(
     setupPopupNotifications();
     setupContextMenu();
     listenFormEmailSubmission();
+    CaptivePortal.init();
```
lissyx: I think we should rather do things differently and make `CaptivePortal` the first thing we hit, return a promise and chain `.then()` for the rest of the Updates and SSO setup

toolkit/components/felt/content/window.js
```diff
@@ -440,6 +644,7 @@ window.addEventListener(
...
     focusEmailOnLoginVisible();
     informAboutPotentialStartupFailure();
     macosActivateApplication();
```
lissyx, line 590:
```suggestion
setBuildVersion();
lazy.FeltErrorReport.init(document);
setupMarionetteEnvironment();
setupPopupNotifications();
setupContextMenu();
informAboutPotentialStartupFailure();
macosActivateApplication();
CaptivePortal.init().then(
  lazy.Updates.init(document);
  listenFormEmailSubmission();
  focusEmailOnLoginVisible();
);
```

toolkit/components/felt/content/window.js
```diff
@@ -26,14 +26,205 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
 Services.obs.notifyObservers(window, "browser-delayed-startup-finished");
 
+/**
+ * Pre-authentication captive-portal sign-in for FELT (bug 2037079). Shows a
+ * banner rather than auto-opening anything, and hosts the portal in an
+ * embedded browser rather than a second top-level window (which would disturb
+ * FELT's window-lifecycle bookkeeping). Dismissal follows
+ * CaptivePortalService's connectivity signal, not the portal's navigation.
+ */
+const CaptivePortal = {
```
lissyx: Move this to a dedicated module `CaptivePortal.sys.mjs` under felt?

toolkit/components/felt/content/window.js
```diff
@@ -26,14 +26,205 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+    return Services.prefs.getCharPref("captivedetect.canonicalURL");
+  },
+
+  // Tracked locally rather than read from _cps.state: CaptivePortalService
+  // sets LOCKED_PORTAL in its own observer of "captive-portal-login", and
+  // observer ordering for a single topic isn't guaranteed, so our observer
+  // can run before the service's -- reading .state at that point would race.
+  isLocked: false,
+
+  init() {
```
lissyx: return a promise that will resolve when either there is no captive portal detected or it's successfully auth'd?

toolkit/components/felt/content/window.js
mkaply: I don't think chaining Updates/SSO off a CaptivePortal promise buys us anything, and it can hurt. The only thing worth gating on is the captive-portal probe finishing, and that adds a delay on every startup regardless of network health, which brings back a checking-to-login flash that happened when we did this. If the promise resolves before the probe instead, it's gating on nothing.

It also wouldn't let us drop any of the observer wiring. The portal can lock, unlock, or drop at any point while the window is open, so we need the captive-portal-login/success/abort observers regardless. A startup promise only covers the initial load, so it's a second coordination path on top of the one we already need.

There's no race to fix either. The load handler is synchronous, so nothing can fire between Updates.init() and CaptivePortal.init(). Both modules read cps.state themselves at launch (Updates already skips the check if it sees LOCKED_PORTAL), so the order doesn't actually matter.

If the concern is evaluating captivity before committing to the update UI, I'm happy to just move CaptivePortal.init() above Updates.init(). Same effect, no signature changes.

toolkit/components/felt/content/window.js
mkaply: The form wiring and focus shouldn't wait on the captive-portal probe. If listenFormEmailSubmission and focusEmailOnLoginVisible only run after CaptivePortal.init() resolves, the email field and sign-in button aren't hooked up until the probe comes back, so the user can be staring at a login form that doesn't respond yet.

It's worse with no network at all. If we're just offline with no portal, the probe never produces a result, so nothing fires and the promise never resolves. Updates.init, the form listener, and focus would never run and the login screen is stuck. That's exactly the case this is supposed to handle.

Gating Updates.init on it also brings back the checking-to-login delay on every startup, since the only thing worth waiting for there is the probe finishing.

We also still need the captive-portal-login/success/abort observers no matter what, because the portal can change state while the window is open. The promise only covers the initial load, so it doesn't let us remove any of that.

Everything in the handler runs synchronously today, so there's no ordering problem to solve. If you want captivity looked at before the update UI, I can move CaptivePortal.init() above Updates.init() and leave the rest as plain calls.

toolkit/components/felt/content/window.js
mkaply: We can, but those two conditions miss the third outcome: no network at all. If there's no portal and no connectivity, the service just keeps retrying quietly and never fires login-success or a "clear" result, so a promise that resolves on those two never resolves. Anything awaiting it hangs until connectivity shows up, which on these networks might be a while, or the whole session.

So it's only safe to await from something that's genuinely fine being deferred forever. Form wiring and focus aren't: the user should be able to type an email and see the field focused right away even with no network. Updates isn't either, it's built to start immediately and time out on its own. The calls that actually need connectivity, device posture and SSO, already handle the offline case with their own timeouts plus a recheck that brings up the banner.

If there's a specific step you think should wait for connectivity, point me at it and I'll look. I just don't think form, focus, or Updates can be the ones.

One implementation note if we do build it: we'd also have to start observing network:captive-portal-connectivity, since the no-portal "clear" result comes through there and today we only watch the login events.

lissyx: and pass the document as a parameter so we can extract this to a different module

```suggestion
  init(doc) {
```

mkaply:
I think a point is missed here.

Captivity isn't just that happens at startup. It can happen at any time during the session which is why we handle it with observers and let network calls timeout.

toolkit/components/felt/content/window.js
```diff
@@ -26,14 +26,205 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+      .querySelector(".felt-browser-error-captive-portal")
+      .classList.add("is-hidden");
+    updateSignInButtonState();
+
+    // Network is back: resume only an attempt the user actually submitted,
+    // not just whatever happens to be sitting in the email field.
+    const email = document.getElementById("felt-form__email").value;
+    if (restored && this._signInPending && email) {
+      connectToConsole(email);
+    }
```
lissyx: I really do not think `CaptivePortal` should interact with this in anyway, this would be fixed with the suggestion from above to move to a component and return a promise that we resolve when we have network available

toolkit/components/felt/content/window.js
```diff
@@ -26,14 +26,205 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+
+  showPortalBrowser() {
+    const url = this.canonicalURL;
+    // Keep https-only from upgrading the plaintext probe. privateBrowsingId
+    // must be 1: this browser's content is forced private by the FELT
+    // window's "private" chrome flag, and the permission must match.
+    const uri = Services.io.newURI(url);
+    const principal = Services.scriptSecurityManager.createContentPrincipal(
+      uri,
+      { privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID }
```
lissyx: this was a discussion we had during the workweek, so I'm not sure it works, but we should have a different ID to avoid sharing cookies between the private session of the portal and the SSO flow

toolkit/components/felt/content/window.js
mkaply: It can't work like that. This isn't just a startup thing that happens once and we're done. You use observers for this, you can't use promises (just like Firefox does).

toolkit/components/felt/content/window.js
lissyx: My major concern is how this is changing a lot of places in ways and interactions that are non trivial to track. I would really prefer that we had a way to have a clear separation.

Flashing is a non problem, it's solved on `Updates.sys.mjs`, we can have the exact same solution. And we will have a delay in any case, since we hit the network to check for updates. Gating so that we do not start anything until captive portal verified things are good should not be a problem.

toolkit/components/felt/content/window.js
lissyx: > The form wiring and focus shouldn't wait on the captive-portal probe. If listenFormEmailSubmission and focusEmailOnLoginVisible only run after CaptivePortal.init() resolves, the email field and sign-in button aren't hooked up until the probe comes back, so the user can be staring at a login form that doesn't respond yet.

We can show a spinner like we do for updates. "Checking network connectivity", and the flashing problem is solvable

> 
> It's worse with no network at all. If we're just offline with no portal, the probe never produces a result, so nothing fires and the promise never resolves. Updates.init, the form listener, and focus would never run and the login screen is stuck. That's exactly the case this is supposed to handle.

We can detect directly that we are offline

> 
> Gating Updates.init on it also brings back the checking-to-login delay on every startup, since the only thing worth waiting for there is the probe finishing.
> 
> We also still need the captive-portal-login/success/abort observers no matter what, because the portal can change state while the window is open. The promise only covers the initial load, so it doesn't let us remove any of that.

True, but then what is the UX here? Captive portal expires during firefox' life, do we want to bring back Felt?
It's going to block communication with the console, this may force shutdown badly.

> 
> Everything in the handler runs synchronously today, so there's no ordering problem to solve. If you want captivity looked at before the update UI, I can move CaptivePortal.init() above Updates.init() and leave the rest as plain calls.

mkaply:
***This is not ready for review again yet. Accidentally force pushed before I tested.***

toolkit/components/felt/content/window.js
mkaply: What do you think about just clearing cookies after portal login? There's no way to use different privateBrowsingIDs

lissyx: I was worried about the second part, thanks for confirming it.

Clearing is definitively something good, but from what I remember as long as we destroy the `browser` element the garbage collection should also get rid of them

toolkit/components/enterprise/modules/CaptivePortal.sys.mjs
lissyx, line 1: I think it should be under `toolkit/components/enterprise/modules`, we're trying to cleanup things from under `felt/content`

toolkit/components/felt/content/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+      .addEventListener("click", () => this.showPortalBrowser());
+
+    // Safety net: never let the gate wedge the login UI. Offline with no portal
+    // the service reports nothing, so connectivity is never confirmed; fall
+    // through after a timeout and let the flow proceed to the login UI. Armed
+    // before the state check below so the already-connected path clears it.
+    const timeoutMs = Services.prefs.getIntPref(
+      "enterprise.felt.captive_portal_gate_timeout_ms",
+      10000
+    );
```
lissyx: We have an API to check for network connectivity at all, no?

lissyx: https://searchfox.org/enterprise-main/rev/3c95c01fd1bd72c5b4f8538aa95d5c1cf67948eb/dom/webidl/NetworkInformation.webidl#11-26

lissyx: OK we may need it enabled for us: https://searchfox.org/enterprise-main/rev/3c95c01fd1bd72c5b4f8538aa95d5c1cf67948eb/modules/libpref/init/StaticPrefList.yaml#4038

lissyx: (it's coming back from B2G, hopefully it's not broken)

We could force enable it only in Felt to limit the scope

toolkit/components/felt/content/window.js
```diff
@@ -435,14 +475,42 @@ window.addEventListener(
...
     informAboutPotentialStartupFailure();
     macosActivateApplication();
+
+    // Gate the update check on captive-portal state being resolved, so we
+    // never start a network update check while stuck behind a portal. The gate
+    // falls through on a timeout, so a portal-free offline start still reaches
+    // the login UI.
+    lazy.CaptivePortal.init(document, {
+      onStateChanged: updateSignInButtonState,
+      abortSso: () => ssoAbort?.(),
```
lissyx: I wish we had the back button already done so this part would not be required

toolkit/components/felt/content/window.js
```diff
@@ -435,14 +475,42 @@ window.addEventListener(
...
+        // instead of starting an update check that can only fail.
+        lazy.FeltErrorReport.updateNetworkError(
+          "netOffline",
+          null,
+          lazy.ERROR_SOURCE.NET
+        );
+        return;
+      }
+      lazy.Updates.init(document);
+    });
```
lissyx, line 619: I need to check what is the behavior wrt this **after** browser has been launched: is there a risk those gets called when felt window is hidden?

toolkit/components/felt/content/felt.xhtml
```diff
@@ -208,6 +208,18 @@
...
+            data-l10n-id="felt-browser-error-captive-portal"
+            type="info"
+          >
+            <moz-button
+              slot="actions"
+              id="felt-open-network-login"
+              data-l10n-id="felt-captive-portal-open-login"
+              size="small"
+            ></moz-button>
+          </moz-message-bar>
```
lissyx, line 270: Please provide screenshots to ease the review (+ link to figma design if applicable)

toolkit/components/felt/content/felt.xhtml
```diff
@@ -255,6 +267,18 @@
...
+        <div class="felt-login__portal is-hidden">
+          <browser
+            xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+            id="portal-browser"
+            type="content"
+            disablehistory="true"
+            disableglobalhistory="true"
+            remote="true"
+            maychangeremoteness="true"
+          ></browser>
```
lissyx: I'm wondering if we should create this element dynamically, would it help ensuring cookies are discarded?

mkaply: It would not because it would be in the same browsing context (private browsing contexts share cookies).

So we would still have to clear cookies (which we do now).

toolkit/components/felt/content/CaptivePortal.sys.mjs
mkaply: It isn't implemented at all on desktop so it would not be helpful for this.

toolkit/components/felt/content/window.js
mkaply: It won't because they are in the private browsing cookie jar.

toolkit/components/felt/content/felt.xhtml
mkaply: Screenshots added to description

toolkit/components/felt/content/window.js
mkaply: The felt window isn't hidden, it's closed which fires unload and CaptivePortal's uninit removes all the observers. There's a test path that keeps it there but in production it would never happen.

toolkit/components/felt/content/CaptivePortal.sys.mjs
mkaply: I've changed things to just remove the timeout. Looking at the code more, the right things happen with regards to timing, so we don't need this,

lissyx: > It isn't implemented at all on desktop so it would not be helpful for this.

yes that was what I wanted to verify but I got dragged into other things, and it's not implemented
but `navigator.online` is, and seems to work, and can send an event `offline`: https://developer.mozilla.org/fr/docs/Web/API/Window/offline_event

mkaply: Knowing whether or not we are online or offline isn't really helpful information. I originally had this and all it does is tell you that you're offline quicker. It only handles the complete lack of a network connection.

The update check will immediately detect if you are offline.

lissyx: > I've changed things to just remove the timeout. Looking at the code more, the right things happen with regards to timing, so we don't need this,

There may still be value in directly showing a network error and handling https://developer.mozilla.org/fr/docs/Web/API/Window/online_event as well ?

It could be a follow up, even though it makes sense in this PR

mkaply: So you want a network error immediately if we are not online? I had that in my patch, but it didn't seem terribly helpful. You get it as soon as you click continue. I'll add it if you want.

lissyx: > So you want a network error immediately if we are not online? I had that in my patch, but it didn't seem terribly helpful. You get it as soon as you click continue. I'll add it if you want.

I agree, I was wondering if there's also a justification to handle the "online" event to trigger update back ? But this start to feel out of scope, so if it makes sense to you, better in a follow up to not block this PR further

toolkit/components/felt/content/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+      // UNKNOWN: probe so we resolve (or surface a banner) promptly.
+      this._cps.recheckCaptivePortal();
+    }
+
+    return this._ready;
+  },
+
+  // offline: true when the gate fell through the safety timeout without any
+  // connectivity being confirmed, i.e. no usable network.
+  _markReady(offline = false) {
```
lissyx: nit: `offline` here can get confusing, we should use a better description. Maybe `networkUsable` is a better description of the state

lissyx: nevermind, it's really an offline state here.

toolkit/components/enterprise/modules/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+        );
+        this.hidePortalBrowser();
+        this.hideBanner(true);
+        // Connectivity restored: release the startup gate.
+        this._markReady();
+        break;
+      case "captive-portal-login-abort":
+        this.isLocked = false;
+        this.hidePortalBrowser();
+        this.hideBanner(false);
```
lissyx, line 149: I'm tempted to say we should clear cookies there as well?

toolkit/components/felt/content/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+        this._markReady();
+        break;
+      case "captive-portal-login-abort":
+        this.isLocked = false;
+        this.hidePortalBrowser();
+        this.hideBanner(false);
+        break;
+      case "network:captive-portal-connectivity":
+        // Fired only once a probe confirms connectivity (no portal, or a
+        // portal just cleared), so it's safe to release the startup gate.
```
lissyx: So we can hit this **and** `"captive-portal-login-success"` from above? That is what I understand form "portal just cleared" ?

lissyx: Checking `_markReady()` looks to be idempotent so it should be fine?

toolkit/components/felt/content/window.js
```diff
@@ -26,14 +27,39 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+// because it belongs to the sign-in flow, not to captive-portal detection.
+let signInPending = false;
+let ssoAbort = null;
+
 async function connectToConsole(email) {
+  signInPending = true;
   let posture;
+  // ConsoleClient's XHR has no built-in timeout, so a portal that silently
+  // drops packets could hang this indefinitely. Abort (not just race) so no
+  // connection is left open in the background.
```
lissyx: Is this really true? https://searchfox.org/enterprise-main/rev/3c95c01fd1bd72c5b4f8538aa95d5c1cf67948eb/toolkit/components/enterprise/modules/ConsoleClient.sys.mjs#284

toolkit/components/felt/content/window.js
```diff
@@ -26,14 +27,39 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+  const devicePostureTimeoutMs = Services.prefs.getIntPref(
+    "enterprise.felt.device_posture_timeout_ms",
+    15000
+  );
+  const devicePostureController = new AbortController();
+  const devicePostureTimeout = setTimeout(
+    () => devicePostureController.abort(),
+    devicePostureTimeoutMs
+  );
```
lissyx: I assume we want to handle this here because of the usecase of reconnecting after launching a first time, and captive portal may have blocked us by then?

Because on first run, we are gated anyway so we would not reach this

toolkit/components/felt/content/window.js
```diff
@@ -26,14 +27,39 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
   try {
-    posture = await lazy.ConsoleClient.sendDevicePosture();
+    posture = await lazy.ConsoleClient.sendDevicePosture({
+      signal: devicePostureController.signal,
+    });
   } catch (err) {
     lazy.log.error(`Failed to send device posture: ${err}`);
+    // Re-probe so a real portal surfaces as the banner, not a dead-end error.
+    lazy.CaptivePortal.recheck();
     await lazy.FeltErrorReport.handleXhrError(err);
```
lissyx, line 131: Don't we risk that we show both a captive portal banner here, as well as the XHR error reported?
The underlying `recheckCaptivePortal` does not return a value, everything is bound to the observers setup in place, and we rely on `showBanner()` to perform the `FeltErrorReport.reset()` call. While unlikely there's still a window of potential race (and I could trigger a lot of races around handling network / XHR errors).

This may lead to weird UX, but I'm not sure it's worth the complication to go further in handling it

toolkit/components/felt/content/window.js
lissyx: I mistyped, I was thinking about the case where it's closed indeed. But `uninit` is called on `unload` so it should be fine

toolkit/components/felt/content/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+
+  uninit() {
+    Services.obs.removeObserver(this, "captive-portal-login");
+    Services.obs.removeObserver(this, "captive-portal-login-success");
+    Services.obs.removeObserver(this, "captive-portal-login-abort");
+    Services.obs.removeObserver(this, "network:captive-portal-connectivity");
+    if (this._readyTimeout) {
+      lazy.clearTimeout(this._readyTimeout);
+      this._readyTimeout = null;
+    }
```
lissyx: Out of safety we may want to `null` back everything:
```
  _onStateChanged: null,
  _onConnectivityRestored: null,
  _abortSso: null,
```

toolkit/components/enterprise/modules/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+    const browser = this._doc.getElementById("portal-browser");
+    this._doc.querySelector(".felt-login__portal").classList.add("is-hidden");
+    // Discard whatever the portal was showing; dismissal is driven by the
+    // connectivity signal, not portal navigation.
+    browser.fixupAndLoadURIString("about:blank", {
+      triggeringPrincipal: Services.scriptSecurityManager.createNullPrincipal(
+        {}
+      ),
+    });
+  },
```
lissyx, line 251: This needs to be called from `FeltErrorReport.update()` to make sure the portal element does not block viewing the errors that may be reported (context: `about:certerror?e=nssBadCert` triggered locally)

toolkit/components/felt/content/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+
+  hidePortalBrowser() {
+    const browser = this._doc.getElementById("portal-browser");
+    this._doc.querySelector(".felt-login__portal").classList.add("is-hidden");
+    // Discard whatever the portal was showing; dismissal is driven by the
+    // connectivity signal, not portal navigation.
+    browser.fixupAndLoadURIString("about:blank", {
+      triggeringPrincipal: Services.scriptSecurityManager.createNullPrincipal(
+        {}
+      ),
```
lissyx: This needs to use the `privateBrowsingId` otherwise we hit `[2117153] Assertion failure: attrs == aOriginAttributes, at docshell/base/nsDocShell.cpp:9197` (more obvious when doing the change to call from `FeltErrorReport` below)

toolkit/components/felt/content/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+    browser.setAttribute(
+      "remoteType",
+      ChromeUtils.predictRemoteTypeForURI(url, { browser })
+    );
+
+    // Untrusted third-party content: use a null principal.
+    browser.fixupAndLoadURIString(url, {
+      triggeringPrincipal: Services.scriptSecurityManager.createNullPrincipal(
+        {}
+      ),
```
lissyx: same, for consistency we need to use the `privateBrowsingId`

toolkit/components/enterprise/modules/CaptivePortal.sys.mjs
```diff
@@ -0,0 +1,281 @@
...
+    const principal = Services.scriptSecurityManager.createContentPrincipal(
+      uri,
+      { privateBrowsingId: lazy.FeltCommon.PRIVATE_BROWSING_ID }
+    );
+    Services.perms.addFromPrincipal(
+      principal,
+      "https-only-load-insecure",
+      Ci.nsIPermissionManager.ALLOW_ACTION,
+      Ci.nsIPermissionManager.EXPIRE_SESSION
+    );
```
lissyx: While the URL for detecting the captive portal needs to be http only, captive portal can be served over https, I'm not sure we really need this permissions here?

toolkit/components/enterprise/modules/CaptivePortal.sys.mjs
lissyx: or `toolkit/components/felt/CaptivePortal.sys.mjs` if it looks like a better place, since it's really limited to felt in fact

toolkit/components/felt/content/window.js
```diff
@@ -435,14 +475,42 @@ window.addEventListener(
   () => {
     setBuildVersion();
     lazy.FeltErrorReport.init(document);
-    lazy.Updates.init(document);
     setupMarionetteEnvironment();
     setupPopupNotifications();
     setupContextMenu();
     listenFormEmailSubmission();
     focusEmailOnLoginVisible();
```
lissyx, line 588: dont forget we need to move this one somewhere else, otherwise the portal browser will miss its focus making it unable to input content into fields

toolkit/components/felt/content/window.js
lissyx: https://github.com/mozilla/enterprise-firefox/pull/1137

Maybe worth waiting on this to land and rebase on top ?

lissyx (approved):
r=me with all the remainders addressed, we verified on slack what was working/broken on my local setup as well

toolkit/components/felt/content/CaptivePortal.sys.mjs
mkaply: yes

toolkit/components/felt/content/window.js
mkaply: No, fixed.

toolkit/components/felt/content/CaptivePortal.sys.mjs
mkaply: done.

toolkit/components/felt/content/CaptivePortal.sys.mjs
mkaply: done.

toolkit/components/felt/content/window.js
mkaply: In theory but if we did it would be a brief flicker and an edge case.

toolkit/components/enterprise/modules/CaptivePortal.sys.mjs
mkaply: It's for HTTPS-First, which is on by default in private browsing (`https_first_pbm`), so the private portal browser upgrades the http probe unless we exempt it. Same thing desktop does in `browser-captivePortal.js` for this button, and it only covers the probe - the portal's own login page can still be https.

toolkit/components/felt/content/window.js
mkaply: gate removed.

PR #1130, opened 2026-07-09
Enterprise beta merge 2026070900

jonathanmendez:
### Description

Bugzilla: NO BUG  
Daily merge from `upstream/beta` to `enterprise-beta`

No merge conflicts.


jonathanmendez:
Closing since the only changes here are translation bumps (which we will pick up in subsequent betas). We can save the build resources since this gives us no real enterprise benefit.

PR #1134, opened 2026-07-10
Bug 2040659 - Rework posture / refresh / login flow

gcp:
### Description 

Bugzilla: Bug-2040659

---

* Depends on:

Console changes as explained in the Bugzilla part

---

### Testing 

* [ ] Added tests


gcp:
```
  ------------------------------------------------------------------------------
   OLD (as landed in Bug 2040418)
  ------------------------------------------------------------------------------

    FELT (launcher + login win)         CONSOLE                 MANAGED FIREFOX
    ---------------------------         -------                 --------------

    (pre-login, no user id yet)
    POST /sso/device_posture   ------------->                  posture carries
       { posture, extensions: null }                           extensions:null
                               <-------------  { posture: <token> }   (phase 1)

    open SSO login (email + posture token)
    GET /sso/login?...&posture=<token> --->
                               <-------------  302 -> /sso/callback
    /sso/callback              <-------------  { access_token, refresh_token }

    setTokens(access, refresh)
    GET /api/browser/whoami    ------------->
                               <-------------  { user_id, email }   (WHOAMI)
    derive profile from user_id
    launch Firefox ------------------------------------------->  starts

                                               GET /api/browser/config
                               <-------------  { ... }   (EDR list HARDCODED
                                                          client-side:
                                                          crowdstrike, cortex-xdr)

                                               POST /api/browser/policies
                                  <----------  { posture, extensions:[...] }
                                                          (phase 2: AddonManager
                               ------------->             ready -> populated)
                                               { policies }
                                                          => extensions reported
                                                             in TWO phases

    token refresh (on 401 / expiry):
    POST /sso/token            ------------->
       { grant_type, refresh_token }
                               <-------------  { access_token, refresh_token }

  ------------------------------------------------------------------------------
   NEW (Bug 2040659)
  ------------------------------------------------------------------------------

    FELT (launcher + login win)         CONSOLE                 MANAGED FIREFOX
    ---------------------------         -------                 --------------

    (pre-login reachability only, NO posture)
    GET  <console base>        ------------->
                               <-------------  any HTTP status = reachable

    open SSO login (email + OS version; NO posture, NO user id yet)
    GET /sso/login?...&osVersion=<os> --->     (OS version lets console tailor
                               <-------------   the posture_elements descriptor)
                                               302 -> /sso/callback
    /sso/callback              <-------------  { one_time_token }   <-- token only

    ONE-TIME-TOKEN EXCHANGE (no posture: profile/extensions unknown until reply)
    POST /sso/exchange         ------------->
       { one_time_token }
                               <-------------  { user_id, email,
                                                 access_token, refresh_token,
                                                 policies,
                                                 config: { posture_elements } }
    setTokens; store descriptor (LOGIN = authoritative, clear-on-omit)
    derive profile from user_id  (getProfileName)
    collect FULL posture from profile-on-disk:
       extensions <- <profile>/extensions.json (XPIDatabase, ONE phase)
       presentEdrs <- probe ONLY console-listed EDR agents (none => none)

    POSTURE-CARRYING REFRESH (before launch)
    POST /sso/token            ------------->
       { grant_type, refresh_token, posture }
                               <-------------  { access_token, refresh_token,
                                                 config: { posture_elements } }
    launch Firefox --------------------------------------------> starts

                                               GET /api/browser/policies   (plain
                                  <----------  GET, NO posture)
                               ------------->  { policies }
                                               on 401 -> posture-carrying
                                                         /sso/token refresh

    FELT posture monitor (setInterval, policy-poll cadence):
       re-collect posture; only if CHANGED vs last submitted:
    POST /sso/token            ------------->
       { grant_type, refresh_token, posture }
                               <-------------  { tokens, config }
       -> setTokens + sendAccessToken to browser

    (standalone GET /api/browser/config still available)

```

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
```diff
@@ -0,0 +1,288 @@
...
+    const readJsonArrayPref = pref => {
+      try {
+        // getStringPref's default only covers an unset pref; a pref set to a
+        // non-string type still throws, so keep the read inside the try.
+        const raw = Services.prefs.getStringPref(pref, "");
+        if (!raw) {
+          return [];
+        }
+        const parsed = JSON.parse(raw);
+        return Array.isArray(parsed) ? parsed : [];
```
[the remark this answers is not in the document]
gcp: Don't care.

toolkit/components/felt/rust/src/edr_checker.rs
```diff
@@ -515,3 +537,34 @@ pub extern "C" fn edr_checker_constructor(
     let obj = EdrCheckerXPCOM::new();
     unsafe { obj.QueryInterface(iid, result) }
 }
+
+#[cfg(test)]
+mod tests {
```
lissyx, line 542: are those rust specific tests running on m-c ? We should add our coverage of them I guess

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -395,11 +410,169 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+   * pref that the pre-launch posture collection reads. The browser gets its own
+   * copy via _applyFirefoxConfigs.
+   *
+   * @param {{edr?: string[]}} [postureElements]
+   */
+  _writePostureElements(postureElements) {
+    Services.prefs.setStringPref(
+      EDR_AGENTS_PREF,
+      this._serializePostureEdr(postureElements)
+    );
```
lissyx: should this pref be locked?

argl:
On the login flow and the posture configuration:

Does it make sense to have the posture configuration dependent on the OS version? I am not sure if we should make the posture configuration itself somehow posture-dependent.

What is the requirement to introduce different posture configurations for different OS versions? I would rather remove this complication if there is no strict need. Even if, for example, different OS versions have different EDR solutions installed, we can probably live with probing for two or three EDR providers.

For reference, this is the (simplified) sequence I have been working with:

argl:
Adding to my own comment: I think it is better to not trust the client to send the correct os/version. 

If we want, for example, different EDR detection lists for different OSes, we just send posture configuration for _all_ 3 platforms to the client. The client (who knows its own platform) picks the correct one then. Same goes for (hypothetical) `os_query` requests. Send settings for all platforms and the client will figure out which one applies.

To sum it up, a configurable posture is a very useful thing to have (selecting the probed EDR vendors is one concrete application), but we should go with a single global posture configuration per customer/company that is sent unconditionally to all clients.

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
```diff
@@ -0,0 +1,378 @@
...
+        mobileEquipmentId,
+        interfaces: networkInterfaces,
+      },
+      extensions,
+      machineId,
+      secureBootEnabled:
+        Services.sysinfo.getPropertyAsBool("secureBootEnabled"),
+      isDomainJoined: Services.sysinfo.getPropertyAsBool("isDomainJoined"),
+      presentEdrs,
+    };
```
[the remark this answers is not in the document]
gcp: Nonsense comment. We only get here in Enterprise builds.

gcp:
Seven commits on top of what you reviewed, each independently revertable:

| commit | title |
| --- | --- |
| `74dfed26864e` | Select the posture-elements platform section client-side |
| `21b49e497e4f` | Run the felt crate's Rust tests in enterprise CI |
| `0ef08b8b02c5` | Document why the browser-driven token refresh carries no posture |
| `bcef92a9014a` | Read the Felt-side add-on list without repointing XPIDatabase |
| `06959454b3b8` | Stop posture reporting when the session is torn down |
| `8d4b27440368` | Apply mid-session posture elements in both processes |
| `f6ab7dc28c55` | Avoid an Instant underflow in the EDR cache test |

### 1. `74dfed2` — Select the posture-elements platform section client-side

@argl's Jul 15 / Jul 16 comments. `posture_elements` is now `{ "<os-name>": { edr: [...] } }`, one section per platform, keyed on `Services.sysinfo.getProperty("name")` (`"Windows_NT"`, `"Darwin"`, `"Linux"`) — the value the client already reports in every posture payload. The client selects its section: `posturePlatform` (`DevicePosture.sys.mjs:34`) and `PostureElements.write()` (`:53`), which writes `"[]"` for a missing section.

`osVersion` is gone from the SSO login URL along with `ConsoleClient._getOsVersion`; `constructSsoLoginURI` sends `target`, `email`, `deviceId`. The descriptor's write side moved from `FeltProcessParent` into `DevicePosture.sys.mjs`, next to the pref it feeds.

@argl: `os_query` belongs in the same per-platform section when it lands.

### 2. `21b49e4` — Run the felt crate's Rust tests in enterprise CI

@lissyx: they were running nowhere. `07180bc` added `felt` to `RUST_TESTS`, which covers local `mach rusttests`, but the only rusttests CI tasks are `linux64-rusttests/{opt,debug}` and those build without `MOZ_ENTERPRISE`.

New `build-linux64-enterprise-rusttests/opt` (`taskcluster/kinds/build/linux.yml:1557`): rusttests variant plus `--enable-enterprise`, treeherder `linux64-enterprise/opt` symbol `BR`. It runs on every `enterprise-firefox` push; upstream gates the equivalent builds on a rusttests *test* task pulling them in, and this branch has none. Green here — task [`f2OTI21J`](https://firefox-ci-tc.services.mozilla.com/tasks/f2OTI21JTYGKtuZO8AuoXQ), `cache_usable_only_when_fresh_and_complete ... ok`; results show up in Treeherder.

### 3. `0ef08b8` — Document why the browser-driven token refresh carries no posture

Documentation only. A 401 on the policy poll refreshes credentials; posture is reported by the Felt monitor on change. Why the split, now recorded at the call site:

- Same cadence either way: `_posturePollMs` is the console's `polling_frequency` (`FeltProcessParent.sys.mjs:430`), the value the browser polls policies on.
- The 401 path is latency-critical — the browser is blocked until FELT returns a token, and posture collection means an EDR probe sweep plus an `extensions.json` read.
- The monitor diffs against `_lastPostureJson` to submit only on change; a second posture-carrying path would race that state.

`refreshTokens()` single-flights, and a caller joining an in-flight refresh gets `postureSubmitted: false` (`ConsoleClient.sys.mjs:509`), so the monitor retries on its next tick.

### 4. `bcef92a` — Read the Felt-side add-on list without repointing XPIDatabase

Pre-launch posture read the target profile's add-ons by pointing the process-global `XPIDatabase` at it. Felt runs `AddonManager` for its own profile, so the repointing outlived the read, and a missing `extensions.json` sent `asyncLoadDB()` into `rebuildDatabase()`, which reconciles the **Felt process's** install locations and saves to `jsonFilePath` — writing a database describing the launcher into the profile it was reading.

`readAddonsForFelt(profileDir)` (`:119`) reads that profile's `extensions.json` with `IOUtils.readJSON` and maps the entries, reporting the fields the browser reports: entries `XPIDatabase` considers `visible`, the `active` flag behind `AddonWrapper.isActive`, and the locale-negotiated name (`localizedAddonName`, `:84`). `getDatabaseForApp` → `getAddonManagerForApp` (`:167`), which is what it returns.

Behaviour change: a profile with no database reports `extensions: null` (unknown), as `DevicePosture`'s typedef already described; `test_felt_device_posture` had assertions accepting only a list. New coverage: an `ADDON_DB` fixture (active, disabled and invisible extensions, plus a theme) and `_assert_addon_db_untouched`, which asserts `extensions.json` is byte-identical after the read.

### 5. `0695945` — Stop posture reporting when the session is torn down

A refresh resolving after `logoutFirefox()` had cleared the tokens went on to `setTokens()` / `sendAccessToken()`, arming a session that was shutting down.

- `_postureRefreshInFlight` became the `_postureRefresh` promise (`FeltProcessParent.sys.mjs:564`): still the single-flight guard, now awaitable.
- `logoutFirefox()` (`:1054`) is async — `logoutReported`, `_stopPostureMonitor()`, `await this._postureRefresh`, server signout, `clearTokens()`, `shutdownFirefox()`.
- The tick re-checks `logoutReported` once its refresh resolves (`:593`) and drops the response; the refresh-failure teardown (`:770`) stops the monitor too.

No automated test: the window is a logout landing inside an in-flight refresh.

### 6. `8d4b274` — Apply mid-session posture elements in both processes

Closes the Copilot threads on the old `:468` and `:742`. Both processes collect posture from their own copy of the probe list; a descriptor arriving after launch reached only Felt's copy, and `_applyFirefoxConfigs` pushed to the browser while leaving Felt's stale.

`_storePostureElements` (`:487`) writes this process's pref and relays the same value, and every config-bearing path goes through it — including the browser-driven refresh (`:521`), which had been destructuring tokens and dropping the `config` alongside them. `PostureElements.write()` returns what it wrote, which retired `serializeEdr`. The login exchange (`:1141`) writes the pref alone, running before the browser exists, and launch forwards the stored value at `:735`; login clears the list on omit, a refresh preserves it.

The mock console now folds `config` into its `/sso/token` response (`felt_tests.py`, `build_config_response()`) — the channel a mid-session descriptor arrives on — and `run_mid_session_descriptor_reaches_browser` drives a refresh from the browser and asserts its probe list is replaced.

### 7. `f6ab7dc` — Avoid an Instant underflow in the EDR cache test

`Instant::now() - CACHE_TTL` panics on a young monotonic clock. The test now advances the `now` passed to `cache_is_usable`.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
gcp: Maybe but it doesn't actually matter, editing it can't subvert anything.

taskcluster/kinds/build/linux.yml
```diff
@@ -1554,6 +1554,52 @@ linux64-rusttests/debug:
     optimization:
         test-inclusive: [rusttests]
 
+linux64-enterprise-rusttests/opt:
+    description: "Linux64 Enterprise Rust Tests Opt"
+    index:
+        product: firefox
+        job-name: linux64-enterprise-rusttests-opt
```
lissyx, line 1592: To surface to github:
```suggestion
        job-name: linux64-enterprise-rusttests-opt
        attributes:
            code-review: true
```

testing/enterprise/felt_tests.py
```diff
@@ -171,6 +171,16 @@ def do_GET(self):
 
 
 class ConsoleHttpHandler(LocalHttpRequestHandler):
+    def build_config_response(self):
+        """The browser config the console folds into the token responses, with a
+        posture-elements descriptor only when the test configured one."""
```
lissyx: The wording is a bit hard to get IMHO

testing/enterprise/felt_tests.py
```diff
@@ -293,6 +305,17 @@ def do_GET(self):
             m = json.dumps({"data": TEST_PRIMARY_SECRET})
             contentType = "application/json"
 
+        elif path == "/api/browser/policies":
+            # Policies are fetched with a plain authenticated GET; posture is
+            # reported separately via the posture-carrying token refresh.
+            if not self.check_auth():
+                return
```
lissyx, line 311: We do not want to issue an error here?

testing/enterprise/felt_tests.py
```diff
@@ -504,12 +519,51 @@ def do_POST(self):
...
+            # is how a descriptor changed mid-session reaches the client.
             m = json.dumps({
                 "access_token": self.server.policy_access_token.value,
                 "token_type": "Bearer",
                 "expires_in": 71999,
                 "refresh_token": self.server.policy_refresh_token.value,
+                "config": self.build_config_response(),
+            })
+
+        elif path == "/sso/exchange":
```
lissyx: Dont we want to get rid of `/sso/token` now or did i missed something in the flow?

gcp: https://github.com/mozilla/enterprise-firefox/pull/1134#issuecomment-5254243977

This is the one-time token early in the flow.

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
```diff
@@ -0,0 +1,378 @@
...
+  "sitepermission",
+  "siteperm_deprecated",
+  "plugin",
+  "mlmodel",
+];
+
+// The name for a serialized add-on, negotiated the way AddonInternal does it:
+// the best match for the requested app locales, falling back to the add-on's
+// default locale, so the Felt-side read reports the same name the browser would.
+function localizedAddonName(addon) {
```
lissyx: I'm a bit unsure if it's a good idea, this assumes the user's locale will always match/admin's one?

How is this value used server-side? It may be nice for displaying in the console but we may want to stick to en-US for logs for example.

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
```diff
@@ -0,0 +1,378 @@
...
+   * @returns {Promise<DeviceAddon[]|null>}
+   */
+  async getExtensions({ profileDir = null, waitForAddons = false } = {}) {
+    try {
+      if (Services.felt.isFeltUI()) {
+        // The profile (and thus its extension list) is only known once SSO has
+        // resolved the user id; without it we cannot report extensions.
+        if (!profileDir) {
+          return null;
+        }
```
lissyx: Is this still true now that we have changed the flow ? I seem to recall we changed the flow so that this was not required anymore, but maybe we need to keep it?

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
gcp: I think that's a good question for the console folks. @fiji-flo or @argl ?

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
gcp: User ID is the outcome of the SSO flow: https://bugzilla.mozilla.org/show_bug.cgi?id=2040659#c1

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -426,11 +453,164 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      // Only record the posture as submitted if this call actually sent it; if
+      // it piggybacked on an in-flight (posture-less) refresh, leave
+      // _lastPostureJson so the next tick retries rather than losing the change.
+      if (postureSubmitted) {
+        this._lastPostureJson = postureJson;
+      }
+    } catch (e) {
+      lazy.log.error("Posture-change refresh failed:", e);
+    }
+  }
```
lissyx: Is there any requirement for this code to live here? I do not see it obviously, if there's none, I'd prefer we place it in `DevicePosture.sys.mjs`

michaelvanstraten:
@gcp We need the **one time token exchange** and the posture submission to happen in **one atomic step** rather then having a `/sso/exchange` route and then doing a `_submitInitialPosture`. I propose the following flow where the `/api/browser/sso/token` endpoint either a one time token or a refresh token.


```mermaid
sequenceDiagram
    autonumber

    participant F as Firefox
    participant C as Console
    participant SSO as SSO Provider

    F->>C: GET /api/browser/sso/login
    C-->>F: Redirect → SSO Provider

    F->>SSO: Perform login flow
    SSO-->>F: Redirect → /sso/callback

    F->>C: GET /sso/callback
    C-->>F: Posture config + user ID + one-time token

    F->>C: POST /api/browser/sso/token<br/>{ posture P₁, refresh token }
    C-->>F: Access token + refresh token

    Note over F: Posture changes<br/>P₁ → P₂

    F->>C: POST /api/browser/sso/token<br/>{ posture P₂, one-time token }
    C-->>F: Access token + refresh token
```

gcp:
@michaelvanstraten your diagram shows the Posture change causing a POST to `api/browser/sso/token` with the one-time token. That seems weird to me - I would expect us to use the refresh token there?

gcp:
```
    FELT (launcher + login window)        CONSOLE
    ------------------------------        -------

    reachability probe, before login
      GET  <console base>             ------->
                                      <-------   any HTTP status = reachable

    login
      GET  /sso/login                 ------->
           ?target&email&deviceId&version
                                      <-------   302 /sso/callback
                                                 { one_time_token, user_id, email,
                                                   config: { posture_elements } }

      version: the login flow this client speaks, so a console can keep
        answering clients that speak another
      posture_elements: ONE descriptor, a section per platform; FELT picks
        its own section (sysinfo "name") -> EDR probe list
      derive the profile from user_id
      collect posture: read <profile>/extensions.json (absent database
        => extensions: null), probe the selected EDR agents

    redeem the token, reporting that posture in the same request: no
    session exists before this returns
      POST /sso/token                 ------->
           { grant_type: one_time_token,
             one_time_token, posture }
                                      <-------   { tokens }
      spawn the browser with the token minted here

    posture monitor, on the policy-poll cadence, when the posture changed
      re-collect posture
      POST /sso/token                 ------->
           { refresh_token, posture }
                                      <-------   { tokens, config }

    credential refresh, requested by the browser on a 401: reports the
    last measurement, re-measured only if older than one interval
      POST /sso/token                 ------->
           { refresh_token, posture }
                                      <-------   { tokens, config }

    logout: stop the monitor, await any in-flight tick, then
      POST /sso/logout                ------->
                                      <-------   failures are logged, logout continues
      clear the tokens and shut the browser down

    an abnormal browser exit relaunches the browser on the session it
    already has; the monitor stops with the process and starts again with
    its replacement. Enough crashes within the crash period stop the
    relaunch and the user is told instead.
```

The browser talks to the console itself for policies (`GET /api/browser/policies`); on a 401 it asks FELT for the credential refresh above. FELT collects every posture the console receives, the extension list included, so an `extensions.json` write by the browser reaches the console on the next tick. Each `config` FELT receives is applied locally and relayed to the browser, keeping the EDR probe list the same in both processes.

gcp:
@michaelvanstraten the flow assumes we get `email` echoed back. This is currently used to trigger what gets stored in the login form (via `whoami`) in Felt, and some of our tests actually assume those can be different, that is, that the console can normalize or return a canonical email:

https://github.com/mozilla/enterprise-firefox/blob/2ae477ce228dbcd4c120fc99816bbe8241ce31f9/testing/enterprise/felt_tests.py#L1044
https://github.com/mozilla/enterprise-firefox/blob/2ae477ce228dbcd4c120fc99816bbe8241ce31f9/testing/enterprise/felt_tests.py#L302

I'm not sure if we should keep support for this or not? (To be clear, keeping it is easier)

toolkit/components/felt/content/FeltProcessParent.sys.mjs
gcp: Moving this

testing/enterprise/felt_tests.py
gcp: OK

taskcluster/kinds/build/linux.yml
gcp: I assume this was mis-indented?

gcp:
### 1. `aa48712` — one request instead of two

`/sso/exchange` is gone. `/sso/token` now takes a one-time-token grant carrying the posture, and the SSO callback supplies `user_id`, `email` and `config` alongside the token, so the profile is derived and its `extensions.json` read before a session exists. The browser is handed the access token minted at login.

A rejected redemption ends the login attempt rather than being retried; the HTTP status is kept on the error to tell a console that refused the posture from one that could not answer.

Assumed in the `felt_tests.py` mock, and to be confirmed:

- `grant_type: "one_time_token"`, mirroring the refresh grant
- the one-time token is single use
- callback fields: `one_time_token`, `user_id`, `email`, `config.posture_elements`
- what a rejection returns
- `extensions: null` is accepted, as a first login creates the profile

### 2. `fcc52fb` — a posture on every token call

The 401 refresh blocks the browser until a token comes back, and a collect can spawn subprocesses (EDR detection, `ioreg` for the machine ID on macOS), so it reports the last measurement and re-measures only if that is older than one poll interval. A posture is stamped with its measurement time rather than its send time, so replaying one cannot extend its own freshness window.

### ~~3. `79075f7` — a crash ends the session~~

~~A relaunched browser has no one-time token to attest itself with, so reusing the tokens of one that just died would hand a fresh browser a session attested for the dead one. Both crash outcomes now clear the tokens and stop posture reporting.~~

~~Abnormal exits are now counted for the Felt process lifetime rather than the actor's. `FeltProcess` is a per-content-process actor, and the sign-in that now follows every crash runs in a fresh content process, so a per-instance counter restarted at zero each time and repeated crashes never reached the limit. `test_felt_browser_crash_restart.py` becomes `test_felt_browser_crash_ends_session.py`.~~

Section 3 is out of the series: the browser relaunch stays, and whether a crash should end the session is decided separately. Sections 1 and 2 were rebased, hence the new hashes.

michaelvanstraten:
> @michaelvanstraten your diagram shows the Posture change causing a POST to `api/browser/sso/token` with the one-time token. That seems weird to me - I would expect us to use the refresh token there?

Yes, sorry about that, copy pasta mistake on my side.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -168,26 +158,87 @@ export const ConsoleClient = {
...
    */
-  async constructSsoLoginURI(email, devicePostureToken) {
+  async constructSsoLoginURI(email) {
     const deviceId = lazy.FeltStorage.getDeviceId();
     const url = await this.consoleBaseURI;
     url.pathname = this._paths.SSO;
     url.searchParams.set("target", "browser");
     url.searchParams.set("email", email);
-    url.searchParams.set("devicePostureToken", devicePostureToken);
     url.searchParams.set("deviceId", deviceId);
```
michaelvanstraten, line 203:
```suggestion
    url.searchParams.set("deviceId", deviceId);
    url.searchParams.set("version", "v2");
```

The console needs to be backwards compatible with old clients, this means we need to include a version parameter here to indicate to the console that this client understands the new login flow.

toolkit/components/felt/content/window.js
```diff
@@ -114,42 +114,34 @@ async function connectToConsole(email) {
...
   const attempt = ++signInGeneration;
   pendingSignInEmail = email;
-  let posture;
+
+  // Probe the console for reachability before starting SSO, so an unreachable or
+  // misconfigured console address surfaces a clear error here, at the point the
+  // user submitted.
   try {
-    posture = await lazy.ConsoleClient.sendDevicePosture();
+    await lazy.ConsoleClient.probeConsoleReachable();
```
michaelvanstraten, line 122: Why do we need this?

taskcluster/kinds/build/linux.yml
```diff
@@ -1554,6 +1554,54 @@ linux64-rusttests/debug:
...
+    attributes:
+        code-review: true
```
lissyx, line 1594: This looks correct

toolkit/components/felt/content/window.js
gcp: It's trying to keep the error handling in the flow the same, not sure it's strictly needed, looking.

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
michaelvanstraten: I agree with Alex's point, we should not assume that the console has the same locale as the user.

michaelvanstraten:
> @michaelvanstraten the flow assumes we get `email` echoed back.

Just echoing back the email you provided in the initial form as a "hint" seams non nonsensical to me. What we can do is give you the canonical email for the user as, that should not be an issue.

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
gcp: Tries to force `en-US` but reports the extension's default if that is not available.

testing/enterprise/felt_tests.py
gcp: If you get here, you already got a `401`.

toolkit/components/felt/content/window.js
gcp: It could be keyed off of the SSO flow perhaps but that then just makes this an even larger change. I'd prefer to just leave it this way.

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
```diff
@@ -0,0 +1,460 @@
...
+      .filter(
+        addon => addon.visible && REPORTED_ADDON_TYPES.includes(addon.type)
+      )
+      .map(addon => ({
+        id: addon.id,
+        name: reportedAddonName(addon),
+        type: addon.type,
+        version: addon.version ?? "",
+        enabled: !!addon.active,
+      }));
```
lissyx, line 150: As much as I remember in my implementation we were relying on addon's code directly to avoid re-implementing here, as I was not confident we may not get badly broken by changes in the addon code.

Of course relying on the addons code would also bring its own problems.

gcp: Yes, that's a tradeoff being made here. This code is a bit simpler and in ways less risky (cfr. the bug that was in here that ended up fetching Felt extensions in error cases), but it risks getting out of sync. On the other hand, this makes posture obtainable without relying in any way on Firefox itself.

I think it's a call we can reconsider independently of this PR afterwards if this turns out to cause frequent breakage itself.

lissyx: Yes, I'm just wondering/hoping our current test infra is enough to catch those and not that we discover that late in dev cycle

testing/enterprise/test_felt_device_posture_elements.py
```diff
@@ -0,0 +1,467 @@
...
+    def _wait_for_string_pref(self, name):
+        self._child_driver.set_context("chrome")
+        try:
+            return self._child_longwait.until(
+                lambda _: self._child_driver.execute_script(
+                    f"return Services.prefs.getStringPref('{name}', '') || null;"
+                )
+            )
+        finally:
+            self._child_driver.set_context("content")
```
lissyx, line 69: This pref comes from the console right? Maybe worth mentionning here in a comment, otherwise it may look weird to wait like that

testing/enterprise/test_felt_device_posture_elements.py
```diff
@@ -0,0 +1,467 @@
...
+        assert json.loads(edr_pref) == self.expected_edr(self.POSTURE_ELEMENTS), (
+            "edr_agents pref reflects this platform's section of the console "
+            f"config, got {edr_pref}"
+        )
+
+    def run_console_driven_probes(self):
+        """DevicePosture.collect probes exactly the console-configured EDR list."""
+        self._child_driver.set_context("chrome")
+        try:
+            rv = self._child_driver.execute_async_script(
```
lissyx: nit:

```suggestion
            collected_edrs = self._child_driver.execute_async_script(
```

testing/enterprise/test_felt_device_posture_elements.py
```diff
@@ -0,0 +1,467 @@
...
+            polls += 1
+            value = self._child_driver.execute_script(
+                "return Services.prefs.getStringPref("
+                "'enterprise.posture.edr_agents', '');"
+            )
+            return value if value and json.loads(value) == expected else None
+
+        self._child_driver.set_context("chrome")
+        try:
+            edr_pref = self._child_longwait.until(refreshed)
```
lissyx: `longwait` is 60s, are we at risk of intermittence here?

testing/enterprise/test_felt_device_posture_elements.py
```diff
@@ -0,0 +1,467 @@
...
+                  );
+                  const { AppConstants } = ChromeUtils.importESModule(
+                    "resource://gre/modules/AppConstants.sys.mjs"
+                  );
+                  return {
+                    alpha1: await getProfileName(alphaId),
+                    beta: await getProfileName(betaId),
+                    alpha2: await getProfileName(alphaId),
+                    fallback: await getProfileName(null),
+                    base: `enterprise-profile-${AppConstants.MOZ_UPDATE_CHANNEL}`,
```
lissyx: Is there a specific reason for `alphaX` / `betaY` naming scheme? I'm not sure I follow and I feed like it's confusing

testing/enterprise/test_felt_device_posture_elements.py
```diff
@@ -0,0 +1,467 @@
...
+        "schemaVersion": 36,
+        "addons": [
+            {
+                "id": "active@example.com",
+                "type": "extension",
+                "version": "1.2",
+                "visible": True,
+                "active": True,
+                "defaultLocale": {"name": "Active Extension"},
+                "locales": [{"locales": ["de"], "name": "Aktive Erweiterung"}],
```
lissyx, line 262: Maybe add more locales to ensure reporting is sound?

testing/enterprise/test_felt_device_posture_elements.py
```diff
@@ -0,0 +1,467 @@
...
+        assert rv["extensions"] is None or isinstance(rv["extensions"], list), (
+            f"extensions degrades gracefully, got {rv['extensions']!r}"
+        )
+        assert rv["hasOs"], "the rest of the posture is still collected"
+        self._assert_addon_db_untouched(rv)
+
+    def _assert_addon_db_untouched(self, rv):
+        assert rv["dbAfter"] == rv["dbBefore"], (
+            "the launching profile's add-on database is left byte-identical"
+        )
```
lissyx, line 646: should we also check `stat` ? Only access time should change

lissyx (approved):
LGTM

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
gcp: I think I can perhaps add a test to have the Firefox code mess with extensions and then check whether the new parser here doesn't encounter any unexpected stuff...

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -168,26 +162,82 @@ export const ConsoleClient = {
...
+  /**
+   * Checks that the configured console is reachable before starting the SSO flow.
+   * Any HTTP response means the host is reachable; only network-level failures
+   * reject, in the shape FeltErrorReport.handleXhrError expects.
+   *
+   * @throws {TypeError} On a network-level failure.
+   * @returns {Promise<void>}
+   */
+  async probeConsoleReachable() {
+    const url = (await this.consoleBaseURI).href;
```
1rneh: `nit`
```suggestion
    const url =  await this.constructURI("");
```

Or `constructURI` could get an empty string as default path.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -905,20 +1011,65 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+
+          // Read the extension list from the profile on disk, before the browser
+          // is spawned and its AddonManager rewrites extensions.json.
+          const { path: profileDir } = await this._resolveProfile();
+          let posture = null;
+          const measuredAt = Date.now();
+          try {
+            posture = await lazy.DevicePosture.collect({ profileDir });
+          } catch (e) {
+            lazy.log.error("Failed to collect the initial device posture:", e);
```
1rneh, line 1112: The docs for `redeemOneTimeToken` state _"the console mints no session without a posture"_. If that is the case, we already abort the login here in case the device posture collection fails.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -545,9 +567,16 @@ export const ConsoleClient = {
         throw new Error(`Token refresh failed: ${text}, Status: ${res.status}`);
       }
 
-      const { access_token, refresh_token, expires_in } = await res.json();
+      const { access_token, refresh_token, expires_in, config } =
+        await res.json();
       const expires_at = Math.floor(Date.now() / 1000) + Number(expires_in);
-      return { access_token, refresh_token, expires_at };
+      return {
```
1rneh, line 642: There is a not so well guarded race potential now that we have multiple refresh initiators (browser-driven and posture monitor). They both go through `ConsoleClient.refreshTokens`, which serializes on `_feltRefreshPromise`. The guard clears as the promise resolves, but the callers set the tokens in felt afterwards — safe today only because there's no `await` in between. If one was added, the other initiator can start a refresh reading the "outdated" refresh token, gets a 401 and tears the session down.

I suggest moving `Services.felt.setTokens(access_token, refresh_token, expires_at);` into `refreshTokens` here, before `.finally` clears the guard.

1rneh (changes requested):

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: This drops any path prefix that the console address might have, and the rest of the code respects that, so we probably shouldn't do this?

See e.g.: https://searchfox.org/enterprise-main/rev/05fa9ca00a2215e7688437e92f3f34d931ca83f8/toolkit/components/enterprise/modules/EnterpriseEndpoints.sys.mjs#61

testing/enterprise/test_felt_device_posture_elements.py
gcp: No special meaning, will rename.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
gcp: Clarified on Slack: path prefixes aren't supported and are dropped in a lot of places in the code.

gcp:
Additional fixes from local review:

-  a997c27 — the resolved profile was cached per actor, which outlives a session: the next user to sign in through it got the previous user's profile and reported their extensions. Now keyed on the user; a callback with no user_id is refused. Also, logout now awaits a browser-driven refresh instead of racing it.
  - 511950d — a 401/403 on the posture monitor's refresh was only logged, so the browser kept running on tokens the console had repudiated. Both refresh initiators now share one teardown; network/5xx there is still just retried.
  - ec16e65 — Firefox deactivates most extensions in safe mode, but the reader reported the active flag straight out of the database, so posture listed them as enabled in a session where they weren't running.
  - 5e94669 — plugin and mlmodel matched nothing since the reader stopped using the AddonManager (GMP comes from prefs, ML models from ModelHub's IndexedDB), so they're commented out.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -905,20 +1011,65 @@ export class FeltProcessParent extends JSProcessActorParent {
...
       case "FeltChild:StartFirefox":
         {
+          const { one_time_token = "", user_id, email, config } = message.data;
+
+          // The profile is derived from the user id, so without one the session
+          // would run in the profile shared by every user.
+          if (!user_id) {
+            lazy.log.error("SSO callback carried no user id");
+            Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLaunchFailure", {
+              errorType: "loginFailed",
```
1rneh, line 1090: Are you expecting some specific message here? I was curious about our error type and we seem to only handle "primarySecret" and "sdrTokenUnlockFailed". All other error types fallback to `felt-browser-error-launch-failure`: https://searchfox.org/enterprise-main/rev/85c8f15ca480b61d6b8bc8326360e42fb23a686a/toolkit/components/felt/Felt.sys.mjs#307-314

1rneh (approved):
r+ on `ConsoleClient` related changes.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
gcp: We already rely on the fallback behavior, as https://searchfox.org/enterprise-main/rev/85c8f15ca480b61d6b8bc8326360e42fb23a686a/toolkit/components/felt/content/FeltProcessParent.sys.mjs#596 just relies on getting fallback to `felt-browser-error-launch-failure` which is also a reasonable message here, although this one points at a console bug or misconfiguration, so it could also use its own message here if we want to (...and that's possible because a different type is passed in).

michaelvanstraten (changes requested):
We discussed this within the console team, and we don’t believe that making posture elements configurable is a good idea, especially not dynamically (e.g., through a token refresh).

If you genuinely want the configuration, I can suggest that I return it to you in the SSO callback. I would then ask you to rename the top-level object to `posture` instead of `config`, and add a subkey called `edr_agents`.

Having the posture elements returned in the `getFirefoxConfigs` makes even less sense to me. This would essentially trigger a token refresh on every startup because you would receive the posture configuration from the SSO callback, submit the posture with the specified agents, fetch the configuration, get a different set of `edr_agents` to detect, and then trigger another refresh.

If we return you a posture configuration on each token refresh, this could essentially lead to the same problem, but even worse. A posture change could trigger a change in the desired `edr_agents`, which would trigger another posture change, and so on.

gcp:
@michaelvanstraten Posture elements were made configurable because @fiji-flo was concerned about privacy issues with them, i.e. not collecting and sending posture elements the admin doesn't need. For the EDR list, there's no point in stomping through the system trying to locate every possible EDR under the sun just to find one, so that's one more reason to want that list to be runtime supplied.

We can make it static, i.e. only supplied during the SSO callback, that just means that admin has to force browser restarts if they want to change posture configuration.

michaelvanstraten:
> We can make it static, i.e. only supplied during the SSO callback, that just means that admin has to force browser restarts if they want to change posture configuration.

Yes, I think that is the best option for now.

michaelvanstraten:
@gcp I understand that you put some effort into this. I would be fine with keeping the current design as long as we agree that the posture elements decision it self can not depend on the posture of the device and is rather just a static configuration for all client. I would ask you however to change the config structure and remove it from the `getFirefoxConfigs`.

gcp:
>I understand that you put some effort into this.

Don't worry, I won't mind deleting complexity from this code again 😁 I'm just slightly concerned about keeping making changes so late.

>as long as we agree that the posture elements decision it self can not depend on the posture of the device and is rather just a static configuration for all client

...but that would be a console-side invariant, not a client side one? If you enforce this on the console, you **would** be able to change this config at runtime?

michaelvanstraten:
> If you enforce this on the console, you **would** be able to change this config at runtime?

Yes, but we would need to agree on that from a product perspective. 

I understand the need to configure this in the Console UI and when we remove the dynamic configuration it become a lot harder to apply this, like you said.

gcp:
```
Wire format now

GET /api/browser/config — no posture field at all:
{"learn_more_url": "…", "company_logo_url": "", "policies": {"polling_frequency": 500},
 "services": {…}, "extra_prefs": [["marionette.port", 0]]}

SSO callback and every token response carry a top-level posture sibling:
{"one_time_token": "…", "user_id": "…", "email": "…",
 "posture": {"edr_agents": {"Windows_NT": ["crowdstrike", "cortex-xdr"],
                            "Darwin": ["crowdstrike"],
                            "Linux": ["sentinelone"]}}}
```

gcp:
The `clientSessionId` we discussed I will add in a follow-up PR to not make this even bigger.

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
```diff
@@ -0,0 +1,499 @@
...
+   * Selects this platform's list and writes it into this process's
+   * EDR_AGENTS_PREF. A missing list writes "[]".
+   *
+   * @param {{[key: string]: string[]}} [edrAgentsByPlatform]
+   * @returns {string} The value written, so callers can relay it to the other
+   *   process.
+   */
+  write(edrAgentsByPlatform) {
+    const edrAgents = JSON.stringify(
+      edrAgentsByPlatform?.[lazy.posturePlatform] ?? []
```
michaelvanstraten: Why do you need this by platform? In the current implementation I just give you all of them, not indexed by the platform https://github.com/mozilla/enterprise-console/blob/dd63a4a8ee95b63c76b2eca4ddd53aa278686a3f/crates/ec_webserver/src/config.rs#L222.

michaelvanstraten (approved):
Looks good from my side, only thing being the os keyed `edr_agents`.

PR #1145, opened 2026-07-13
Bug 2039308 - Avoid trying to access CrashReporter ui when not ready in auto-submit flow

jonathanmendez:
### Description 

Bugzilla: Bug-2039308

On linux (and likely mac), the code that is trying to read the comment field accesses the UI before it is ready. The comment field is disabled for the policy auto-submit flow, so we don't need to read it. The code already was guarded for the non-UI auto-submit flow, so I added a guard for the policy auto-submit flow as well.

---

### Testing 

* [x] Manual testing performed on wsl, untested on mac

### DIFF
--- a/toolkit/crashreporter/client/app/src/logic.rs
+++ b/toolkit/crashreporter/client/app/src/logic.rs
@@ -700,7 +700,10 @@ impl ReportCrash {
     /// Form the extra data, taking into account user input.
     fn current_extra_data(&self) -> serde_json::Value {
         let include_address = self.settings.borrow().include_url;
-        let comment = if !self.config.auto_submit {
+        let get_comment = !self.config.auto_submit;
+        #[cfg(feature = "enterprise")]
+        let get_comment = get_comment && !self.config.policy_auto_submit;
+        let comment = if get_comment {
             self.ui().wait(|r| r.comment.get())
         } else {
             Default::default()


toolkit/crashreporter/client/app/src/logic.rs
```diff
@@ -700,7 +700,10 @@ impl ReportCrash {
     /// Form the extra data, taking into account user input.
     fn current_extra_data(&self) -> serde_json::Value {
         let include_address = self.settings.borrow().include_url;
-        let comment = if !self.config.auto_submit {
+        let get_comment = !self.config.auto_submit;
```
lissyx, line 703: nit:
```suggestion
        #[cfg(not(feature = "enterprise"))]
        let get_comment = !self.config.auto_submit;
```

lissyx (approved):

toolkit/crashreporter/client/app/src/logic.rs
afranchuk: This would make the following line not work (it reads `get_comment`).

afranchuk (approved):

toolkit/crashreporter/client/app/src/logic.rs
lissyx: I misread sorry

lissyx:
Looks like it works under linux?
image

PR #1149, opened 2026-07-14
Bug 2051323 - Exclude the enterprise console address when a Proxy policy is set

jporter-dev:
### Description 

Bugzilla: Bug-2051323

A broken and locked `Proxy` policy can break networking with no in-browser recovery path since all traffic is routed through the proxy by default. Updated policies are fetched from the enterprise console, so a bad proxy config could break the console connection, leaving the browser unable to retrieve the fix. 

This PR excludes the console endpoint from proxying so the console remains accessible. 

- Adds `ProxyPolicies.excludeConsoleFromProxy`, which appends the console host to `network.proxy.no_proxies_on`
- The console address is provisioned asynchronously at startup, so the implementation uses `ConsoleClient.consoleBaseURI` instead of reading the pref directly

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. Configure a proxy policy with an invalid `HTTProxy`, e.g.:
```
{
  "Mode": "manual",
  "HTTPProxy": "192.168.1.10:3128",
  "UseHTTPProxyForAllProtocols": true
}
``` 
2. Access the console URL

**Expected result:**

Console URL loads as expected and no proxy error is shown.

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -3500,14 +3500,18 @@ export var Policies = {
   },
 
   Proxy: {
-    onBeforeAddons(manager, param) {
+    async onBeforeAddons(manager, param) {
       if (param.Locked) {
         manager.disallowFeature("changeProxySettings");
       }
-      lazy.ProxyPolicies.configureProxySettings(
-        param,
-        lazy.PoliciesUtils.setDefaultPref.bind(lazy.PoliciesUtils)
+      let setDefaultPref = lazy.PoliciesUtils.setDefaultPref.bind(
+        lazy.PoliciesUtils
       );
+      lazy.ProxyPolicies.configureProxySettings(param, setDefaultPref);
+
+      if (AppConstants.MOZ_ENTERPRISE) {
+        await lazy.ProxyPolicies.excludeConsoleFromProxy(param, setDefaultPref);
+      }
     },
     onRemove(manager, oldParams) {
       if (oldParams.Locked) {
@@ -3517,6 +3521,10 @@ export var Policies = {
         oldParams,
         lazy.PoliciesUtils.setDefaultPref.bind(lazy.PoliciesUtils)
       );
+
+      if (AppConstants.MOZ_ENTERPRISE) {
+        lazy.PoliciesUtils.unsetDefaultPref("network.proxy.no_proxies_on");
+      }
     },
   },
 
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -3500,17 +3500,20 @@ export var Policies = {
   },
 
   Proxy: {
-    async onBeforeAddons(manager, param) {
+    onBeforeAddons(manager, param) {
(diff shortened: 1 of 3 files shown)


browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3463,14 +3463,18 @@ export var Policies = {
   },
 
   Proxy: {
-    onBeforeAddons(manager, param) {
+    async onBeforeAddons(manager, param) {
```
lissyx: I'm not sure this is allowed because there is no await on policies code side. You should verify this with @1rneh

lissyx (approved):
make sure `async` is safe to use here

browser/components/enterprisepolicies/tests/xpcshell/test_proxy.js
```diff
@@ -68,6 +68,54 @@ add_task(async function test_proxy_socks_and_passthrough() {
...
+  calls = [];
+  await ProxyPolicies.excludeConsoleFromProxy(
+    { Mode: "autoConfig", Passthrough: "a, b" },
+    setPref
+  );
+  Assert.deepEqual(calls, [
+    { name: NO_PROXIES, value: "a, b, console.example.com", locked: undefined },
+  ]);
+});
+
```
1rneh: This must only run in enterprise builds.

browser/components/enterprisepolicies/Policies.sys.mjs
1rneh: Nothing awaits callbacks although one would think that because there is a lot of async code to be found in policy callbacks. But it seems the policy engine is fine with that. @jporter-dev and I agreed that we should rather use the `then` method, as adding `async` to the function head could lead to the false assumption that the policy engine will await the returned promise.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3463,14 +3463,18 @@ export var Policies = {
...
       if (param.Locked) {
         manager.disallowFeature("changeProxySettings");
       }
-      lazy.ProxyPolicies.configureProxySettings(
-        param,
-        lazy.PoliciesUtils.setDefaultPref.bind(lazy.PoliciesUtils)
+      let setDefaultPref = lazy.PoliciesUtils.setDefaultPref.bind(
```
1rneh: `nit`
```suggestion
      const setDefaultPref = lazy.PoliciesUtils.setDefaultPref.bind(
```

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3480,6 +3484,10 @@ export var Policies = {
         oldParams,
         lazy.PoliciesUtils.setDefaultPref.bind(lazy.PoliciesUtils)
       );
+
+      if (AppConstants.MOZ_ENTERPRISE) {
+        lazy.PoliciesUtils.unsetDefaultPref("network.proxy.no_proxies_on");
+      }
     },
```
1rneh, line 3531: Just a note: The live support for the `Proxy` policy will be implemented and verified as part of Bug-2046454.

1rneh (approved):

PR #1150, opened 2026-07-14
Bug 2054968 - Revert "Bug 2050550 - Use leaf name without path for remoting."

lissyx:
Bugzilla: Bug-2054968


lissyx:
Reverting for regression

mkaply (approved):

mkaply:
The fix is straightforward. After doing a deeper dig, this should have been Windows only.

Let's do that instead of backing this out.

lissyx:
Fix in #1151

PR #1151, opened 2026-07-14
Make patch for 2050550 Windows specific.

mkaply:
The %TEMP% instability this works around is Windows-specific. macOS and Linux resolve their temp directory differently and don't need it.

### Description

Bugzilla: Bug-2054968

The leafName patch for remoting on 2050550 should be Windows specific.

### DIFF
--- a/toolkit/xre/nsAppRunner.cpp
+++ b/toolkit/xre/nsAppRunner.cpp
@@ -5522,15 +5522,16 @@ int XREMain::XRE_mainStartup(bool* aExitFlag,
 #  endif
 
     if (NS_SUCCEEDED(rv)) {
-#  if defined(MOZ_ENTERPRISE)
+#  if defined(MOZ_ENTERPRISE) && defined(XP_WIN)
       if (is_felt_ui()) {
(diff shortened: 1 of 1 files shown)


lissyx (approved):

mkaply:
The 2050550 fix keyed FELT remoting off the profile's leaf name instead of its full path, using a hardcoded name that includes the update channel (`felt-<channel>`), so the key stays stable no matter how the launch context resolves `%TEMP%`. That's fine on Windows, but macOS remoting (`CFMessagePort`) uses a global name namespace, and CI runs multiple jobs concurrently on the same machine. Two FELT-UI processes on the same channel, even from unrelated test files, now register the same port name and collide. Full-path keying never had this problem because each job's temp profile directory was unique.

Windows doesn't have that global-namespace issue — `FindWindowW` lookups are scoped per session, so concurrent CI jobs don't collide the way they do on Mac. This patch only restricts the leaf-name behavior to `XP_WIN`; the Windows code path is unchanged from before and after 2050550. So Windows keeps the fix that solved the original `%TEMP%` instability, and any flakiness there now is unrelated to this key logic.

PR #1152, opened 2026-07-14
Bug 2055023 - Fix intermittent neterror reset race in FELT

jporter-dev:
### Description 

Bugzilla: Bug-2055023

Bug-2031400 caused this regression by resetting the SSO pane on all `resetToLoginPage` calls, as well as making this function async. This broke the `ERROR_SOURCE.NET` reset + error paths by introducing a race between that path and the `ERROR_SOURCE.RESET` path. 

This PR implements the suggested refactor not applied that splits `resetToLoginPage` into two separate calls - one with the error and one without. Additionally, it removes the SSO reset from `resetToLoginPage` and reset the SSO browser and calls `clearSsoSessionData()` explicitly in the back button handler, which in turn removes the need for making `resetToLoginPage` async.

### DIFF
--- a/toolkit/components/felt/content/window.js
+++ b/toolkit/components/felt/content/window.js
@@ -37,26 +37,23 @@ function clearSsoSessionData() {
   });
 }
 
-async function resetToLogin(errorType = null, details = null, cause = null) {
+function resetToLoginPage() {
   cancelActiveSso?.();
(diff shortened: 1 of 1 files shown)


lissyx (approved):
r=me with green tests :)

PR #1156, opened 2026-07-15
Bug 2026616 - Fix session loss on browser restart

jporter-dev:
### Description 

Bugzilla: Bug-2026616

When a FELT-managed browser requests a restart, `FeltProcessParent` force-kills the browser process immediately and then relaunches it. This races the browser's graceful shutdown before its session store has flushed to disk, so the relaunch comes up looking like a crash instead of restoring the previous session.

This PR updates the `felt-firefox-restarting` branch in `FeltProcessParent.sys.mjs` to wait for the child browser's `proc.exitPromise` to resolve rather than calling `proc.kill()` before relaunching a new child browser. This allows the graceful shutdown to write the session to disk before relaunching. 

In addition to fixing the session persistence, this PR also moves `AppShutdown::MaybeDoRestart` guard from wrapping the call in `nsAppRunner.cpp` to guarding directly in the `AppShutdown::MaybeDoRestart` function itself in `AppShutdown.cpp`. A [second unguarded call](https://searchfox.org/enterprise-main/source/xpcom/base/AppShutdown.cpp#258) to this method was causing a second browser window to appear on restarts.

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1.  In `about:preferences`, ensure "Open previous windows and tabs" is unchecked
2. Start the FELT browser and open a few tabs
3. Trigger a restart. Options:
    - Type "Restart" into the urlbar and click "Restart Firefox Enterprise"
    - Navigate to `about:restartrequired` and click "Restart Firefox Enterprise"
    - Navigate to `about:profiles` and click "Restart Normally"
    - In the browser console:
   `Services.startup.quit(Ci.nsIAppStartup.eRestart | Ci.nsIAppStartup.eAttemptQuit);`
    - In the browser console: `Services.obs.notifyObservers(null, "felt-update-ready");`, then click "Update Available - Restart Now" 
5. Wait for the browser to reopen

**Expected result:**

The browser exits cleanly and relaunches once, with the previous session/tabs restored, and no extra browser windows or tabs.

### DIFF
--- a/testing/enterprise/test_felt_restart_persists_session.py
+++ b/testing/enterprise/test_felt_restart_persists_session.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_tests import FeltTests
+from marionette_driver.errors import (
+    NoSuchWindowException,
+    UnknownException,
+)
+
+
+class AppRestartPersistsSession(FeltTests):
+    """A FELT restart must preserve the open tabs.
+
+    Seeds tabs, restarts, and reconnects to the relaunched browser
+    to confirm it resumed the session with those tabs open.
+    """
+
+    def test_restart_persists_session(self):
+        super().run_felt_base()
+
+        session_tabs = [
+            "about:support",
+            "about:policies",
+            f"http://localhost:{self.console_port}/ping",
+        ]
+
+        self.connect_child_browser()
+        self._browser_pid = self._child_driver.session_capabilities["moz:processID"]
+
+        self._open_child_tabs(session_tabs)
+        opened = self._child_tab_urls()
+        assert set(session_tabs).issubset(opened), (
+            f"Tabs were not open before the restart: {opened}"
+        )
+
+        self._child_longwait.until(
+            lambda _: set(session_tabs).issubset(self._child_session_urls()),
+            message="SessionStore did not record the tabs before restart",
+        )
+
+        self.issue_child_restart()
+        self.wait_process_exit(self._browser_pid)
+
+        self._logger.info("Connecting to new browser")
+        self.connect_child_browser()
+        new_browser_pid = self._child_driver.session_capabilities["moz:processID"]
+        assert new_browser_pid != self._browser_pid, (
+            f"Expected a new process, still {new_browser_pid}"
+        )
+
+        self._child_longwait.until(
+            lambda _: set(session_tabs).issubset(set(self._child_tab_urls())),
+            message="Relaunched browser did not restore the session tabs",
+        )
+
+    def issue_child_restart(self):
+        try:
+            self._logger.info("Issuing restart being done by felt")
+            self._child_driver.set_context("chrome")
+            self._child_driver.execute_script(
+                "Services.startup.quit(Ci.nsIAppStartup.eRestart | Ci.nsIAppStartup.eAttemptQuit);"
+            )
+        except UnknownException:
+            self._logger.info("Received expected UnknownException")
+        except NoSuchWindowException:
+            self._logger.info("Received expected NoSuchWindowException")
+        except OSError:
+            self._logger.info(
+                "Firefox quit before execute_script returned, no data received over Marionette socket"
+            )
+
+    def _open_child_tabs(self, urls):
+        for url in urls:
+            self.open_tab_child(url)
+
+    def _child_tab_urls(self):
+        self._child_driver.set_context("chrome")
+        return self._child_driver.execute_script(
+            "return Array.from(gBrowser.tabs).map(t => t.linkedBrowser.currentURI.spec);"
+        )
+
+    def _child_session_urls(self):
+        self._child_driver.set_context("chrome")
+        return self._child_driver.execute_script(
+            """
+            const { SessionStore } = ChromeUtils.importESModule(
+              "resource:///modules/sessionstore/SessionStore.sys.mjs"
+            );
+            const state = JSON.parse(SessionStore.getBrowserState());
+            return state.windows.flatMap(w =>
+              w.tabs.flatMap(t => (t.entries || []).map(e => e.url))
+            );
+            """
+        )
--- a/testing/enterprise/test_felt_restart_works.py
+++ b/testing/enterprise/test_felt_restart_works.py
@@ -8,6 +8,7 @@ import sys
 
 sys.path.append(os.path.dirname(__file__))
 
+import psutil
 from felt_tests import FeltTests
 from marionette_driver.errors import (
     NoSuchWindowException,
@@ -16,7 +17,7 @@ from marionette_driver.errors import (
 
 
 class AppRestartWorks(FeltTests):
-    def test_app_resart_works(self):
+    def test_app_restart_works(self):
         super().run_felt_base()
         self.run_felt_perform_restart()
         self.run_felt_restart_new_process()
@@ -61,8 +62,31 @@ class AppRestartWorks(FeltTests):
             f"PID changed from {self._browser_pid} to {new_browser_pid}"
         )
 
+        felt_ui_pid = self._driver.session_capabilities["moz:processID"]
+        target_exe = psutil.Process(felt_ui_pid).exe()
+        felt_browser_pids = set(self._get_felt_browser_pids(target_exe))
+        unexpected = felt_browser_pids - {new_browser_pid}
+        assert not unexpected, (
+            f"Extra FELT browser process(es) after restart (double relaunch?): {unexpected}"
+        )
+        assert new_browser_pid in felt_browser_pids, (
+            f"Relaunched browser PID {new_browser_pid} is not running"
+        )
+
         self._logger.info(f"Closing new browser with PID {new_browser_pid}")
         self._child_driver.set_context("chrome")
         self._child_driver.execute_script(
             "Services.startup.quit(Ci.nsIAppStartup.eForceQuit);"
         )
+
+    def _get_felt_browser_pids(self, target_exe):
+        pids = []
+        for proc in psutil.process_iter(["pid", "exe", "cmdline"]):
+            try:
+                cmdline = proc.info["cmdline"] or []
+                # Check for FELT browser process with -felt.
+                if proc.info["exe"] == target_exe and "-felt" in cmdline:
+                    pids.append(proc.info["pid"])
+            except (psutil.NoSuchProcess, psutil.ZombieProcess, psutil.AccessDenied):
+                continue
+        return pids
--- a/toolkit/components/felt/content/FeltProcessParent.sys.mjs
+++ b/toolkit/components/felt/content/FeltProcessParent.sys.mjs
@@ -16,6 +16,8 @@ ChromeUtils.defineESModuleGetters(lazy, {
     "resource://gre/modules/enterprise/EnterpriseCommon.sys.mjs",
   FeltCommon: "chrome://felt/content/FeltCommon.sys.mjs",
   FeltStorage: "resource://gre/modules/enterprise/FeltStorage.sys.mjs",
+  setTimeout: "resource://gre/modules/Timer.sys.mjs",
+  clearTimeout: "resource://gre/modules/Timer.sys.mjs",
 });
 
 if (lazy.isBuildAppBrowser()) {
@@ -157,6 +159,12 @@ export class FeltProcessParent extends JSProcessActorParent {
           }
 
           case "felt-firefox-restarting": {
+            if (gFeltProcessParentInstance) {
+              gFeltProcessParentInstance.restartReported = true;
+              gFeltProcessParentInstance.firefox = null;
+            }
+
+            const proc = gFeltProcessParentInstance?.proc;
             const restartDisabled = Services.prefs.getBoolPref(
               "enterprise.disable_restart",
               false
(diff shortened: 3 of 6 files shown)


toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -324,6 +315,94 @@ export class FeltProcessParent extends JSProcessActorParent {
     };
   }
 
+  /**
+   * Stops the current browser for a restart once it has persisted its session.
+   * We wait for the browser's felt-firefox-session-flushed signal (with a
+   * bounded fallback) rather than a full clean shutdown, which can hang, then
+   * stop the process so the relaunch is deterministic and the session is never
+   * lost. Resolves once the stopped process has been reaped.
```
lissyx: That comment makes me worried: why would a full clean shutdown hang? There are no good reason.

toolkit/xre/nsAppRunner.cpp
```diff
@@ -7275,14 +7275,7 @@ int XREMain::XRE_main(int argc, char* argv[], const BootstrapConfig& aConfig) {
...
 
-#if defined(MOZ_ENTERPRISE)
-  // Do not restart when running as Felt client (i.e. Enterprise browser)
-  if (!is_felt_browser()) {
-#endif
-    mozilla::AppShutdown::MaybeDoRestart();
-#if defined(MOZ_ENTERPRISE)
-  }
-#endif
+  mozilla::AppShutdown::MaybeDoRestart();
```
lissyx, line 7284: I'm certain I did the change here because there were cases that were not handled with your solution

testing/enterprise/test_felt_restart_persists_session.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class AppRestartPersistsSession(FeltTests):
+    """A FELT restart must preserve the open tabs.
+
+    Seeds tabs, restarts, and reconnects to the relaunched browser
+    to confirm it resumed the session with those tabs open.
+    """
+
+    SESSION_TABS = ["about:support", "about:policies"]
```
lissyx: nit: `about:` pages may have specific behavior, we may want to have a normal page (`/ping` from the console mock?)

toolkit/xre/nsAppRunner.cpp
lissyx: @jporter-dev can you elaborate why this part is required ?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -188,19 +193,18 @@ export class FeltProcessParent extends JSProcessActorParent {
...
-                  gFeltProcessParentInstance.restartReported = true;
-                  gFeltProcessParentInstance.firefox = null;
+                const proc = gFeltProcessParentInstance?.proc;
+                if (proc) {
                   lazy.log.debug(
-                    `ParentProcess: Killing Firefox PID=${gFeltProcessParentInstance.proc.pid}`
+                    `ParentProcess: Waiting for Firefox PID=${proc.pid} to exit for restart`
                   );
-                  gFeltProcessParentInstance.proc
-                    .kill()
```
lissyx, line 200: As much as I can recall, the `kill()` here was required because we were not performing a restart anymore due to the guarding we had in place on `MaybeDoRestart()` before `XRE_DeinitCommandLine();`

lissyx: (and there were cases where the process would stay around?)

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -227,11 +231,16 @@ export class FeltProcessParent extends JSProcessActorParent {
                       }
                     })
                     .catch(err => {
-                      lazy.log.debug(`ParentProcess: Kill failed: ${err}`);
+                      lazy.log.debug(
+                        `ParentProcess: Wait for exit failed: ${err}`
+                      );
```
lissyx, line 257: This is likely a condition we want to do more than just `debug()` because if `exitPromise` failed we are likely in a very bad posittion ?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -227,11 +231,16 @@ export class FeltProcessParent extends JSProcessActorParent {
...
                     });
                 } else {
-                  lazy.log.debug(`ParentProcess: No proc to kill!`);
+                  lazy.log.debug(`ParentProcess: No proc to wait for!`);
                 }
+              })
+              .catch(err => {
+                lazy.log.debug(`ParentProcess: Restart failed: ${err}`);
```
lissyx: SImilar, it's worth more than just a `debug` level, and likely requires more handling?

testing/enterprise/test_felt_restart_works.py
```diff
@@ -61,8 +62,38 @@ def run_felt_restart_new_process(self):
...
+        felt_ui_pid = self._driver.session_capabilities["moz:processID"]
+        target_exe = psutil.Process(felt_ui_pid).exe()
+        felt_browser_pids = set(self._get_felt_browser_pids(target_exe))
+        unexpected = felt_browser_pids - {new_browser_pid}
+        assert not unexpected, (
+            f"Extra FELT browser process(es) after restart (double relaunch?): {unexpected}"
+        )
+        assert new_browser_pid in felt_browser_pids, (
+            f"Relaunched browser PID {new_browser_pid} is not running"
+        )
```
lissyx, line 74: Is it out of safety or did you hit regressions ?

testing/enterprise/test_felt_restart_works.py
```diff
@@ -61,8 +62,38 @@ def run_felt_restart_new_process(self):
...
+    def _get_felt_browser_pids(self, target_exe):
+        pids = []
+        for proc in psutil.process_iter(["pid", "exe", "cmdline"]):
+            try:
+                if proc.info["exe"] != target_exe:
+                    continue
+                cmdline = proc.info["cmdline"]
+                if not cmdline:
+                    continue
+                if "-contentproc" in cmdline:
```
lissyx: While `-contentproc` seems to be there always, it may not be true. But `-parentPid` looks better

```suggestion
                if "-parentPid" in cmdline:
```

toolkit/xre/nsAppRunner.cpp
jporter-dev: The only other call site I could find was in [AppShutdown.cpp](https://searchfox.org/enterprise-main/source/xpcom/base/AppShutdown.cpp#258) for fast shutdowns, so I _think_ that'd be the only other case? 

The ungated fast shutdown path above is what was triggering the second browser relaunch when removing the `.kill()` to allow the graceful shutdown without relaunch. This was suppressed by FELT force-killing the child browser without letting it complete its shutdown process

testing/enterprise/test_felt_restart_works.py
```diff
@@ -61,8 +62,38 @@ def run_felt_restart_new_process(self):
...
+            f"Relaunched browser PID {new_browser_pid} is not running"
+        )
+
         self._logger.info(f"Closing new browser with PID {new_browser_pid}")
         self._child_driver.set_context("chrome")
         self._child_driver.execute_script(
             "Services.startup.quit(Ci.nsIAppStartup.eForceQuit);"
         )
+
+    def _get_felt_browser_pids(self, target_exe):
```
lissyx, line 82: The naming is confusing: are we querying for felt browser processes, or for felt processes ? From the name I get we want felt browser. But the checks `if "-felt" not in cmdline:` would exclude felt browser main process .... ?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: Yeah good call, the logs in the `catch` here and below should probably be a `error()` and we can `proc.kill()`?

toolkit/xre/nsAppRunner.cpp
lissyx: > The ungated fast shutdown path above is what was triggering the second browser relaunch

The code path it was applying was supposed to terminate gracefully the process. You are stating it was in fact performing a relaunch in our back?

When you say "second browser relaunch", it means we had:
 - a relaunch from felt that detected the shutdown,
 - a relaunch from the browser itself,

Right?

It would be nice to know **where** this second relaunch was coming from. Did you traced it to an explicite call to `mozilla::AppShutdown::MaybeDoRestart()` ?

testing/enterprise/test_felt_restart_persists_session.py
jporter-dev: Good idea, I can add that to the list of tabs to open/restore

testing/enterprise/test_felt_restart_persists_session.py
```diff
@@ -0,0 +1,107 @@
...
+
+        self._logger.info("Connecting to new browser")
+        self.connect_child_browser()
+        new_browser_pid = self._child_driver.session_capabilities["moz:processID"]
+        assert new_browser_pid != self._browser_pid, (
+            f"Expected a new process, still {new_browser_pid}"
+        )
+
+        # Assert against SessionStore state, not live tab URIs: lazily restored
+        # background tabs keep about:blank as currentURI until selected.
```
lissyx: The wording is super confusing. It's just saying that until a restored tab is opened it's URI is `about:blank` ?

testing/enterprise/test_felt_restart_persists_session.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_restart_persists_session(self):
+        super().run_felt_base()
+
+        self.connect_child_browser()
+        self._browser_pid = self._child_driver.session_capabilities["moz:processID"]
+
+        self._open_child_tabs(self.SESSION_TABS)
+        opened = self._child_tab_urls()
+        assert set(self.SESSION_TABS).issubset(opened), (
+            f"Seeded tabs were not open before the restart: {opened}"
```
lissyx:
```suggestion
            f"Tabs were not open before the restart: {opened}"
```

testing/enterprise/test_felt_restart_persists_session.py
```diff
@@ -0,0 +1,107 @@
...
+        except UnknownException:
+            self._logger.info("Received expected UnknownException")
+        except NoSuchWindowException:
+            self._logger.info("Received expected NoSuchWindowException")
+        except OSError:
+            self._logger.info(
+                "Firefox quit before execute_script returned, no data received over Marionette socket"
+            )
+        finally:
+            self._manually_closed_child = True
```
lissyx: We do not need this ?

testing/enterprise/test_felt_restart_persists_session.py
```diff
@@ -0,0 +1,107 @@
...
+        self._child_longwait.until(
+            lambda _: set(self.SESSION_TABS).issubset(set(self._child_session_urls())),
+            message="Relaunched browser did not restore the session tabs",
+        )
+
+        self._logger.info(f"Closing new browser with PID {new_browser_pid}")
+        self._child_driver.set_context("chrome")
+        self._child_driver.execute_script(
+            "Services.startup.quit(Ci.nsIAppStartup.eForceQuit);"
+        )
```
lissyx: Once `self._manually_closed_child = True` is removed this is also not removed anymore.

testing/enterprise/test_felt_restart_persists_session.py
```diff
@@ -0,0 +1,107 @@
...
+        assert new_browser_pid != self._browser_pid, (
+            f"Expected a new process, still {new_browser_pid}"
+        )
+
+        # Assert against SessionStore state, not live tab URIs: lazily restored
+        # background tabs keep about:blank as currentURI until selected.
+        self._child_longwait.until(
+            lambda _: set(self.SESSION_TABS).issubset(set(self._child_session_urls())),
+            message="Relaunched browser did not restore the session tabs",
+        )
```
lissyx, line 61: As much as I can recall, we should be presented with a message asking to restore tabs, right?

I'd prefer the test to completely exercize it: find the message, perform restore, navigate the opened tabs and verify all the expected ones are restored ?

testing/enterprise/test_felt_restart_persists_session.py
jporter-dev: This comment is old, I'll remove. Previously was checking state directly because the second browser window made it untestable

testing/enterprise/test_felt_restart_persists_session.py
jporter-dev: I think that is an artifact of us doing the `proc.kill()`. I tested the browser restart paths in the PR description in nightly, and those paths didn't show that message. 

I think since `proc.kill()` was racing the clean shutdown, resulting in an unclean shutdown, it made it look like we were crashing during shutdown so that's why we got the "recover session" window.

toolkit/xre/nsAppRunner.cpp
jporter-dev: Yes, FELT triggered a relaunch and the [ungated call](https://searchfox.org/enterprise-main/source/xpcom/base/AppShutdown.cpp#258) in `AppShutdown.cpp` to `MaybeDoRestart` launched the additional child. 

Previously this behavior was masked by the `proc.kill()` since it raced and ultimately interrupted the restart/shutdown.

The additional call to `MaybeDoRestart` that I linked above is for fast shutdowns, controlled by `toolkit.shutdown.fastShutdownStage` (defaults to 1). Reverting my change here, but continuing to await the `proc.exitPromise` in `FeltProcessParent`, and setting `toolkit.shutdown.fastShutdownStage=0` forces it to take the original gated path in `nsAppRunner.cpp`, resulting in no additional window on restart.

testing/enterprise/test_felt_restart_works.py
jporter-dev: The regression was the second browser window launching

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: Do you remember the cases where the process would stay around? I wonder if it could be related to the other `MaybeDoRestart` call? Happy to test out these other cases.

testing/enterprise/test_felt_restart_works.py
jporter-dev: Updated this to be more clear by explicitly checking for `exe` and `-felt` to only grab felt browser processes

testing/enterprise/test_felt_restart_persists_session.py
jporter-dev: Thanks, removed! Was leftover from when I was testing with the second window launch bug

testing/enterprise/test_felt_restart_persists_session.py
lissyx: You're right but why do we check the list of URLs here again with `self._child_session_urls()` ? We're poking into the session restore code, instead of checking that the tabs are reopened. We should just iterate on all open tabs and verify we have what we expected ?

testing/enterprise/test_felt_restart_works.py
lissyx: ok fair enough

toolkit/components/felt/content/FeltProcessParent.sys.mjs
lissyx: Except we'd need a different error message, or no error message at all ? We are supposed to restart anyway. We should just kill indeed. I cannot remember if `exitPromise` has a timeout or if it can block though ...

toolkit/xre/nsAppRunner.cpp
lissyx: That's weird the call diagram shows `MaybeFastShutdown` is called from content process ?

lissyx: I'd really like to have a stack to understand what was the trigger point here

toolkit/components/felt/content/FeltProcessParent.sys.mjs
lissyx: Unfortunately no, I cannot remember

lissyx (approved):
r=me with the fix of the test to verify the URL loaded and not the ccontent of session store, making sure `exitPromise` will at some point return and of course handling the error message suggestion

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: `exitPromise` only resolves, it never rejects :( and there was no timeout. I ended up adding a force-kill timeout as a safety net for when `exitPromise` doesn't resolve. I set it longer than `toolkit.asyncshutdown.crash_timeout` so it only acts as a last-resort failsafe.

PR #1162, opened 2026-07-16
Bug 2021342 - Keep the enterprise-managed SDR unlocked and fail closed on unlock failure

beurdouche:
The enterprise storage-encryption feature sets the NSS internal token's primary password to the console-supplied primarySecret and unlocks it once at startup. Two gaps let a primary-password prompt reach the user for a secret they cannot supply (about:logins, password import, Sync on startup):

- EnterpriseStorageEncryption.load set the password on a fresh token via changePassword("", primarySecret) but never logged the token in (PK11_ChangePW does not authenticate), and its only error handling was console.error. Make the login explicit and symmetric across the fresh and returning branches, verify pk11token.isLoggedIn, and fail the launch with a dedicated exit code (FeltEncryptionExitCode_UnlockFailed) if any step fails instead of continuing half-unlocked.

- Any CLEAR_AUTH_TOKENS clear (Clear Recent History, sanitize, site-data) routed through SecretDecoderRing::Logout / LogoutAndTeardown ran PK11_LogoutAll, re-locking the internal token mid-session with no re-login. Skip the token logout while the primary password is enterprise-managed; the rest of the teardown (TLS cache, connection pruning) still runs.

### DIFF
--- a/browser/components/BrowserGlue.sys.mjs
+++ b/browser/components/BrowserGlue.sys.mjs
@@ -1120,72 +1120,65 @@ BrowserGlue.prototype = {
             false
           ),
         task: async () => {
-          // Get the primary secret from the console backend
-          // The API returns { data: "secret_value" }
+          // A managed browser that cannot unlock its encrypted storage must not
+          // keep running: it would prompt for a primary secret the user does not
+          // know, or operate against a profile it cannot decrypt. Fail the launch
+          // with a dedicated exit code (Bug 2021342), mirroring the launcher-side
+          // abort in FeltProcessParent.
+          const fail = (msg, e) => {
+            console.error(
+              `EnterpriseStorageEncryption.load: ${msg}${e ? ": " + e : ""}`
+            );
+            Services.startup.quit(
+              Ci.nsIAppStartup.eForceQuit,
+              Ci.nsIFelt.FeltEncryptionExitCode_SdrTokenUnlockFailed
+            );
+          };
+
+          // The API returns { data: "secret_value" }.
           let primarySecret;
           try {
-            const payload = await lazy.ConsoleClient.getPrimarySecret();
-            primarySecret = payload.data;
-            if (!primarySecret) {
-              console.error(
-                "EnterpriseStorageEncryption.load: No data field in payload:",
-                payload
-              );
-              return;
-            }
+            primarySecret = (await lazy.ConsoleClient.getPrimarySecret()).data;
           } catch (e) {
-            console.error(
-              "EnterpriseStorageEncryption.load: Failed to get primary secret:",
-              e
-            );
+            fail("Failed to get primary secret", e);
+            return;
+          }
+          if (!primarySecret) {
+            fail("No primary secret in payload");
             return;
           }
 
-          // Load the PK11 token
           let pk11token;
           try {
             pk11token = Cc[
               "@mozilla.org/security/internalkeytoken;1"
             ].createInstance(Ci.nsIPKCS11Token);
           } catch (e) {
-            console.error(
-              "EnterpriseStorageEncryption.load: Error getting PK11 token: " + e
-            );
+            fail("Error getting PK11 token", e);
             return;
           }
 
-          // Check if the PK11 token needs initialization
+          // Ensure the internal token's password is the primarySecret.
           if (!pk11token.hasPassword) {
-            // Token doesn't need login (empty password), set it to primarySecret
             try {
               await pk11token.changePassword("", primarySecret);
             } catch (e) {
-              console.error(
-                "EnterpriseStorageEncryption.load: Failed to change password from empty to primarySecret: " +
-                  e
-              );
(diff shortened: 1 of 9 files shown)


browser/components/BrowserGlue.sys.mjs
```diff
@@ -1119,72 +1119,64 @@ BrowserGlue.prototype = {
...
+              return fail("Failed to set the primary secret on the token", e);
             }
+          }
 
-            if (!isPasswordValid) {
-              console.error(
-                "EnterpriseStorageEncryption.load: Password against the PK11 token is not valid"
-              );
+          // Setting the password (PK11_ChangePW) does not leave the token logged
+          // in, and on a returning profile the token starts this session logged
```
simon-friedberger: not sure what "on a returning profile" is supposed to mean

simon-friedberger: I would probably throw out this comment. It's clear what the code is trying,.

security/manager/ssl/SecretDecoderRing.cpp
```diff
@@ -308,16 +308,34 @@ SecretDecoderRing::Login(const nsACString& password, bool* success) {
...
+  }
   mozilla::net::SSLTokensCache::ClearSessionCacheAndTokens();
   return NS_OK;
 }
 
 NS_IMETHODIMP
 SecretDecoderRing::LogoutAndTeardown() {
-  PK11_LogoutAll();
+  if (!EnterpriseManagesPrimaryPassword()) {
+    PK11_LogoutAll();
```
simon-friedberger, line 337: I'm a little concerned that something could break if we stop logging out the token here. That's a big change in functionality. Have you run the tests?

simon-friedberger: I can't find any uses of this outside of tests but it looks like it will definitely break some tests.

simon-friedberger (approved):
lgtm as long as the tests all pass, I suspect they don't so those would need fixing

browser/components/BrowserGlue.sys.mjs
```diff
@@ -1119,72 +1119,64 @@ BrowserGlue.prototype = {
             false
           ),
         task: async () => {
-          // Get the primary secret from the console backend
-          // The API returns { data: "secret_value" }
+          // A managed browser that cannot establish its SDR unlock must not keep
```
lissyx: What does `SDR` stands for?

toolkit/components/felt/rust/nsIFelt.idl
```diff
@@ -10,6 +10,10 @@
 interface nsIFelt : nsISupports {
   const unsigned long FeltEncryptionExitCode_Archive = 0x93;
   const unsigned long FeltEncryptionExitCode_Delete = 0x94;
+  // The spawned browser could not establish its SDR unlock (no primarySecret,
+  // or the internal token could not be logged in). Exit with this code rather
+  // than run against a profile it cannot decrypt (Bug 2021342).
+  const unsigned long FeltEncryptionExitCode_UnlockFailed = 0x95;
```
lissyx: This should be handled gracefully within https://searchfox.org/enterprise-main/rev/7a857888019ca1ba7dc9181679ebbe4068810c73/toolkit/components/felt/content/FeltProcessParent.sys.mjs#574

Especially we should show a clear error message of what happened.

browser/components/BrowserGlue.sys.mjs
simon-friedberger: SecretDecoderRing https://searchfox.org/firefox-main/source/security/manager/ssl/SecretDecoderRing.cpp

beurdouche:
Looking into this now.

security/manager/ssl/SecretDecoderRing.cpp
beurdouche: Good call to double-check. The guard is double-gated — #if defined(MOZ_ENTERPRISE) and StaticPrefs::security_storage_encryption_enabled(), which defaults false with no enterprise override (StaticPrefList.yaml). No test in the SDR / cleardata / passwordmgr / enterprise suites sets that pref true except a policy pref-lock test that never calls logout, so in every current run EnterpriseManagesPrimaryPassword() returns false and PK11_LogoutAll() still runs.

I've also (a) narrowed the guard to LogoutAndTeardown() only and reverted Logout() (no production caller), (b) added an enterprise-gated regression test test_sdr_enterprise_logout.js (run-if = ["enterprise"]) asserting the token stays logged in through logoutAndTeardown on an enterprise build, and (c) fixed a consistent-return lint error in the task. Relying on PR CI for the full run.

toolkit/components/felt/rust/nsIFelt.idl
beurdouche: Handled the error

lissyx:
gtest failures are unrelated

PR #1137, opened 2026-07-10
Bug 2031400 - Add back to login button to FELT

jporter-dev:
### Description 

Bugzilla: Bug-2031400

Adds a "Back to login" button to the FELT login window. It appears in a new footer bar (alongside the version and "powered by" text) once the user reaches the SSO pane, and is hidden everywhere else.

Clicking it abandons the in-flight SSO attempt and returns to the email pane. The SSO navigation is cancelled, the private-browsing session is cleared (so a retry starts fresh rather than skipping to MFA), and the reused `<browser>` is
reset to `about:blank`. This reset logic is shared with the existing error/timeout path.

> Note: Final UX for this is still pending, but changes should be minor if the button remains in a newly added footer.

---

### Screenshots 

Initial email pane

_Initial email pane_

SSO login pane with back button

_SSO login pane with back button_

---

### Testing 

* [x] Added tests
* [x] Manual testing performed

**Steps to verify changes:**
1. Launch FELT and enter an email to reach the SSO pane
2. Click "Back to login"
3. Enter an email to reach the SSO pane again

**Expected result:**

Upon pressing the back button, the SSO MFA is cancelled, SSO pane is reset, and the next login attempt starts fresh

PR #1163, opened 2026-07-17
Bug 2054803 - Update FELT login and back button UX

jporter-dev:
### Description 

Bugzilla: Bug-2054803

Refresh the FELT login window layout and improve the back button UX based on the updated design.

- Reverts UX changes from #1137 
- Fixes the right pane visibly widening when moving from the email step to the SSO step due to added card margins
- Adds a `--background-color-tabstrip` token based on design

---

* Depends on: #1137 

---

### Screenshots 

FELT Login Page

_FELT Login Page_

FELT SSO Page

_FELT SSO Page_

felt-back-button-ux-errors

_FELT with multiple errors_

felt-back-button-ux-updater

_FELT with update checker_

felt-back-button-ux-updater2

_FELT with update downloader_

felt-back-button-ux-updates-message

_FELT with updated message_

### DIFF
--- a/toolkit/components/felt/content/felt.xhtml
+++ b/toolkit/components/felt/content/felt.xhtml
@@ -179,68 +179,6 @@
               >
               </span>
             </div>
-            <div class="felt-browser-error">
-              <moz-message-bar
-                class="felt-browser-error-multiple-crashes is-hidden"
-                data-l10n-id="felt-browser-error-multiple-crashes2"
-                type="error"
-                dismissable="true"
-              ></moz-message-bar>
-              <moz-message-bar
-                class="felt-browser-error-token-refresh-failed is-hidden"
-                data-l10n-id="felt-browser-error-token-refresh-failed"
-                type="info"
-              ></moz-message-bar>
-              <moz-message-bar
-                class="felt-browser-error-launch-failure is-hidden"
-                data-l10n-id="felt-browser-error-launch-failure"
-                type="error"
-                dismissable="true"
-              ></moz-message-bar>
-              <moz-message-bar
-                class="felt-error-primary-secret is-hidden"
-                data-l10n-id="felt-error-primary-secret"
-                type="error"
-                dismissable="true"
-              ></moz-message-bar>
-              <moz-message-bar
-                class="felt-browser-error-no-network is-hidden"
-                data-l10n-id="felt-browser-error-no-network"
-                type="error"
-                dismissable="true"
-              >
-                <span slot="message" class="felt-browser-error-details"></span>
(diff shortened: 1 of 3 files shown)


toolkit/components/felt/content/styles/felt-tokens.css
```diff
@@ -22,3 +22,7 @@
     --felt-update-checkmark-img-url: url("chrome://felt/content/assets/enterprise-update-checkmark-hcm.svg");
   }
 }
+
+:root {
+  --background-color-tabstrip: light-dark(var(--color-gray-20), var(--color-gray-90));
```
1rneh: `nit:` Can we add a `--felt-` prefix?

toolkit/components/felt/content/styles/felt.css
1rneh: Setting `background` is redundant as they are both nested in `.felt-card` which already sets the background to `var(--background-color-box)`.

toolkit/components/felt/content/felt.xhtml
```diff
@@ -115,155 +115,157 @@
...
+              </form>
+            </div>
+            <div class="felt-updates-message is-hidden">
+              <span
+                class="felt-updates-uptodate"
+                data-l10n-id="felt-updates-uptodate"
+              >
+              </span>
+            </div>
+            <div class="felt-browser-error">
```
1rneh: I think `.felt-browser-error` shouldn't be nested in `.felt-card` as it should remain anchored at the top of `.felt-main`.

1rneh (changes requested):
Thank you, really happy about the new back button!

The updated design displays all chrome content inside of `.felt-card`. To me this makes the chrome content look like remote website content. It feels a bit counter intuitive to me, but as talked about before I think I'm looking at this a bit too much through the development lens :)

Have you verified that all chrome content nested in `.felt-card` remains properly aligned? (E.g. see my comment about `.felt-browser-error`)

toolkit/components/felt/content/styles/felt-tokens.css
jporter-dev: Yeah that makes sense to me. I had just taken the name directly from Figma.

toolkit/components/felt/content/felt.xhtml
jporter-dev: Ah didn't notice the overlap with the error. Thanks, moved it! Also updated screenshots with the errors shown as well as a couple of the updater pieces.

1rneh (approved):
Looks good to me now! And thanks for all the additional screenshots!

PR #1167, opened 2026-07-17
Bug 2054570 - Add progress indicator to FELT SSO browser panel

jporter-dev:
### Description 

Bugzilla: Bug-2054570

Adds a status panel to the bottom left corner of the FELT SSO login pane that shows the network progress (ex: `Connecting to...`, `Transferring data from...`) while the SSO browser is loading a page. This mirrors Firefox's behavior and gives users feedback when page load is slower.

---

### Screenshots 

felt-statuspanel

_FELT with status panel_

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. Launch FELT
2. Progress through login flow

**Expected result:**

Statusbar in bottom left shows progress updates as they fire.

### DIFF
--- a/toolkit/components/felt/content/window.js
+++ b/toolkit/components/felt/content/window.js
@@ -28,6 +28,28 @@ Services.obs.notifyObservers(window, "browser-delayed-startup-finished");
 
 let cancelActiveSso = null;
 
+// define as a var so it is exposed as window.FeltStatusPanel for the marionette test
+var FeltStatusPanel = {
+  get _panel() {
+    return document.getElementById("felt-statuspanel");
+  },
+  get _label() {
+    return document.getElementById("felt-statuspanel-label");
+  },
+  update(text) {
+    if (text) {
+      this._label.textContent = text;
+      this._panel.classList.remove("is-hidden");
+    } else {
+      this._panel.classList.add("is-hidden");
+      this._label.textContent = "";
+    }
+  },
+  clear() {
+    this.update("");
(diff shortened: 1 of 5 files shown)


toolkit/components/felt/content/window.js
```diff
@@ -28,6 +28,28 @@ Services.obs.notifyObservers(window, "browser-delayed-startup-finished");
 
 let cancelActiveSso = null;
 
+// define as a var so it is exposed as window.FeltStatusPanel for the marionette test
+var FeltStatusPanel = {
```
1rneh: Although we don't enforce the `no-var` lint rule, we could still prefer `const` where possible. Hence we could declare `FeltStatusPanel` as `const` and explicitly attach it with `window.FeltStatusPanel = FeltStatusPanel;`. This still creates a global (a property on window), but the intent is explicit rather than relying on the implicit "top-level var becomes a global" behavior. It's worth it to keep the short comment noting it's exposed for the marionette test.

1rneh (approved):
Apart from my comment below, it's good to go!

toolkit/components/felt/content/window.js
jporter-dev: I had the same thought, wasn't sure how to best handle it, but `const` + `window` is nicer to allow us to prefer `const`. 

Looks like we actually expose a few other things to `window` in `setupMarionetteEnvironment` so I've moved it there :)

PR #1175, opened 2026-07-21
Bug 2056556 - Require local opt-in to sync passwords and payment methods

beurdouche:
In enterprise builds the Sync key is provisioned by the management console, so synced data is encrypted at rest but not end-to-end encrypted (the organization can access it).

Prevent the Sync enterprise policy from force-enabling the passwords and payment-method engines. They now default off in enterprise builds and can only be turned on by an explicit local choice; the policy may still disable (lock off) these engines. Other engines are unaffected.

Add a red warning in about:preferences - in both the redesigned and legacy Sync panes and in the "Choose what to sync" dialog - stating that synced data is encrypted at rest but not end-to-end encrypted, with passwords and payment methods called out as opt-in only.

Also fixes stale helper imports in SyncPolicy.sys.mjs and adds xpcshell and browser tests (gated to enterprise builds).

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -3022,6 +3022,16 @@ pref("browser.crashReports.cleanupCheck.enabled", true);
 // Checkbox in sync options for credit card data sync service
 pref("services.sync.engine.creditcards.available", true);
 
+#ifdef MOZ_ENTERPRISE
+// In enterprise builds, Sync data is encrypted at rest but not end-to-end
+// encrypted (the management console holds the key), so passwords and payment
+// methods must never sync unless the user explicitly opts in locally. Default
+// both engines off (upstream defaults passwords on); leave `.available` on so
+// the user can still turn them on themselves.
+pref("services.sync.engine.passwords", false);
+pref("services.sync.engine.creditcards", false);
+#endif
+
 // Whether or not to restore a session with lazy-browser tabs.
 pref("browser.sessionstore.restore_tabs_lazily", true);
 
--- a/browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
+++ b/browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
@@ -32,6 +32,13 @@ const ENGINE_PREFS = {
 
 const SYNC_FEATURE = "change-sync-state";
 
+// Engines whose data is especially sensitive and, in enterprise builds, is
+// encrypted at rest but not end-to-end encrypted (the management console holds
+// the key). Policy may turn these off (following the usual `Locked` handling),
+// but must never turn them on: enabling them requires an explicit local opt-in
+// from the user.
+const CONSENT_REQUIRED_ENGINES = new Set(["Passwords", "PaymentMethods"]);
+
 /**
  * Customizes Sync settings (all settings are optional):
  *    - Whether sync is enabled/disabled
@@ -61,8 +68,13 @@ export const SyncPolicy = {
    * @property {boolean} [Bookmarks] Whether syncing bookmarks should be enabled
    * @property {boolean} [History] Whether syncing history should be enabled
    * @property {boolean} [OpenTabs] Whether syncing open tabs should be enabled
-   * @property {boolean} [Passwords] Whether syncing passwords should be enabled
-   * @property {boolean} [PaymentMethods] Whether syncing payment methods should be enabled
+   * @property {boolean} [Passwords] May only disable syncing passwords (false).
+   *                                 A request to enable it is ignored: enabling
+   *                                 requires an explicit local opt-in.
+   * @property {boolean} [PaymentMethods] May only disable syncing payment
+   *                                      methods (false). A request to enable it
+   *                                      is ignored: enabling requires an
+   *                                      explicit local opt-in.
    * @property {boolean} [Settings] Whether syncing settings should be enabled
    */
 
@@ -100,6 +112,18 @@ export const SyncPolicy = {
 
     for (const [type, value] of Object.entries(typeSettings)) {
       const pref = ENGINE_PREFS[type];
+
+      // Passwords and payment methods can only ever be turned off by policy.
+      // Enabling them is reserved to an explicit local choice by the user, so a
+      // request to turn them on is ignored regardless of `Locked`. Turning them
+      // off falls through to the regular handling below.
+      if (CONSENT_REQUIRED_ENGINES.has(type) && value !== false) {
+        lazy.log.warn(
+          `Ignoring policy request to enable ${type} (${pref}); syncing ${type} requires explicit local user consent.`
+        );
+        continue;
+      }
+
       if (isIgnoringUserPreferences) {
         lazy.log.debug(`Setting and locking ${type}: ${pref} : ${value}`);
         lazy.PoliciesUtils.setAndLockPref(pref, value);
--- a/browser/locales/en-US/browser/preferences/preferences.ftl
+++ b/browser/locales/en-US/browser/preferences/preferences.ftl
@@ -1408,6 +1408,14 @@ prefs-syncing-button-2 =
     .label = Syncing…
     .title = Sync now
 
+# Shown in enterprise builds, where Sync data is stored encrypted but the
+# management console holds the key (so it is not end-to-end encrypted). The
+# passwords and payment method checkboxes are in the "Choose what to sync"
+# dialog rather than on this pane, which is why this avoids saying "here".
+sync-managed-pane =
+    .heading = Your organization manages Sync on this device
+    .message = Your synced data is encrypted, and your organization can access it. Passwords and payment methods only sync if you choose to turn them on.
+
 ## The list of things currently syncing.
 
 sync-syncing-across-devices-heading = You are syncing these items across all your connected devices:
@@ -1447,6 +1455,18 @@ sync-choose-what-to-sync-dialog4 =
     .buttonaccesskeyaccept = S
     .buttonlabelextra2 = Disconnect…
     .buttonaccesskeyextra2 = D
+
+# Shown in enterprise builds, where Sync data is stored encrypted but the
+# management console holds the key (so it is not end-to-end encrypted).
+sync-managed-dialog =
+    .heading = Your organization manages Sync on this device
+    .message = Your synced data is encrypted, and your organization can access it. Passwords and payment methods only sync if you turn them on here.
+
+# Same as sync-managed-dialog, but shown when policy has locked the passwords and
+# payment method checkboxes off, so the user cannot turn them on.
+sync-managed-dialog-locked =
+    .heading = Your organization manages Sync on this device
+    .message = Your synced data is encrypted, and your organization can access it.
 
 sync-engine-bookmarks =
     .label = Bookmarks
--- a/services/sync/tests/unit/head_appinfo.js
+++ b/services/sync/tests/unit/head_appinfo.js
@@ -22,6 +22,25 @@ Services.prefs.setStringPref(
   "http://token-server"
 );
 
+// Enterprise builds default the passwords engine off, since syncing it requires
+// an explicit local opt-in, but these tests assume the upstream default. This
+// has to go on the default branch rather than be a user pref, because some
+// tests clear every user pref under services.sync. between tasks.
+// AppConstants is imported inline because several tests in this directory
+// declare `const AppConstants` themselves, and head files share their global.
+if (
+  ChromeUtils.importESModule("resource://gre/modules/AppConstants.sys.mjs")
+    .AppConstants.MOZ_ENTERPRISE
+) {
+  const syncEngineDefaults = Services.prefs.getDefaultBranch(
+    "services.sync.engine."
+  );
+  syncEngineDefaults.setBoolPref("passwords", true);
+  registerCleanupFunction(() => {
+    syncEngineDefaults.setBoolPref("passwords", false);
+  });
+}
+
 // Make sure to provide the right OS so crypto loads the right binaries
 function getOS() {
   switch (mozinfo.os) {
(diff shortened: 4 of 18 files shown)


browser/app/profile/firefox.js
```diff
@@ -3010,6 +3001,16 @@ pref("browser.crashReports.cleanupCheck.enabled", true);
 // Checkbox in sync options for credit card data sync service
 pref("services.sync.engine.creditcards.available", true);
 
+#ifdef MOZ_ENTERPRISE
+// In enterprise builds, Sync data is encrypted at rest but not end-to-end
+// encrypted (the management console holds the key), so passwords and payment
+// methods must never sync unless the user explicitly opts in locally. Default
+// both engines off (upstream defaults passwords on); leave `.available` on so
+// the user can still turn them on themselves.
```
lissyx, line 3030: Can we ensure those are the only data that requires this ? We may want a follow up with UX to make it obvious from the UI of the risks involved here

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -2,12 +2,8 @@
...
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import {
-  PREF_LOGLEVEL,
-  setAndLockPref,
-  unsetAndUnlockPref,
-  PoliciesUtils,
-} from "resource:///modules/policies/Policies.sys.mjs";
+import { PREF_LOGLEVEL } from "resource:///modules/policies/Policies.sys.mjs";
+import { PoliciesUtils } from "resource://gre/modules/PoliciesHelpers.sys.mjs";
```
lissyx: @jporter-dev had a fix for this you need to rebase

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
lissyx, line 1: @1rneh or @jporter-dev will likely want to check the `SyncPolicy`

browser/app/profile/firefox.js
lissyx: Is this related to `"syncNotE2EWarning"` later ? If so, disregard this comment

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
```diff
@@ -5,10 +5,8 @@
 policy-AccessConnector2 = Configure the { -enterprise-feature-access-connector } for proxying web traffic.
 policy-AIChatbot = Configure available AI chatbot providers, default provider, and prompt features.
 policy-BlocklistDomainBrowsedTelemetry = Enable and configure security logging/telemetry when { -brand-short-name } blocks a visit to a blocklisted domain.
-policy-DisableLocalPolicies = Disable all local policy sources (policies.json, Windows GPO and macOS plist).
 policy-DownloadTelemetry = Enable and configure security logging/telemetry when a download is triggered.
 policy-EnterpriseStorageEncryption = Enable enterprise-managed primary password for encrypted storage.
 policy-PrintPageTelemetry = Enable and configure security logging/telemetry when a page is printed.
-policy-Sync = Enable or disable sync and define which data to include.
+policy-Sync = Enable or disable sync and define which data to include. Passwords and payment methods can only be turned off by policy; the user must opt in locally to sync them.
```
lissyx: Localization would require a new identifier, e.g. `policy-Sync2`

lissyx (approved):
LGTM but I'd like other's input as well.

jporter-dev (approved):
LGTM as well once this is rebased with the import fixes from #1212!

browser/app/profile/firefox.js
beurdouche: Yes it is.

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
beurdouche: Fixed

1rneh (commented):
Hey @beurdouche, thanks for the PR, and for raising the concern about an admin being able to force-enable password and credit-card sync without end-to-end encryption - that's a real issue worth addressing.

Could you share any more context on how we got here? I don't recall a conversation or a decision on the approach, and the bug doesn't have much to go on.

beurdouche:
So far the agreement with @rtestard is to disable now to avoid issues and have a longer discussion later to define how we handle sync of such data long term.

There are a few options, I have been running an experiment with private E2EE vaults for sync which I like quite a lot but we might go for personnal profiles instead, TBD. (This is not what is implemented in this PR to be clear :) )

Screenshot 2026-05-25 at 09 54 14

beurdouche:
Made a minor update to fine tune the UI presentation. If the CI returns green, the PR can land.

browser/locales/en-US/browser/enterprise/enterprise-policies-descriptions.ftl
```diff
@@ -5,12 +5,12 @@
...
-policy-ContentAnalysisTelemetry = Enable and configure security logging/telemetry when a data loss prevention (DLP) rule is triggered.
-policy-DataLossPrevention = Enable and configure built-in data loss prevention (DLP) engine.
+policy-ContentAnalysisTelemetry = Enable and configure security logging/telemetry when a DLP rule is triggered.
+policy-DataLossPrevention = Enable and configure built-in data-loss-prevention engine.
 policy-DisableLocalPolicies = Disable all local policy sources (policies.json, Windows GPO and macOS plist).
 policy-DownloadTelemetry = Enable and configure security logging/telemetry when a download is triggered.
 policy-EnterpriseStorageEncryption = Enable enterprise-managed primary password for encrypted storage.
 policy-PrintPageTelemetry = Enable and configure security logging/telemetry when a page is printed.
-policy-Sync = Enable or disable sync and define which data to include.
+policy-Sync-NoE2EE = Enable or disable sync and define which data to include. Passwords and payment methods can only be turned off by policy; the user must opt in locally to sync them.
```
1rneh: I don't think this change is needed. 

> Passwords and payment methods can only be turned off by policy; the user must opt in locally to sync them.

This part needs to go into the admin docs rather than here. It's config details.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -100,6 +111,22 @@ export const SyncPolicy = {
...
     for (const [type, value] of Object.entries(typeSettings)) {
       const pref = ENGINE_PREFS[type];
+
+      if (CONSENT_REQUIRED_ENGINES.has(type)) {
+        // Passwords and payment methods can only ever be disabled by policy.
+        // Enabling them is reserved to an explicit local choice by the user,
+        // so a request to turn them on is ignored regardless of `Locked`.
+        if (value === false) {
+          lazy.log.debug(`Locking ${type} off: ${pref}`);
+          lazy.PoliciesUtils.setAndLockPref(pref, false);
```
1rneh: This ignores the value of `Locked` and will also lock the preferences in case `Locked===false`, but should call `lazy.PoliciesUtils.setDefaultPref(pref, value, false);` instead.

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -2980,7 +2970,7 @@
     "Sync": {
       "type": "object",
       "x-category": "Local data storage",
-      "description": "Enable or disable sync and define which data to include.",
+      "description": "Enable or disable sync and define which data to include. Passwords and payment methods can only be turned off by policy; the user must opt in locally to sync them.",
```
1rneh: Same here.

services/sync/tests/unit/head_appinfo.js
```diff
@@ -22,6 +22,14 @@ Services.prefs.setStringPref(
   "http://token-server"
 );
 
+// Enterprise builds default the passwords engine off, since syncing it requires
+// an explicit local opt-in, but these tests assume the upstream default. This
+// has to go on the default branch rather than be a user pref, because some
+// tests clear every user pref under services.sync. between tasks.
+Services.prefs
+  .getDefaultBranch("services.sync.engine.")
+  .setBoolPref("passwords", true);
```
1rneh: Please make sure this is not set unconditionally but only in enterprise builds and the test cleans it up.

browser/locales/en-US/browser/preferences/preferences.ftl
```diff
@@ -1448,6 +1448,12 @@ sync-choose-what-to-sync-dialog4 =
     .buttonlabelextra2 = Disconnect…
     .buttonaccesskeyextra2 = D
 
+# Shown in enterprise builds, where Sync data is stored encrypted but the
+# management console holds the key (so it is not end-to-end encrypted).
+sync-choose-not-e2e-warning =
+    .heading = Synced data is not end-to-end encrypted
+    .message = Everything you sync is encrypted at rest but not end-to-end encrypted, so your organization could access it. Passwords and payment methods only sync if you turn them on here.
+
```
1rneh, line 1470: I checked this locally to see the warning in context:

Image

I think the wording and presentation need another pass with Matt. Has that review happened yet?

Two thoughts:
- The red `type="error"` styling reads as "something's wrong" rather than an informational disclosure. 
- I'd like to see if the wording can be improved. As it stands I think it could push users away from Sync entirely, which probably isn't the outcome we want.

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -32,6 +32,12 @@ const ENGINE_PREFS = {
 
 const SYNC_FEATURE = "change-sync-state";
 
+// Engines whose data is especially sensitive and, in enterprise builds, is
+// encrypted at rest but not end-to-end encrypted (the management console holds
+// the key). Policy may force these off, but must never force them on: enabling
+// them requires an explicit local opt-in from the user.
+const CONSENT_REQUIRED_ENGINES = new Set(["Passwords", "PaymentMethods"]);
```
1rneh, line 40: Should `"History"`also require the same local opt-in, as it's also very sensitive user data?

1rneh (changes requested):
> So far the agreement with @rtestard is to disable now to avoid issues and have a longer discussion later to define how we handle sync of such data long term.

Thanks for the additional context!

browser/locales/en-US/browser/preferences/preferences.ftl
beurdouche: The users don't really have a choice as this is set by policy (except now for CC and passwords). 

With respect to design, I am happy to see if Matt has an improved design but we can do this as a follow-up, I have pinged him on slack

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
beurdouche: In this case don't think it would be meaningful as the administrators can collect through other means and that would sort of confuse the users into thinking that if it is disabled the backend doesn't have access to that info.

1rneh: Through which other means can the admin collect the entire browsing history?

browser/locales/en-US/browser/preferences/preferences.ftl
1rneh: > The users don't really have a choice as this is set by policy (except now for CC and passwords).

They don't have a choice if we force-enable all types by policy. In every other case (not locked or policy not active) they do have a choice. That's why I'd say it's preferable to be careful about the presentation of the warning and only land this with an approved UX.

browser/locales/en-US/browser/preferences/preferences.ftl
```diff
@@ -1448,6 +1448,12 @@ sync-choose-what-to-sync-dialog4 =
     .buttonlabelextra2 = Disconnect…
     .buttonaccesskeyextra2 = D
 
+# Shown in enterprise builds, where Sync data is stored encrypted but the
+# management console holds the key (so it is not end-to-end encrypted).
+sync-choose-not-e2e-warning =
+    .heading = Synced data is not end-to-end encrypted
+    .message = Everything you sync is encrypted at rest but not end-to-end encrypted, so your organization could access it. Passwords and payment methods only sync if you turn them on here.
```
bsmth:
```suggestion
    .heading = Your organization manages Sync on this device
    .message = Your synced data is encrypted, and your organization can access it. Passwords and payment methods only sync if you turn them on here.
```

browser/locales/en-US/browser/preferences/preferences.ftl
bsmth: See suggestion above -> https://github.com/mozilla/enterprise-firefox/pull/1175/changes#r3720075444

beurdouche:
Per discussion. Relinquished to @gcp as is for further modifications.

gcp:
As said, there are policy questions here that need to be answered, especially with relation to History sync, and where this sits in priority compared to all the P1/P2 issues.

browser/locales/en-US/browser/preferences/preferences.ftl
gcp: "Your synced data is encrypted, **but** your organization can access it."

Small nuance, but gets more to the point why these are carved out?

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
gcp: Forcing a proxy for example, although your point is that this only collects history *going forward*, not the past.

The nuance that I think @beurdouche has managed to explain is that a compromise of both history or passwords by an adversary would be catastrophic, but passwords are definitely **even more** catastrophic as they allow escalating the attack very quickly.

browser/locales/en-US/browser/preferences/preferences.ftl
bsmth: no objection for `and` -> `but` here.

PR #1176, opened 2026-07-21
Bug 2040302 - Disable primary password controls in the Settings Redesign for enterprise-managed profiles

beurdouche:
In enterprise builds the primary password is managed transparently and the user must not be able to add, change or remove it. The legacy about:preferences pane disables the controls imperatively in _initMasterPasswordUI(), but the Settings Redesign renders them through the declarative Preferences.addSetting() framework, whose disabled state is driven solely by each setting's disabled() callback.

The addPrimaryPassword, changePrimaryPassword and turnOffPrimaryPassword settings never consulted LoginHelper.isEnterpriseManagedPrimaryPassword(), so the redesign left them enabled. Add the enterprise check to those callbacks, restore it on the shared _initMasterPasswordUI() button (dropped in an earlier refactor), and cover the redesign behaviour with a browser-chrome test.

### DIFF
--- a/browser/components/preferences/config/passwords-autofill.mjs
+++ b/browser/components/preferences/config/passwords-autofill.mjs
@@ -162,7 +162,7 @@ export class PasswordSettingHelpers {
 
     var button = document.getElementById(buttonId);
     if (button) {
-      button.disabled = noMP;
+      button.disabled = noMP || isEnterpriseManagedPrimaryPassword;
     }
 
     var checkbox = document.getElementById(checkboxId);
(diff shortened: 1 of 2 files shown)


beurdouche:
Rebased onto `enterprise-main` @ `5c1d1f8f` (the branch was 3688 commits behind). The rebase was conflict-free and the diff is unchanged: 2 files, +66/-3.

CI failures look independent from the patch.

beurdouche:
Relinquished to @gcp as is for further modifications.

PR #1186, opened 2026-07-23
Bug 2056598 - Add live support for policy Cookies

1rneh:
### Description 

Bugzilla: Bug-2056598

- Removes the two `EXPIRE_POLICY` modification guards in `PermissionManager::AddInternal` (`eOperationRemoving` / `eOperationChanging`), so policy-set permissions can be removed and modified live.
- Adds a `Cookies` `onRemove` callback that clears the permissions it added, unsets and unlocks prefs, and clears the `clearCookiesForBlockedHosts` run-once. It also reconciles permissions on a narrowing live update.
- Adds a browser test covering apply, update, removal, and reconciliation of the `Cookies` policy.

_Note for reviewers:
The `EXPIRE_POLICY` guards assume the policy engine only runs at startup, so they reject all live modifications, which live policies break. Rather than adding a bypass, this patch drops the guards: `EXPIRE_POLICY` keeps only its non-persistence role (see `IsPersistentExpire`), and I've filed a bug to proposed the user-modification lock to be reintroduced as an explicit, decoupled lock (see Bug-2058217)._

---
Testing 

- [x] Added tests
- [X] Manual testing performed


netwerk/base/nsIPermissionManager.idl
```diff
@@ -199,6 +199,14 @@ interface nsIPermissionManager : nsISupports
    */
   void removeByType(in ACString type);
 
+  /**
+   * Clear all permissions of the passed type that were set by enterprise
+   * policy (i.e. with EXPIRE_POLICY expiry). This is the only way to remove
+   * policy permissions at runtime; Parent process only, and intended for the
+   * enterprise policy engine when a policy is updated or removed live.
+   */
+  void removePolicyPermissionsByType(in ACString type);
```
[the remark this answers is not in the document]
jonathanmendez: This appears to be outdated advice. Because of the way things are built, any callers should get rebuilt together without needing to bump the uuid (for example, several other methods have been added to this same interface without bumping it).

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -1204,6 +1204,15 @@ export var Policies = {
         param.Locked
       );
     },
+    onRemove() {
+      Services.perms.removePolicyPermissionsByType("cookie");
+      Services.perms.removePolicyPermissionsByType("persist-data-on-shutdown");
```
jporter-dev:
```suggestion
      if (oldParams?.Allow?.length &&
        !manager.getActivePolicies()?.SanitizeOnShutdown?.Exceptions?.length
      ) {
        Services.perms.removePolicyPermissionsByType("persist-data-on-shutdown");
      }
```

This is cleared unconditionally here, but in the `onBeforeUIStartup` it is gated by a check on both `param.Allow` and `!SanitizeOnShutdown.Exceptions`. Since `Cookies.Allow` only set this when `SanitizeOnShutdown.Exceptions` is **not** set, the same condition to clear it in `onRemove` seems appropriate.

jporter-dev: Worth adding a test for this as well

netwerk/base/nsIPermissionManager.idl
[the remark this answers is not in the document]
1rneh: Thanks for double-checking this! It also says to in the [XPIDL docs](https://firefox-source-docs.mozilla.org/xpcom/xpidl.html#uuid):

> The internal unique identifier for the interface. it must be unique, and the uuid must be generated when creating the interface. After that, it doesn’t need to be changed any more.

I'll revert [1b41635](https://github.com/mozilla/enterprise-firefox/pull/1186/commits/1b41635d0453480317dd47de16cd719ae64641a9).

lissyx: @1rneh I think it's worth getting a quick feedback from the netwerk peers (they were rather reactive for me) on adding this method and/or how much they may want this to be upstreamed as well

lissyx: (the method is simple enough and we would gain upstream coverage of tests, this might be a quick fix)

lissyx (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
1rneh: Good point, thanks for catching this! 

I made the removal logic respect the presence of the `SanitizeOnShutdown` policy and added test coverage.

1rneh: I changed my approach.. I think we can leave the `persist-data-on-shutdown` entries in place on policy removal. They come from the deprecated `Cookies.Allow` shim (to be removed in one of the next releases). Removing those entries cleanly is complicated since `removePolicyPermissionsByType` is coarse, and we can't tell which policy owns an entry, and `SanitizeOnShutdown` may own the same type. I think it's not worth adding per-origin removal logic for a shim that's being removed.

netwerk/base/nsIPermissionManager.idl
1rneh: Good idea!

1rneh: After a conversation with @mkaply I've changed my approach and removed the `EXPIRE_POLICY` guards instead. I also filed Bug-2058217 to add an explicit permission lock mechanism that is independent of `EXPIRE_POLICY`.

1rneh:
**Note to reviewers:** I updated the PR description to reflect the new approach for removing policy-set permissions live.

jporter-dev (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -1204,6 +1204,35 @@ export var Policies = {
...
+            `Unable to remove cookie permission - ${origin.href || origin}`
+          );
+        }
+      }
+      // persist-data-on-shutdown entries added by the deprecated Allow shim
+      // (see onBeforeUIStartup) are left in place as the shim is being removed
+      // in one of the next releases, and SanitizeOnShutdown.Exceptions
+      // owns these entries going forward.
+      lazy.clearRunOnceModification("clearCookiesForBlockedHosts");
+      lazy.PoliciesUtils.unsetDefaultPref("network.cookie.cookieBehavior");
```
jonathanmendez, line 1262: nit: I'm just curious, why use `unsetDefaultPref` rather than `unsetAndUnlockPref`? They are the same today; is this choice implying they might drift in implementation later?

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policy_Cookies.js
```diff
@@ -0,0 +1,260 @@
...
+  Assert.equal(
+    getCurrentCookiePermission(droppedOrigin),
+    UNKNOWN,
+    "dropped origin's stale entry was reconciled away on live update"
+  );
+});
+
+// The deprecated Cookies.Allow shim mirrors Allow origins into
+// persist-data-on-shutdown; onRemove intentionally leaves those entries in
+// place as the shim is being removed (see Bug 2051574).
```
jonathanmendez, line 221: Is it the case that when the shim is removed, this test will start to fail because persist-data-on-shutdown will stop being set at all?

jonathanmendez (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
1rneh: No, this could also call `unsetAndUnlockPref` instead. In fact, it's not really useful to have two methods if one just calls the other without passing it. I guess it was just mirroring what [setAndLockPref](https://searchfox.org/enterprise-main/rev/95146bb1afd35f0802332c5fd0d1011f44c7349e/toolkit/components/enterprisepolicies/PoliciesHelpers.sys.mjs#149-164) does without passing anything unique to `unsetDefaultPref`.

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policy_Cookies.js
1rneh: Yes, this test just serves the purpose to fail once the shim is removed, so we can get rid of anything in `Cookies` that we've added that is related to `persist-data-on-shutdown`.

PR #1198, opened 2026-07-28
Bug 2043142 - send telemetry for DLP events in enterprise builds

gregstoll:
### Description 

Bugzilla: Bug-2043142

Add a telemetry event whenever a DLP result is returned (only in Enterprise builds). Also allows configuring via policy:
- whether to report full URLs, just the domain, or nothing
- whether to send an event for every DLP result, or only ones that are not "allow". (per our existing telemetry, around 99.9% of requests return "allow" results, so this would greatly cut down on recorded events)

---

### Testing

* [x] Added tests

### DIFF
--- a/browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
+++ b/browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
@@ -106,16 +106,6 @@ export const ContentAnalysis = {
   requestTokenToRequestInfo: new Map(),
 
   /**
-   * Telemetry context for warn responses, kept around between the initial
-   * "warn" rule_triggered event and the "dlp-warn-resolved" notification
-   * that reports how the warn was resolved.
-   *
-   * @type {Map<string, {operation: string, url: string,
-   *   analysisType: string, reason: string}>}
-   */
-  _pendingWarnTelemetryInfo: new Map(),
-
-  /**
    * Whether we are resolving warn dialogs because the application is quitting,
    * in which case the resolutions aren't choices the user made.
    *
@@ -203,7 +193,7 @@ export const ContentAnalysis = {
     if (this.isInitialized) {
       this.isInitialized = false;
       this.requestTokenToRequestInfo.clear();
-      this._pendingWarnTelemetryInfo.clear();
+      lazy.ContentAnalysisTelemetry.reset();
       this.userActionToBusyDialogMap.clear();
       this.uninitializeObservers();
     }
@@ -300,8 +290,8 @@ export const ContentAnalysis = {
         // Clear this first so the handler showing the dialog will know not
         // to call respondToWarnDialog() again.
         this.warnDialogRequestTokens = new Set();
-        // Tells _recordWarnResolutionTelemetry() that these resolutions are
-        // not choices the user made.
+        // Tells ContentAnalysisTelemetry.recordWarnResolution() that these
+        // resolutions are not choices the user made.
         this._isRespondingToWarnDialogsForQuit = true;
         try {
           for (let warnDialogRequestToken of requestTokensToCancel) {
@@ -381,7 +371,7 @@ export const ContentAnalysis = {
         }
         this.requestTokenToRequestInfo.delete(response.requestToken);
         this._removeSlowCAMessage(response.userActionId, response.requestToken);
-        this._maybeRecordRuleTriggeredTelemetry(requestInfo, response);
+        lazy.ContentAnalysisTelemetry.recordVerdict(requestInfo, response);
         if (
           requestInfo.resourceNameOrOperationType?.operationType ===
           Ci.nsIContentAnalysisRequest.eDownload
@@ -413,7 +403,11 @@ export const ContentAnalysis = {
             "Got dlp-warn-resolved message but no response object was passed"
           );
         }
-        this._recordWarnResolutionTelemetry(response, aData);
+        lazy.ContentAnalysisTelemetry.recordWarnResolution(
+          response,
+          aData,
+          this._isRespondingToWarnDialogsForQuit
+        );
         break;
       }
     }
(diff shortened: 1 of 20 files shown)


browser/components/contentanalysis/content/ContentAnalysisTelemetry.sys.mjs
```diff
@@ -0,0 +1,32 @@
...
+  console.warn(
+    "[ContentAnalysisTelemetry] Enterprise implementation not available, using no-op shim. Error:",
+    ex.message
+  );
+  ContentAnalysisTelemetryImpl = {
+    recordRuleTriggered: () => {},
+  };
+}
+
+export const ContentAnalysisTelemetry = ContentAnalysisTelemetryImpl;
```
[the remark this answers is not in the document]
gregstoll: Oh, I was wondering if this was possible, thanks Copilot!

toolkit/components/contentanalysis/metrics.yaml
```diff
@@ -294,3 +294,89 @@ content_analysis:
...
+          ContentAnalysisTelemetry policy's UrlLogging setting.
+        type: string
+      action:
+        description: >
+          The DLP verdict: "block", "warn", "allow", "report_only", or
+          "canceled". "canceled" means the operation did not proceed but no
+          block dialog was shown, which is what content analysis reports when
+          a request fails while browser.contentanalysis.default_result is
+          "block", and when the user cancels the operation themselves; see
+          the cancel_error extra key to tell those apart.
```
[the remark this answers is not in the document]
gregstoll: I don't think this is necessary because "unspecified" and "unknown:<number>" should never happen.

toolkit/components/contentanalysis/metrics.yaml
```diff
@@ -294,3 +294,89 @@ content_analysis:
...
+  rule_triggered:
+    type: event
+    description: Recorded when content analysis (DLP) makes a decision
+     about a user action (paste, upload, download, print, etc), subject
+     to the ContentAnalysisTelemetry enterprise policy's filtering.
+     Only collected when MOZ_ENTERPRISE is defined and enabled.
+    bugs:
+      - https://bugzil.la/2043142
+    data_reviews:
+      - https://bugzil.la/TBD
```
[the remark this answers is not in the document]
gregstoll: I saw this done in another change: @jonathanmendez, do you know what the story is re data review for enterprise-specific telemetry?

toolkit/components/contentanalysis/metrics.yaml
```diff
@@ -294,3 +294,89 @@ content_analysis:
...
+      - highly_sensitive
+    expires: never
+    notification_emails:
+      - gstoll@mozilla.com
+      - jmendez@mozilla.com
+      - enterprise-telemetry@mozilla.org
+    send_in_pings:
+      - enterprise
+    extra_keys:
+      operation:
```
jonathanmendez, line 318: What is the difference between 'operation' and 'reason'? Is it a 1:1 mapping with operation being human-readable and reason being the proto enum? Or is it not 1:1? Either way, it may be worth adding extra clarity about the differences between these (and analysis_type) since they seem quite similar.

toolkit/components/contentanalysis/metrics.yaml
```diff
@@ -294,3 +294,89 @@ content_analysis:
...
+    data_sensitivity:
+      - highly_sensitive
+    expires: never
+    notification_emails:
+      - gstoll@mozilla.com
+      - jmendez@mozilla.com
+      - enterprise-telemetry@mozilla.org
+    send_in_pings:
+      - enterprise
+    extra_keys:
```
jonathanmendez, line 317: Probably will want to add a rule_name field to capture which rule we're using for the verdict. (Since we will eventually use the 'message' from the rule, I assume we'll resolve to a single rule rather than multiple).

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -807,6 +807,32 @@
...
+      ],
+      "properties": {
+        "Enabled": {
+          "type": "boolean"
+        },
+        "UrlLogging": {
+          "type": "string",
+          "enum": ["full", "domain", "none"]
+        },
+        "RecordEvents": {
```
jonathanmendez: I'm not sure that we have a use case for recording 'allow' events (that could basically end up sending an event e.g. for every single time a user pastes something in the browser, which would be a lot). So it may be better to just remove the `RecordEvents` part of the policy. @bsmth to weigh in?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -1076,6 +1076,58 @@ export var Policies = {
...
+        ) {
+          lazy.PoliciesUtils.setAndLockPref(
+            "browser.contentanalysis.enterprise.telemetry.recordEvents",
+            param.RecordEvents
+          );
+        }
+      }
+    },
+    onRemove(manager, oldParams) {
+      if (oldParams && typeof oldParams === "object") {
```
jonathanmendez: We don't need to inspect `oldParams` for simple pref-based policies. Basically, the state after `onRemove` is done should be as if the policy had never been applied. So this should just always call `unsetAndUnlockPref(...)` for each of the policy-related prefs (even if it hadn't been set previously, `unsetAndUnlockPref` won't mess anything up).

browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
jonathanmendez: nit: this name no longer matches the actual data since now there is more... `requestInfo` is obviously very generic but maybe it's appropriate.

browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
```diff
@@ -580,6 +627,168 @@ export const ContentAnalysis = {
...
+   * Records rule_triggered telemetry for a content analysis response. The
+   * enterprise policy-configured ContentAnalysisTelemetry module decides
+   * whether the event is actually recorded (e.g. it may filter out allow
+   * verdicts).
+   *
+   * @param {object} aRequestInfo The cached request info for this response's
+   *   requestToken (see requestTokenToRequestInfo).
+   * @param {nsIContentAnalysisResponse} aResponse
+   */
+  _maybeRecordRuleTriggeredTelemetry(aRequestInfo, aResponse) {
```
jonathanmendez: I think most of this, `_recordWarnResolutionTelemetry`, and the string maps, could move to `ContentAnalysisTelemetry.enterprise.sys.mjs`? I think maybe the main thing holding them here is the `_pendingWarnTelemetryInfo` lifetime, but I wonder if that could also be moved.

If these need to stay here, it could be good to guard with `AppConstants.MOZ_ENTERPRISE` so we don't bother gathering all this data if we don't need to.

jonathanmendez (changes requested):
Looks good overall!

toolkit/components/contentanalysis/metrics.yaml
gregstoll: Yeah, I went back and forth on that. "operation", "reason", "analysis_type" are all 1:1 with the ContentAnalysisRequest fields. I added a note to this one similar to the others.

browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
gregstoll: Yeah, all that stuff moves pretty well, thanks!

toolkit/components/contentanalysis/metrics.yaml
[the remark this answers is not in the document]
jonathanmendez: I'm not aware of anything we've done with this yet. As context, it's inherently different since the idea is these metrics will never be sent to Mozilla but only to the enterprise's admin console. @gcp is another person who may have looked into it.

browser/components/enterprisepolicies/schemas/policies-schema.json
bsmth: Thanks, I agree with removing the `RecordEvents` - this essentially means we  ignore the `allow` events and don't give the ability to configure which ones to forward, right?

gregstoll: Yes, that's correct. (and we'll still record `warn` events even if the user decides to allow them)

jonathanmendez (approved):
Approved for its current form. Re-request if the changes to remove the `RecordEvents` property are too invasive.

PR #1204, opened 2026-07-29
Bug 2060545 - Prepare for more locales

lissyx:
### Description 

Bugzilla: Bug-2060545

It was reported that more locales can be useful in some contexts (demo). This change adds production of locales data on CI. They will be later picked by repacks definition once Bug-2060543 is fixed and repack repos are updated to request more locales.

To make sure things are not broken when this happens (there is no per-branch repacks, it would be too many configurations to manage), let's land this change that makes sure the data is there at CI time.

### DIFF
--- a/browser/locales/enterprise-l10n-changesets.json
+++ b/browser/locales/enterprise-l10n-changesets.json
@@ -1,5 +1,93 @@
 {
+    "de": {
+        "pin": false,
+        "platforms": [
+            "linux64",
+            "linux64-enterprise",
+            "linux64-aarch64",
+            "linux64-aarch64-enterprise",
+            "linux64-aarch64-devedition",
+            "linux64-devedition",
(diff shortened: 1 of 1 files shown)


lissyx:
https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-pr&revision=1651fee86738c5c94e7acadc8f4b60d71eadc442

(failures on hosted repacks are expected because this repack does not exists in the staging manifest repo)

lissyx:
I was worried about our tooling but reading https://github.com/jonathanmendez/enterprise-merge-scripts/blob/ec2bb2f9f1fc6f129a9fbdefa417739bb9fec3d5/enterprise_merge_lib.py#L140-L204 it should work (I have not been able to verify myself yet)

1rneh (approved):

PR #1207, opened 2026-07-29
Bug 2057225 - Use a direct connection to the enterprise console for all proxy configurations

jporter-dev:
### Description 

Bugzilla: Bug-2057225

Bug-2051323 added support for forcing the console address into the `no_proxies_on` configuration when a `Proxy` configuration was applied. However, this only applied when the proxy was set via policy, and left a gap where a user could misconfigure a proxy configuration (if not set/locked by policy) in `about:preferences` and break connection to the console. 

This PR reverts the policy-driven `no_proxies_on` approach from #1149 in favor of registering a `nsIProtocolProxyFilter` that forces a direct connection to the console, bypassing any other attempts at proxying.

---

* Depends on: #1149 

---

### Testing 

* [x] Added tests

##### Proxy by Policy

1. Configure a proxy policy with an invalid `HTTProxy`, e.g.:
```
{
  "Mode": "manual",
  "HTTPProxy": "badproxy.enterfox.eu:8124",
  "UseHTTPProxyForAllProtocols": true,
  "Passthrough": ".firefox.com"
}
``` 
2. Access the console URL or `firefox.com`

##### User-configured Proxy

1. Open proxy settings in `about:preferences#connectionSecurity`
2. Select "Manual proxy configuration
3. Enter an invalid proxy host, e.g.: `badproxy.enterfox.eu:8124`
5. Check "Also use this proxy for HTTPS"
6. Add ".firefox.com" to the "No proxy for" field
7. Access the console URL or `firefox.com`

**Expected result:**

Console URL and `firefox.com` both load as expected and no proxy error is shown. Non-console URLs should give a "Unable to find the proxy server" neterror page.

### DIFF
--- a/browser/components/BrowserComponents.manifest
+++ b/browser/components/BrowserComponents.manifest
@@ -23,6 +23,7 @@ category browser-before-ui-startup resource:///modules/AccountsGlue.sys.mjs Acco
 category browser-before-ui-startup moz-src:///browser/modules/ObserverForwarder.sys.mjs ObserverForwarder.init
 #ifdef MOZ_ENTERPRISE
 category browser-before-ui-startup resource://gre/modules/enterprise/EnterpriseEndpoints.sys.mjs EnterpriseEndpoints.init
+category browser-before-ui-startup resource://gre/modules/enterprise/ConsoleProxyBypass.sys.mjs ConsoleProxyBypass.init
 #endif
 #ifdef XP_WIN
 category browser-before-ui-startup moz-src:///browser/components/shell/StartupOSIntegration.sys.mjs StartupOSIntegration.applyCustomIconOnStartup
--- a/browser/components/BrowserComponents.manifest
+++ b/browser/components/BrowserComponents.manifest
@@ -23,7 +23,7 @@ category browser-before-ui-startup resource:///modules/AccountsGlue.sys.mjs Acco
 category browser-before-ui-startup moz-src:///browser/modules/ObserverForwarder.sys.mjs ObserverForwarder.init
 #ifdef MOZ_ENTERPRISE
 category browser-before-ui-startup resource://gre/modules/enterprise/EnterpriseEndpoints.sys.mjs EnterpriseEndpoints.init
-category browser-before-ui-startup resource://gre/modules/enterprise/ConsoleProxyBypass.sys.mjs ConsoleProxyBypass.init
+category browser-before-ui-startup resource://gre/modules/enterprise/ConsoleProxyBypassFilter.sys.mjs ConsoleProxyBypassFilter.init
 #endif
 #ifdef XP_WIN
 category browser-before-ui-startup moz-src:///browser/components/shell/StartupOSIntegration.sys.mjs StartupOSIntegration.applyCustomIconOnStartup
--- a/browser/components/BrowserComponents.manifest
+++ b/browser/components/BrowserComponents.manifest
@@ -23,7 +23,6 @@ category browser-before-ui-startup resource:///modules/AccountsGlue.sys.mjs Acco
 category browser-before-ui-startup moz-src:///browser/modules/ObserverForwarder.sys.mjs ObserverForwarder.init
 #ifdef MOZ_ENTERPRISE
 category browser-before-ui-startup resource://gre/modules/enterprise/EnterpriseEndpoints.sys.mjs EnterpriseEndpoints.init
-category browser-before-ui-startup resource://gre/modules/enterprise/ConsoleProxyBypassFilter.sys.mjs ConsoleProxyBypassFilter.init
 #endif
 #ifdef XP_WIN
 category browser-before-ui-startup moz-src:///browser/components/shell/StartupOSIntegration.sys.mjs StartupOSIntegration.applyCustomIconOnStartup
--- a/toolkit/components/enterprise/modules/ConsoleProxyBypass.sys.mjs
+++ b/toolkit/components/enterprise/modules/ConsoleProxyBypass.sys.mjs
@@ -1,0 +1,115 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { XPCOMUtils } from "resource://gre/modules/XPCOMUtils.sys.mjs";
+
+const lazy = XPCOMUtils.declareLazy({
+  CONSOLE_ADDRESS_PREF:
+    "resource://gre/modules/enterprise/ConsoleClient.sys.mjs",
+  createEnterpriseLogger:
+    "resource://gre/modules/enterprise/EnterpriseCommon.sys.mjs",
+  log: () => lazy.createEnterpriseLogger("ConsoleProxyBypass"),
+  ProxyService: {
+    service: "@mozilla.org/network/protocol-proxy-service;1",
+    iid: Ci.nsIProtocolProxyService,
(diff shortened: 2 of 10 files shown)


toolkit/components/enterprise/modules/ConsoleProxyBypass.sys.mjs
```diff
@@ -0,0 +1,115 @@
...
+    }
+  },
+
+  /**
+   * Parses and caches the console hostname from the console address pref.
+   */
+  _updateConsoleHost() {
+    const address = Services.prefs.getStringPref(lazy.CONSOLE_ADDRESS_PREF, "");
+    this._consoleHost = URL.parse(address)?.hostname || null;
+  },
```
lissyx: I think it may be better to rely on `ConsoleClient`'s exposure of the address here?

toolkit/components/enterprise/modules/ConsoleProxyBypass.sys.mjs
```diff
@@ -0,0 +1,115 @@
...
+    "resource://gre/modules/enterprise/EnterpriseCommon.sys.mjs",
+  log: () => lazy.createEnterpriseLogger("ConsoleProxyBypass"),
+  ProxyService: {
+    service: "@mozilla.org/network/protocol-proxy-service;1",
+    iid: Ci.nsIProtocolProxyService,
+  },
+});
+
+// Run after other registered filters, so the console bypass takes precedence.
+const FILTER_POSITION = 100;
```
lissyx: where does the value comes from?
https://searchfox.org/enterprise-main/rev/95146bb1afd35f0802332c5fd0d1011f44c7349e/netwerk/base/nsIProtocolProxyService.idl#263-268

jporter-dev: It's basically arbitrary I guess, [the other filters](https://searchfox.org/enterprise-main/search?q=%5C.register.*Filter%5C%28&path=&case=true&regexp=true) I could see sit at lower positions, so this had to be higher. 

Maybe using `const MAX_UINT32 = 0xffffffff;` instead of an arbitrary `const FILTER_POSITION = 100;` would be better

toolkit/components/enterprise/modules/ConsoleProxyBypass.sys.mjs
1rneh: I agree that the address should come from `ConsoleClient.consoleBaseURI`.

toolkit/components/enterprise/modules/ConsoleProxyBypass.sys.mjs
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * @see nsIObserver
+   * @param {nsISupports} _subject The pref branch or subject of the notification.
+   * @param {string} topic Either "nsPref:changed" or "xpcom-shutdown".
+   */
+  observe(_subject, topic) {
+    switch (topic) {
+      case "nsPref:changed":
+        this._updateConsoleHost();
+        break;
```
1rneh: We can remove the observer again here after the first preference change as the console address will only be set once.

1rneh (approved):
Two small requests, otherwise it looks good to me :)

toolkit/components/enterprise/modules/ConsoleProxyBypassFilter.sys.mjs
```diff
@@ -0,0 +1,112 @@
...
+   */
+  async init() {
+    if (this._initialized) {
+      return;
+    }
+    this._initialized = true;
+
+    lazy.ProxyService.registerFilter(this, MAX_UINT32);
+    Services.obs.addObserver(this, "xpcom-shutdown");
+    await this._updateConsoleHost();
```
lissyx: Is it fine to register the filter before we have the console host?

Generally speaking, I'm wondering how it is going to behave at startup as well

browser/components/BrowserComponents.manifest
```diff
@@ -23,6 +23,7 @@ category browser-before-ui-startup resource:///modules/AccountsGlue.sys.mjs Acco
 category browser-before-ui-startup moz-src:///browser/modules/ObserverForwarder.sys.mjs ObserverForwarder.init
 #ifdef MOZ_ENTERPRISE
 category browser-before-ui-startup resource://gre/modules/enterprise/EnterpriseEndpoints.sys.mjs EnterpriseEndpoints.init
+category browser-before-ui-startup resource://gre/modules/enterprise/ConsoleProxyBypassFilter.sys.mjs ConsoleProxyBypassFilter.init
```
lissyx: I'm wondering if we do not want this at `app-startup` ?

toolkit/components/enterprise/modules/ConsoleProxyBypassFilter.sys.mjs
jporter-dev: It no-ops until we have it. I'll do a bit more testing with this and the manifest category you suggested locally. I have a local proxy set up that denies console connections so it's only accessible via proxy.

browser/components/BrowserComponents.manifest
jporter-dev: Removed the separate startup entry, instead now `ConsoleClient` handles the registering the filter once the console address resolves.

toolkit/components/enterprise/modules/ConsoleProxyBypassFilter.sys.mjs
jporter-dev: No longer an issue since now registration is triggered by `ConsoleClient`

toolkit/components/enterprise/modules/ConsoleProxyBypassFilter.sys.mjs
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * @param {string} consoleHost Hostname of the enterprise console.
+   */
+  register(consoleHost) {
+    this._consoleHost = consoleHost;
+    if (this._registered) {
+      return;
+    }
+    this._registered = true;
+    lazy.ProxyService.registerFilter(this, MAX_UINT32);
```
lissyx, line 41: nit: explicit why `MAX_UINT32` in a comment here

toolkit/components/enterprise/modules/ConsoleProxyBypassFilter.sys.mjs
```diff
@@ -0,0 +1,84 @@
...
+  },
+
+  /**
+   * Unregisters the filter and clears the console host.
+   */
+  unregister() {
+    if (!this._registered) {
+      return;
+    }
```
lissyx, line 50: Are there legit cases where we can have unbalanced register/unregister? If not, we should throw and generate errors that we can find

lissyx (approved):
r=me with comments addressed

toolkit/components/enterprise/modules/ConsoleProxyBypassFilter.sys.mjs
jporter-dev: No, this was more defensive but I suppose it's not necessary since `ProxyService.registerFilter` unregisters before registering.

https://searchfox.org/enterprise-main/source/netwerk/base/nsProtocolProxyService.cpp#1806-1812

toolkit/components/enterprise/modules/ConsoleProxyBypassFilter.sys.mjs
jporter-dev: Had a comment about this at the `MAX_UINT32` const def but moved it here for better context

PR #1218, opened 2026-07-31
Bug 2058222 - Add live support for ContentAnalysis and DataLossPrevention policies

jonathanmendez:
### Description 

Bugzilla: Bug-2058222 

- Add ability to change ContentAnalysis backend depending on which policy is active
   - Cancels any requests in flight at the moment of the change and old backend becomes inert to avoid late responses
- Adjust some existing ContentAnalysis infrastructure to be able to react to CA becoming disabled at runtime (previously not possible).
   - Any interception point that still tries to make a request will get filtered in ShouldCheckReason
   - Interception points should be more aware of IsEnabled

---

### Testing 

* [x] Added tests

### DIFF
--- a/browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
+++ b/browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
@@ -160,11 +160,11 @@ export const ContentAnalysis = {
    * @param {Window} window - The window to monitor
    */
   initialize(window) {
-    if (!this.contentAnalysis.isActive) {
-      this.uninitialize();
-      return;
-    }
-    let doc = window.document;
+    // Register unconditionally (not gated on isActive): Content Analysis can be
+    // turned on or off at runtime by an enterprise-policy update, so the
+    // observers must already be listening when a later update activates it. The
+    // observers are process-global and registered once; the handlers no-op when
+    // nothing is pending. Teardown happens on quit-application.
     if (!this.isInitialized) {
       this.isInitialized = true;
       this.initializeObservers();
@@ -177,16 +177,41 @@ export const ContentAnalysis = {
       });
     }
 
-    // Do this even if initialized so the icon shows up on new windows, not just the
-    // first one.
-    for (let indicator of doc.getElementsByClassName(
-      "content-analysis-indicator"
-    )) {
-      doc.l10n.setAttributes(indicator, "content-analysis-indicator-tooltip", {
-        agentName: lazy.agentName,
-      });
+    // Reflect the current active state in this window's indicator. Done for
+    // every window (not just the first) so the icon shows up on new windows,
+    // and re-run for all windows on a policy update (see observe()).
+    this.updateWindowContentAnalysisState(window);
+  },
+
+  /**
+   * Show or hide the Content Analysis indicator in a window to match the
+   * service's current active state.
+   *
+   * @param {Window} window - The window to update
+   */
+  updateWindowContentAnalysisState(window) {
+    let doc = window.document;
+    // Check the enabled pref before touching this.contentAnalysis so an
+    // unrelated policy update doesn't construct the service (and its backend)
+    // for a profile that never turns Content Analysis on. isActive still
+    // requires the pref, so this short-circuit changes nothing when active.
+    let isActive =
+      Services.prefs.getBoolPref("browser.contentanalysis.enabled", false) &&
+      this.contentAnalysis.isActive;
+    if (isActive) {
+      for (let indicator of doc.getElementsByClassName(
+        "content-analysis-indicator"
+      )) {
+        doc.l10n.setAttributes(
+          indicator,
+          "content-analysis-indicator-tooltip",
+          { agentName: lazy.agentName }
+        );
+      }
+      doc.documentElement.setAttribute("contentanalysisactive", "true");
+    } else {
+      doc.documentElement.removeAttribute("contentanalysisactive");
     }
-    doc.documentElement.setAttribute("contentanalysisactive", "true");
   },
 
   async uninitialize() {
@@ -209,6 +234,7 @@ export const ContentAnalysis = {
     Services.obs.addObserver(this, "quit-application");
     Services.obs.addObserver(this, "quit-application-granted");
     Services.obs.addObserver(this, "quit-application-requested");
+    Services.obs.addObserver(this, "EnterprisePolicies:PolicyUpdatesApplied");
   },
 
   /**
@@ -221,6 +247,10 @@ export const ContentAnalysis = {
     Services.obs.removeObserver(this, "quit-application");
     Services.obs.removeObserver(this, "quit-application-granted");
     Services.obs.removeObserver(this, "quit-application-requested");
+    Services.obs.removeObserver(
+      this,
+      "EnterprisePolicies:PolicyUpdatesApplied"
+    );
   },
 
   // nsIObserver
@@ -307,6 +337,14 @@ export const ContentAnalysis = {
       }
       case "quit-application": {
         this.uninitialize();
+        break;
+      }
+      // An enterprise-policy update may have activated or deactivated Content
+      // Analysis; refresh every window's indicator to match the new state.
+      case "EnterprisePolicies:PolicyUpdatesApplied": {
+        for (let window of lazy.BrowserWindowTracker.orderedWindows) {
+          this.updateWindowContentAnalysisState(window);
+        }
         break;
       }
       case "dlp-request-made":
--- a/browser/components/enterprisepolicies/helpers/ContentAnalysisPolicies.sys.mjs
+++ b/browser/components/enterprisepolicies/helpers/ContentAnalysisPolicies.sys.mjs
@@ -53,11 +53,53 @@ const CA_PLAIN_TEXT_POINTS = [
   ["DragAndDrop", "drag_and_drop"],
 ];
 
+// Every browser.contentanalysis.* pref either provider may set, apart from the
+// per-interception-point ones handled alongside these below.
+const CA_SHARED_PREFS = [
+  "enabled",
+  "use_wasm_backend",
+  "wasm_module_extension_require_signature",
+  "default_result",
+  "timeout_result",
+  "dlp_rules",
+  "allow_url_regex_list",
+  "deny_url_regex_list",
+  "agent_timeout",
+  "show_blocked_result",
+  "bypass_for_same_tab_operations",
+  "agent_name",
+  "pipe_path_name",
+  "client_signature",
+  "max_connections",
+  "is_per_user",
+];
+
+// Return every shared pref to its built-in default, unlocked.
+function releaseContentAnalysisPrefs() {
+  for (let suffix of CA_SHARED_PREFS) {
+    lazy.PoliciesUtils.unsetAndUnlockPref(caPrefName(suffix));
+  }
+  for (let [, suffix] of CA_INTERCEPTION_POINTS) {
+    lazy.PoliciesUtils.unsetAndUnlockPref(
+      caPrefName(`interception_point.${suffix}.enabled`)
+    );
+  }
+  for (let [, suffix] of CA_PLAIN_TEXT_POINTS) {
+    lazy.PoliciesUtils.unsetAndUnlockPref(
+      caPrefName(`interception_point.${suffix}.plain_text_only`)
+    );
+  }
+}
+
 export const ContentAnalysisPolicies = {
   reconcileContentAnalysis(manager) {
     let active = manager.getActivePolicies() ?? {};
     let caParam = active.ContentAnalysis;
     let dlpParam = active.DataLossPrevention;
+
+    // Release every shared pref before applying the winner's configuration.
+    // (Avoid stale prefs during a switch)
+    releaseContentAnalysisPrefs();
 
     // The external agent is authoritative only when it is actually enabled; a
     // ContentAnalysis block that does not enable an agent doesn't suppress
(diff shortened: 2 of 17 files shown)


browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
```diff
@@ -177,16 +177,41 @@ export const ContentAnalysis = {
...
+   */
+  updateWindowContentAnalysisState(window) {
+    let doc = window.document;
+    // Check the enabled pref before touching this.contentAnalysis so an
+    // unrelated policy update doesn't construct the service (and its backend)
+    // for a profile that never turns Content Analysis on. isActive still
+    // requires the pref, so this short-circuit changes nothing when active.
+    let isActive =
+      Services.prefs.getBoolPref("browser.contentanalysis.enabled", false) &&
+      this.contentAnalysis.isActive;
```
gregstoll, line 200: This is probably OK but `isActive` should only be called from the parent process on the main thread. (we have assertions for this in debug mode, so if we're not hitting those it should be fine) If we needed to change it we could use `mightBeActive`, which should work from anywhere.

browser/components/enterprisepolicies/helpers/ContentAnalysisPolicies.sys.mjs
```diff
@@ -53,22 +53,67 @@ const CA_PLAIN_TEXT_POINTS = [
...
     } else if (dlpParam) {
       applyBuiltinDlp(dlpParam);
-    } else {
-      disableContentAnalysis(caParam);
+    } else if (caParam) {
+      // Match previous behavior of locking all prefs when policy present
+      // and Enabled=false
+      activelyDisableContentAnalysis(caParam);
     }
+    // Else, neither provider is configured: the release above is the whole job.
```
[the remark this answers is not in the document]
gregstoll: I don't think this is necessary because `markContentAnalysisPolicyControlled()` will always set `isSetByEnterprisePolicy` correctly.

toolkit/components/contentanalysis/tests/xpcshell/test_wasm_backend.js
```diff
@@ -424,3 +472,113 @@ add_task(async function test_succeeds_with_signature_check_if_signed() {
...
+  );
+}
+
+// Whether a request was allowed, treating a failure to analyze as "not allowed"
+// so a fail-closed error and an explicit block are both reported as blocked.
+async function uploadWasAllowed() {
+  try {
+    return (await uploadToExample()).shouldAllowContent;
+  } catch (e) {
+    return false;
```
gregstoll, line 503: Do we expect this to ever fail? If not we should just throw here or make the test fail in some way.

toolkit/components/contentanalysis/tests/xpcshell/test_wasm_backend.js
```diff
@@ -424,3 +472,113 @@ add_task(async function test_succeeds_with_signature_check_if_signed() {
...
+    {
+      Name: "block-example-uploads",
+      Enabled: true,
+      Actions: ["FileUpload"],
+      Domains: ["example.com"],
+      Type: "block",
+    },
+  ]);
+  Assert.ok(
+    !(await uploadWasAllowed()),
```
gregstoll, line 578: It seems like we should test both allowed and not allowed after restoring valid rules.

toolkit/components/contentanalysis/ContentAnalysis.cpp
```diff
@@ -801,6 +809,97 @@ void ContentAnalysis::Close() {
...
+                          pipePathName);
+  nsString clientSignature;
+  Preferences::GetString("browser.contentanalysis.client_signature",
+                         clientSignature);
+  return pipePathName != mExternalPipePathName ||
+         clientSignature != mExternalClientSignature ||
+         StaticPrefs::browser_contentanalysis_is_per_user() !=
+             mExternalIsPerUser ||
+         StaticPrefs::browser_contentanalysis_max_connections() !=
+             mExternalMaxConnections;
```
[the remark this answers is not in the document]
gregstoll: I think not doing this is also fine, up to you.

toolkit/components/contentanalysis/ContentAnalysis.cpp
```diff
@@ -1161,6 +1281,9 @@ RefPtr<ContentAnalysis> ContentAnalysis::GetContentAnalysisFromService() {
 }
 
 static bool ShouldCheckReason(nsIContentAnalysisRequest::Reason aReason) {
+  if (!mozilla::StaticPrefs::browser_contentanalysis_enabled()) {
```
gregstoll: just so I remember, we discussed doing something better than this in a followup, right?

toolkit/components/contentanalysis/ContentAnalysis.h
```diff
@@ -314,6 +314,22 @@ class ContentAnalysis final : public nsIContentAnalysis,
   // Destroy the service.  Happens during xpcom-shutdown-threads.
   void Close();
 
+  // Re-select and refresh the backend after a live enterprise-policy update
+  // (driven by the EnterprisePolicies:PolicyUpdatesApplied notification). Swaps
```
gregstoll: you can probably delete everything after the first sentence here

toolkit/components/contentanalysis/ContentAnalysisUtils.sys.mjs
```diff
@@ -69,15 +69,21 @@ export const ContentAnalysisUtils = {
...
     let caEventChecker = async event => {
+      // Fetch the service (and re-check isActive) per event rather than once at
+      // setup: Content Analysis can be turned on or off at runtime by an
+      // enterprise-policy update, so an element wired while inactive must still
+      // enforce once activated (and stop once deactivated). Fetching per event
+      // also avoids caching a stale mock across tests.
+      const contentAnalysis = Cc["@mozilla.org/contentanalysis;1"].getService(
+        Ci.nsIContentAnalysis
+      );
+      if (!contentAnalysis.isActive) {
```
gregstoll, line 84: I think this is wrong - pretty sure we will not be on the main thread here (unlike before when we were in initialize() code). Can you test this and see what happens when this call is done? You should be able to just paste into a `prompt()` dialog to trigger this (or paste into the GenAI sidebar)

gregstoll (commented):
(haven't finished looking yet, here's what I have so far)

browser/components/enterprisepolicies/helpers/ContentAnalysisPolicies.sys.mjs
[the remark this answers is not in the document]
jonathanmendez: This was a good catch/comment that I have since fixed differently by always calling `markContentAnalysisPolicyControlled(bool)`

browser/components/enterprisepolicies/helpers/ContentAnalysisPolicies.sys.mjs
```diff
@@ -209,58 +244,18 @@ function applyBuiltinDlp(dlpParam) {
     caPrefName("dlp_rules"),
     JSON.stringify({ DLPRules: { Rules: rules } })
   );
```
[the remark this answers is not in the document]
jonathanmendez: This isn't true, both in that the code is calling [setPrefIfPresentAndLock](https://searchfox.org/firefox-main/source/toolkit/components/enterprisepolicies/PoliciesHelpers.sys.mjs#122), which always locks, and in that the applyBuiltInDlp implementation exactly matches the applyContentAnalysisConfig implementation.

toolkit/components/contentanalysis/ContentAnalysisUtils.sys.mjs
jonathanmendez: I think this is actually working maybe? I believe I'm in a debug build (`ac_add_options --enable-debug`) and I didn't run into problems pasting into a prompt on a test page (w3cschools) and into the GenAI sidebar (which correctly triggered a warn for my rule for chatgpt). Your analysis seems correct though so let me know if I did something obviously wrong in testing.

But I ran into what I think is potentially a pre-existing bug, testing by calling `prompt()` in a (content) console: I got a (parent/browser) console error with `Content analysis request returned error:  NS_ERROR_FAILURE: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIContentAnalysis.getURIForBrowsingContext]`. The [comment just after that call](https://searchfox.org/firefox-main/source/toolkit/components/contentanalysis/ContentAnalysisUtils.sys.mjs#115-117) says that the browsing context might be null, but the first thing the called function does is [dereference the browsing context](https://searchfox.org/firefox-main/source/toolkit/components/contentanalysis/ContentAnalysis.cpp#3227). I'm not sure if there's any real-world scenario that matches this setup, but if it's worth a bug/fix I can file that.

gregstoll: Yeah I don't know how realistic that is but if we're being careful we might as well be effectively careful, please file a bug :-)

toolkit/components/contentanalysis/ExternalAgentBackend.cpp
```diff
@@ -941,8 +944,13 @@ void ExternalAgentBackend::HandleResponseFromAgent(
   NS_DispatchToMainThread(NS_NewRunnableFunction(
       __func__,
       [self = RefPtr{this}, aResponse = std::move(aResponse)]() mutable {
+        AssertIsOnMainThread();
```
gregstoll, line 947: hypernit: we're calling `NS_DispatchToMainThread()` to get here so I don't think this is helpful

toolkit/components/contentanalysis/ExternalAgentBackend.cpp
```diff
@@ -463,6 +463,9 @@ ExternalAgentBackend::~ExternalAgentBackend() {
 void ExternalAgentBackend::Shutdown() {
   AssertIsOnMainThread();
 
+  // Drop any agent response that arrives after this point (see mInert).
```
gregstoll: hypernit: "(see mInert)" is a bit confusing here, I'd just remove it :-)

toolkit/components/contentanalysis/WasmModuleBackend.cpp
```diff
@@ -171,6 +182,11 @@ nsresult WasmModuleBackend::Analyze(
                            contentBytes = std::move(contentBytes),
                            rules = std::move(rules), userActionId,
                            aAutoAcknowledge]() mutable {
+                  AssertIsOnMainThread();
```
gregstoll, line 185: hypernit: as above, probably not worth keeping this assert

toolkit/components/contentanalysis/WasmModuleBackend.cpp
```diff
@@ -331,7 +357,11 @@ WasmModuleBackend::GetDiagnosticInfo() {
   return DiagnosticInfoPromise::CreateAndResolve(info, __func__);
 }
 
-void WasmModuleBackend::Shutdown() {}
+void WasmModuleBackend::Shutdown() {
+  AssertIsOnMainThread();
+  // Drop any runner promise that resolves after this point (see mInert).
```
gregstoll: hypernit: "(see mInert)" is redundant

toolkit/components/contentanalysis/ExternalAgentBackend.cpp
gregstoll: Oh, we probably need this because `mInert` is guarded by `sMainThreadCapability`, my bad

toolkit/components/contentanalysis/WasmModuleBackend.cpp
gregstoll: Oh, we probably need this because `mInert` is guarded by `sMainThreadCapability`, my bad

toolkit/components/contentanalysis/ContentAnalysisUtils.sys.mjs
gregstoll: (and if this seems to be working this seems fine then)

gregstoll (approved):
just some minor stuff, thanks!

toolkit/components/contentanalysis/ContentAnalysis.cpp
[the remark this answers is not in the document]
jonathanmendez: Yeah, I think I'll keep the stronger rebuild of the backend here because policy will change the _default_ pref which won't notify the pref observer, so I'd have to handle it here anyway. Just rebuilding is simpler than special casing a check that _only_ max connections changed.

toolkit/components/contentanalysis/ExternalAgentBackend.cpp
jonathanmendez: yes exactly that, we get warnings otherwise

toolkit/components/contentanalysis/WasmModuleBackend.cpp
jonathanmendez: yes exactly that, we get warnings otherwise

browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
gregstoll: Per below this seems fine

toolkit/components/contentanalysis/tests/xpcshell/test_wasm_backend.js
gregstoll: Sorry, missed the comment that mentions that we do actually test this.

toolkit/components/contentanalysis/tests/xpcshell/test_wasm_backend.js
gregstoll: eh, this seems fine on further review

toolkit/components/contentanalysis/ContentAnalysis.cpp
jonathanmendez: I found a better version of this by putting a better explicit check at the start of `AnalyzeContentRequestsCallback`

PR #1228, opened 2026-08-04
Bug 2060543 - Build repack list from github instead of hardcoding

lissyx:
### Description 

Bugzilla: Bug-2060543

### DIFF
--- a/taskcluster/gecko_taskgraph/util/partners.py
+++ b/taskcluster/gecko_taskgraph/util/partners.py
@@ -157,8 +157,11 @@ def get_token(params):
     if os.environ.get("GITHUB_API_TOKEN"):
         return os.environ["GITHUB_API_TOKEN"]
 
+    if not "source" in params:
+        params["source"] = "releng/gecko/build"
+
     # The 'usual' method - via taskClusterProxy for decision tasks
-    url = "{secret_root}/project/releng/gecko/build/level-{level}/partner-github-api".format(
+    url = "{secret_root}/project/{source}/level-{level}/partner-github-api".format(
         secret_root=TASKCLUSTER_PROXY_SECRET_ROOT, **params
     )
     try:
@@ -203,7 +206,7 @@ def get_repo_params(repo):
         return repo.split(":")[-1].split("/")
 
 
-def get_partners(manifestRepo, token):
+def get_partners(manifestRepo, token, use_path_for_name=False):
     """Given the url to a manifest repository, retrieve the default.xml and parse it into a
     list of partner repos.
     """
@@ -237,15 +240,25 @@ def get_partners(manifestRepo, token):
                 "repo": child.attrib["name"],
                 "revision": child.attrib["revision"],
             }
-            partners[child.attrib["name"]] = partner_url
+            # on enterprise, the partner name is later used to build treeherder
+            # symbol that are limited to 25 chars. but "name" from the manifest
+            # is the name of the repository, which we do not want to change.
+            # however "path" is where the checkout happens and this can be more
+            # flexible, so make use of it
+            partner_name = (
+                child.attrib["path"].split("/")[-1]
+                if use_path_for_name
+                else child.attrib["name"]
+            )
+            partners[partner_name] = partner_url
             log.debug(
-                "Added partner %s at revision %s"
-                % (partner_url["repo"], partner_url["revision"])
+                "Added partner %s repo %s at revision %s"
+                % (partner_name, partner_url["repo"], partner_url["revision"])
             )
     return partners
 
 
-def parse_config(data):
+def parse_config(data, platform_mapping):
     """Parse a single repack.cfg file into a python dictionary.
     data is contents of the file, in "foo=bar\nbaz=buzz" style. We do some translation on
     locales and platforms data, otherwise passthrough
@@ -263,9 +276,9 @@ def parse_config(data):
             l = str(l)
             key, value = l.split("=", 1)
             value = value.strip("'\"").rstrip("'\"")
-            if key in TC_PLATFORM_PER_FTP.keys():
+            if key in platform_mapping.keys():
                 if value.lower() == "true":
-                    config["platforms"].append(TC_PLATFORM_PER_FTP[key])
+                    config["platforms"].append(platform_mapping[key])
                 continue
             if key not in ALLOWED_KEYS:
                 continue
@@ -276,7 +289,7 @@ def parse_config(data):
     return config
 
 
-def get_repack_configs(repackRepo, token):
+def get_repack_configs(repackRepo, token, platform_mapping):
     """For a partner repository, retrieve all the repack.cfg files and parse them into a dict"""
     log.debug("Querying for configs in %s", repackRepo)
     query = REPACK_CFG_QUERY % repackRepo
@@ -289,7 +302,7 @@ def get_repack_configs(repackRepo, token):
         for file in sub_config["object"].get("entries", []):
             if file["name"] != "repack.cfg":
                 continue
-            configs[name] = parse_config(file["object"]["text"])
+            configs[name] = parse_config(file["object"]["text"], platform_mapping)
     return configs
 
 
@@ -315,7 +328,14 @@ def get_attribution_config(manifestRepo, token):
     return yaml.safe_load(raw_manifest["data"]["repository"]["object"]["text"])
 
 
-def get_partner_config_by_url(manifest_url, kind, token, partner_subset=None):
+def get_partner_config_by_url(
+    manifest_url,
+    kind,
+    token,
+    partner_subset=None,
+    platform_mapping=TC_PLATFORM_PER_FTP,
+    use_path_for_name=False,
+):
     """Retrieve partner data starting from the manifest url, which points to a repository
     containing a default.xml that is intended to be drive the Google tool 'repo'. It
     descends into each partner repo to lookup and parse the repack.cfg file(s).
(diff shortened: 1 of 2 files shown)


taskcluster/gecko_taskgraph/target_tasks.py
```diff
@@ -504,6 +504,9 @@ def target_tasks_enterprise_firefox_with_tests(
     )
 
     def filter(task):
+        if "shippable" in task.label:
+            return True
+
```
lissyx: temp hack for checking PR with shippable builds

taskcluster/gecko_taskgraph/util/verify.py
```diff
@@ -192,7 +192,7 @@ def verify_task_graph_no_shippable_enterprise_level_not_1(
 
     level = int(parameters["level"])
     if level < 3 and "enterprise" in task.label and "shippable" in task.label:
-        raise Exception(
+        print (
```
lissyx: temp hack for checking PR with shippable builds

lissyx:
For similar reasons as #1204 we likely want to uplift this on all branches

lissyx:
Once approved, this will need to be uplifted and merged at the same time in all branches and at the time enterprise repacks repos and manifests repos are merged as well

lissyx:
Also this shows it on Thunderbird: https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-pr&revision=c0a793b15e7b70b918bb2d17347ee023a75692a2&selectedTaskRun=cIBr3inUQiGkrcHtv-aSag.0

There we use `root=thunderbird/` with https://github.com/lissyx/sample-configs/commit/bec72f04b52f52b6a758746bc95a28cf42ec5ce4

taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -621,32 +636,30 @@ def _pad_macos_attribution_code(attribution_string):
...
-def make_enterprise_repack(platforms):
-    repacks = deepcopy(ENTERPRISE_REPACKS)
-    for repack_repo in repacks.keys():
-        for repack_name in repacks[repack_repo]:
-            repacks[repack_repo][repack_name].update({"platforms": deepcopy(platforms)})
-    return repacks
+def make_enterprise_repack_from_url(parameters, graph_config, kind, platform_mapping):
+    token = get_token({"level": parameters["level"], "source": "enterprise"})
+    partner_url_config = get_partner_url_config(parameters, graph_config)
+    get_partner_config_by_url(
```
jporter-dev, line 649: `get_partner_config_by_url` returns `partner_configs[kind]`. This function can assign a local var to be returned instead of manipulating the global `partner_configs`? it's a little bit confusing what's happening since it throws out the return value of `get_partner_config_by_url`, and it's not clear that `get_partner_config_by_url` is manipulating that global as well.

This also prevents having to walk and remap all of `partner_configs` on every call here and instead just apply to the specified `kind`.

for example

```
    configs = get_partner_config_by_url(
        manifest_url=partner_url_config["enterprise-repack"],
        kind=kind,
        token=token,
        platform_mapping=platform_mapping,
    )

    for old_key, new_key in ENTERPRISE_REPACKS_REPO_MAPPING.items():
        if old_key in configs:
            configs[new_key] = configs.pop(old_key)

    return configs
```

taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -621,32 +636,30 @@ def _pad_macos_attribution_code(attribution_string):
...
+        manifest_url=partner_url_config["enterprise-repack"],
+        kind=kind,
+        token=token,
+        platform_mapping=platform_mapping,
+    )
+
+    for first_level_key, second_level_dict in partner_configs.items():
```
jporter-dev: This returns `partner_configs[kind]` but walks and remaps all `partner_configs`, not sure if this is intended?

taskcluster/gecko_taskgraph/util/partners.py
lissyx: The problem here is that the rest of the code makes use of `partner_configs`, like you discovered with `get_partner_config_by_url`.

Your idea was my first implementation but then it showed to break silently things.

lissyx: > This also prevents having to walk and remap all of `partner_configs` on every call here and instead just apply to the specified `kind`.

given how `get_partner_config_by_url` can do things, I think it's safer we rewrite always

taskcluster/gecko_taskgraph/util/partners.py
lissyx: It is indended, confere how it's used below, where we make a call for each `kind`, so we need to return the kind's value.

taskcluster/gecko_taskgraph/util/partners.py
jporter-dev: yeah, I was worried it would have unintended side effects since other things manipulate the global

jporter-dev (approved):

taskcluster/gecko_taskgraph/util/partners.py
lissyx: it does but we indeed have incorrect behavior if we do not respect it :)

taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -68,7 +68,7 @@
 # Returns the contents of desktop/*/repack.cfg for a partner repository
 REPACK_CFG_QUERY = """query{
   repository(owner:"%(owner)s", name:"%(repo)s") {
-    object(expression: "%(revision)s:desktop/"){
+    object(expression: "%(revision)s:%(root)s"){
```
lissyx: This change may not be needed in fact ; I had hoped it would allow to distinguish products but it does not according to https://treeherder.mozilla.org/logviewer?job_id=583119453&repo=enterprise-firefox-pr&task=R2f88hH5QOqHacyXvgMcQQ.0&lineNumber=391-393:
```
[task 2026-08-05T13:01:11.115+00:00] 13:01:11     INFO -  2026-08-05 13:01:11,114 - DEBUG - Found partner config: ../../workspace/partners/moz/desktop/stageGCP ['distribution'] repack.cfg
[task 2026-08-05T13:01:11.115+00:00] 13:01:11     INFO -  2026-08-05 13:01:11,115 - DEBUG - Found partner config: ../../workspace/partners/moz/desktop/prodGCP ['distribution'] repack.cfg
[task 2026-08-05T13:01:11.115+00:00] 13:01:11     INFO -  2026-08-05 13:01:11,115 - DEBUG - Found partner config: ../../workspace/partners/moz/thunderbird/sbxTb ['distribution'] repack.cfg
```

Unfortunately I think it means we will have to separate repacks repo and manifests per product?

lissyx: (correction, we can keep the same repos but use different branches)

lissyx: I'm checking it right now with Thunderbird. Only a bit painful point is that we will need to have a dedicated thunderbird manifest repo.

taskcluster/gecko_taskgraph/util/partners.py
lissyx: Turns out we have the information we need when getting the manifest, and it's cleaner to derive it directly rather than adding an extra manual mapping, so we should be more flexible now

taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -157,8 +157,11 @@ def get_token(params):
     if os.environ.get("GITHUB_API_TOKEN"):
         return os.environ["GITHUB_API_TOKEN"]
 
+    if not "source" in params:
+        params["source"] = "releng/gecko/build"
+
     # The 'usual' method - via taskClusterProxy for decision tasks
-    url = "{secret_root}/project/releng/gecko/build/level-{level}/partner-github-api".format(
+    url = "{secret_root}/project/{source}/level-{level}/partner-github-api".format(
```
lissyx, line 164: https://bugzilla.mozilla.org/show_bug.cgi?id=2060549

taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -237,15 +240,20 @@ def get_partners(manifestRepo, token):
                 "repo": child.attrib["name"],
                 "revision": child.attrib["revision"],
             }
-            partners[child.attrib["name"]] = partner_url
+            partner_name = (
+                child.attrib["path"].split("/")[-1]
+                if use_path_for_name
+                else child.attrib["name"]
+            )
+            partners[partner_name] = partner_url
```
ahal, line 253: Where does this value come from? Is it possible to just set `child.attrib["name"]` correctly at the source instead?

ahal (approved):
Seems fine, don't forget to drop the temporary commits

taskcluster/gecko_taskgraph/util/partners.py
lissyx: `name` in the manifest used by repo is the real path to use to build the repo url, e.g. `git@github.com:mozilla-partners/<name>.git` so if we change it there we need to rename repositories on github. I am not sure how this is used elsewhere but for us the partner's name is being used to construct the treeherder name, so we need something that fits within the 25 chars limit, assuming we concatenate this name with repack name and locale later.

The only source of liberty we have here is short name like `partners/moz` so we extract `moz` for later. I'll add a comment

PR #1229, opened 2026-08-04
Bug 2059734 - Hide helpSwitchDevice button from help menu on enterprise builds

jporter-dev:
### Description 

Bugzilla: Bug-2059734

This PR removes the "Switching to a New Device" button in both Help menus. The "Getting a new device soon" Nova promo in the hamburger menu -> `Help and Report` menu is also removed by this since it replaces the "Switching to a New Device" button in that context. 

---

### Screenshots 

helpmenu-before

_Before_

helpmenu-after

_After_

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. Open the hamburger menu
2. Click "Help and Report"
3. Check for missing menu
4. On MacOS, click "Help" from the top toolbar

**Expected result:**

Neither menu displays the "Switching to a New Device" menu option, or the "Getting a new device soon" promo banner.


1rneh (approved):
Nice, thanks! Have you checked if there's a test that tests the presence of the "Switch Device" menu item which we could simply mark with `skip-if = ["enterprise"]` in one go?

jporter-dev:
> Nice, thanks! Have you checked if there's a test that tests the presence of the "Switch Device" menu item which we could simply mark with `skip-if = ["enterprise"]` in one go?

Thanks, I added a skip to the `novaHelpSwitchDevicePromo` test! That's the only one that's testing against the `#helpSwitchDevice` element that's being excluded.

PR #1233, opened 2026-08-05
Bug 2060885 - Non-ReauthRequiredError on failed token refresh needs to bring felt to foreground

1rneh:
### Description 

Bugzilla: Bug-2060885

Fixes that a non-ReauthRequiredError (e.g. 5xx) on a token refresh doesn't leave felt in a windowless state after shutting down the browser.

Splits the refresh-failure teardown into a new `FirefoxSessionInterrupted` message that brings FELT to the foreground with a "session interrupted" error bar, distinguishing re-auth (`tokenRefreshExpired`) from a transient failure (`tokenRefreshFailed`). In both cases we now end up with the felt window and a dedicated error message bar.

_Note to reviewer: more sophisticated 5xx handling will be implemented in Bug-2060868. This PR only addresses the poor client-side UX observed during the latest console deployment._

---

### Screenshots

Screenshot 2026-08-05 at 16 17 55

---

### Testing 

* [X] Added tests
* [X] Manual testing performed


testing/enterprise/test_felt_token_refresh_5xx.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        super().run_felt_base()
+        self.connect_child_browser()
+        self.assert_user_signed_in(env=Environment.FIREFOX)
+
+        browser_pid = self._child_driver.session_capabilities["moz:processID"]
+
+        # The tokens FELT currently holds are the ones the console issued.
+        expected_access = self.policy_access_token.value
+        expected_refresh = self.policy_refresh_token.value
```
argl:
```suggestion
        expected_access = getattr(self.policy_access_token, "value", None)
        expected_refresh = getattr(self.policy_refresh_token, "value", None)
```
Looks a bit cleaner if `self.policy_access_token` itself is botched. But this is the javascript part of my brain ringing the alarm. If this is a pattern we use throughout the python tests, I am fine with leaving this as is.

testing/enterprise/test_felt_token_refresh_5xx.py
```diff
@@ -0,0 +1,107 @@
...
+
+        super().run_felt_base()
+        self.connect_child_browser()
+        self.assert_user_signed_in(env=Environment.FIREFOX)
+
+        browser_pid = self._child_driver.session_capabilities["moz:processID"]
+
+        # The console starts returning 5xx for the live policy poll.
+        self.policies_fail_request.value = 1
+        time.sleep(5)
```
argl, line 45: Is this tied to the policy polling frequency? Isn't this set to a different value than normal under testing conditions? This looks potentially brittle, but if it is the pattern we use in these tests, fair enough.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -320,31 +320,24 @@ export class FeltProcessParent extends JSProcessActorParent {
...
-                  Services.felt.shutdownFirefox();
-                  return;
-                }
-                // At this point, we need to reauthenticate.
-                lazy.log.error("token refresh failed, reauthenticate", error);
+                lazy.log.error(
+                  `token refresh failed (${error.name}), shutting down Firefox`,
+                  error
+                );
                 Services.felt.clearTokens();
```
argl, line 332: Clearing the tokens under any circumstance looks wrong.

argl (changes requested):

testing/enterprise/test_felt_token_refresh_5xx.py
1rneh: The policy polling interval is 500ms. This is to wait out that the browser wasn't shutdown when returned a 5xx server error. So I agree, it's quite brittle, but there is no assertion/condition to await.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
1rneh: Their is no path to recover yet, so these shouldn't linger around. I filed Bug-2060929 to investigate the browser recovering path.

testing/enterprise/test_felt_token_refresh_5xx.py
1rneh: These are always `SharedString`/`Value` instances set in [`setUp`](https://searchfox.org/enterprise-main/rev/74e2555ace490082007078509503ace5dc90fbb6/testing/enterprise/felt_tests.py#744-752), so `.value` is guaranteed, and direct `.value` access is the pattern used throughout the tests.

argl (approved):

testing/enterprise/felt_tests.py
```diff
@@ -482,6 +482,11 @@ def do_POST(self):
             )
             parsed_payload = json.loads(payload)
 
+            # Simulate a console that is transiently unavailable
+            if self.server.token_fail_request.value:
+                self.reply("", 500, "Internal Server Error", "application/json")
```
lissyx: `""` would be `null` in json right? I remember weird behavior could popup depending on json parsing, I'm not sure if we want to be super strict here?

lissyx (approved):

testing/enterprise/felt_tests.py
1rneh: `""` isn't JSON at all. It reflects a real error response because e.g. a reverse proxy might send back an html or an empty body anyway..  So either I could drop the wrong content-type header `"application/json"` or return real JSON. I'm going with dropping it and also replace other occurrences in the file that have the same pattern.

Note: The ConsoleClient reads the errors with `res.text()` instead of `res.json()`.

lissyx: No `""` isn't json at all, but I remember there was some conversion happening down the line and making it `null` at the end, and it was a source of bug somewhere in our code because we were expecting some more valid JSON than what we were receiving, and our mock vs real console was behaving differently.

So making sure this works well even in the weirdest case sounds good to me, I dont really have a preference on how, and we cannot start checking all possibilities anyway

PR #1235, opened 2026-08-05
Bug 2059889 - Control "Tabs from other devices" visibility based on Sync policy

jporter-dev:
### Description 

Bugzilla: Bug-2059889

The "Tabs from other devices" views in both the sidebar and `about:firefoxview` do not behave as expected when `Sync` is disabled by policy. The sidebar item is always visible, and depending on prefs/policies, it displays different messages which are not useful if Sync is disabled by policy. The item in Firefox View acts mostly the same, but does hide under the `DisableAccounts` policy (which we do not support).

This PR adds the ability to control the visibility of the Firefox View and sidebar tools based on the Sync enterprise policy. When Sync is locked and disabled by policy, or the OpenTabs engine is disabled with Sync enabled, the tools are hidden rather than surfacing a panel that can only show an error. When Sync is not locked by policy, the tools remain visible since they can show "Turn on Sync" messaging.

---

### Screenshots 

before-sync-disabled-locked

_Before: Sync disabled and locked_

before-sync-enabled-locked-opentabs-on

_Before: Sync enabled and locked with OpenTabs on_

before-sync-enabled-locked-opentabs-off

_Before: Sync enabled and unlocked with OpenTabs off_

before-disableaccounts-true

_Before: DisableAccounts true (not really relevant since unsupported, just showing one more state)_

after_opentabs_sync_disabled

_After: Sync disabled and locked (or enabled with OpenTabs disabled)_

after_opentabs_sync_enabled

_After: Sync enabled and locked with OpenTabs on_

after-sync-enabled-unlocked-opentabs-false

_After: Sync enabled and unlocked with OpenTabs off_

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. Launch Firefox with a policy that locks Sync off and disables OpenTabs, e.g.:
`{ "policies": { "SyncSettings": { "Enabled": false, "TypeSettings": { "OpenTabs": false } } } }`
2. Open Firefox View and confirm the "Tabs from other devices" section is hidden. 
3. Open the sidebar customize/tools list and confirm the "Tabs from other devices" tool is absent and does not show in the sidebar settings.
4. Unlock the `Sync` policy or enable Sync and confirm "Tabs from other devices" is visible in both Firefox View and the sidebar.

**Expected result:**

"Tabs from other devices" is not present in either the sidebar or Firefox View when `Sync` policy:

- Is disabled and locked
- Is enabled and locked and `OpenTabs` is false

"Tabs from other devices" is present in either the sidebar or Firefox View when `Sync` policy:

- Is enabled and locked and `OpenTabs` is true
- Is enabled and unlocked and `OpenTabs` is true or false

### DIFF
--- a/browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
+++ b/browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
@@ -30,7 +30,8 @@ const ENGINE_PREFS = {
   Settings: "services.sync.engine.prefs",
 };
 
-const SYNC_FEATURE = "change-sync-state";
+const SYNC = "sync";
+const SYNC_TABS = "sync-tabs";
 
 /**
  * Customizes Sync settings (all settings are optional):
@@ -84,6 +85,11 @@ export const SyncPolicy = {
     } = params;
 
     if (isIgnoringUserPreferences) {
+      // "Tabs from other devices" needs both Sync and its tabs engine unlocked.
+      if (shouldEnableSync === false || typeSettings.OpenTabs === false) {
+        manager.disallowFeature(SYNC_TABS);
+      }
+
       const isSyncCurrentlyEnabled = this.isSyncCurrentlyEnabled();
       if (shouldEnableSync && !isSyncCurrentlyEnabled) {
         lazy.log.debug("Enable Sync");
@@ -107,7 +113,7 @@ export const SyncPolicy = {
 
     // Only lock the Sync feature if 'Enabled' is configured
     if (isIgnoringUserPreferences && shouldEnableSync !== undefined) {
-      manager.disallowFeature(SYNC_FEATURE);
+      manager.disallowFeature(SYNC);
     }
   },
 
@@ -117,8 +123,11 @@ export const SyncPolicy = {
    * @param {EnterprisePoliciesManager} manager
    */
   async restoreSettings(manager) {
-    if (!Services.policies.isAllowed(SYNC_FEATURE)) {
-      manager.allowFeature(SYNC_FEATURE);
+    if (!Services.policies.isAllowed(SYNC)) {
+      manager.allowFeature(SYNC);
+    }
+    if (!Services.policies.isAllowed(SYNC_TABS)) {
+      manager.allowFeature(SYNC_TABS);
     }
     for (const pref of Object.values(ENGINE_PREFS)) {
       lazy.log.debug(`Unsetting ${pref}`);
--- a/browser/components/sidebar/browser-sidebar.js
+++ b/browser/components/sidebar/browser-sidebar.js
@@ -185,6 +185,16 @@ var SidebarController = {
       ],
     ]);
 
+    // Hide synced tabs when accounts or Sync are disabled by policy, rather
+    // than surfacing a panel that can only show an error.
+    const syncedTabsSidebar = this._sidebars.get("viewTabsSidebar");
+    Object.defineProperty(syncedTabsSidebar, "visible", {
+      get: () =>
+        Services.prefs.getBoolPref("identity.fxaccounts.enabled", false) &&
+        Services.policies.isAllowed("sync-tabs"),
+      configurable: true,
+    });
+
     this.registerPrefSidebar(
       "browser.ml.chat.enabled",
       "viewGenaiChatSidebar",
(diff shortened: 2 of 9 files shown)


browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -109,9 +115,8 @@ export const SyncPolicy = {
       lazy.PoliciesUtils.setDefaultPref(pref, value, false);
     }
 
-    // Only lock the Sync feature if 'Enabled' is configured
```
1rneh, line 108: Oh, could you add back this comment?

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -88,6 +89,11 @@ export const SyncPolicy = {
     } = params;
 
     if (isIgnoringUserPreferences) {
+      // "Tabs from other devices" needs both Sync and its tabs engine on.
```
1rneh: `nit`
```suggestion
      // "Tabs from other devices" needs both Sync and its tabs engine unlocked.
```

browser/components/enterprisepolicies/helpers/SyncPolicy.sys.mjs
```diff
@@ -30,7 +30,8 @@ const ENGINE_PREFS = {
   Settings: "services.sync.engine.prefs",
 };
 
-const SYNC_FEATURE = "change-sync-state";
+const SYNC_FEATURE_LOCKED = "change-sync-state";
+const SYNC_FEATURE_TABS = "synced-tabs";
```
1rneh: `nit:` This should describe the feature that will be allowed/disallowed.
```suggestion
const SYNC = "sync";
const SYNC_TABS = "sync-tabs";
```

browser/components/sidebar/browser-sidebar.js
```diff
@@ -628,6 +638,18 @@ var SidebarController = {
       Services.obs.addObserver(this, "sessionstore-single-window-restored");
       this._windowRestoredObserverAdded = true;
     }
+    if (!this._policiesObserverAdded) {
+      Services.obs.addObserver(this, "EnterprisePolicies:AllPoliciesApplied");
```
1rneh: `EnterprisePolicies:AllPoliciesApplied` will never be observed here, because it is fired when the policy initialization completes and that's before any window/sidebar is built. On the first window/sidebar the policy state is already final. Let's drop it and keep only `EnterprisePolicies:PolicyUpdatesApplied` to observe policy updates.

1rneh (changes requested):
We currently have zero test coverage for the Sync policy so I added basic test coverage in #1293. Once that has landed, could you help move the tab syncing tests from `browser_policy_synced_tabs_hidden.js` to the newly added browser test and also make sure it updates correctly on live updates?

toolkit/components/enterprisepolicies/tests/xpcshell/live/test_live_policy_Sync.js
```diff
@@ -205,7 +205,7 @@ add_task(async function test_locked_disabled_disconnects_and_locks_feature() {
 });
 
 // An unlocked Sync policy only changes the engine setting's default values;
-// the preference stay unlocked and the sync feature stays allowed.
+// the preferences stay unlocked and the sync feature stays allowed.
```
1rneh, line 208: Thanks!

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policy_Sync.js
```diff
@@ -174,3 +231,99 @@ add_task(async function test_sync_controls_reflect_feature_lock() {
...
+  await updatePolicies({ policies: {} });
+  checkSyncTabsFeatureAllowed(true);
+  await checkSyncedTabsTool(false);
+  await checkFirefoxViewSyncedTabsHidden(false);
+});
+
+// DisableAccounts turns off Firefox accounts, which also hides the synced tabs
+// surfaces. Both an already-open Firefox View and the sidebar tool react live
+// because they observe identity.fxaccounts.enabled.
+add_task(async function test_synced_tabs_hidden_by_disable_accounts_live() {
```
1rneh: `DisableAccounts` is not supported in enterprise builds. I'm currently working on not applying those not supported policies which will break this test. Can we remove it?

1rneh (approved):
Lgtm!

toolkit/components/enterprisepolicies/tests/browser/live/browser_live_policy_Sync.js
jporter-dev: Removed it, thanks!

PR #1242, opened 2026-08-06
Bug 2021342 - Load encryption key from console earlier

lissyx:
### Description 

Bugzilla: Bug-2021342

Session restore of about:logins could expose a race condition where trying to access the primary password that should be filled from the console's sent primary secret is done but the code loading the secret from the console has not been ran yet.

Moving to browser-before-ui-startup startup phase.

### Testing 

**Preconditions:**

    Have passwords saved in about:logins
    Have the browser set to open tabs from the last session

**Steps**

    Open about:logins
    While on the about:logins page, sign out of the browser
    Reopen FxE

**Expected result**

    The browser opens up on the about:logins page

### DIFF
--- a/toolkit/components/enterprise/modules/EnterpriseStorageEncryption.sys.mjs
+++ b/toolkit/components/enterprise/modules/EnterpriseStorageEncryption.sys.mjs
@@ -1,0 +1,89 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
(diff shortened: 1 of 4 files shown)


Not-Nik (approved):

toolkit/components/enterprise/modules/EnterpriseStorageEncryption.sys.mjs
```diff
@@ -0,0 +1,89 @@
...
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  ConsoleClient: "resource://gre/modules/enterprise/ConsoleClient.sys.mjs",
+});
+
+export const EnterpriseStorageEncryption = {
+  async init(_window) {
```
jporter-dev: nit: I don't think `_window` is needed here?

jporter-dev (approved):

toolkit/components/enterprise/modules/EnterpriseStorageEncryption.sys.mjs
lissyx: not really yes

PR #1256, opened 2026-08-11
Bug 2059588 - Add console bearer authentication to crash reports/pings

jonathanmendez:
### Description 

Bugzilla: Bug-2059588

Include authentication header with access token from Felt for crashes. For crashreporter app (for app crashes), pass access token into the spawned process's environment.

---

### Testing 

* [x] Added tests

### DIFF
--- a/toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -749,12 +749,30 @@ export const ConsoleClient = {
       Services.obs.addObserver(this, "felt-firefox-access-token-refreshed");
       Services.obs.addObserver(this, "felt-firefox-shutdown");
 
+      // Seed the crash reporter with any token already available at startup.
+      this._syncCrashReporterAuthToken();
+
       this.consoleBaseURI.then(
         ({ hostname }) => lazy.ConsoleProxyBypassFilter.register(hostname),
         e => lazy.log.error("Failed to register console proxy bypass:", e)
       );
     }
     return this;
+  },
+
+  /**
+   * Hand the current access token to the crash reporter so it can authenticate
+   * crash report and crash ping uploads with the console.
+   * Called on every token update in the browser process.
+   */
+  _syncCrashReporterAuthToken() {
+    try {
+      Services.appinfo.setAuthToken(
+        Services.felt.getAccessTokenIfValid() || ""
+      );
+    } catch (e) {
+      lazy.log.warn("Failed to sync crash reporter auth token", e);
+    }
   },
 
   observe(_, topic) {
@@ -781,6 +799,8 @@ export const ConsoleClient = {
         this._refreshResolve?.();
         // The `finally()` block of our promise chain will
         // reset/nullify the promise.
+        // Keep the crash reporter's inherited token in sync.
+        this._syncCrashReporterAuthToken();
         break;
       }
       case "nsPref:changed": {
--- a/toolkit/crashreporter/CrashSubmit.sys.mjs
+++ b/toolkit/crashreporter/CrashSubmit.sys.mjs
@@ -2,6 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
+const lazy = {};
+
+// ConsoleClient only exists in enterprise builds (toolkit/components/enterprise
+// is gated on MOZ_ENTERPRISE), so only define the getter there.
+if (AppConstants.MOZ_ENTERPRISE) {
+  ChromeUtils.defineESModuleGetters(lazy, {
+    CONSOLE_ADDRESS_PREF:
+      "resource://gre/modules/enterprise/ConsoleClient.sys.mjs",
+    ConsoleClient: "resource://gre/modules/enterprise/ConsoleClient.sys.mjs",
+  });
+}
+
 const SUCCESS = "success";
 const FAILED = "failed";
 const SUBMITTING = "submitting";
@@ -46,6 +60,31 @@ async function writeSubmittedReportAsync(crashID, viewURL) {
   }
 
   await writeFileAsync("submitted", `${crashID}.txt`, data);
+}
+
+// Whether the given upload URL is the configured enterprise console, the only
+// destination the Felt bearer token may be sent to.
+function isConsoleURL(url) {
+  if (!AppConstants.MOZ_ENTERPRISE) {
+    return false;
+  }
+
+  let consoleAddress = Services.prefs.getStringPref(
+    lazy.CONSOLE_ADDRESS_PREF,
+    ""
+  );
+  if (!consoleAddress) {
+    return false;
+  }
+
+  try {
+    const origin = new URL(url).origin;
+    // "null" is the opaque origin of URLs like data: or file:, which must never
+    // compare equal to anything.
+    return origin !== "null" && origin === new URL(consoleAddress).origin;
+  } catch {
+    return false;
+  }
 }
 
 // the Submitter class represents an individual submission.
@@ -139,8 +178,10 @@ Submitter.prototype = {
       serverURL = envOverride;
     }
 
-    let xhr = new XMLHttpRequest();
-    xhr.open("POST", serverURL, true);
+    // Only ever send authentication token to the console URL.
+    let shouldAuthenticate = isConsoleURL(serverURL);
+
+    let didAuthRetry = false;
 
     let formData = new FormData();
 
@@ -197,8 +238,30 @@ Submitter.prototype = {
     let manager = Services.crashmanager;
     let submissionID = manager.generateSubmissionID();
 
-    xhr.addEventListener("readystatechange", () => {
+    // Called when the send itself fails (e.g. building or opening the request
+    // throws, or the promise chain below rejects) rather than producing a
+    // response.
+    let failSubmission = err => {
+      this.notifyStatus(FAILED, `failed to send crash report: ${err}`);
+      this.cleanup();
+    };
+
+    let onReadyStateChange = event => {
+      const xhr = event.target;
       if (xhr.readyState == 4) {
+        // Enterprise: if the console rejected the token, refresh it and retry
+        // once before falling through to the normal response handling below.
+        if (
+          AppConstants.MOZ_ENTERPRISE &&
+          shouldAuthenticate &&
+          !didAuthRetry &&
+          (xhr.status === 401 || xhr.status === 403) &&
+          Services.felt?.isFeltBrowser?.()
+        ) {
+          didAuthRetry = true;
+          send(true).catch(failSubmission);
+          return;
+        }
         let ret =
           xhr.status === 200 ? this.parseResponse(xhr.responseText) : {};
         let failmsg;
@@ -208,6 +271,12 @@ Submitter.prototype = {
               case 400:
                 return xhr.responseText;
 
+              case 401:
+                return "Discarded=unauthorized";
+
+              case 403:
+                return "Discarded=forbidden";
+
               case 413:
                 return "Discarded=post_body_too_large";
 
@@ -216,10 +285,11 @@ Submitter.prototype = {
             }
           };
           let err = xhrStatus(xhr.status);
-          if (err.length && err.startsWith("Discarded=")) {
+          const discardedPrefix = "Discarded=";
+          if (err.length && err.startsWith(discardedPrefix)) {
             // Place the error code after, otherwise JS will complain we start
             // with a number when dealing with the telemetry value on JS side
-            const errMsg = `${err.split("Discarded=")[1]}_${xhr.status}`;
+            const errMsg = `${err.substring(discardedPrefix.length)}_${xhr.status}`;
             Glean.crashSubmission.collectorErrors[errMsg].add();
             failmsg = `received bad response: ${xhr.status} ${err}`;
           }
@@ -257,7 +327,27 @@ Submitter.prototype = {
           }
         });
       }
-    });
+    };
+
+    // Enterprise auth as a thin wrapper around the send: fetch the bearer token
+    // (refreshing it when retrying), build the request, attach the (upstream)
+    // response handler, and send. The handler above calls this again with
+    // `refresh = true` to retry once after the console rejected the token.
+    let send = async refresh => {
+      let token = null;
+      if (shouldAuthenticate) {
+        token = refresh
+          ? await this.refreshAuthToken()
+          : await this.getAuthToken();
+      }
+      const xhr = new XMLHttpRequest();
+      xhr.open("POST", serverURL, true);
+      if (token) {
+        xhr.setRequestHeader("Authorization", `Bearer ${token}`);
+      }
+      xhr.addEventListener("readystatechange", onReadyStateChange);
+      xhr.send(formData);
+    };
 
     let p = Promise.all(promises);
     let id = this.id;
@@ -267,10 +357,37 @@ Submitter.prototype = {
         return manager.addSubmissionAttempt(id, submissionID, new Date());
       });
     }
-    p.then(() => {
-      xhr.send(formData);
-    });
+    p.then(() => send(false)).catch(failSubmission);
     return true;
+  },
+
+  // Returns the Felt access token used to authenticate uploads to the
+  // enterprise admin console, or null for non-enterprise builds (where uploads
+  // remain unauthenticated). Never throws: on failure the report is sent
+  // unauthenticated and recorded as a failed submission.
+  async getAuthToken() {
+    if (!AppConstants.MOZ_ENTERPRISE || !Services.felt?.isFeltBrowser?.()) {
+      return null;
+    }
+    try {
+      return await lazy.ConsoleClient.getAccessToken();
+    } catch {
+      return null;
+    }
+  },
+
+  // Force a Felt token refresh and return the new access token, or null on
+  // failure. Never throws; used to retry a submission rejected with 401/403.
+  async refreshAuthToken() {
+    if (!AppConstants.MOZ_ENTERPRISE || !Services.felt?.isFeltBrowser?.()) {
+      return null;
+    }
+    try {
+      await lazy.ConsoleClient._refreshSession();
+      return await lazy.ConsoleClient.getAccessToken();
+    } catch {
+      return null;
+    }
   },
 
   // `ret` is determined based on `status`:
--- a/toolkit/crashreporter/client/app/src/glean.rs
+++ b/toolkit/crashreporter/client/app/src/glean.rs
@@ -181,23 +181,41 @@ mod uploader {
     impl PingUploader for Uploader {
         fn upload(&self, upload_request: CapablePingUploadRequest) -> UploadResult {
             let upload_request = upload_request.capable(|cap| cap.is_empty()).unwrap();
-            let request_builder = http::RequestBuilder::Post {
-                body: upload_request.body.as_slice(),
-                headers: upload_request.headers.as_slice(),
-            };
 
-            let do_send = move || match request_builder.build(upload_request.url.as_ref()) {
-                Err(e) => {
-                    log::error!("failed to build request for glean ping: {e}");
-                    UploadResult::unrecoverable_failure()
-                }
-                Ok(request) => match request.send() {
-                    Err(e) => {
-                        log::error!("failed to send glean ping: {e:#}");
-                        UploadResult::recoverable_failure()
+            let do_send = move || {
+                cfg_if::cfg_if! {
+                    if #[cfg(feature = "enterprise")] {
+                        let mut pairs = upload_request.headers.clone();
+                        let auth = crate::net::auth::enterprise_authorization_header(
+                            upload_request.url.as_ref(),
+                        );
+                        if let Some(header) = auth {
+                            pairs.push(header);
+                        }
+                        let headers = http::header_map_from_pairs(pairs);
+                    } else {
+                        let headers = http::header_map_from_pairs(upload_request.headers.clone());
                     }
-                    Ok(_) => UploadResult::http_status(200),
-                },
+                }
+
+                let request_builder = http::RequestBuilder::Post {
+                    body: upload_request.body.as_slice(),
+                    headers,
+                };
+
+                match request_builder.build(upload_request.url.as_ref()) {
+                    Err(e) => {
+                        log::error!("failed to build request for glean ping: {e}");
+                        UploadResult::unrecoverable_failure()
+                    }
+                    Ok(request) => match request.send() {
+                        Err(e) => {
+                            log::error!("failed to send glean ping: {e:#}");
+                            UploadResult::recoverable_failure()
+                        }
+                        Ok(_) => UploadResult::http_status(200),
+                    },
+                }
             };
 
             #[cfg(mock)]
--- a/toolkit/crashreporter/client/app/src/logic.rs
+++ b/toolkit/crashreporter/client/app/src/logic.rs
@@ -650,6 +650,10 @@ impl ReportCrash {
             dump_file: self.config.dump_file(),
             memory_file: memory_file.as_deref(),
             url,
+            #[cfg(feature = "enterprise")]
+            auth_headers: net::http::header_map_from_pairs(
+                net::auth::enterprise_authorization_header(url),
+            ),
         };
 
         let report_response = async_scoped_thread(|| report.send())
--- a/toolkit/crashreporter/client/app/src/net/auth.rs
+++ b/toolkit/crashreporter/client/app/src/net/auth.rs
@@ -1,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Best-effort Felt bearer token for authenticating enterprise crash uploads.
+//!
+//! In Firefox Enterprise, crash reports and crash pings are uploaded to the
+//! admin console, which requires the same bearer token used for other console
+//! communication. The crashing Firefox process exports its current access token
+//! into its environment as `MOZ_CRASHREPORTER_AUTH_TOKEN`; the crash reporter
+//! client (and the `crashreporterNetworkBackend` background task it may spawn)
+//! inherit that environment and attach it as an `Authorization` header.
+//!
+//! This is best-effort: only the access token that was valid at crash time is
+//! available (the refresh token never leaves the Felt UI process), so there is
+//! no way to refresh it here. If the token is missing or the server rejects it,
+//! the upload is sent unauthenticated (and a pending report is retried later by
+//! an authenticated in-process session).
+
+/// Return an `Authorization: Bearer` header for an upload to `url`, built from
+/// the access token that the crashing process exported into the environment.
+///
+/// Returns `None` if no token is present (non-enterprise builds, or a crash
+/// before sign-in), or if `url` is not on the enterprise console.
+pub fn enterprise_authorization_header(url: &str) -> Option<(String, String)> {
+    let token = crate::std::env::var(ekey!("AUTH_TOKEN")).ok()?;
+    if token.is_empty() {
+        return None;
+    }
+    if !crate::enterprise_prefs::is_console_url(url) {
+        log::warn!("not authenticating the upload: {url} is not the enterprise console");
+        return None;
+    }
+    Some(("Authorization".to_owned(), format!("Bearer {token}")))
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::std::{
+        env::{MockCurrentExe, MockEnv},
+        fs::{MockFS, MockFiles},
+        mock,
+    };
+
+    const CONSOLE_URL: &str = "https://console.example.com/api/browser/crash-reports/submit";
+
+    fn run_with_env(value: Option<&str>, url: &str) -> Option<(String, String)> {
+        let files = MockFiles::new();
+        files.add_dir("work_dir").add_file(
+            "work_dir/firefox.cfg",
+            crate::test::enterprise_autoconfig("https://console.example.com"),
+        );
+
+        let mut builder = mock::builder();
+        builder
+            .set(MockFS, files)
+            .set(MockCurrentExe, "work_dir/crashreporter".into());
+        if let Some(value) = value {
+            builder.set(MockEnv(ekey!("AUTH_TOKEN").into()), value.to_owned());
+        }
+        builder.run(|| enterprise_authorization_header(url))
+    }
+
+    #[test]
+    fn token_present_yields_header() {
+        let header = run_with_env(Some("abc"), CONSOLE_URL).expect("expected a header");
+        assert_eq!(header.0, "Authorization");
+        assert_eq!(header.1, "Bearer abc");
+    }
+
+    #[test]
+    fn empty_token_yields_none() {
+        assert!(run_with_env(Some(""), CONSOLE_URL).is_none());
+    }
+
+    #[test]
+    fn missing_token_yields_none() {
+        assert!(run_with_env(None, CONSOLE_URL).is_none());
+    }
+
+    #[test]
+    fn non_console_url_yields_none() {
+        assert!(run_with_env(Some("abc"), "https://evil.example.com/submit").is_none());
+    }
+}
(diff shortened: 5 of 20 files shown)


toolkit/crashreporter/client/app/src/net/http.rs
```diff
@@ -266,7 +269,10 @@ impl<'a> RequestBuilder<'a> {
         cmd.args(["--user-agent", user_agent()]);
 
         match self {
-            Self::MimePost { parts } => {
+            Self::MimePost { parts, headers } => {
+                for (k, v) in headers.iter() {
+                    cmd.args(["--header", &format!("{k}: {v}")]);
+                }
```
afranchuk: Copilot had a suggestion to sanitize the bearer header value, but I think it'd be even better to sanitize headers here (and in the libcurl variant) to eliminate the potential of incorrectly invoking curl.

afranchuk: It's perhaps overkill, but I wouldn't be opposed to pulling in the `http` crate to use `HeaderName`/`HeaderValue` for validation.

toolkit/crashreporter/client/app/src/net/report.rs
```diff
@@ -35,6 +35,9 @@ pub struct CrashReport<'a> {
     pub dump_file: &'a Path,
     pub memory_file: Option<&'a Path>,
     pub url: &'a str,
+    /// Extra HTTP headers (e.g. an enterprise `Authorization` header). Empty
+    /// for non-enterprise uploads.
+    pub auth_headers: Vec<(String, String)>,
```
afranchuk: This could be behind a `cfg`, but it's not a big deal.

toolkit/crashreporter/client/app/src/glean.rs
```diff
@@ -181,23 +181,40 @@ mod uploader {
...
+                #[cfg(feature = "enterprise")]
+                let headers = {
+                    let mut headers = upload_request.headers.clone();
+                    if let Some(header) = crate::net::auth::enterprise_authorization_header() {
+                        headers.push(header);
+                    }
+                    headers
+                };
+                #[cfg(not(feature = "enterprise"))]
+                let headers = &upload_request.headers;
```
afranchuk: I'd prefer a form that uses `cfg_if!`.

toolkit/crashreporter/client/app/src/logic.rs
```diff
@@ -645,11 +645,16 @@ impl ReportCrash {
 
         // Send the report to the server.
         let memory_file = self.config.memory_file();
+        #[cfg(feature = "enterprise")]
+        let auth_headers = net::auth::enterprise_authorization_headers();
+        #[cfg(not(feature = "enterprise"))]
+        let auth_headers = Vec::new();
```
afranchuk: Again, `cfg_if!` might be nicer. Or make the `auth_headers` field enterprise-only.

toolkit/crashreporter/networking/BackgroundTask_crashreporterNetworkBackend.sys.mjs
```diff
@@ -28,6 +28,7 @@ async function createRequestInit(requestBuilder) {
       }
       return {
         method: "POST",
+        headers: Object.fromEntries(requestBuilder.headers ?? []),
         body: formData,
       };
```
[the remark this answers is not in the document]
afranchuk: Not really a concern since we control the headers, but it doesn't hurt to do.

toolkit/crashreporter/CrashSubmit.sys.mjs
```diff
@@ -197,80 +200,132 @@ Submitter.prototype = {
...
+        }
+      });
+      xhr.send(formData);
     });
+  },
 
-    let p = Promise.all(promises);
-    let id = this.id;
+  async sendRequest(serverURL, formData, promises, manager, submissionID) {
+    await Promise.all(promises);
```
afranchuk: This feels weird here. Maybe it should just be
`Promise.all(promises).then(() => this.sendRequest(...))` at the `sendRequest` call site.

toolkit/crashreporter/CrashSubmit.sys.mjs
```diff
@@ -197,80 +200,132 @@ Submitter.prototype = {
...
+
+    await this.handleResponse(xhr, manager, submissionID);
+  },
+
+  async handleResponse(xhr, manager, submissionID) {
+    let ret = xhr.status === 200 ? this.parseResponse(xhr.responseText) : {};
+    let failmsg;
+    if (xhr.status !== 200) {
+      const xhrStatus = code => {
+        switch (code) {
```
afranchuk: `401`/`403` are possibilities now. Add a more descriptive discard reason for them.

toolkit/crashreporter/CrashSubmit.sys.mjs
```diff
@@ -197,80 +200,132 @@ Submitter.prototype = {
...
+
+          default:
+            return "Discarded=unknown_error";
+        }
+      };
+      let err = xhrStatus(xhr.status);
+      if (err.length && err.startsWith("Discarded=")) {
+        // Place the error code after, otherwise JS will complain we start
+        // with a number when dealing with the telemetry value on JS side
+        const errMsg = `${err.split("Discarded=")[1]}_${xhr.status}`;
```
afranchuk: This was existing code, but it could probably just use `substring`.
```suggestion
      const discardedPrefix = "Discarded=";
      if (err.length && err.startsWith(discardedPrefix)) {
        // Place the error code after, otherwise JS will complain we start
        // with a number when dealing with the telemetry value on JS side
        const errMsg = `${err.substring(discardedPrefix.length)}_${xhr.status}`;
```

afranchuk (changes requested):

toolkit/crashreporter/CrashSubmit.sys.mjs
```diff
@@ -197,80 +200,132 @@ Submitter.prototype = {
...
+
+  // Force a token refresh, then return the new access token. Used to retry a
+  // submission that was rejected with a 401/403.
+  async refreshAuthToken() {
+    if (!Services.felt?.isFeltBrowser?.()) {
+      return null;
+    }
+    await lazy.ConsoleClient._refreshSession();
+    return lazy.ConsoleClient.getAccessToken();
+  },
```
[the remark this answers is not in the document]
jonathanmendez: This is now caught with a null token returned.

toolkit/crashreporter/client/app/src/net/auth.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Return an `Authorization: Bearer` header built from the access token that
+/// the crashing process exported into the environment, or `None` if no token is
+/// present (non-enterprise builds, or a crash before sign-in).
+pub fn enterprise_authorization_header() -> Option<(String, String)> {
+    let token = crate::std::env::var(ekey!("AUTH_TOKEN")).ok()?;
+    if token.is_empty() {
+        return None;
+    }
+    Some(("Authorization".to_owned(), format!("Bearer {token}")))
+}
```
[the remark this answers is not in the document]
jonathanmendez: Current implementation uses HeaderName/HeaderValue which sanitizes these issues.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -816,6 +827,23 @@ export const ConsoleClient = {
...
+  _syncCrashReporterAuthToken() {
+    try {
+      Services.env.set(
+        CRASHREPORTER_AUTH_TOKEN_ENV,
+        Services.felt.getAccessTokenIfValid() || ""
+      );
+    } catch (e) {
+      lazy.log.warn("Failed to sync crash reporter auth token", e);
+    }
+  },
```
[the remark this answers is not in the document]
jonathanmendez: Trying to solve this added a lot of code in sensitive areas. I'm not super familiar with the other platforms so I'll need to spend more time checking for correctness there.

toolkit/crashreporter/client/app/src/net/http.rs
jonathanmendez: I pulled in http and used HeaderName/HeaderValue for the actual fields, not just for sanitizing. If this is too much of a change, let me know and I'll simplify to just sanitizing.

toolkit/crashreporter/client/app/src/net/report.rs
jonathanmendez: Good call, done.

toolkit/crashreporter/CrashSubmit.sys.mjs
jonathanmendez: I restructured to better match the original code (which will help with merging from upstream as well). It hopefully reads less weirdly overall now.

toolkit/crashreporter/CrashSubmit.sys.mjs
jonathanmendez: Let me know if what I added was ok.

toolkit/crashreporter/client/app/src/net/http.rs
afranchuk: This is good!

toolkit/crashreporter/client/app/src/net/http.rs
```diff
@@ -112,21 +115,73 @@ std::mock::mocked_static! {
...
+}
+
+/// Format a header as a `name: value` line for curl/libcurl, or `None` if the
+/// value is not representable as a string. The value is already guaranteed
+/// free of CR/LF and control characters by [`HeaderValue`], so the resulting
+/// line cannot inject additional headers.
+fn curl_header_line(name: &HeaderName, value: &HeaderValue) -> Option<String> {
+    value
+        .to_str()
+        .ok()
```
afranchuk: We use `from_str` to create the `HeaderValue`, so `to_str` should succeed (and thus this could be an `unwrap`/`expect`, and `curl_header_line` could return `String`).

toolkit/crashreporter/CrashSubmit.sys.mjs
```diff
@@ -200,131 +209,139 @@ Submitter.prototype = {
...
-  // which requires the same Felt bearer token used for other console
-  // communication. Returns the access token, or null for non-enterprise
-  // builds (where uploads remain unauthenticated).
-  async getAuthToken() {
-    if (!Services.felt?.isFeltBrowser?.()) {
-      return null;
-    }
-    return lazy.ConsoleClient.getAccessToken();
-  },
+    let onReadyStateChange = () => {
```
afranchuk: I sort of liked the way this logic was refactored into separate async functions earlier, but now it's gone back a bit (albeit it's still separate lambdas, but not async lambdas for whatever reason).

toolkit/crashreporter/nsExceptionHandler.cpp
```diff
@@ -1238,6 +1293,28 @@ static bool LaunchProgram(const XP_CHAR* aProgramPath,
...
+      static const size_t kMaxEnvEntries = 512;
+      char* childEnv[kMaxEnvEntries + 2];
+      size_t n = 0;
+      while (environ[n] && n < kMaxEnvEntries) {
+        childEnv[n] = environ[n];
+        ++n;
+      }
+      if (!environ[n]) {  // the environment fit within our bound
+        childEnv[n++] = const_cast<char*>(gAuthTokenEnvEntry.get());
+        childEnv[n] = nullptr;
```
afranchuk: Move the "copy environment and add values" logic to a separate function (or multiple platform-specific functions, I don't care either way). Also, share the logic between mac/linux (they are the same aside from how the original environment is obtained). For mac/linux, we want to keep things on the stack, so the function will have to accept that as an argument (well, it could return it too, and rely on the compiler doing the right thing to avoid a copy, which it ought to).

toolkit/crashreporter/nsExceptionHandler.cpp
```diff
@@ -2626,6 +2704,19 @@ nsresult SetServerURL(const nsACString& aServerURL) {
...
+nsresult SetAuthToken(const nsACString& aToken) {
+  // Pre-format the full environment entry so that LaunchProgram (which runs on
+  // the crash path) only has to reference it without allocating. The token is
+  // handed to the crash reporter child alone, not set in our own environment.
+  if (aToken.IsEmpty()) {
+    gAuthTokenEnvEntry.Truncate();
+  } else {
+    gAuthTokenEnvEntry.AssignLiteral("MOZ_CRASHREPORTER_AUTH_TOKEN=");
+    gAuthTokenEnvEntry.Append(aToken);
+  }
```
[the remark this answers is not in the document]
afranchuk: Even though we control the input at the one call site, since this is on appinfo and accessible in JS, it's probably best to validate.

toolkit/crashreporter/CrashSubmit.sys.mjs
afranchuk: I kinda liked the new structure with async functions. Maybe we could have async lambdas at least to make it less awkward?

toolkit/crashreporter/CrashSubmit.sys.mjs
afranchuk: Looks good!

afranchuk (changes requested):
I think there's just a few more things I'd like changed.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -808,6 +808,9 @@ export const ConsoleClient = {
       Services.obs.addObserver(this, "felt-firefox-access-token-refreshed");
       Services.obs.addObserver(this, "felt-firefox-shutdown");
 
+      // Seed the crash reporter with any token already available at startup.
+      this._syncCrashReporterAuthToken();
+
```
lissyx, line 754: Do we really have any at that moment?

toolkit/crashreporter/client/app/src/net/auth.rs
```diff
@@ -0,0 +1,60 @@
...
+//! communication. The crashing Firefox process exports its current access token
+//! into its environment as `MOZ_CRASHREPORTER_AUTH_TOKEN`; the crash reporter
+//! client (and the `crashreporterNetworkBackend` background task it may spawn)
+//! inherit that environment and attach it as an `Authorization` header.
+//!
+//! This is best-effort: only the access token that was valid at crash time is
+//! available (the refresh token never leaves the Felt UI process), so there is
+//! no way to refresh it here. If the token is missing or the server rejects it,
+//! the upload is sent unauthenticated (and a pending report is retried later by
+//! an authenticated in-process session).
```
lissyx, line 18: What is the compromise here: why is it better to send authenticated? is it the only way we can link a crash report to a user ?

toolkit/crashreporter/client/app/src/net/auth.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Return an `Authorization: Bearer` header built from the access token that
+/// the crashing process exported into the environment, or `None` if no token is
+/// present (non-enterprise builds, or a crash before sign-in).
+pub fn enterprise_authorization_header() -> Option<(String, String)> {
+    let token = crate::std::env::var(ekey!("AUTH_TOKEN")).ok()?;
```
lissyx, line 26: `AUTH_TOKEN` may be way too much generic, we could get into clashes. `CRASHREPORTER_AUTH_TOKEN` might be better scopped?

toolkit/crashreporter/client/app/src/net/report.rs
```diff
@@ -87,7 +90,14 @@ impl CrashReport<'_> {
...
 
-            request = Some(http::RequestBuilder::MimePost { parts }.build(self.url)?);
+            cfg_if::cfg_if! {
+                if #[cfg(feature = "enterprise")] {
+                    let headers = self.auth_headers.clone();
+                } else {
+                    let headers = Default::default();
+                }
+            };
+            request = Some(http::RequestBuilder::MimePost { parts, headers }.build(self.url)?);
```
lissyx, line 100: nit: can we write rather something like this? Just for ease of reading
```suggestion
            let headers = cfg_if::cfg_if! {
                if #[cfg(feature = "enterprise")] {
                    self.auth_headers.clone()
                } else {
                    Default::default()
                }
            };
            request = Some(http::RequestBuilder::MimePost { parts, headers }.build(self.url)?);
```

lissyx: also we run no risk with `Default::default` as a default value?

toolkit/crashreporter/client/app/src/glean.rs
```diff
@@ -181,23 +181,38 @@ mod uploader {
...
-                    Ok(_) => UploadResult::http_status(200),
-                },
+                    Ok(request) => match request.send() {
+                        Err(e) => {
+                            log::error!("failed to send glean ping: {e:#}");
+                            UploadResult::recoverable_failure()
+                        }
+                        Ok(_) => UploadResult::http_status(200),
+                    },
+                }
```
lissyx, line 218: LGTM but do we want to get feedback from glean people maybe?

toolkit/crashreporter/client/app/src/net/auth.rs
afranchuk: `ekey!` expands to `MOZ_CRASHREPORTER_AUTH_TOKEN`.

toolkit/crashreporter/client/app/src/test.rs
```diff
@@ -1378,7 +1378,8 @@ fn background_task_network_backend() {
                             "value": platform_path("data_dir/pending/minidump.memory.json.gz"),
                         },
                     }
-                ]
+                ],
+                "headers": []
```
lissyx, line 1408: shouldn't we update with a variant with real headers?

toolkit/crashreporter/client/app/src/test.rs
```diff
@@ -1537,7 +1538,8 @@ fn background_task_curl_fallback() {
                                 "value": platform_path("data_dir/pending/minidump.dmp"),
                             },
                         }
-                    ]
+                    ],
+                    "headers": []
```
lissyx, line 1875: same ?

toolkit/crashreporter/client/app/src/net/auth.rs
lissyx: oof, I missed it

maybe worth updating the comment?

toolkit/crashreporter/CrashSubmit.sys.mjs
```diff
@@ -197,8 +209,20 @@ Submitter.prototype = {
     let manager = Services.crashmanager;
     let submissionID = manager.generateSubmissionID();
 
-    xhr.addEventListener("readystatechange", () => {
+    let onReadyStateChange = () => {
```
lissyx: this is making `xhr` scope less obvious

lissyx:
```suggestion
    let onReadyStateChange = (event) => {
```

and then use `event.target` to get back to `xhr` in the body?

lissyx: (even though there are other uses of `xhr.` below outside of the callback)

toolkit/crashreporter/CrashSubmit.sys.mjs
```diff
@@ -139,8 +151,8 @@ Submitter.prototype = {
       serverURL = envOverride;
     }
 
-    let xhr = new XMLHttpRequest();
-    xhr.open("POST", serverURL, true);
+    let xhr;
+    let didAuthRetry = false;
```
lissyx, line 184: I do not see any other use than setting this var, what is its role?

toolkit/crashreporter/nsExceptionHandler.cpp
lissyx, line 1: Did @afranchuk had a look here?

toolkit/crashreporter/nsExceptionHandler.cpp
```diff
@@ -1205,13 +1216,38 @@ static bool LaunchProgram(const XP_CHAR* aProgramPath,
...
+  DWORD creationFlags =
+      NORMAL_PRIORITY_CLASS | CREATE_NO_WINDOW | CREATE_BREAKAWAY_FROM_JOB;
+  LPVOID envBlock = nullptr;  // null => inherit our environment
+  // Storage for a custom environment block; must outlive the CreateProcess
+  // call.
+  nsAutoString childEnv;
+  if (!gAuthTokenEnvEntry.IsEmpty()) {
+    // Build a child-only environment block (a copy of ours plus the auth token)
+    // so the token is passed to the crash reporter alone, rather than living in
+    // our environment where every child process would inherit it.
```
lissyx: I dont know on Windows but on Linux, messing around with the env may have weird outcomes, we have to be very careful here

toolkit/crashreporter/nsExceptionHandler.cpp
```diff
@@ -1227,6 +1263,25 @@ static bool LaunchProgram(const XP_CHAR* aProgramPath,
...
+  // If we have an auth token, pass it to the crash reporter through a
+  // child-only copy of the environment rather than setting it in our own
+  // environment (where every child process would inherit it). Built on the
+  // stack to avoid allocation on the crash path.
+  static const size_t kMaxEnvEntries = 512;
+  char* childEnv[kMaxEnvEntries + 2];
+  if (env && !gAuthTokenEnvEntry.IsEmpty()) {
+    size_t n = 0;
+    while (env[n] && n < kMaxEnvEntries) {
+      childEnv[n] = env[n];
```
lissyx: We're just copying pointers here?

lissyx (approved):

toolkit/crashreporter/client/app/src/net/report.rs
jonathanmendez: I had originally wanted to do something like that, but `cfg_if!` is [not supported for use in an expression](https://github.com/rust-lang/cfg-if/issues/43), only for statements.

I don't imagine that there's a realistic risk from `Default::default` here; a change of the default implementation to a non-empty map would be a pretty drastic and unreasonable change that I don't see happening.

toolkit/crashreporter/client/app/src/net/auth.rs
jonathanmendez: Sending it authenticated is the only way the console will accept it (security review flagged that it wouldn't be acceptable to leave the endpoint unauthenticated). If anything, I could try to update the code to just fail early rather than try to send it unauthenticated.

toolkit/crashreporter/CrashSubmit.sys.mjs
jonathanmendez: It's used in `onReadyStateChange` to determine whether to call `refreshAuthToken` and try sending again. If it does another attempt it sets `didAuthRetry` to `true` to only do one extra attempt. Overall this is similar to [`_didRefresh` in ConsoleClient](https://searchfox.org/enterprise-main/source/toolkit/components/enterprise/modules/ConsoleClient.sys.mjs#389-422).

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
jonathanmendez: From my reading of the code, no. I'll remove this call.

jonathanmendez: Actually I was wrong, and it was important to keep this line 😛

toolkit/crashreporter/CrashSubmit.sys.mjs
jonathanmendez: As discussed elsewhere, I think trying to keep the structure similar enough is worthwhile for making merge conflicts easier, but I've updated to some async lambdas, let me know if I got it in enough of a better shape.

toolkit/crashreporter/nsExceptionHandler.cpp
```diff
@@ -1199,6 +1199,67 @@ static void AnnotateMemoryStatus(AnnotationTable&) {
...
+// `aCapacity` slots) and append the auth token. Returns the null-terminated
+// `aBuffer`, or nullptr when there is no token or the environment did not fit,
+// in which case the caller launches with the inherited environment.
+//
+// The caller supplies the buffer so the result lives on the caller's stack: on
+// Linux this runs post-fork in a signal-handler context, where heap allocation
+// is unsafe. macOS and Linux share this logic and differ only in how `aSource`
+// is obtained.
+static char** CopyEnvWithAuthToken(char** aSource, char** aBuffer,
+                                   size_t aCapacity) {
```
afranchuk, line 1211: You could technically omit the capacity and hard-code the buffer size, but leaving it as-is is definitely more conventional. Also, to make the compiler actually enforce the array size, you have to pass a pointer (e.g. `char* (*aBuffer)[kChildEnvCapacity]`). For whatever reason it won't enforce it as just `char* aBuffer[kChildEnvCapacity]`. Terrible language.

So anyway, it's fine as-is :)

afranchuk (approved):
This looks great, especially the tests!

toolkit/crashreporter/CrashSubmit.sys.mjs
jonathanmendez: Thanks for the suggestion, I've updated in a way that xhr scope is much clearer.

toolkit/crashreporter/client/app/src/glean.rs
```diff
@@ -181,23 +181,38 @@ mod uploader {
...
+                    body: upload_request.body.as_slice(),
+                    headers,
+                };
+
+                match request_builder.build(upload_request.url.as_ref()) {
                     Err(e) => {
-                        log::error!("failed to send glean ping: {e:#}");
-                        UploadResult::recoverable_failure()
+                        log::error!("failed to build request for glean ping: {e}");
+                        UploadResult::unrecoverable_failure()
```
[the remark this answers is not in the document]
jonathanmendez: Bad diff analysis/lineup, this logic didn't change, the diff renderer is just mismatching lines.

toolkit/crashreporter/client/app/src/net/http.rs
```diff
@@ -349,22 +438,37 @@ static REQUEST_NUM: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUs
...
+        // component so that the path cannot be guessed by another process (the contents may
+        // include sensitive headers).
         let path = std::env::temp_dir().join(format!(
-            "{}{}-request{}.json",
+            "{}{}-{kind}{}-{}.{extension}",
             env!("CARGO_PKG_NAME"),
             std::process::id(),
-            REQUEST_NUM.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+            REQUEST_NUM.fetch_add(1, std::sync::atomic::Ordering::Relaxed),
+            uuid::Uuid::new_v4().simple()
```
afranchuk: If there's a random component, there's no need for the process id nor process-unique id.

toolkit/crashreporter/nsExceptionHandler.cpp
```diff
@@ -1188,6 +1194,72 @@ static void AnnotateMemoryStatus(AnnotationTable&) {
...
+    return;
+  }
+  for (LPWCH v = curEnv; *v; v += wcslen(v) + 1) {
+    aBlock.Append(nsDependentString(reinterpret_cast<const char16_t*>(v)));
+    aBlock.Append(char16_t(0));
+  }
+  FreeEnvironmentStringsW(curEnv);
+  AppendUTF8toUTF16(gAuthTokenEnvEntry, aBlock);
+  aBlock.Append(char16_t(0));  // terminate the token entry
+  aBlock.Append(char16_t(0));  // terminate the block
```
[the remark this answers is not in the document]
afranchuk: TIL windows environment blocks are miserable. Also the only doc that explicitly mentions the sort order requirement is https://learn.microsoft.com/en-us/windows/win32/procthread/changing-environment-variables. https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables doesn't mention it. 🤦

afranchuk (changes requested):

toolkit/crashreporter/nsExceptionHandler.cpp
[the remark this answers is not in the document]
jonathanmendez: Oh I forgot to comment here that I found https://nullprogram.com/blog/2023/08/23/ that empirically found the sort requirement to apparently be bunk

toolkit/crashreporter/client/app/src/net/http.rs
```diff
@@ -108,22 +111,72 @@ std::mock::mocked_static! {
...
+/// We only construct HeaderValues with from_str so to_str is safe to unwrap.
+fn curl_header_line(name: &HeaderName, value: &HeaderValue) -> String {
+    format!("{}: {}", name.as_str(), value.to_str().unwrap())
+}
+
+/// Serialize a [`HeaderMap`] as an array of `[name, value]` pairs, matching the
+/// JSON format expected by `BackgroundTask_crashreporterNetworkBackend.sys.mjs`.
+fn serialize_headers<S>(headers: &HeaderMap, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
```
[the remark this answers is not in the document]
jonathanmendez: Using a UUID in the file name and create_new to avoid an attack via an existing file. The file gets overwritten (with the response) by the spawned background process and deleted when dropped here after reading the response, so the risk is fairly isolated.

toolkit/crashreporter/client/app/src/logic.rs
```diff
@@ -650,6 +650,10 @@ impl ReportCrash {
             dump_file: self.config.dump_file(),
             memory_file: memory_file.as_deref(),
             url,
+            #[cfg(feature = "enterprise")]
+            auth_headers: net::http::header_map_from_pairs(
+                net::auth::enterprise_authorization_header(),
+            ),
```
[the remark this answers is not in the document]
jonathanmendez: I'll add origin validation once https://github.com/mozilla/enterprise-firefox/pull/1076 lands

toolkit/crashreporter/client/app/src/net/auth.rs
```diff
@@ -0,0 +1,60 @@
...
+    if token.is_empty() {
+        return None;
+    }
+    Some(("Authorization".to_owned(), format!("Bearer {token}")))
```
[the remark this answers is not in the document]
jonathanmendez: I'll add origin validation once https://github.com/mozilla/enterprise-firefox/pull/1076 lands

toolkit/crashreporter/client/app/src/net/http.rs
```diff
@@ -266,14 +319,17 @@ impl<'a> RequestBuilder<'a> {
         cmd.args(["--user-agent", user_agent()]);
 
         match self {
-            Self::MimePost { parts } => {
+            Self::MimePost { parts, headers } => {
+                for (name, value) in headers.iter() {
+                    cmd.args(["--header", &curl_header_line(name, value)]);
```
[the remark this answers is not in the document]
jonathanmendez: Now passed via a randomized temporary file.

xpcom/system/nsICrashReporter.idl
```diff
@@ -196,4 +196,10 @@ interface nsICrashReporter : nsISupports
    * @throws NS_ERROR_NOT_INITIALIZED if crash reporting is disabled.
    */
   void saveMemoryReport();
+
+  /**
+   * Set the bearer token used to authenticate crash report and crash ping
+   * uploads to the enterprise console. Pass an empty string to clear it.
+   */
+  void setAuthToken(in ACString token);
```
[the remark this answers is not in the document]
jonathanmendez: outdated advice

toolkit/crashreporter/client/app/src/net/http.rs
```diff
@@ -349,22 +438,34 @@ static REQUEST_NUM: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUs
...
+    fn create(kind: &str, extension: &str) -> std::io::Result<Self> {
+        // Incorporate a random component to avoid conflicts and so that the path cannot be
+        // guessed by another process (the contents may include sensitive headers).
         let path = std::env::temp_dir().join(format!(
-            "{}{}-request{}.json",
-            env!("CARGO_PKG_NAME"),
-            std::process::id(),
-            REQUEST_NUM.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+            "{kind}{}-{}.{extension}",
+            REQUEST_NUM.fetch_add(1, std::sync::atomic::Ordering::Relaxed),
```
afranchuk: We don't need `REQUEST_NUM` anymore.

```suggestion
            "{kind}-{}.{extension}",
```

afranchuk (changes requested):

afranchuk (approved):

PR #1259, opened 2026-08-11
Bug 2062592 - Fix schema validation for AIChatbot policy custom provider

jporter-dev:
### Description 

Bugzilla: Bug-2062592

The policy schema was recently updated upstream, but the `AIChatbot` policy which is only in the `enterprise-firefox` repo contained old type definitions. 

This PR updates the type of `AIChatbot.Providers.Add.url` to use the new schema definition and adds a simple xpcshell test for this case.

---
### Screenshots

image

_Before fix: AIChatbot policy shows errors and custom provider is not present in sidebar_

image

_After fix: AIChatbot policy with no errors and custom provider present in sidebar_

---

### Testing 

* [x] Added tests

**Steps to verify changes:**
1. Set a valid `AIChatbot` policy
2. Navigate to `about:policies`
3. Open the AIChatbot sidebar

**Expected result:**

`about:policies` displays no errors, and the custom chatbot appears in the AI Chatbot sidebar tool.

### DIFF
--- a/browser/components/enterprisepolicies/schemas/policies-schema.json
+++ b/browser/components/enterprisepolicies/schemas/policies-schema.json
@@ -132,9 +132,7 @@
               "items": {
                 "type": "object",
                 "properties": {
-                  "url": {
-                    "type": "URL"
-                  },
+                  "url": { "$ref": "#/definitions/url" },
                   "name": {
                     "type": "string"
                   },
--- a/browser/components/enterprisepolicies/tests/xpcshell/test_aichatbot.js
+++ b/browser/components/enterprisepolicies/tests/xpcshell/test_aichatbot.js
@@ -1,0 +1,48 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+"use strict";
+
+const { CHAT_PROVIDERS_DEFAULT } = ChromeUtils.importESModule(
+  "resource:///modules/GenAI.sys.mjs"
+);
+
+add_task(async function test_aichatbot_custom_provider_added() {
+  await setupPolicyEngineWithJson({
+    policies: {
+      AIChatbot: {
+        Providers: {
+          Add: [
+            {
+              id: "my-private-ai",
+              name: "My Private AI",
+              url: "https://example.com/chat",
+              iconUrl: "https://example.com/icon.png",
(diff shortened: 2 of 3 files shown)


1rneh (approved):

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -132,9 +132,7 @@
               "items": {
                 "type": "object",
                 "properties": {
-                  "url": {
-                    "type": "URL"
-                  },
+                  "url": { "$ref": "#/definitions/url" },
```
lissyx, line 135: Assuming this is related to the freeze that romain was experiencing (I could not repro at all), how come this was not caught in CI ?

jporter-dev: I couldn't repro either - I didn't get any freezing but I did get the `about:policies` error and the policy was not applied. 

re: CI - looks like this test should handle it but I don't think we show xpcshell tests in PRs: https://treeherder.mozilla.org/logviewer?job_id=584331358&repo=enterprise-main&task=W1WbsFjKQC2R1RfooYHkSw.0&lineNumber=8298-8525

@bsmth I think a bunch of policies that live only in this repo need updating to the new schema changes if we want to surface this test for PRs. From what I could tell, the AIChatbot bug is the only "breaking" failure.

lissyx: > I couldn't repro either - I didn't get any freezing but I did get the `about:policies` error and the policy was not applied.
> 
> re: CI - looks like this test should handle it but I don't think we show xpcshell tests in PRs: https://treeherder.mozilla.org/logviewer?job_id=584331358&repo=enterprise-main&task=W1WbsFjKQC2R1RfooYHkSw.0&lineNumber=8298-8525

We need to ensure they are green outside of our failures before showing on PRs

From looking at this link, it looks like we are super close to it at least on linux, so maybe we should fix those remainders of failure and turn `xpcshell` into something we monitor closely

jporter-dev: Created bug-2062723 for that test and left a comment with this context on bug-2012260 so this doesn't get lost since monitoring for policy breakage is rather important. Thanks for calling this out!

PR #1261, opened 2026-08-11
Bug 2058036 - Combine the *Telemetry policies into a single SecurityLogging policy

gcp:
### Description

Bugzilla: [Bug 2058036](https://bugzilla.mozilla.org/show_bug.cgi?id=2058036)

Consolidates `DownloadTelemetry`, `PrintPageTelemetry`, and
`BlocklistDomainBrowsedTelemetry` into a single `SecurityLogging` policy with a
sub-object for each security event. A separate second commit adds control over
the existing enterprise add-on installation event.

```json
{
  "policies": {
    "SecurityLogging": {
      "AddonInstall": { "Enabled": true },
      "BlocklistDomainBrowsed": { "Enabled": true, "UrlLogging": "full" },
      "Download": {
        "Enabled": true,
        "UrlLogging": "full",
        "FileLogging": "metadata"
      },
      "PrintPage": { "Enabled": true, "UrlLogging": "domain" },
      "UnsafeDownload": { "Enabled": true, "UrlLogging": "full" },
      "UnsafeSiteVisit": { "Enabled": true, "UrlLogging": "full" }
    }
  }
}
```

The original telemetry preferences remain unchanged.
`AddonInstall.Enabled` uses a new preference and preserves existing collection
behavior when the policy is absent. The three superseded policy names were not
customer-facing and are removed without compatibility aliases.

### Implementation

`SecurityLoggingPolicy.sys.mjs` maps explicitly configured settings to locked
preferences after schema validation. Live updates remove settings from the
previous policy before applying the new configuration, so unrelated or omitted
preferences remain untouched.

The schema supports enterprise version 156 and marks the policy as not requiring
a restart. Nested objects work with the existing Windows GPO and macOS plist
parsers.

### Testing

Tests configure values through the enterprise policy engine and cover:

- applying and removing every `SecurityLogging` setting;
- live remote-policy updates and their effect on download telemetry;
- enabled, disabled, and removed add-on installation states through a real
  Addon Manager installation flow; and
- WebsiteFilter telemetry using the consolidated policy.

Validation on the final rebased stack:

- `./mach format`
- `./mach lint` (0 errors; one unrelated existing warning)

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -31,6 +31,8 @@ ChromeUtils.defineESModuleGetters(lazy, {
   FileUtils: "resource://gre/modules/FileUtils.sys.mjs",
   ProxyPolicies: "resource:///modules/policies/ProxyPolicies.sys.mjs",
   SearchService: "moz-src:///toolkit/components/search/SearchService.sys.mjs",
+  SecurityLoggingPolicy:
+    "resource:///modules/policies/SecurityLoggingPolicy.sys.mjs",
   AIChatbotPolicies: "resource:///modules/policies/AIChatbotPolicies.sys.mjs",
   QuickSuggest: "moz-src:///browser/components/urlbar/QuickSuggest.sys.mjs",
   WatermarkPolicy: "resource:///modules/policies/WatermarkPolicy.sys.mjs",
@@ -605,47 +607,6 @@ export var Policies = {
     onRemove(manager, _oldParams) {
       lazy.unblockAboutPage(manager, "about:support");
       manager.allowFeature("aboutSupport");
-    },
-  },
-
-  BlocklistDomainBrowsedTelemetry: {
-    onBeforeAddons(manager, param) {
-      if (param && typeof param === "object") {
-        // Enable/disable blocklist domain browsed telemetry
-        if (typeof param.Enabled === "boolean") {
-          lazy.PoliciesUtils.setAndLockPref(
-            "browser.policies.enterprise.telemetry.blocklistDomainBrowsed.enabled",
-            param.Enabled
-          );
-        } else {
-          lazy.PoliciesUtils.unsetAndUnlockPref(
-            "browser.policies.enterprise.telemetry.blocklistDomainBrowsed.enabled"
-          );
-        }
-
-        // Set URL logging level
-        if (
-          typeof param.UrlLogging === "string" &&
-          ["full", "domain", "none"].includes(param.UrlLogging)
-        ) {
-          lazy.PoliciesUtils.setAndLockPref(
-            "browser.policies.enterprise.telemetry.blocklistDomainBrowsed.urlLogging",
-            param.UrlLogging
-          );
-        } else {
-          lazy.PoliciesUtils.unsetAndUnlockPref(
-            "browser.policies.enterprise.telemetry.blocklistDomainBrowsed.urlLogging"
-          );
-        }
-      }
-    },
-    onRemove(_manager, _oldParams) {
-      lazy.PoliciesUtils.unsetAndUnlockPref(
-        "browser.policies.enterprise.telemetry.blocklistDomainBrowsed.enabled"
-      );
-      lazy.PoliciesUtils.unsetAndUnlockPref(
-        "browser.policies.enterprise.telemetry.blocklistDomainBrowsed.urlLogging"
-      );
     },
   },
 
@@ -1711,61 +1672,6 @@ export var Policies = {
         "browser.download.useDownloadDir",
         true
       );
-    },
-  },
-
-  DownloadTelemetry: {
-    onBeforeAddons(manager, param) {
-      if (param && typeof param === "object") {
-        // Enable/disable download telemetry
-        if (typeof param.Enabled === "boolean") {
-          lazy.PoliciesUtils.setAndLockPref(
-            "browser.download.enterprise.telemetry.enabled",
-            param.Enabled
-          );
-        }
-
-        // Set URL logging level
-        if (
-          typeof param.UrlLogging === "string" &&
-          ["full", "domain", "none"].includes(param.UrlLogging)
-        ) {
-          lazy.PoliciesUtils.setAndLockPref(
-            "browser.download.enterprise.telemetry.urlLogging",
-            param.UrlLogging
-          );
-        }
-
-        // Set file logging level
-        if (
-          typeof param.FileLogging === "string" &&
-          ["full", "metadata", "none"].includes(param.FileLogging)
-        ) {
-          lazy.PoliciesUtils.setAndLockPref(
-            "browser.download.enterprise.telemetry.fileLogging",
-            param.FileLogging
-          );
-        }
-      }
-    },
-    onRemove(manager, oldParams) {
-      if (oldParams && typeof oldParams === "object") {
-        if ("Enabled" in oldParams) {
-          lazy.PoliciesUtils.unsetAndUnlockPref(
-            "browser.download.enterprise.telemetry.enabled"
-          );
-        }
-        if ("UrlLogging" in oldParams) {
-          lazy.PoliciesUtils.unsetAndUnlockPref(
-            "browser.download.enterprise.telemetry.urlLogging"
-          );
-        }
-        if ("FileLogging" in oldParams) {
-          lazy.PoliciesUtils.unsetAndUnlockPref(
-            "browser.download.enterprise.telemetry.fileLogging"
-          );
-        }
-      }
     },
   },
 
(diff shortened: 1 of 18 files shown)


browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -3894,6 +3364,122 @@
...
+      "properties": {
+        "BlocklistDomainBrowsed": {
+          "type": "object",
+          "properties": {
+            "Enabled": {
+              "type": "boolean"
+            },
+            "UrlLogging": {
+              "type": "string",
+              "enum": ["full", "domain", "none"]
```
bsmth: Could you please use the form here for enums:

https://hg-edge.mozilla.org/mozilla-central/rev/07becdf2c615

i.e.,

```diff
-              "enum": ["full", "domain", "none"]
+              "oneOf": [
+                {
+                  "const": "full",
+                  "title": "Full URLs",
+                  "description": "The full URLs…"
+                },
+                {
+                  "const": "domain",
+                  "title": "Domains only",
+                  "description": "The domains…"
+                }
+                …
```

Andi is using this to drive the UI/policy editor

bsmth (approved):
One comment about enums in the schema, but otherwise looking good, thanks 👍🏻

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -7,6 +7,26 @@
       "type": "string",
       "anyOf": [{ "format": "uri" }, { "maxLength": 0 }]
     },
+    "urlLoggingLevel": {
+      "type": "string",
+      "oneOf": [
```
bsmth, line 12: Looks better, tnx

bsmth (approved):
Another +1 from me after the updates from main - thank you 👍🏻

browser/components/enterprisepolicies/tests/xpcshell/xpcshell.toml
```diff
@@ -83,6 +83,8 @@ skip-if = ["enterprise"] # Bug 2065476: skip tests for enterprise builds ; re-en
 
 ["test_security_devices.js"]
 
+["test_security_logging.js"]
```
1rneh, line 86:
```suggestion
["test_security_logging.js"]
run-if = ["enterprise"]
```

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -57,6 +57,99 @@ ChromeUtils.defineESModuleGetters(lazy, {
 export const PREF_LOGLEVEL = "browser.policies.loglevel";
 const BROWSER_DOCUMENT_URL = AppConstants.BROWSER_CHROME_URL;
 
+const URL_LOGGING_LEVELS = ["full", "domain", "none"];
```
1rneh: Could any security logging related code be extracted to its own helper module `SecurityLoggingPolicy.sys.mjs`?

toolkit/components/enterprisepolicies/PolicySchemaValidator.sys.mjs
```diff
@@ -17,7 +17,8 @@
  *     normalized value using only standard JSON Schema keywords (type, format,
```
1rneh, line 18: The `trimInvalidObjectProperties` support added here is generic and non-enterprise-specific. More policies than `SecurityLogging` could benefit from it. Could this be opened against upstream instead?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -57,6 +57,99 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
+ * the schema would have rejected, are skipped entirely: their prefs may be
+ * managed by another policy (Preferences, say), and unsetting or unlocking
+ * them here would clobber it.
+ *
+ * @param {object} param
+ *        SecurityLogging policy parameters, or null.
+ * @param {Function} callback
+ *        Called as callback(prefName, value).
+ */
+function forEachSecurityLoggingSetting(param, callback) {
```
1rneh: If I understand this correctly, `forEachSecurityLoggingSetting` re-validates things that the `PolicySchemaValidator` already does. If it gets parsed, it should be valid, no? 

The re-validation seems to be the only reason for the `values` key in `SECURITY_LOGGING_EVENTS` + all the additional test coverage:
- `test_schema_matches_implementation`
- `test_invalid_values_are_ignored`
- `test_logging_levels_are_not_interchangeable`
 
In my opinion, the re-evaluation + the related additional code. In my opinion, this is not needed or am I missing something?

browser/components/enterprisepolicies/tests/xpcshell/test_security_logging.js
```diff
@@ -0,0 +1,311 @@
...
+  checkUnsetPref("browser.download.enterprise.telemetry.urlLogging");
+  checkUnsetPref("browser.download.enterprise.telemetry.fileLogging");
+
+  Policies.SecurityLogging.onRemove(null, params);
+
+  checkUnsetPref("browser.download.enterprise.telemetry.urlLogging");
+  checkUnsetPref("browser.download.enterprise.telemetry.fileLogging");
+});
+
+add_task(async function test_update_removes_dropped_setting() {
```
1rneh: Policy updates should be moved to the live test suite and actually run it through the policy engine. Simulating the update logic in tests will break if the remove-then-reapply update logic changes (which is in my backlog see Bug-2063350).

browser/components/enterprisepolicies/tests/xpcshell/test_security_logging.js
```diff
@@ -0,0 +1,311 @@
...
+  Policies.SecurityLogging.onBeforeAddons(null, newParams);
+
+  checkLockedPref("browser.download.enterprise.telemetry.enabled", true);
+  checkUnsetPref("browser.download.enterprise.telemetry.urlLogging");
+
+  Policies.SecurityLogging.onRemove(null, newParams);
+  checkUnsetPref("browser.download.enterprise.telemetry.enabled");
+});
+
+add_task(async function test_removing_policy_unsets_what_it_applied() {
```
1rneh: Same here. This should go into the live test suite.

browser/components/enterprisepolicies/tests/xpcshell/test_security_logging.js
```diff
@@ -0,0 +1,311 @@
...
+"use strict";
+
+const { Policies, SECURITY_LOGGING_EVENTS } = ChromeUtils.importESModule(
+  "resource:///modules/policies/Policies.sys.mjs"
+);
+const { schema } = ChromeUtils.importESModule(
+  "resource:///modules/policies/schema.sys.mjs"
+);
+
+// Focused handler tests call Policies.SecurityLogging directly to isolate its
```
1rneh: All other tests set the configs though the policy engine instead of directly calling policy callbacks. I'd prefer to stick to this "convention".

toolkit/components/enterprisepolicies/tests/xpcshell/live/test_live_policy_SecurityLogging.js
```diff
@@ -0,0 +1,73 @@
+/* Any copyright is dedicated to the Public Domain.
```
1rneh, line 1: Can we test here that the preferences that the policy is controlling have listeners that will change the telemetry recording behavior? The current live tests only test the live support half-way, i.e. they would pass if nothing listened to the preference changes.

1rneh (changes requested):
Praise for having a single `SecurityLogging` policy to set instead of multiple ones. Thanks for working on this!

Do we plan to integrate `ContentAnalysisTelemetry` into `SecurityLogging` as well?

gcp:
https://searchfox.org/enterprise-main/rev/4246e65997b8878463d72352b87cc6466010aad9/toolkit/mozapps/extensions/metrics.yaml#327

This should probably have a policy controlling it as well.

gcp:
- Extracted the policy implementation into `SecurityLoggingPolicy.sys.mjs`.
- Removed the generic `trimInvalidObjectProperties` validator changes so they can be handled upstream separately.
- Removed duplicate validation and the tests that existed solely for it; the policy now relies on schema validation.
- Reworked the tests to configure values through the policy engine rather than calling policy callbacks directly.
- Moved update coverage into the live-policy suite. It now verifies that live changes affect actual download telemetry behavior, including enabled state, URL detail, file detail, and restoration after policy removal.
- Updated the WebsiteFilter test to configure telemetry through `SecurityLogging`.
- Kept `ContentAnalysisTelemetry` outside this consolidation.
- Added add-on installation control as a separate commit, tested through a real Addon Manager installation flow for enabled, disabled, and removed policy states.
- Reused the generic `EnterprisePolicyTesting.checkPolicyPref` helper and removed incidental shared-helper and metric-documentation changes.

browser/components/enterprisepolicies/tests/xpcshell/test_security_logging.js
```diff
@@ -0,0 +1,90 @@
...
+  for (const [pref, value] of Object.entries(SECURITY_LOGGING_PREFS)) {
+    checkLockedPref(pref, value);
+  }
+
+  await setupPolicyEngineWithJson("");
+
+  for (const pref of Object.keys(SECURITY_LOGGING_PREFS)) {
+    checkUnsetPref(pref);
+  }
+});
```
1rneh: Is this trying to simulate a live update? Because `setupPolicyEngineWithJson` restarts the entire engine:
https://searchfox.org/firefox-main/rev/543bce9ee58f21e87c5fba0e1865d3b3c6aa27bd/toolkit/components/enterprisepolicies/tests/EnterprisePolicyTesting.sys.mjs#66

 If you want to test anything live policy update related, please move it to the live test suite. Although the two "policy update" pref-based checks seem to be covered already by `toolkit/components/enterprisepolicies/tests/xpcshell/live/test_live_policy_SecurityLogging.js`, hence they can be removed here.

browser/components/enterprisepolicies/tests/xpcshell/test_security_logging.js
```diff
@@ -0,0 +1,90 @@
...
+      },
+    });
+
+    checkLockedPref(downloadPref, true);
+    checkLockedPref(externalPref, true);
+
+    await setupPolicyEngineWithJson("");
+
+    checkUnsetPref(downloadPref);
+    checkLockedPref(externalPref, true);
```
1rneh: Same here.

toolkit/components/printing/metrics.yaml
```diff
@@ -109,9 +109,7 @@ printing:
     extra_keys:
       source_url:
         description: >
-          The printed page source URL. Collection level configurable via
-          enterprise policy DownloadTelemetryUrlLogging: "full" (default),
-          "domain", or "none".
+          The printed page source URL.
```
1rneh, line 112: Was the _"Collection level configurable via enterprise policy SecurityLogging [...]"_ information intentionally removed from the majority of the metrics descriptions?

1rneh (approved):
r+

My only request is to place make sure that each individual security event handler listens to the policy-related preference and that we have a test in place to proof that.

toolkit/components/printing/metrics.yaml
gcp: Yes, it felt completely superfluous, repetitive and inconsistent to me.

PR #1270, opened 2026-08-12
Bug 2063400 - Console-driven relaunch deadline

gcp:
Draft for inspection. Bug 2063400 / follow-up to the `RelaunchRequired` work.

## Why

`RelaunchRequired` computes its restart schedule **once**, locally, from a local time of day, and arms a timer. Two problems:

- **It isn't console-driven.** The console supplies static scheduling parameters; the client alone decides when to restart.
- **Client-side timing is unreliable, and being console-driven is the fix.** Bug 2028364 and bug 2042370 are a long chain of "it did not restart" reports rooted in client timing: `DeferredTask` waiting on browser idle, `nsITimer` not accruing across suspend, `Date.now()` jumping. `ScheduledTask` patched the worst of it, but the design is still a fire-and-forget local timer. It also has no persistence and a permanent one-shot guard (`if (!deferredRestartTasks)`), so the schedule is lost on restart and can never be revised mid-session.

## Approach

Invert the authority. The console reports **how much time the session has left** and re-reports it on every policy poll, so the poll becomes a self-healing heartbeat: a suspend, a clock jump, or a missed timer is corrected on the next tick. An unreachable console leaves the last local deadline armed, so the feature fails closed.

A freshly launched session also gets a grace period to restart itself rather than being killed the moment it comes up past its deadline.

## Contract

A `relaunch` key riding along in the existing `GET /api/browser/policies` response:

```json
{
  "policies": { "...": "unchanged" },
  "relaunch": {
    "MinutesRemaining": 45,
    "HardMinutesRemaining": 180,
    "GracePeriodMinutes": 10
  }
}
```

| Field | Required | Default | Meaning |
| --- | --- | --- | --- |
| `MinutesRemaining` | yes | — | Soft budget left as of this response. `0` or negative means already past. |
| `HardMinutesRemaining` | no | none | Budget before the grace period stops applying. Clamped to `>= MinutesRemaining`. |
| `GracePeriodMinutes` | no | `10` | Time a freshly launched session is granted to restart itself. |

Every field is capped at 30 days (43200 minutes), the longest budget the client supports: `nsITimer` takes a 32-bit millisecond delay, so a longer wait wraps and fires early.

`restartAt = min(hardAt, max(softAt, sessionStart + grace))`.

Absence of the key cancels any pending restart. An omitted or `null` optional field takes its default; a field the console sends that fails to parse cancels the pending restart, because a restart drops the user's work and a malfunctioning console should not be able to cause that. A failed poll, or a response carrying no `policies`, delivers nothing and leaves the armed deadline standing.

Relative rather than absolute deliberately: nothing in the enterprise stack consumes a server absolute timestamp today, so this introduces no clock-skew dependency.

## Why it rides along instead of being a policy

It is delivered in the policy response but handed to the enforcer **unconditionally**, outside the policy engine:

- The engine's change detection is wrong in **both** directions for a heartbeat. An unchanged payload is deliberately *not* re-applied, so a console repeating a value produces no tick. A changed payload runs `onRemove` then re-applies synchronously, which would tear down and rebuild the warning bar on every poll.
- `about:policies` visibility and GPO/plist authorability are actively unwanted for a transient directive. Suppressing them meant inventing an `x-hidden` schema flag and patching `aboutPolicies.js`.
- What remained — validating three numeric fields, and teardown — is a few lines done directly.

`/api/browser/config` was also considered and rejected: applied once at launch, in the Felt process, mirrored through pref IPC. Wrong process, wrong cadence, more plumbing.

## UX

Reuses the existing ASRouter universal-infobar pattern: `type: "universal"`, `dismissable: false`, a "Restart now" button, two escalating phases.

| Phase | When | `priority` |
| --- | --- | --- |
| `ENTERPRISE_RELAUNCH_WARNING` | > 5 min out, shows the target time | `PRIORITY_INFO_HIGH` |
| `ENTERPRISE_RELAUNCH_IMMINENT` | <= 5 min out, counts down minutes | `PRIORITY_CRITICAL_HIGH` |

A second `ScheduledTask` fires the escalation, since polls are a minute apart but the boundary is minute-precise.

## Scope

`RelaunchRequired` and `UpdatePolicyEnforcer.sys.mjs` are **untouched** — not enabled in release, to be removed in a later patch. No deprecation warnings, no shared-code extraction.

## For review

- **A remote message can still suppress or evict the warning, and the restart happens regardless.** `InfoBar` serves one message at a time and yields the slot only to a message naming the incumbent, so `canReplace` lists the in-tree infobar messages and the bar takes the slot from them. A message from Nimbus or Remote Settings is not enumerable, and `_maybeReplaceActiveInfoBar` runs both ways — a remote message naming our id evicts our bar. Appending to `gNotificationBox` directly instead of going through ASRouter would close both gaps.
- **This touches one file outside the enterprise directories.** `<remote-text>` gains a `setVariable()` so both phases update the bar that is up, keeping focus on its button instead of rebuilding the bar. Re-showing under an id `InfoBar` still tracks would leave the new bar out of `_universalInfobars` and so unremovable, since the outgoing bar's `removed` callback fires during the append and matches on `message.id`; teardown removes ours from each window by hand for the same reason.
- **A restart while the console is unreachable is not covered end to end.** Forcing it during an outage also breaks the relaunched browser's startup policy fetch, which is fatal by design.
- The console side of this contract does not exist yet; the field names and semantics above are the proposal.

## Not restarting before session restore

A hard deadline caps the grace period, so an exhausted hard budget can come due during startup: the first poll runs inside `_initialize()` at `policies-startup`, well before `sessionstore-windows-restored`. Quitting there drops the tabs the warning promised would reopen, so the restart waits for that topic. `getStartupInfo().sessionRestored` is recorded exactly when it fires, so this also answers for a session restored before the module loaded.


toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -0,0 +1,521 @@
...
+// delay, so a wait past 2^32 - 1 ms, about 49.7 days, wraps and fires early.
+const MAX_BUDGET_MINUTES = 30 * 24 * 60;
+
+const WARNING_ID = "ENTERPRISE_RELAUNCH_WARNING";
+const IMMINENT_ID = "ENTERPRISE_RELAUNCH_IMMINENT";
+
+// InfoBar serves one message at a time, and yields the slot only to a message
+// naming the incumbent. These are the in-tree infobar messages; a message from
+// Nimbus or Remote Settings holds the slot until the next poll retries. Keep it
+// in step with the ids the in-tree message providers author.
```
gcp, line 38: I know this will get out of sync but any other solution requires more invasive surgery on an upstream piece.

browser/locales/en-US/browser/policy-messages.ftl
```diff
@@ -4,3 +4,13 @@
...
+
+# $datetime (Date) - The time the browser will be restarted at.
+enterprise-relaunch-warning-message = <strong>Your administrator requires { -brand-short-name } to restart.</strong> It will restart at { DATETIME($datetime, dateStyle: "short", timeStyle: "short") }. Tabs will reopen.
+# $minutes (Number) - How many minutes are left before the browser restarts.
+enterprise-relaunch-imminent-message =
+    { $minutes ->
+        [one] <strong>{ -brand-short-name } will restart in { $minutes } minute.</strong> Save your work now. Tabs will reopen.
+       *[other] <strong>{ -brand-short-name } will restart in { $minutes } minutes.</strong> Save your work now. Tabs will reopen.
+    }
+enterprise-relaunch-restart-now = Restart now
```
lissyx: If it's specific to enterprise it should live in `enterprise.ftl`

lissyx: Or we introduce a new ftl file but this may require coordination with l10n ?

testing/enterprise/manifest.toml
```diff
@@ -158,6 +158,11 @@ run-if = [
 
 ["test_felt_init_failures.py"]
 
+["test_felt_relaunch_deadline.py"]
```
lissyx: nit: make sure the filename helps identify the scope
```suggestion
["test_felt_browser_relaunch_deadline.py"]
```

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
lissyx: Should not be a big deal as long as:
 - we have test coverage
 - our suites are green enough we can spot regressions

browser/components/asrouter/content/components/remote-text.js
```diff
@@ -45,6 +45,21 @@
...
+     * @param {string|number} value - The value to substitute.
+     */
+    setVariable(name, value) {
+      this.setAttribute(`fluent-variable-${name}`, value);
+      if (this._content) {
+        this.render();
+        RemoteL10n.l10n.translateFragment(this._content);
+      }
+    }
+
```
lissyx, line 60: Maybe we should directly upstream this?

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -0,0 +1,521 @@
...
+ * session is overdue. A negative grace period is not a budget but a nonsense
+ * bound, and is rejected with the rest of a payload that fails to parse.
+ *
+ * @param {object} options
+ * @param {number} options.now - Date.now() at the moment the budget arrived.
+ * @param {number} options.sessionStart - Epoch ms this process started.
+ * @param {object|null} options.params - The console's `relaunch` payload.
+ * @returns {{restartAt: number}|null} null means nothing is pending.
+ */
+export function computeRestartTime({ now, sessionStart, params }) {
```
lissyx: I'd prefer `now` was optional and only overriden in tests, it looks weird to pass the value except in tests

lissyx: Similar, `sessionStart` seems to be useful really only during testing

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -0,0 +1,521 @@
...
+  );
+
+  const softAt = now + softMinutes * MS_PER_MINUTE;
+  const graceEnd = sessionStart + graceMinutes * MS_PER_MINUTE;
+
+  // The grace period floors the deadline, a hard deadline caps it.
+  return {
+    restartAt: Math.min(hardAt ?? Infinity, Math.max(softAt, graceEnd)),
+  };
+}
```
lissyx: Shouldn't this be wholy part of `RelaunchEnforcer` and handle `now` and `sessionStart` as optional arguments? When not provided they can default to `Date.now()` and `this._sessionStarted`

I do not see any good reason to keep this function out of the object

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -0,0 +1,521 @@
...
+      if (relaunch) {
+        lazy.log.error(
+          `Ignoring malformed relaunch budget: ${JSON.stringify(relaunch)}`
+        );
+      }
+      this.cancel();
+      return;
+    }
+
+    this._schedule = schedule;
```
lissyx, line 176: Might be worth logging for debugging purpose in the future

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -0,0 +1,521 @@
...
+    this._restartTask = null;
+    this._escalationTask?.disarm();
+    this._escalationTask = null;
+    this._countdownTask?.disarm();
+    this._countdownTask = null;
+  },
+
+  /**
+   * Advances the countdown from the local deadline, so the minutes the bar
+   * shows stay true while the console is unreachable.
```
lissyx: The comment is a bit confusing with the naming of the function

lissyx: It's not just advancing, it's also arming?

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -0,0 +1,521 @@
...
+        this._shownDeadlineMinute = deadlineMinute;
+        if (isImminent) {
+          this._armCountdown(minutes);
+        }
+        return;
+      }
+    }
+
+    const message = {
+      weight: 100,
```
lissyx: Why this value, how is it computed?

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -0,0 +1,521 @@
...
+      id: phase,
+      content: {
+        priority: isImminent
+          ? win.gNotificationBox.PRIORITY_CRITICAL_HIGH
+          : win.gNotificationBox.PRIORITY_INFO_HIGH,
```
lissyx, line 363:
```

const PriorityLevels = {
  PRIORITY_INFO_LOW: 1,
  PRIORITY_INFO_MEDIUM: 2,
  PRIORITY_INFO_HIGH: 3,
  // Type NEW should be used to highlight new features, and should be more
  // eye-catchy than INFO level notifications.
  PRIORITY_NEW: 4,
  PRIORITY_WARNING_LOW: 5,
  PRIORITY_WARNING_MEDIUM: 6,
  PRIORITY_WARNING_HIGH: 7,
  PRIORITY_CRITICAL_LOW: 8,
  PRIORITY_CRITICAL_MEDIUM: 9,
  PRIORITY_CRITICAL_HIGH: 10,
  PRIORITY_CRITICAL_BLOCK: 11,
};
```

I think the `PRIORITY_INFO_HIGH` may be rather `PRIORITY_CRITICAL_MEDIUM` because we really want this and maybe the `isImminent` should use `PRIORITY_CRITICAL_BLOCK` ?

toolkit/components/enterprise/tests/xpcshell/xpcshell.toml
```diff
@@ -0,0 +1,7 @@
+[DEFAULT]
+firefox-appdir = "browser"
```
lissyx, line 2: what is it used for? I find no reference to it

toolkit/components/enterprise/tests/xpcshell/test_relaunch_enforcer.js
lissyx, line 1: We need to green `xpcshell` suites to make sure this can be tracked correctly

toolkit/components/enterprise/tests/xpcshell/test_relaunch_enforcer.js
```diff
@@ -0,0 +1,186 @@
...
+      restartAt: at(30 * DAY),
+    },
+  ];
+
+  for (const { what, sessionStart, params, restartAt } of cases) {
+    const schedule = computeRestartTime({ now: NOW, sessionStart, params });
+    Assert.ok(schedule, `${what}: a schedule is produced`);
+    Assert.equal(
+      schedule.restartAt,
+      restartAt,
```
lissyx: I think this will make reading the above cases much more natural
```suggestion
      expectedRestartAt,
```

testing/enterprise/test_felt_browser_relaunch_deadline.py
```diff
@@ -0,0 +1,379 @@
...
+
+
+class BrowserRelaunchDeadline(FeltTests):
+    """The console drives the restart deadline through the policy poll.
+
+    Covers the two ends of the contract: a comfortable budget warns the user,
+    and an exhausted budget with no grace left restarts the browser.
+    """
+
+    def test_relaunch_deadline_warns_then_restarts(self):
```
lissyx: There may be bad interactions between different executions of `test_*` methods, we probably want to have each scoped to its own file

testing/enterprise/test_felt_relaunch_deadline.py
```diff
@@ -0,0 +1,379 @@
...
+            """
+            const win = Services.wm.getMostRecentBrowserWindow();
+            const bar = win.gNotificationBox.getNotificationWithValue(
+                arguments[0]
+            );
+            return {
+                sameBar: bar === win._relaunchTestBar,
+                focused: win.document.activeElement === win._relaunchTestButton,
+            };
+            """,
```
lissyx: nit: should this cleanup `win._relaunchTestButton` and `win._relaunchTestBar` ?

testing/enterprise/test_felt_relaunch_deadline.py
```diff
@@ -0,0 +1,379 @@
...
+                restartArmed: !!RelaunchEnforcer._restartTask?.isArmed,
+            };
+            """
+        )
+
+    def process_start(self):
+        self._child_driver.set_context("chrome")
+        return self._child_driver.execute_script(
+            "return Services.startup.getStartupInfo().process.getTime();"
+        )
```
lissyx: nit: we should revert `context` here back to content, for consistency

lissyx (approved):

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
gcp: More about how much state you want to supply to this function compared to how much internals it sees. I think I'll keep this.

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
gcp: I think there's enough logging in the consumers, like `_arm`.

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
gcp: https://searchfox.org/firefox-main/rev/b0bc20aac45f7cb55b18adbf4d2313457659d140/toolkit/content/widgets/notificationbox.js#154

`PRIORITY_CRITICAL_BLOCK` is not accepted there. These don't control what gets shown (hence the `REPLACEABLE_IDS` hack), just the visual styling.

So I think the current ones are fine.

toolkit/components/enterprise/tests/xpcshell/xpcshell.toml
gcp: Required for `resource:///modules/asrouter/InfoBar.sys.mjs`.

https://github.com/mozilla/enterprise-firefox/pull/1270/changes#diff-cbef6dae1f84125d94eead4c84b9ab5c1574dba6e31926597d6bb3cb866041f8R10

browser/components/asrouter/content/components/remote-text.js
```diff
@@ -45,6 +45,21 @@
...
+     *
+     * @param {string} name - The variable name, without the attribute prefix.
+     * @param {string|number} value - The value to substitute.
+     */
+    setVariable(name, value) {
+      this.setAttribute(`fluent-variable-${name}`, value);
+      if (this._content) {
+        this.render();
+        RemoteL10n.l10n.translateFragment(this._content);
+      }
```
Mossop: As far as I can tell you only need to call `this.render()` here. That function already checks whether there is content and then uses the l10n function to set the attributes which should trigger translation automatically.

Mossop (approved):

browser/components/asrouter/content/components/remote-text.js
gcp: This doesn't work.

Apparently the reason is: "_Automatic translation requires a root connected to that DOMLocalization; this standalone RemoteL10n instance has none."_

browser/components/asrouter/modules/InfoBar.sys.mjs
```diff
@@ -694,9 +695,11 @@ export const InfoBar = {
...
             );
-            const nextNotification = nextEntry?.notification;
-            InfoBar._activeInfobar = nextNotification
-              ? { message, dispatch, nextNotification }
+            // _universalInfobars holds the bar elements; the active infobar
+            // holds the notification that owns them all, so hand that over
+            // rather than an element out of the list.
+            InfoBar._activeInfobar = nextEntry
+              ? { message, dispatch, notification }
               : null;
```
Mossop: I don't think this is correct, `notification` here should come from `nextEntry.notification`.

Mossop (changes requested):

browser/components/asrouter/modules/InfoBar.sys.mjs
gcp: It's a type mixup: `nextEntry.notification` is a `<notification-message>` DOM element, whereas `_activeInfobar.notification` expects `InfoBarNotification`

gcp: Half agreed, but I don't think `nextEntry.notification` is the value either.

There are two different things called `notification` here: `_universalInfobars` entries hold `this.notification`, the DOM element returned by `appendNotification()` (`:203`, `:236`), while `_activeInfobar.notification` is the `InfoBarNotification` instance (`:660`, `:681`). Both consumers of `_activeInfobar.notification` need the instance: `:614` calls `removeUniversalInfobars()`, which is a method on the class, and `:617` calls `.notification.dismiss()`. Storing the element leaves `?.removeUniversalInfobars()` hitting the optional chain and silently no-opping, which is the stacking this commit is fixing. (The pre-patch code also named the property `nextNotification`, so `_activeInfobar.notification` was `undefined` regardless.)

That said, you are right that the line is wrong. The assignment shouldn't be there at all. The listener is registered per window and closes over its own `message`, so it can outlive the infobar it was registered for:

1. A (universal) is shown from win2, registering win2's unload listener over A.
2. B replaces A. `_activeInfobar` is now B, and B's bars are in `_universalInfobars`.
3. win2 closes. The stale listener finds a live entry (B's bar in win1) and writes A back into `_activeInfobar`.

From there a message that replaces B is refused, because the active id is A. My version writes A's instance, yours writes A's message with B's element; both clobber B. So I propose dropping the handover:

```js
win.addEventListener(
  "unload",
  () => {
    // Remove this window's stale entry
    InfoBar._universalInfobars = InfoBar._universalInfobars.filter(
      ({ box }) => box.documentGlobal !== win
    );

    // This listener is registered per window and closes over the message
    // it was shown for, so it can outlive its own infobar: a replacement
    // takes over _activeInfobar while this window stays open. Only the
    // message that still owns the active infobar may clear it.
    if (InfoBar._activeInfobar?.message !== message) {
      return;
    }

    // A universal infobar survives in the windows that are still open;
    // closing one of them leaves the notification that owns them all in
    // place. Anything else goes with its window.
    const survives =
      isUniversal &&
      InfoBar._universalInfobars.some(({ box }) => !box.documentGlobal?.closed);
    if (!survives) {
      InfoBar._activeInfobar = null;
    }
  },
  { once: true }
);
```

Closing a window never changes who owns the bars, so there is nothing to hand over: either the message still owns `_activeInfobar` and keeps it, or the last bar went away and it clears. There is precedent for the guard at `:744`, which already checks `InfoBar._activeInfobar.message !== message` before acting.

I'll add a regression test for the stale-closure case (show A from win2, replace with B from win1, close win2, assert `_activeInfobar` is still B and that a message replacing B is still accepted). The existing `universal_replaced_after_originating_window_closes` passes unchanged, since the unload no longer touches `_activeInfobar`. Lint is clean; I have not run the mochitest against this variant yet.

Also worth a follow-up regardless of which way this goes: `_universalInfobars[].notification` (element), `_activeInfobar.notification` (instance) and `InfoBarNotification.notification` (element) are three meanings for one name, which is what made this hunk ambiguous in the first place.

gcp:
Filed upstream as https://bugzilla.mozilla.org/show_bug.cgi?id=2066128

Note that if that gets fixed upstream it will likely generate merge conflicts here.

Mossop (approved):

PR #1274, opened 2026-08-13
Bug 2056600 - Add live support for policy ExtensionSettings

1rneh:
### Description 

Bugzilla: Bug-2056600

Add an `ExtensionSettings.onRemove` callback that reverts the policy's reversible state (gating data, locked prefs, and feature locks) to the no-policy baseline on a live removal or update, without undoing some one-way actions (installs, uninstalls and permission revocations).

---

### Testing 

* [X] Added tests
* [X] Manual testing performed

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -2375,6 +2375,60 @@ export var Policies = {
       // drops the blocked permission re-enables the addon automatically.
       lazy.AddonManagerPrivate.updateAddonAppDisabledStates();
     },
+    onRemove(manager, oldParam) {
+      // Revert to the no-policy baseline: clear the settings object and host
+      // guards, unlock the prefs, and release the feature locks it set.
+      // Note: It does not undo the policy's one-way actions. Uninstalled extensions are not
+      // reinstalled and permissions revoked under blocked_permissions are not
+      // re-granted.
+      manager.setExtensionSettings({});
+      try {
+        lazy.applyExtensionGuards({});
+      } catch (e) {
+        lazy.log.error(
+          `Could not clear ExtensionSettings guards: ${e.message}`
+        );
+      }
+
+      lazy.PoliciesUtils.unsetDefaultPref("extensions.getAddons.showPane");
+      lazy.PoliciesUtils.unsetDefaultPref(
+        "extensions.htmlaboutaddons.recommendations.enabled"
+      );
+      lazy.PoliciesUtils.unsetDefaultPref(
+        "extensions.webextensions.restrictedDomains"
+      );
+
+      let activePolicies = manager.getActivePolicies();
+
+      // Don't re-allow installTemporaryAddon if it's still
+      // disallowed by the InstallAddonsPermission policy.
+      if (
+        oldParam["*"]?.installation_mode == "blocked" &&
+        activePolicies?.InstallAddonsPermission?.Default !== false
+      ) {
+        manager.allowFeature("installTemporaryAddon");
+      }
+
+      // Don't re-allow uninstall-/disable-extension:<id> if it's
+      // still disallowed by the Extensions policy.
+      let lockedByExtensions = new Set(
+        activePolicies?.Extensions?.Locked ?? []
+      );
+      for (let extensionID in oldParam) {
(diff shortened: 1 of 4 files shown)


browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -2375,6 +2375,60 @@ export var Policies = {
...
+        if (extensionID == "*" || lockedByExtensions.has(extensionID)) {
+          continue;
+        }
+        let mode = oldParam[extensionID].installation_mode;
+        if (mode == "force_installed" || mode == "normal_installed") {
+          manager.allowFeature(`uninstall-extension:${extensionID}`);
+          if (mode == "force_installed") {
+            manager.allowFeature(`disable-extension:${extensionID}`);
+          }
+        }
```
lissyx, line 2427: The matching code in `onBeforeUIStartup` would perform addons installation / uninstallation depending on the `mode` case. 

The current implementation would just switch back the `uninstall-extension:X` or `disable-extension:Y` features, but not interact with installation itself.

Would it make sense that e.g. for `force_installed` and `normal_installed` extensions that `onBeforeUIStartup()` would have installed via `lazy.installAddonFromURL()` or `lazy.installAddonFromRepository(extensionID)` the `onRemove` implementation would uninstall in this case ? I think it might make sense, except if it was discussed somewhere and I missed that (but I did not found it on bugzilla).

Similar, `onBeforeUIStartup` can do some addon uninstallation, would we want to install them back ?

lissyx (approved):

browser/components/enterprisepolicies/Policies.sys.mjs
1rneh: I tried to match the non-live behavior of setting `ExtensionSettings` in `policies.json`, removing/updating it on the next restart:
- Installed add-ons are not uninstalled
- Uninstalled add-ons are not re-installed
- Revoked permissions are not re-granted

All of this would be achievable live, but it would mean carrying extra state:
- To re-install an add-on, we'd need a very accurate snapshot of the add-on storage so we bring it back in exactly the state it was in before (same version, source/URL, granted permissions, etc), persisted to disk so it survives a restart.
- To uninstall a `force_installed`/`normal_installed` add-on on removal, we'd first need to be sure the user didn't already have it installed before the policy, so we don't remove the user's own installed add-ons.

An admin who wants a force-installed add-on gone still has the means to do it — e.g. switching it to blocked. So I think handing control of the add-ons back to the user on removal is good enough here, and it matches the non-live behavior.

But I'm deferring to @bsmth, as it seems to be a product decision to me.

lissyx: If we are unsure, this can be handled as a follow up IMHO, the current basic behavior is sane and clear at least.

1rneh: I've cross-posted this to Bug-2056600 so we can continue exploring any follow-ups there.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -2375,6 +2375,60 @@ export var Policies = {
...
+      let activePolicies = manager.getActivePolicies();
+
+      // Don't re-allow installTemporaryAddon if it's still
+      // disallowed by the InstallAddonsPermission policy.
+      if (
+        oldParam["*"]?.installation_mode == "blocked" &&
+        activePolicies?.InstallAddonsPermission?.Default !== false
+      ) {
+        manager.allowFeature("installTemporaryAddon");
+      }
```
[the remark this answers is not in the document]
1rneh: Filed Bug-2063350 as this is not a specific to `ExtensionSettings`, but rather a risky behavior of the current live policy update logic.

browser/components/enterprisepolicies/Policies.sys.mjs
bsmth: Thanks both, I read the thread a few times to make sure I'm following and I agree that we shouldn't be restoring pre-policy state when the policy is gone.

>I think handing control of the add-ons back to the user on removal is good enough here

Agreed. We may eventually have feedback that this is needed/requested by users, but it's not necessary now, IMO. Thanks!

PR #949, opened 2026-05-22
Bug 2019572 - Remove firefox-enterprise.js

1rneh:
### Description 

Bugzilla: Bug-2019572

- Moves setting `browser.profiles.enabled` for enterprise builds in `firefox.js`
- Repurposes `firefox-enterprise.js` to `enterprise-defaults.js`
- Moves seeting branding related pref `extensions.activeThemeID` for enterprise builds in `firefox-branding.js`

PR #1276, opened 2026-08-13
Bug 2061810 - Disable the about:addons recommendations in enterprise …

mz-darnold:
…builds

Default extensions.getAddons.showPane and
extensions.htmlaboutaddons.recommendations.enabled to false in the enterprise branding prefs, hiding the Recommendations tab and the recommendation cards in the extension and theme list views. Neither is relevant where add-ons are managed by policy.

They are defaults rather than locked values so that the mochitest profile can restore them on the user branch, which keeps the upstream discovery pane and list recommendation tests running.

Add an enterprise-only test that clears those overrides and checks the UI: the Recommendations category is hidden, about:addons falls back to the extension list, and neither list view renders recommendation cards.

### Description 

Bugzilla: Bug-2061810

---

* Depends on:

---

### Screenshots 

---

### Testing 

* [x] Added tests

### DIFF
--- a/testing/profiles/mochitest/user.js
+++ b/testing/profiles/mochitest/user.js
@@ -44,9 +44,3 @@ user_pref("dom.push.serverURL", "wss://push.services.mozilla.com/");
 // Set a breakpad URL for mochitest runs. In enterprise builds the
 // default is an empty string, but push tests need a non-empty value.
 user_pref("breakpad.reportURL", "https://crash-stats.mozilla.org/report/index/");
-
-// Show the about:addons recommendations for mochitest runs. In enterprise
-// builds both default to false, but the discovery pane and list view
-// recommendation tests need them enabled.
-user_pref("extensions.getAddons.showPane", true);
-user_pref("extensions.htmlaboutaddons.recommendations.enabled", true);
--- a/testing/profiles/mochitest/user.js
+++ b/testing/profiles/mochitest/user.js
@@ -44,3 +44,9 @@ user_pref("dom.push.serverURL", "wss://push.services.mozilla.com/");
 // Set a breakpad URL for mochitest runs. In enterprise builds the
 // default is an empty string, but push tests need a non-empty value.
 user_pref("breakpad.reportURL", "https://crash-stats.mozilla.org/report/index/");
+
+// Show the about:addons recommendations for mochitest runs. In enterprise
+// builds both default to false, but the discovery pane and list view
+// recommendation tests need them enabled.
+user_pref("extensions.getAddons.showPane", true);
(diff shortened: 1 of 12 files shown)


mkaply (approved):

browser/branding/enterprise/pref/firefox-branding.js
```diff
@@ -48,4 +48,10 @@ pref("app.update.badgeWaitTime", 345600);
...
 
 // Default enterprise theme
-pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
\ No newline at end of file
+pref("extensions.activeThemeID", "firefox-enterprise-light@mozilla.org");
+
+// Hide the about:addons Recommendations tab (showPane) and the recommendations
+// in its extension and theme list views. AMO recommendations are not relevant
+// where add-ons are managed by policy.
+pref("extensions.getAddons.showPane", false);
```
gcp: https://github.com/mozilla/enterprise-firefox/pull/949#discussion_r3302638771

I think these shouldn't be in branding based on past discussion. @1rneh @lissyx ?

gcp (changes requested):

browser/branding/enterprise/pref/firefox-branding.js
mkaply: do we have another JS file just for Enterprise? I'm the one that suggested it because there were other non-branding prefs there

gcp: All the prefs here are branding? The update ones are all channel specific so they'd count. 

See the discussion linked above on where we landed on this, basically put them next to the original prefs with `MOZ_ENTEPRISE` guards.

1rneh: > See the discussion linked above on where we landed on this, basically put them next to the original prefs with MOZ_ENTEPRISE guards.

@gcp Thanks for catching this and pointing to our previous conversation. @lissyx and I had agreed that any enterprise-specific overrides should be set alongside the non-enterprise preference. As there is no default set for `extensions.getAddons.showPane` it should just go into firefox.js.

mz-darnold: I think this is fixed now. Thanks.

testing/profiles/mochitest/user.js
lissyx: But then we run tests on `enterprise` builds with the wrong values, I think it needs to be addressed at test harness level.

This is what we discussed in the past right @1rneh ?

1rneh: Yes, please set an cleanup the preferences in the tests that require them.

1rneh (approved):
Thank you!

gcp (approved):

PR #1286, opened 2026-08-14
Bug 2063728 - Drain launched browser's stdout and stderr

lissyx:
### Description 

Bugzilla: Bug-2063728

The Subprocess module creates a pipe for stdout, in which the launched browser will write. If not drained, it may end up full and thus blocking the main thread of the launched browser on e.g. dump() call from JS.

### DIFF
--- a/testing/enterprise/test_felt_app_pipe.py
+++ b/testing/enterprise/test_felt_app_pipe.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import platform
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_tests import FeltTests
+
+
+class FeltAppPipe(FeltTests):
+    def test_felt_app_pipe(self):
+        read_fd, write_fd = os.pipe()
+        read_bytes = self.get_pipe_capacity(read_fd)
+        write_bytes = self.get_pipe_capacity(write_fd)
+        print(f"Platform: {platform.system()}")
+        print(f"Read-end capacity:  {read_bytes}")
+        print(f"Write-end capacity: {write_bytes}")
+        os.close(read_fd)
+        os.close(write_fd)
+
+        self.run_felt_base()
+        self.connect_child_browser()
+        self.dump_kilobytes(sys.stdout.fileno(), 2 * write_bytes)
+        self.dump_kilobytes(sys.stderr.fileno(), 2 * write_bytes)
+
+    def get_pipe_capacity(self, fd):
+        """
+        Return the capacity (in bytes) of the pipe referred to by file
+        descriptor `fd`, as best as each platform allows it to be queried.
+
+        Returns None if the platform provides no way to determine it.
+        """
+        system = platform.system()
+
+        if system == "Linux":
+            return self._linux_pipe_capacity(fd)
+        elif system == "Darwin":
+            return self._macos_pipe_capacity()
+        elif system == "Windows":
+            return self._windows_pipe_capacity(fd)
+        else:
+            raise NotImplementedError(f"Unsupported platform: {system}")
+
+    def _linux_pipe_capacity(self, fd):
+        import fcntl
+
+        F_GETPIPE_SZ = 1032  # from <linux/fcntl.h>, stable since kernel 2.6.35
+        return fcntl.fcntl(fd, F_GETPIPE_SZ)
+
+    def _macos_pipe_capacity(self):
+        # No F_GETPIPE_SZ equivalent on macOS/BSD. PC_PIPE_BUF is the closest
+        # standard value the OS will report (usually 512), and it reflects the
+        # atomic-write limit, not the actual kernel buffer size (often larger,
+        # commonly 16 KiB on modern XNU, but that number isn't queryable).
+        # This 16 KiB limit gets pushed dynamically by the kernel according to
+        # various sources, usually to 64 KiB. Some mentions that "writing a lot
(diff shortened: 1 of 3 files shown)


toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -788,7 +792,6 @@ export class FeltProcessParent extends JSProcessActorParent {
     const firefoxRun = {
       command: firefoxBin,
       arguments: firefoxRunArgs,
-      stdout: "stdout",
       stderr: "stderr",
```
gcp: `stderr: pipe` or remove? (same bug)

toolkit/components/felt/content/FeltProcessParent.sys.mjs
gcp, line 801: Get this in a local that you then pass to the two callbacks below (and also store it in `this.proc` like you do now).

Else they will misattribute on a relaunch I think, when the global changes but they are still writing out buffers for the old pid.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -802,9 +805,72 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+        lineBuffer += decoder.decode(buffer, { stream: true });
+
+        // Split by lines so logProcess receives complete log messages
+        let lines = lineBuffer.split("\n");
+        // Keep the last incomplete line in the buffer
+        lineBuffer = lines.pop();
+
+        for (let line of lines) {
+          if (line.trim()) {
+            callback(line);
```
gcp: If this throws your code must survive, so just catch everything here.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -802,9 +805,72 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+          }
+        }
+      }
+
+      // Flush remaining buffer & decoder tail when stream closes
+      lineBuffer += decoder.decode();
+      if (lineBuffer.trim()) {
+        callback(lineBuffer);
```
gcp: Same.

gcp (changes requested):

toolkit/components/felt/content/FeltProcessParent.sys.mjs
lissyx: `stdout` has no purpose at all here, hence removal. `"stderr": "pipe"` would be better indeed

toolkit/components/felt/content/FeltProcessParent.sys.mjs
lissyx: we have a global `try` on top of it, so this should be already protected ?

lissyx: (it may break the loop though)

gcp: Yes, the loop has to keep going past the errors?

browser/branding/enterprise/distribution.ini
```diff
@@ -1,10 +1,8 @@
...
 [Global]
-id=enterprise-local
+id=enterprise-test
 version=1.0
-about=Mozilla Firefox Enterprise Local Build
+about=Mozilla Firefox Enterprise Test Build
 
 [Preferences]
-enterprise.console.address=https://stage.fx-enterprise.nonprod.webservices.mozgcp.net
+enterprise.console.address=http://localhost:37447
```
gcp: Nope

testing/enterprise/test_felt_app_pipe.py
```diff
@@ -0,0 +1,118 @@
...
+            raise ctypes.WinError(ctypes.get_last_error())
+
+        # For os.pipe() pipes, the read end reports in_buf_size and the write
+        # end reports out_buf_size; take whichever is nonzero/larger.
+        return max(out_buf_size.value, in_buf_size.value)
+
+    def dump_kilobytes(self, target_bytes, line_size: int = 1024) -> int:
+        num_lines = max(1, target_bytes // line_size)
+
+        self._logger.info(f"Writing {target_bytes} bytes to stdout/stderr")
```
gcp: Where is this writing to stderr?

testing/enterprise/test_felt_app_pipe.py
```diff
@@ -0,0 +1,118 @@
...
+        import fcntl
+
+        F_GETPIPE_SZ = 1032  # from <linux/fcntl.h>, stable since kernel 2.6.35
+        return fcntl.fcntl(fd, F_GETPIPE_SZ)
+
+    def _macos_pipe_capacity(self, fd):
+        # No F_GETPIPE_SZ equivalent on macOS/BSD. PC_PIPE_BUF is the closest
+        # standard value the OS will report (usually 512), and it reflects the
+        # atomic-write limit, not the actual kernel buffer size (often larger,
+        # commonly 16 KiB on modern XNU, but that number isn't queryable).
```
gcp, line 59: So this test doesn't achieve anything? Why not just return a random large number?

gcp (changes requested):

testing/enterprise/test_felt_app_pipe.py
lissyx: It's not, sorry

testing/enterprise/test_felt_app_pipe.py
lissyx: Well I could repro some freeze with this at first, but I think I made a mistake since it does not anymore. Most of the references I find on the interweb shows there's a default kernel-size limit of 16KiB that dynamically gets increased to 64KiB. There are mentions of higher limit if the application writes "a lot at once".

gcp (approved):
This can land but

- I think there's a typo in `_driver` vs `_child_driver`
- `ruff` lint is failing
- `stderr` isn't tested, I sent you some suggestions in Slack to sidestep that entirely.

PR #1287, opened 2026-08-14
Bug 2063734 - Remove app.update.auto from basic checks

lissyx:
### Description 

Bugzilla: Bug-2063734

Forcing the pref true on our builds was removed in bug 2063598 but the test was not updated.


1rneh (approved):
And that's why it's a bad idea to skip CI. My bad - sorry!  Now this needs to be uplifted to enterprise-beta and enterprise-release as well.

PR #1304, opened 2026-08-19
Bug 2064658 - Move badge and panel in toolkit/

lissyx:
### Description 

Bugzilla: Bug-2064658

Moving from browser/ to make it available to Thunderbird, which in itself requires a couple of porting changes to sync how some UI elements diverged between Firefox and Thunderbird, but the changes are small and in the end the company badge / panel works decently with this change.


toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
1rneh, line 1: I think moving the entire `EnterpriseHandler` in the first place wasn't a clean move as it's littered with many browser dependencies. 

In my opinion a better solution would be to separate browser dependent code from the one that can live well in toolkit, e.g. url validation, fetching user data and populating the panel and badge with it and give it its own module under toolkit, i.e. `EnterpriseBadge`. And `initiateShutdown` could go into `EnterpriseCommon`?

1rneh (changes requested):
Moving the assets and css files to toolkit looks fine to me.

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
lissyx: It's a bit more work but sounds reasonnable

lissyx: Less painful that I was worried about, tests looks good locally, let's hope CI agrees

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,324 @@
...
+    try {
+      const { name, email, picture } =
+        await lazy.ConsoleClient.getLoggedInUserInfo();
+
+      let validPicture = picture;
+      // Thunderbird's CSP is stricter, make use of a data URL
+      // img-src chrome: data: moz-icon: https://addons.thunderbird.net;
+      if (AppConstants.MOZ_APP_NAME == "thunderbird") {
+        validPicture = await this.urlToDataUrl(picture);
+      }
```
lissyx: @1rneh this is what I was mentionning on Matrix. We could move also Firefox to use it, but if we are worried, I think it's not too complicated to handle both like that.

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,324 @@
...
+   */
+  _isInitialized: false,
+
+  /**
+   * Handles the enterprise state for each new browser window.
+   * On first call:
+   *    - Make a request to the console to retrieve the user information of the signed in user.
+   *
+   * @param {Window} window chrome window
+   */
```
1rneh, line 130: Let's update the docs.
```suggestion
  /**
   * Initializes the enterprise badge: Fetches the information of the signed-in 
   * user and populates the badge with the user information.
   *
   * @param {Window} window chrome window
   */
```

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
1rneh: I'd prefer for Firefox to use it as so that the shared toolkit code stays clear of any consumer specific branching.

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,324 @@
...
+  async urlToDataUrl(url) {
+    const response = await fetch(url);
+    if (!response.ok) {
+      throw new Error(`Failed to fetch ${url}: ${response.status}`);
+    }
+    const blob = await response.blob();
+
+    return new Promise((resolve, reject) => {
+      const reader = new FileReader();
+      reader.onloadend = () => resolve(reader.result); // "data:image/png;base64,...."
```
1rneh: `loadent` is fired on read completion, [no matter if successful or not](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/loadend_event).
```suggestion
      reader.onload = () => resolve(reader.result); // "data:image/png;base64,...."
```

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,324 @@
...
+      reader.onerror = reject;
```
1rneh: Let's pass the actual exception.
```suggestion
      reader.onerror = () => reject(reader.error);
```

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,324 @@
...
+  async initUser() {
+    try {
+      const { name, email, picture } =
+        await lazy.ConsoleClient.getLoggedInUserInfo();
+
+      let validPicture = picture;
+      // Thunderbird's CSP is stricter, make use of a data URL
+      // img-src chrome: data: moz-icon: https://addons.thunderbird.net;
+      if (AppConstants.MOZ_APP_NAME == "thunderbird") {
+        validPicture = await this.urlToDataUrl(picture);
```
1rneh: Multiple thoughts here:
- The IdP might not send a `picture` which would throw in `urlToDataUrl`. Missing `picture` fallbacks are handled in `_updateUserIcon`, so let's move calling `_updateUserIcon` there.
- The company logo we additionally validate (`validateDataUrl` is called in `_updateLogo`), we should likely do the same for the user picture data url.
- The method `urlToDataUrl` doesn't depend on `this`, so it can move out of the `EnterpriseBadge` const.

1rneh (changes requested):
Thanks! That's already a much better separation!

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
lissyx: Ok, cool that's also what I prefer, but I was unsure about changing it on Firefox that late in the cycle.

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
1rneh: * Missing picture fallbacks are handled in `_updateUserIcon`, so let's move calling `urlToDataUrl` there.

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,330 @@
...
+      lazy.log.warn(
+        "Unable to update email in enterprise panel without user information"
+      );
+      return;
+    }
+
+    if (!email.textContent) {
+      email.textContent = this._signedInUser.email;
+    }
+  },
```
1rneh, line 297: Right now, `EnterpriseBadge` still depends on browser specific machinery: `_setupLearnMoreLink` and `openPanel` utilises `PanelUI`, `PanelMultiView`, `lazy.BrowserUtils.whereToOpenLink`, etc. Hence the controller to open the panel doesn't belong into the shared toolkit component.

1rneh (changes requested):
Sorry, my most important comment remark from my last review has gone lost...

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,330 @@
...
+/**
+ * Produce a data URL from an non data URL to hold the user picture
+ *
+ * @param {string} url - The URL to fetch the picture from
+ * @returns {Promise} Promise that performs the data URL conversion
+ */
+async function urlToDataUrl(url) {
+  const response = await fetch(url);
+  if (!response.ok) {
+    throw new Error(`Failed to fetch ${url}: ${response.status}`);
```
1rneh: If we encounter a failure, we should rely on our icon fallback instead of throwing and breaking any following badge init flow.

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,330 @@
...
+  async initUser() {
+    try {
+      const { name, email, picture } =
+        await lazy.ConsoleClient.getLoggedInUserInfo();
+      this._signedInUser = { name, email, pictureUrl: picture };
+    } catch (e) {
+      // TODO: Bug 2000864 - Handle unsuccessful GET /WHOAMI
+      lazy.log.warn("Unable to initialize enterprise user: ", e);
+    }
+  },
```
1rneh, line 162:
```suggestion
  async initUser() {
    try {
      const { name, email, picture } =
        await lazy.ConsoleClient.getLoggedInUserInfo();
      let pictureUrl = null;
      if (picture) {
        try {
          pictureUrl = validateDataUrl(await urlToDataUrl(picture))?.href ?? null;
        } catch (e) {
          lazy.log.warn("Unable to fetch user picture: ", e);
        }
      }
      this._signedInUser = { name, email, pictureUrl };
    } catch (e) {
      // TODO: Bug 2000864 - Handle unsuccessful GET /WHOAMI
      lazy.log.warn("Unable to initialize enterprise user: ", e);
    }
  }
```

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,330 @@
...
+    const wrapper = window.document.getElementById(
+      "enterprise-user-icon__wrapper"
+    );
+    const { name, pictureUrl } = this._signedInUser;
+
+    let validPictureUrl = null;
+    if (pictureUrl) {
+      urlToDataUrl(pictureUrl).then(pictureDataUrl => {
+        validPictureUrl = validateDataUrl(pictureDataUrl);
+      });
```
1rneh: This is broken. `urlToDataUrl(pictureUrl)` is asynchronous and not awaited so the `if (validPictureUrl)` check executes synchronously before the `.then` callback runs. `validPictureUrl` is always null at that point.

On a second thought, maybe we should keep `_updateUserIcon` synchronously and put the url handling back into `initUser`. This way, if we ever reuse the user icon anywhere we don't need to recompute the data url (see my code suggestion above).

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,330 @@
...
+      return;
+    }
+
+    if (!email.textContent) {
+      email.textContent = this._signedInUser.email;
+    }
+  },
+
+  uninit() {
+    this._signedInUser = {};
```
1rneh: Because `if ({})` is truthy.
```suggestion
    this._signedInUser = null;
```

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,330 @@
...
+   */
+  async init(window) {
+    if (Services.felt.isFeltUI()) {
+      // Nothing to setup for the felt window
+      return;
+    }
+    if (!this._isInitialized) {
+      lazy.log.debug("Initializing...");
+      await this.initUser();
+      this._isInitialized = true;
```
1rneh: `nit` To avoid a duplicated fetch in case there is two browser windows opening.
```suggestion
      this._isInitialized = true;
      await this.initUser();
```

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -137,29 +59,12 @@ export const EnterpriseHandler = {
       return;
     }
     if (!this._isInitialized) {
-      lazy.log.debug("Initializing...");
-      await this.initUser();
       this._isInitialized = true;
     }
```
1rneh: Now `this._isInitialized` doesn't guard anything anymore, so this can be removed.

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: Can you please update the docs? `EnterpriseHandler` doesn't take care of the user info retrieval anymore. Therefore the `async` can be removed as well.

1rneh (changes requested):

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
lissyx: `PanelUI` and `PanelMultiView` are available in Thunderbird, I'll check the `BrowserUtils.whereToOpenLink`

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
lissyx: (re-use was my concern for doing it like that at first)

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
lissyx: `BrowserUtils` is part of `toolkit/` (`gre/modules`) and is also already used by Thunderbird: https://searchfox.org/comm-central/rev/5631a57ef8994fce81d10552da2e640f1de90020/mail/actors/LinkClickHandlerChild.sys.mjs#8-10

lissyx: https://searchfox.org/comm-central/rev/5631a57ef8994fce81d10552da2e640f1de90020/mail/components/customizableui/content/panelUI.js

lissyx: https://searchfox.org/comm-central/rev/5631a57ef8994fce81d10552da2e640f1de90020/mail/components/customizableui/PanelMultiView.sys.mjs

1rneh: Oh, my bad! I checked where the panel related file live on `firefox/main` and they were under `browser/*`. Thanks for double-checking!

lissyx: They are, indeed Thunderbird has its own implementation :)

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: Can you add that it initializes the url button on every call. 

> Therefore the async can be removed as well.

This is still unaddressed.

toolkit/components/enterprise/modules/EnterpriseBadge.sys.mjs
```diff
@@ -0,0 +1,329 @@
...
+ *
+ * @param {string} url - The URL to fetch the picture from
+ * @returns {Promise} Promise that performs the data URL conversion
+ */
+async function urlToDataUrl(url) {
+  const response = await fetch(url);
+  if (!response.ok) {
+    lazy.log.error(`Failed to fetch ${url}: ${response.status}`);
+    return null;
+  }
```
1rneh, line 102: Now that the `urlToDataUrl` caller is inside a try-catch we can throw again `if (!response.ok)`, which is more consistent with how we reject here: `reader.onerror = () => reject(reader.error);`.

1rneh (approved):
r+ with my two comments addressed.

browser/components/enterprise/EnterpriseHandler.sys.mjs
lissyx: Ah I missed the mention of the `async`

PR #1308, opened 2026-08-19
Bug 2064395 - Lock the FELT session on browser close

jporter-dev:
### Description 

Bugzilla: Bug-2064395

Adds session locking for FELT SSO sessions to require OS auth to resume. Instead of always signing out when the browser closes, FELT can now *lock* the session, controlled by policy: the refresh token is encrypted via OSKeyStore and persisted, then resumed on next launch after the user passes an OS auth prompt. If locking is disabled or unavailable, we fall back to the existing sign-out-on-close behavior.

When a session is "locked" the user will be prompted with the OS-level authentication prompt (Windows Hello, TouchID, password prompt) instead of SSO when proceeding to sign in from FELT with their email address.

- Adds `enterprise.locking.on_close` pref, plus a `SignOut` enterprise policy to control it.
- `FeltLocking` module handles encrypt/store, update, and unlock. The token is only ever stored encrypted in the keystore, and is dropped whenever locking is disabled or the user signs out.

---

### Screenshots 

| | |
|---|---|
| felt-locking-unlock-button<br>_FELT login with unlock button_||
| felt-locking-macos-touchid-2<br>_Unlock prompt on MacOS with TouchID_ | felt-locking-macos-password-2<br>_Unlock prompt on MacOS without TouchID_ |
| felt-locking-macos-keychain-permission-prompt<br>_Keychain Permission Request on MacOS_ | felt-locking-ubuntu-2<br>_Unlock prompt on Ubuntu_ |
| felt-locking-windows-pin<br>_Unlock prompt on Windows with pin_ | felt-locking-windows-password<br>_Unlock prompt on Windows with password_ |
| felt-locking-signout-prompt-without-tabs<br>_Signout prompt with locking_ | felt-locking-signout-prompt-with-tabs<br>_Signout prompt with locking and tabs_ |

---

### Testing 

* [x] Added tests

**Steps to verify changes:**

1. Set `enterprise.locking.browser_close` to true or configure a `SignOut` policy with `{"BrowserClose": {"Action": "lock"}}`
    - `enterprise.locking.browser_close` is locked by default to prevent a user from toggling this. Unlock it in `firefox.js` before building if you want to test using that path.
3. Sign in then trigger a browser exit, confirming the "Lock" dialog (normally the Signout dialog).
4. Relaunch, felt should now show "Unlock" instead of "Continue" when your email is entered.
5. Click "Unlock"
6. Authenticate with the OS auth prompt and resume the session without redoing SSO.

**Expected result:**

Session resumes after OS auth. With locking off, the browser signs out on close as before.


lissyx:
@jporter-dev When does the prompt you shared in screenshots gets displayed? What is the user input flow to it?

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Encrypts the refresh token and persists it for the given user via
+ * FeltStorage, so the value never lands in a plaintext pref / about:config.
+ *
+ * @param {string} email
+ * @param {string} token The plaintext refresh token.
+ * @returns {Promise<void>}
+ */
+async function storeToken(email, token) {
+  const encryptedRefreshToken = await lazy.OSKeyStore.encrypt(token);
```
lissyx: I assume `OSKeyStore` are the calls triggering the ask to unlock key store that are presented in screenshot. Can we make sure there's an explicit message shown in felt when we know we will ask for the key store credentials ? I believe in term of UX security it's not a good practice that we accustom people to accept "random" request to unlock system key store on our behalf, without a good explanation before.

jporter-dev:
> @jporter-dev When does the prompt you shared in screenshots gets displayed? What is the user input flow to it?

When a user attempts to sign-in using an email that has a stored token. I'll add more detail and a recording of the flow to the PR.

toolkit/components/felt/content/FeltLocking.sys.mjs
jporter-dev: I think many applications show a "Unlock with TouchID" button or indicator. May be worth exploring the UX on this a bit.

browser/app/profile/firefox.js
```diff
@@ -32,6 +32,8 @@ pref("enterprise.prompt_on_signout", true);
 //              or is it fine to apply it to any enterprise build?
 pref("app.update.checkOnlyInstance.enabled", false);
 pref("app.update.background.enabled", true);
+// Allow locking the session (persist behind OS auth) instead of signing out.
+pref("enterprise.locking.browser_close", false);
```
lissyx: Should this pref be locked?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -4063,6 +4063,25 @@ export var Policies = {
...
+      for (const [key, pref] of configuredTriggers) {
+        const { Action, Locked } = param[key];
+        lazy.PoliciesUtils.setDefaultPref(
+          pref,
+          Action === "lock",
+          Locked ?? false
+        );
+      }
+    },
+  },
```
lissyx, line 3870: if it's not live should it use the new `x-restart-required` from #1331  ?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -4063,6 +4063,25 @@ export var Policies = {
...
   },
 
+  SignOut: {
+    onBeforeAddons(manager, param) {
+      const TRIGGERS = {
+        BrowserClose: "enterprise.locking.browser_close",
+      };
+      const configuredTriggers = Object.entries(TRIGGERS).filter(
+        ([key]) => param[key]
+      );
```
lissyx: This looks way too complex for nothing right now

testing/enterprise/test_felt_browser_lock_on_close.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+        self._accept_close_dialog()
+        self._settle_after_close(browser_pid)
+
+        assert self.signout_count.value == 1, (
+            f"Expected exactly 1 signout request, got {self.signout_count.value}"
+        )
+        assert not self._felt_has_locking_token(), (
+            "Signing out must not leave a resume token behind"
+        )
```
lissyx: This looks to be mostly the same code, with just a few variations on some values and expectations, I think we can factorize much more

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -371,13 +373,21 @@ export const EnterpriseHandler = {
         id: `enterprise-close-prompt-title-with-tabcount${warnSuffix}`,
         args: { tabCount },
       };
+      // Titles are action-neutral; only the message reflects lock vs sign-out.
+      const lockSuffix = warnOnSignout ? "-and-lock-warning" : "-lock";
       messageId = {
-        id: `enterprise-close-prompt-message-with-tabcount${warnSuffix}`,
+        id: `enterprise-close-prompt-message-with-tabcount${
+          lockOnClose ? lockSuffix : warnSuffix
+        }`,
```
lissyx: let's make sure l10n checks those so we do not land incorrect things :)

toolkit/components/enterprise/modules/EnterpriseHandler.sys.mjs
```diff
@@ -558,6 +572,9 @@ export const EnterpriseHandler = {
   },
 
   initiateShutdown() {
+    // Suppress the quit prompt the felt-driven shutdown would re-trigger, for
+    // callers that reach here without going through showSignoutPrompt.
+    this._skipSignoutPrompt = true;
```
lissyx: That's far from clear, make sure the comment is relevant here

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -568,6 +585,36 @@ export const EnterpriseHandler = {
...
+  lockOrSignOut() {
+    this._skipSignoutPrompt = true;
+    if (this.willLockOnClose) {
+      try {
+        Services.felt.performLock();
+      } catch (e) {
+        // performLock could not be dispatched, fall back to the normal sign-out-on-close quit.
+        lazy.log.error(`Unable to lock, falling back to signout: ${e}`);
+        Services.startup.quit(Ci.nsIAppStartup.eAttemptQuit);
+      }
```
lissyx: Looking at the rust code, the only error is `NS_ERROR_FAILURE` and it's returned when we have no client, I'm not sure it's a situation we should handle so "gracefully"

Are we here closing Felt ? Or the browser ?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -568,6 +585,36 @@ export const EnterpriseHandler = {
...
+   */
+  get willLockOnClose() {
+    return Services.prefs.getBoolPref(LOCK_ON_CLOSE_PREF, false);
+  },
+
+  /**
+   * Ends the FELT session on browser close by either locking it (persisting it
+   * behind OS auth to resume later) or signing out.
+   */
+  lockOrSignOut() {
```
lissyx, line 394: We should make sure this is called from the right context, i.e. verify for felt or browser.

toolkit/components/felt/content/FeltLocking.sys.mjs
lissyx: From only looking at the code, it's not clear what precisely is the flow here. Will I input my email, hiyt "login" and be presented with the OS-level prompt asking me for my password?

For an uneducated folk, how can this be distinguished from trying to steal their information ?

toolkit/components/felt/FeltStorage.sys.mjs
```diff
@@ -68,6 +68,42 @@ export const FeltStorage = {
     this._feltStorage.saveSoon();
   },
 
+  /**
+   * Gets the stored (encrypted) refresh token for a locked session, if any.
```
lissyx:
```suggestion
   * Gets the encrypted stored refresh token for a locked session, if any.
```

toolkit/components/felt/FeltStorage.sys.mjs
```diff
@@ -68,6 +68,42 @@ export const FeltStorage = {
...
+   *
+   * @param {string} email
+   * @returns {string | undefined} The encrypted refresh token.
+   */
+  getLockingToken(email) {
+    return this._feltStorage.data?.lockingTokens?.[email];
+  },
+
+  /**
+   * Stores the (already encrypted) refresh token for a locked session.
```
lissyx:
```suggestion
   * Stores the refresh token for a locked session. Encryption needs to be performed before calling this
```

toolkit/components/felt/FeltStorage.sys.mjs
```diff
@@ -68,6 +68,42 @@ export const FeltStorage = {
...
+   *
+   * @param {string} email
+   * @param {string} token The encrypted refresh token.
```
lissyx: How can we assert that the token **is** encrypted? I feel unsafe relying only on the caller to encrypt it

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,166 @@
...
+   * @returns {Promise<boolean>} Whether the session was successfully unlocked.
+   */
+  tryUnlock: async (email, browser) => {
+    // A stored token exists only because a browser-authorized lock created it,
+    // so its presence is the authorization to resume: the locking pref lives in
+    // the browser process, which the Felt UI process cannot read.
+    const token = lazy.FeltStorage.getLockingToken(email);
+    if (token) {
+      const { authenticated } = await lazy.OSKeyStore.ensureLoggedIn(
+        "Trying to unlock existing session",
```
lissyx: The wording is weird on macOS where it already says "Firefox is trying to ...", this should be per-platform

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,166 @@
...
+   * @param {Element} browser
+   * @returns {Promise<boolean>} Whether the session was successfully unlocked.
+   */
+  tryUnlock: async (email, browser) => {
+    // A stored token exists only because a browser-authorized lock created it,
+    // so its presence is the authorization to resume: the locking pref lives in
+    // the browser process, which the Felt UI process cannot read.
+    const token = lazy.FeltStorage.getLockingToken(email);
+    if (token) {
+      const { authenticated } = await lazy.OSKeyStore.ensureLoggedIn(
```
lissyx, line 51: Cannot `OSKeyStore` expose this string when unlocking keyring on linux? This should be filed as a follow up against OSKeyStore, and if it's supposed to work we need to find why it's not showing up

toolkit/components/felt/content/FeltLocking.sys.mjs
lissyx: I suspect this should be l10n string as well

toolkit/components/felt/content/window.js
```diff
@@ -112,6 +113,20 @@ async function connectToConsole(email) {
     return;
   }
   signInInFlight = true;
+
+  const browser = document.getElementById("browser");
+  try {
+    if (await lazy.FeltLocking.tryUnlock(email, browser)) {
```
lissyx: It should be stated here that upon success, the whole flow is being bypassed and the actor is directly called.

But this design is making me a bit worried about the extra complexity handled on `FeltLocking` side

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,166 @@
...
+
+        // Tokens are committed; from here a failure is a launch failure, not
+        // a reason to fall back to SSO, so let it propagate to the caller.
+        const parentActor =
+          browser.browsingContext.currentWindowGlobal.domProcess.getActor(
+            "FeltProcess"
+          );
+        await parentActor.receiveMessage({
+          name: "FeltChild:StartFirefox",
+          data: {},
```
lissyx: Why do not we pass the token information here like in https://searchfox.org/enterprise-main/rev/c20638f063b74a5e3a22c90a6285a07cc143e1c6/toolkit/components/felt/content/FeltWindowChild.sys.mjs#59-62 ?

lissyx: We could keep the logic above that makes sure the tokens are uptodate but avoid the `setTokens` calls and rely on the existing code path: https://searchfox.org/enterprise-main/rev/c20638f063b74a5e3a22c90a6285a07cc143e1c6/toolkit/components/felt/content/FeltProcessParent.sys.mjs#982-1010

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * The email of the currently signed-in user, used as the key under which a
+ * locked session's refresh token is stored. Read from the cached value rather
+ * than the network so locking cannot hang or fail at shutdown.
+ *
+ * @returns {string | undefined} email
+ */
+function currentEmail() {
+  return lazy.FeltStorage.getLastSignedInUser();
+}
```
lissyx: nit: not convinced of the added value of the extra function call `currentEmail()` here

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -322,6 +328,14 @@ export class FeltProcessParent extends JSProcessActorParent {
...
                 );
                 Services.felt.sendAccessToken();
+                // Keep the persisted token in sync with the rotated refresh
+                // token; a keystore failure here must not tear down a healthy
+                // session.
+                lazy.FeltLocking.updateStoredToken(refresh_token).catch(err => {
+                  lazy.log.warn(
+                    `Failed to update the stored locked-session token on refresh: ${err}`
+                  );
+                });
```
lissyx: So any `setTokens` needs to be tied to an `updateStoredToken` ? (except when dealing with the `OSKeyStore` itself?)

I'm wondering if we should not have a callback defined somewhere so that `setTokens` would call it anytime a token is set to handle the update of the key store ?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -333,6 +347,7 @@ export class FeltProcessParent extends JSProcessActorParent {
                   `token refresh failed (${error.name}), shutting down Firefox`,
                   error
                 );
+                lazy.FeltLocking.clear();
                 Services.felt.clearTokens();
```
lissyx: same here, looks like we have two things that needs to happen in parallel

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -973,20 +988,79 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+      await lazy.FeltLocking.store(Services.felt.getRefreshToken());
+    };
+
+    persistLock()
+      .catch(err => {
+        // If we cannot persist the session there is nothing to unlock later,
+        // so fall back to a server signout rather than leaving a dangling
+        // session behind, and drop any stale stored token.
+        lazy.log.error(`Locking failed, falling back to signout: ${err}`);
+        lazy.FeltLocking.clear();
```
lissyx: Don't we need a `Services.felt.clearTokens()` out of safety?

lissyx: OK it's below

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -973,20 +988,79 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+        return lazy.ConsoleClient.performServerSignout().catch(e => {
+          lazy.log.error(`Server signout failed: ${e}`);
+        });
+      })
+      .finally(() => {
+        Services.felt.clearTokens();
+        Services.felt.shutdownFirefox();
+        gFeltProcessParentInstance.proc.exitPromise.then(_ => {
+          Services.cpmm.sendAsyncMessage("FeltParent:FirefoxLogoutExit", {
+            reason: "lock",
```
lissyx: Should we really send "lock" when the `persistLock()` failed above?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -973,20 +988,79 @@ export class FeltProcessParent extends JSProcessActorParent {
...
-          const expires_at = Math.floor(Date.now() / 1000) + Number(expires_in);
-          Services.felt.setTokens(access_token, refresh_token, expires_at);
+          // An unlock resumes with the refresh token already set on the FELT
+          // side (see FeltLocking.tryUnlock), so message.data is empty; a fresh
+          // SSO login carries the tokens here.
+          const isUnlock = !message.data?.refresh_token;
+          if (isUnlock) {
+            if (!Services.felt.getRefreshToken()) {
+              throw new Error("No token!");
+            }
```
lissyx: With my above suggestion we could avoid this

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -999,7 +1073,7 @@ export class FeltProcessParent extends JSProcessActorParent {
           lazy.log.debug(`Collected cookies: ${ssoCollectedCookies.length}`);
           // When a restart was reported we assume cookies were stored properly on the
           // browser side?
-          if (!ssoCollectedCookies.length) {
+          if (!isUnlock && !ssoCollectedCookies.length) {
```
lissyx: If you implement my suggestion we can add a `isUnlock` explicit parameter that we pass from `FeltLocking.tryUnlock`

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,166 @@
...
+  tryUnlock: async (email, browser) => {
+    // A stored token exists only because a browser-authorized lock created it,
+    // so its presence is the authorization to resume: the locking pref lives in
+    // the browser process, which the Felt UI process cannot read.
+    const token = lazy.FeltStorage.getLockingToken(email);
+    if (token) {
+      const { authenticated } = await lazy.OSKeyStore.ensureLoggedIn(
+        "Trying to unlock existing session",
+        "Firefox Enterprise"
+      );
```
lissyx, line 54: I'm not sure this is working as we would like?

https://searchfox.org/enterprise-main/rev/c20638f063b74a5e3a22c90a6285a07cc143e1c6/toolkit/modules/OSKeyStore.sys.mjs#186-233

Specifically there's an explicit check if `reauth` is a string that this only works on local builds.

So I'm unsure what is the exact behavior of your changes on a non local build ?

lissyx: never mind the indentation tricked me (I wish searchfox would highlight braces)

toolkit/components/felt/rust/nsIFelt.idl
```diff
@@ -102,6 +102,9 @@ interface nsIFelt : nsISupports {
   [must_use]
   void performSignout();
 
+  [must_use]
+  void performLock();
```
lissyx: nit: we have been bad at documentation, can you include proper doc here and file a follow up blocking `firefox-client-mvp` and assign it to me to document more?

browser/components/enterprisepolicies/Policies.sys.mjs
jporter-dev: I had originally planned for `browser_crash`, `browser_restart`, and `network_loss` triggers but decided to keep this PR focused on  `browser_close`. If preferred I can simplify now and expand it in the next PR.

toolkit/components/felt/rust/nsIFelt.idl
jporter-dev: Sure I can do that

toolkit/components/felt/tests/xpcshell/head.js
```diff
@@ -0,0 +1,21 @@
...
+  const appData = profileDir.clone();
+  appData.append("felt-app-data");
+  if (!appData.exists()) {
+    appData.create(Ci.nsIFile.DIRECTORY_TYPE, 0o755);
+  }
+  const props = Services.dirsvc.QueryInterface(Ci.nsIProperties);
+  if (!props.has("UAppData")) {
+    props.set("UAppData", appData);
+  }
+})();
```
lissyx: Shouldn't his be in a `add_setup(...)` rather than this call ?

lissyx: There's already `makeFakeAppDir()`: https://searchfox.org/enterprise-main/rev/c20638f063b74a5e3a22c90a6285a07cc143e1c6/testing/modules/AppData.sys.mjs#8-94

toolkit/components/felt/tests/xpcshell/xpcshell.toml
```diff
@@ -0,0 +1,7 @@
+[DEFAULT]
+head = "head.js"
+prefs = ["enterprise.log_level=Debug"]
+
```
lissyx, line 3:
```suggestion
[DEFAULT]
head = "head.js"
prefs = ["enterprise.log_level=Debug"]

run-if = ["enterprise"]
```

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -316,8 +310,13 @@ export class Felt {
       }
 
       case "FeltParent:FirefoxLogoutExit": {
-        Services.felt.makeBackgroundProcess(false);
-        this.showWindow();
+        // Lock quits to resume next launch, sign-out returns to sign-in.
```
lissyx: Please reformulate

lissyx (changes requested):

toolkit/components/felt/rust/nsIFelt.idl
jporter-dev: https://bugzilla.mozilla.org/show_bug.cgi?id=2066208

toolkit/components/felt/FeltStorage.sys.mjs
jporter-dev: Since this should always be encrypted, I went ahead and moved the encryption responsibility here instead of relying on the caller

toolkit/components/felt/content/FeltLocking.sys.mjs
jporter-dev: Idk how I didn't notice this 🤦‍♂️ Updated this to use the `PLATFORM()` variant, same way this os auth prompt handles it: https://searchfox.org/enterprise-main/source/browser/locales/en-US/browser/aboutLogins.ftl#146-150

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: This shouldn't really happen with the addition of the `isFeltBrowser()` check that was added, but the fallback is to close the browser

browser/components/enterprisepolicies/Policies.sys.mjs
jporter-dev: It's meant to be live, updated the policy with that flag as well

toolkit/components/felt/content/FeltLocking.sys.mjs
jporter-dev: I've added an "Unlock" button to FELT when it's meant to unlock instead of go to SSO signin. Pressing "Unlock" will prompt with the OS auth.

felt-locking-unlock-button

toolkit/components/felt/content/FeltLocking.sys.mjs
jporter-dev: I think these comments/bugs address this gap: https://searchfox.org/firefox-main/source/toolkit/modules/OSKeyStore.sys.mjs#74 and https://searchfox.org/firefox-main/source/security/manager/ssl/OSReauthenticator.cpp#399-417.

We can file an additional follow-up or poke bug-1527745 in case there's been any progress/changes on this on the Linux front based on the possible paths they considered here: https://bugzilla.mozilla.org/show_bug.cgi?id=1494478#c8

jporter-dev:
@lissyx thanks for all of the feedback! I think I addressed everything, and made some other changes prompted by the posture rework and the signout on close handling.

I ended up reworking this quite a bit, I think it's a bit simpler now - instead of a custom "lockAndClose" event, the lock intent is set based on the pref and is passed with the exit. This allows the browser to go through the regular shutdown flow (same way we do for "signoutAndClose"), and keeps it compatible with our existing prompt and close logic and makes the "lockOnRestart" and "lockOnCrash" paths easier to add.

I also updated the FELT UX when unlocking to make it more explicit that it'll be unlocking, updated the OS auth messaging to be OS-dependent, and updated unlocking to support posture collection.

browser/components/enterprise/EnterpriseHandler.sys.mjs
lissyx: @jporter-dev is it OK ?

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -150,9 +174,9 @@ export const EnterpriseHandler = {
     ] = await lazy.localization.formatValues([
       titleId,
       messageId,
-      { id: "enterprise-close-prompt-primary-btn-label" },
-      { id: "enterprise-close-prompt-message-reauth" },
-      { id: "enterprise-close-prompt-checkbox-label" },
+      { id: `enterprise-close-prompt-primary-btn-label${lockSuffix}` },
+      { id: `enterprise-close-prompt-message${lockSuffix}-reauth` },
+      { id: `enterprise-close-prompt-checkbox-label${lockSuffix}` },
```
lissyx: I remember doing something similar and flod reporting it was not a good idea, this should get vetted by l10n

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -4176,6 +4164,43 @@
...
+          "properties": {
+            "Action": {
+              "type": "string",
+              "enum": ["signout", "lock"]
+            }
+          },
+          "required": ["Action"]
+        }
+      }
+    },
```
lissyx, line 4285: Please make either a port of this for Thunderbird if it is a policy that should be handled there, or update allowed differences of https://searchfox.org/comm-central/source/mail/components/enterprisepolicies/tests/browser/browser_policies_differences.js

Or at least ping me when you land so I can adapt it

Also update/adapt https://docs.google.com/spreadsheets/d/1h-UmrLEBVmLVEl9eIVQlNkrP-NkgGXfxpzN5tJWCWv4/edit?pli=1&gid=1904079787#gid=1904079787 ?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -3713,6 +3712,27 @@ export var Policies = {
...
+      if (oldParams.BrowserClose) {
+        lazy.PoliciesUtils.unsetAndUnlockPref(
+          "enterprise.locking.browser_close"
+        );
+        // unsetAndUnlockPref restores the build default but never re-locks;
+        // re-lock to match the locked default the enterprise build ships.
+        Services.prefs.lockPref("enterprise.locking.browser_close");
+      }
+    },
+  },
```
lissyx, line 3870: Same comment regarding Thunderbird

browser/components/enterprisepolicies/schemas/policies-schema.json
lissyx: Or at least file a bug blocking `thunderbird-client-poc`

browser/components/BrowserComponents.manifest
```diff
@@ -239,6 +233,7 @@ category browser-quit-application-granted moz-src:///browser/components/ipprotec
 
 #ifdef MOZ_ENTERPRISE
 category browser-quit-application-granted resource://gre/modules/enterprise/EnterpriseBadge.sys.mjs EnterpriseBadge.uninit
+category browser-quit-application-granted resource:///modules/enterprise/EnterpriseHandler.sys.mjs EnterpriseHandler.uninit
```
lissyx, line 262: Please make sure to file a porting bug blocking `thunderbird-client-poc` or better provide a patch against `comm-central`

lissyx: Nevermind, this component is browser specific

toolkit/components/felt/content/FeltLocking.sys.mjs
lissyx: If it's really the same upstream issue, we should maybe report it should need investigation further, and/or think about how it can impact our product and maybe make it blocking v1 or something ?

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,201 @@
...
+        return true;
+      }
+    }
+    return false;
+  },
+
+  /**
+   * Persist the (encrypted) refresh token so the session can later be unlocked.
+   * Only ever reached via an explicit, browser-authorized lock, so it does not
+   * consult the locking pref (which the Felt UI process cannot read).
```
lissyx: This is confusing me: token are managed by felt anyway, so why do we mention it here? Are there concurrent accesses between browser and felt on this ?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -50,6 +51,12 @@ const PROCESS_START_REASON = {
   CRASH: "crash",
 };
 
+const FIREFOX_EXIT_MESSAGE = {
+  LOCKED: "FeltParent:FirefoxLockExit",
+  CLOSED: "FeltParent:FirefoxNormalExit",
+  SIGNED_OUT: "FeltParent:FirefoxLogoutExit",
+};
```
lissyx: I'm always skeptical of adding `const` like that for something used once. Can we avoid?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -348,6 +370,18 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+                    // Awaited so the gBrowserRefresh drain covers the write
+                    // and teardown cannot race it. A keystore failure must
+                    // not tear down an otherwise healthy session.
+                    await lazy.FeltLocking.updateStoredToken(
+                      result.refresh_token
+                    );
+                  } catch (err) {
+                    lazy.log.warn(
+                      `Failed to update the stored locked-session token on refresh: ${err}`
+                    );
```
lissyx: That sounds like much more of a fatal error than just a warning to me?

lissyx: It means we have a stale token stored, so when we try to re-use it and we expect it to be good, it will fail. Are we sure this side effect is also correctly handled?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -823,6 +862,24 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+   * @returns {Promise<{posture: DevicePosture, measuredAt: number}>}
+   */
+  async collectLaunchPosture(userId) {
+    const { path: profileDir } = await lazy.resolveManagedProfile({
+      id: userId,
+    });
+    const measuredAt = Date.now();
+    const posture = await lazy.DevicePosture.collect({ profileDir });
+    return { posture, measuredAt };
+  }
```
lissyx: This should be in `DevicePosture.sys.mjs`, similar to `postureForRefresh` no ? https://searchfox.org/enterprise-main/rev/b43f65e4868d2595f2b8c11a8d1df99292742dc8/toolkit/components/enterprise/modules/DevicePosture.sys.mjs#427-444

toolkit/components/felt/content/FeltLocking.sys.mjs
jporter-dev: Overly detailed comment, and a bit outdated after I reworked things. No concurrent access, FELT owns the locking process and the policy-controlled-pref is synced from the browser for gating token storage. 

I'll clean up this docstring.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: yep, good call. adding a line here to clear the stored token and in `tryUnlock` to ensure the spent token is cleared if we fail to store the new one.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: Done, removed. I was torn between adding the const or inlining the strings :)

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Thanks for pointing this out, the previous `warnSuffix` must have slipped by a l10n review and I just built upon that. Fixed it up to have the full strings in the code instead of interpolating the suffix.

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: https://firefox-source-docs.mozilla.org/l10n/fluent/review.html#avoid-generating-identifiers-at-run-time

for future reference :) hadn't seen this detail before but it was advised to be more verbose when dealing with l10n

browser/components/enterprisepolicies/schemas/policies-schema.json
jporter-dev: I think this is the TB version of this: https://bugzilla.mozilla.org/show_bug.cgi?id=2060195

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: This is more like the block here, for collecting fresh posture: https://searchfox.org/enterprise-main/source/toolkit/components/felt/content/FeltProcessParent.sys.mjs#1104-1117

I moved this logic directly into FeltLocking instead of defining this function here.

browser/components/enterprisepolicies/schemas/policies-schema.json
lissyx: So we will have a different policy? Can you make sure you have a patch in the meantime to allow for the missing policy? Otherwise tests will fail and thunderbird people or myself will have to make the fix

jporter-dev: Possibly, I guess this is a good time to sort out policy key naming. Right now the plan was to have...

```
{
  SignOut: {
    BrowserClose: {},
    BrowserCrashed: {}
    BrowserRestart: {},
    NetworkLoss: {},
  }
}
```

Would it make more sense to just remove the `Browser` prefix from those subkeys to keep it generic for both FxE and TbE? Or would it be okay to use the same policy above for TbE? I'm happy with either, but would like to keep it in line with how things are handled on the Thunderbird side to avoid needing a difference in policy. 

Either way, I made bug-2069549 and have a local patch for exempting the `SignOut` policy from testing, the same way we do for policies like `AccessConnector`. Just let me know who to flag as the reviewer (you? or tobyp?) and I can submit to prep for landing this PR.

browser/app/profile/firefox.js
```diff
@@ -32,6 +32,8 @@ pref("enterprise.prompt_on_signout", true);
 //              or is it fine to apply it to any enterprise build?
 pref("app.update.checkOnlyInstance.enabled", false);
 pref("app.update.background.enabled", true);
+// Lock the session instead of signing out on browser close.
+pref("enterprise.locking.browser_close", false, locked);
```
1rneh: `nit`
```suggestion
pref("enterprise.locking.browser_close.enabled", false, locked);
```

browser/components/enterprise/EnterpriseHandler.sys.mjs
```diff
@@ -24,6 +24,7 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
 
 const PROMPT_ON_SIGNOUT_PREF = "enterprise.prompt_on_signout";
 const WARN_ON_CLOSE_PREF = "browser.tabs.warnOnClose";
+const LOCK_ON_CLOSE_PREF = "enterprise.locking.browser_close";
```
1rneh: Above I suggested to append `.enabled` to make it clear by the name that it's a boolean preference. But looking at this preference name salad it looks like that would even add more variety...

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -4176,6 +4176,43 @@
...
+      "x-category": "Authentication",
+      "x-compatibility": {
+        "firefox": {
+          "version_added": false
+        },
+        "firefox_esr": {
+          "version_added": false
+        },
+        "firefox_enterprise": {
+          "version_added": "156"
```
1rneh: The version needs to be updated once this is close to landing.

1rneh (commented):
On a first glimpse this looks sound. I will give it a more thorough review on Monday.

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,211 @@
...
+        let lockingToken, tokenData;
+        try {
+          lockingToken = await lazy.FeltStorage.getLockingToken(email);
+        } catch (err) {
+          lazy.log.warn(
+            `tryUnlock: decrypt failed, falling back to sign-in: ${err}`
+          );
+        }
+        if (!lockingToken) {
+          Services.felt.setTokens("", "", 0);
```
lissyx: Shouldn't we clear?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -538,18 +566,10 @@ export class FeltProcessParent extends JSProcessActorParent {
   }
 
   async startFirefox(startReason, ssoCollectedCookies = []) {
-    // Finish refreshes from the previous browser before replacing its session
-    // id and clearing the posture baseline.
-    if (startReason !== PROCESS_START_REASON.INITIAL_START) {
-      await lazy.PostureMonitor.idle();
-      await gBrowserRefresh;
-      lazy.ClientSession.renew();
-      lazy.PostureMonitor.forget();
```
lissyx, line 548: If we remove all this aren't we regressing posture monitor?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -1085,82 +1154,135 @@ export class FeltProcessParent extends JSProcessActorParent {
     switch (message.name) {
       case "FeltChild:StartFirefox":
         {
+          // An unlock resumes tokens committed through this message (see
+          // FeltLocking.tryUnlock); a fresh SSO login carries a one-time token
+          // the parent redeems below.
+          const { isUnlock = false } = message.data;
```
lissyx: Why isn't it part of the rest of the reading below?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -1085,82 +1154,135 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+              access_token = "",
+              refresh_token = "",
+              expires_in,
+              expires_at,
+            } = message.data;
+            Services.felt.setTokens(
+              access_token,
+              refresh_token,
+              expires_at ??
+                Math.floor(Date.now() / 1000) + Number(expires_in ?? 0)
```
lissyx: Are we sure about this value?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -1085,82 +1154,135 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+            );
+            // Resume into the per-user profile the locked session used.
+            this.loggedInUserInfo = { id: user_id, email };
+            lazy.FeltStorage.updateLastSignedInUserEmail(email);
+            // Clear-on-omit like a login: the resuming refresh restarts the
+            // session, so its response is authoritative, unlike mid-session
+            // refreshes which preserve on omit (see _storeEdrAgents).
+            lazy.EdrAgents.write(postureConfig?.edr_agents);
```
lissyx: Isn't this call redundant with the one after?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -1085,82 +1154,135 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+            } = message.data;
+            Services.felt.setTokens(
+              access_token,
+              refresh_token,
+              expires_at ??
+                Math.floor(Date.now() / 1000) + Number(expires_in ?? 0)
+            );
+            // Resume into the per-user profile the locked session used.
+            this.loggedInUserInfo = { id: user_id, email };
+            lazy.FeltStorage.updateLastSignedInUserEmail(email);
```
lissyx: Similar, looks like we do it here and again just below

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: Somehow lost this during rebase :( restoring it.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -1085,82 +1154,135 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+            lazy.PostureMonitor.record(posture, measuredAt);
+          }
 
           const ssoCollectedCookies = this.getAllCookies();
           lazy.log.debug(`Collected cookies: ${ssoCollectedCookies.length}`);
           // When a restart was reported we assume cookies were stored properly on the
           // browser side?
-          if (!ssoCollectedCookies.length) {
+          if (!isUnlock && !ssoCollectedCookies.length) {
             throw new Error("Not enough cookies!!");
```
lissyx, line 1294: We should update this to know why. We may need a follow up to make sure the error properly surfaces correctly to the user via existing error reporting tooling

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: Rebase issue, fixed.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: same as here: https://searchfox.org/enterprise-main/source/toolkit/components/felt/content/FeltProcessParent.sys.mjs#1151

moved the default 0 back to the `expires_in` destructuring to use the same pattern though.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: removed this duplication and added private helpers to tidy up this added complexity

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jporter-dev: Added a clarifying comment here - we don't have cookies since unlocks just refresh the token and do no full web-based SSO flow.

toolkit/components/felt/FeltStorage.sys.mjs
```diff
@@ -97,6 +98,85 @@ export const FeltStorage = {
...
+   *
+   * @param {string} email
+   * @returns {Promise<string | undefined>} The plaintext refresh token, or
+   *   undefined if none is stored.
+   */
+  async getLockingToken(email) {
+    const ciphertext = this._feltStorage.data?.lockingTokens?.[email]?.token;
+    if (ciphertext === undefined) {
+      return undefined;
+    }
```
lissyx: Cannot we ?

```suggestion
    if (!this.hasLockingToken(email)) {
      return undefined;
    }
```

toolkit/components/felt/Felt.sys.mjs
```diff
@@ -264,26 +266,28 @@ export class Felt {
...
         );
 
+        // Clean exit is the sign-out path, drop any locked tokens.
         lazy.ConsoleClient.performServerSignout()
           .catch(err => {
             console.error(`Failed to post signout on exit: ${err}`);
           })
           .finally(() => {
+            lazy.FeltLocking.clear();
             Services.felt.clearTokens();
```
lissyx: There are still so many places those calls are together, I'm always worried when I see one of those alone. Can we make sure we have a proper factorization here and that all places where it needs to be called in sequences are using an appropriate wrapper?

Could be `FeltLocking.clearLockAndTokens()`

toolkit/components/felt/rust/src/message.rs
```diff
@@ -69,7 +69,7 @@ pub enum FeltMessage {
     RestartForced,
     Restarting,
     LogoutShutdown,
-    Exiting,
+    Exiting(bool),
```
lissyx, line 72: Would it be cleaner to have `ExitingWithLock` instead?

toolkit/components/felt/content/FeltProcessParent.sys.mjs
lissyx: Right, but we should separate the conditions, it would be easier to follow.

lissyx (approved):
r=me with the comments addressed

toolkit/components/felt/Felt.sys.mjs
jporter-dev: Thanks for the suggestion, I had missed this call with a helper I added in this file. I also moved the helper to FeltLocking.

toolkit/components/felt/content/FeltLocking.sys.mjs
```diff
@@ -0,0 +1,224 @@
...
+   *   the same per-user profile.
+   * @throws {Error} If no signed-in user, refresh token, or user id is known,
+   *   so the caller can fall back to signing out instead of persisting a record
+   *   that cannot resume a session.
+   * @returns {Promise<void>}
+   */
+  store: async (refresh_token, userId) => {
+    const email = lazy.FeltStorage.getLastSignedInUser();
+    if (!email) {
+      throw new Error(
```
1rneh, line 196: In this case we can also try one more thing and retrieve the email by initiating a request to `/api/browser/whoami` before throwing. Same for all other email usages in this module. Maybe introduce a new method `getUserEmail()` and implement the multiple steps in there?

browser/components/enterprisepolicies/schemas/policies-schema.json
1rneh: I think it makes sense to already make the keys app name agnostic by removing the `Browser` prefix if this policy is used by TB as well. In general, it's preferable to avoid any foreseeable policy schema updates since the policy schema is additive-only.

browser/components/enterprise/EnterpriseHandler.sys.mjs
1rneh: If we strip the `Browser` prefix from the policy keys, then we should also make the prefs names app agnostic 
```suggestion
const LOCK_ON_CLOSE_PREF = "enterprise.locking.on_close";
```

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -4176,6 +4176,43 @@
...
+          }
+        }
+      ],
+      "properties": {
+        "BrowserClose": {
+          "type": "object",
+          "properties": {
+            "Action": {
+              "type": "string",
+              "enum": ["signout", "lock"]
```
1rneh: The action "signout" seems to be our current default implementation and isn't explicitly read anywhere. Is the plan that it divergences from the default at some point so that "signout" needs to be set explicitly through a policy? Do we expect another action to be added? 

As of now each key answers the question: _"Should we lock or not lock on path x?"_. If this will remain, I'm wondering if this policy is not actually about locking and the schema can be simplified to something like:
```
SessionLocking: {
   OnClose: true|false,
   OnCrash: true|false,
   OnRestart: true|false,
   OnNetworkLoss: true|false,
   [...]
}
```

toolkit/components/felt/tests/xpcshell/xpcshell.toml
1rneh: `toolkit/components/felt/` is build in enterprise builds only:
https://searchfox.org/enterprise-main/rev/73feaaf3ca5b5936d44e3244c981881f73c4e3c6/toolkit/components/moz.build#165-16. So nothing included in this module needs guarding.

1rneh (changes requested):
Thanks for the ping! I'm happy to discuss my suggested changes on the current `SignOut` policy schema (more context in the comments).

browser/components/enterprise/EnterpriseHandler.sys.mjs
jporter-dev: Agree, this makes makes sense, thanks for pointing it out!

browser/components/enterprisepolicies/schemas/policies-schema.json
jporter-dev: I went back and forth on the policy shape a bit and had this same thought originally, and the current version is what I settled on with input from @bsmth - SignOut is easily understood by users/admins, and having the underlying child object allows for additional configuration keys that may be needed. For example, `OnNetworkLoss` would need a configurable timer to determine how long to wait before triggering a signout. This also allows for additional configuration in the future, such as adding the possibility of enabling session restore (bug-2001029) on these signout/lock paths. Keeping each value as an object would enable further expansion of the policy.

As for renaming the policy, personally I think `SessionLocking` is pretty descriptive, but this would require all locking paths we are adding to enforce signout by default. Close/restart/crash all make sense and currently enforce signout already. `NetworkLoss` is a new one, and if that and any others would force a signout by default, then renaming would make more sense.

An example with `SessionLocking` and `Enabled` subbed for `SignOut` and `Action` along with some possible extra params.

```
SessionLocking: {
   OnClose: { Enabled: true|false, RestoreSession: true|false },
   OnCrash: { Enabled: true|false, RestoreSession: true|false  },
   OnRestart: { Enabled: true|false, RestoreSession: true|false  },
   OnNetworkLoss: { Enabled: true|false, GracePeriod: 60, RestoreSession: true|false  },
   [...]
}
```

@bsmth would love your thoughts here!

toolkit/components/felt/content/FeltLocking.sys.mjs
jporter-dev: Thanks, added this fallback and wrapped it in a helper method

browser/components/enterprisepolicies/schemas/policies-schema.json
bsmth: Thanks both! It's not an easy one, for sure. Some thoughts:

* `SignOut`: Indeed, we looked at "session locking" and similar concepts for the policy name, but decided that "signing out" was terminology that would be quicker to to understand at-a-glance, so I think we shouldn't change it based on that.

* `OnClose` v `Close` v `BrowserClose`: I think it's a good call to get rid of `Browser`. One convention we already have is to say "App", like `AppAutoUpdate`, but there may be better alternatives. Should we instead say `Shutdown`, `Crash`, `Restart` and `NetworkLoss`? I don't think we have `On*` in the existing policy conventions.

In a nutshell, I think this is a good shape:

```
"SignOut": {
  "Shutdown": {
    "Action": "lock"
  }
}
```

Then we have sub properties for each event, and an `action` on each event. I like the `SessionLocking` suggestions, but I think this example might be more forward-compatible if we need to make additions, do you agree?

bsmth: And Janika has some good questions, which we can consider:

> Do we expect another action to be added?

Not sure right now, which then makes `SessionLocking` to be about only that use case, which might not be ideal.

>Is the plan that (signout) divergences from the default at some point so that "signout" needs to be set explicitly through a policy?

Good question, I think this is something to cover in docs, that omitting an action for e.g., `Shutdown` would be the same as `Action: signout` (documenting the defaults). We can also express this through the schema?

browser/components/enterprisepolicies/schemas/policies-schema.json
```diff
@@ -4176,6 +4176,43 @@
...
+          }
+        }
+      ],
+      "properties": {
+        "OnClose": {
+          "type": "object",
+          "properties": {
+            "Action": {
+              "type": "string",
+              "enum": ["signout", "lock"]
```
bsmth: Please also check for enum v oneOf -> https://github.com/mozilla/enterprise-firefox/blob/5649786c47540379fa35b6b472b73ca68ed9debc/browser/components/enterprisepolicies/schemas/policies-schema.json#L12. `oneOf` with title and description is better for the console UI generation.

browser/components/enterprisepolicies/schemas/policies-schema.json
jporter-dev: Thanks for the input Brian! 

I have no strong opinion on removing `Browser` and using Shutdown/Restart/Crash/NetworkLoss or `s/Browser/App`.

I agree that having sub-properties for each event, and using `Action: lock|signout` is more forward-compatible in general despite not having an immediate third action.

jporter-dev: I updated this PR to use `Shutdown` and swapped to `oneOf` instead of `enum`. The proposed shape for the full policy will be:

```
"SignOut": {
  "Shutdown": {
    "Action": [signout|lock]
  },
  "Restart": {
    "Action": [signout|lock]
  },
  "Crash": {
    "Action": [signout|lock]
  },
  "NetworkLoss": {
    "Action": [signout|lock],
    "GracePeriodMinutes": 5 // or Seconds?
  }
}
```

1rneh (approved):
Lgtm! Thanks!

PR #1312, opened 2026-08-20
Bug 2066101 - Add release promotion for Enterprise 

lissyx:
### Description 

Bugzilla: Bug-2066101

Add the TaskCluster action required to perform release promotion

---

### Testing 

* [ ] Added tests

Push PR / push to try, once decision task has ran, use `Custom Push Actions` from Treeherder, select `enterprise-release-promotion` and use the flavor you need


python/mozrelease/mozrelease/partner_repack.py
```diff
@@ -431,12 +431,22 @@ def cleanup(self):
...
+        log.info("START [self.repackBuild(] %s %s build %s"% (self.ftp_platform, self.locale, self.build))
         self.repackBuild()
+        log.info("END [self.repackBuild(] %s %s build %s"% (self.ftp_platform, self.locale, self.build))
+        log.info("START [self.stage(] %s %s build %s"% (self.ftp_platform, self.locale, self.build))
         self.stage()
         os.chdir(self.base_dir)
+        log.info("END [self.stage(] %s %s build %s"% (self.ftp_platform, self.locale, self.build))
+        log.info("START [self.rmRecursive(] %s %s build %s"% (self.ftp_platform, self.locale, self.build))
         rmdirRecursive(self.working_dir)
+        log.info("END [self.rmRecursive(] %s %s build %s"% (self.ftp_platform, self.locale, self.build))
```
lissyx: This will be removed

python/mozrelease/mozrelease/partner_repack.py
```diff
@@ -483,7 +493,7 @@ def repackBuild(self):
             tar_flags = "rvf"
         tar_cmd = "tar %s %s %s" % (tar_flags, self.uncompressed_build, self.dest_dir)
         shellCommand(tar_cmd)
-        compress_cmd = "xz -f -z -e -9 %s" % self.uncompressed_build
+        compress_cmd = "xz -f -z -e -9 -T0 --block-size=16MiB %s" % self.uncompressed_build
```
lissyx: This will be removed and maybe landed upstream

taskcluster/kinds/enterprise-repack/kind.yml
```diff
@@ -51,7 +51,7 @@ task-defaults:
     attributes:
         shippable: true
         build_type: opt
-        artifact_prefix: project/enterprise/repacks
+        artifact_prefix: public/project/enterprise/repacks
```
lissyx, line 56: This will be removed and may require us to add some more scopes @ahal

taskcluster/gecko_taskgraph/actions/enterprise_release_promotion.py
```diff
@@ -0,0 +1,116 @@
...
+    context=[],
+    available=is_enterprise_release_promotion_available,
+    cb_name="enterprise-release-promotion",
+    schema=lambda graph_config: {
+        "type": "object",
+        "properties": {
+            "release_promotion_flavor": {
+                "type": "string",
+                "description": "The flavor of release promotion to perform.",
+                "default": get_default_enterprise_flavor(graph_config),
```
jonathanmendez: Is it meaningful/good to populate a default here? I don't know if there's a risk someone accidentally just hits "ok" before checking the flavor. `release_promotion.py` uses "FILL ME OUT"

taskcluster/gecko_taskgraph/transforms/enterprise_release_definition.py
```diff
@@ -0,0 +1,345 @@
...
+        releases = _collect_releases(get_dependencies(config, task))
+
+        if not releases and config.params["release_enable_enterprise_repack"]:
+            raise Exception(
+                "No enterprise release definition could be built: none of the "
+                f"{sorted(BLOB_SPECS)} dependencies yielded a repack id. Without "
+                "this check the kind would silently emit no tasks at all."
+            )
+
+        for (partner, sub_config), (variants, labels) in sorted(releases.items()):
```
jonathanmendez: I think the data structures in this file are way too complicated. I know that nesting of dicts is pretty common in taskcluster code but it feels a bit extreme here. I don't know what to specifically suggest (maybe some typed class objects with mutation methods?) But I'm barely able to grasp the happy path in here, much less keep a review lens for things that could go wrong.

taskcluster/gecko_taskgraph/transforms/require_deps.py
```diff
@@ -0,0 +1,28 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""
+Drop tasks whose ``from-deps`` selection matches nothing. ``from_deps`` raises
```
jonathanmendez: I think this comment can be improved to clarify intent _and_ context.

From my understanding, the context is "The `complete` kind uses `from_deps` with a configuration of `group-by: all`. However, a parameter for a task graph generation (specifically for an enterprise release promotion?) may make it so that the `complete` task's deps are empty, and in that case the run of `from_deps` will raise/error."

And so the intent is to pre-check whether the relevant deps are empty so we can remove the `complete` task so that it won't go through `from_deps` and trigger the error.

taskcluster/gecko_taskgraph/util/verify.py
```diff
@@ -192,7 +192,7 @@ def verify_task_graph_no_shippable_enterprise_level_not_1(
 
     level = int(parameters["level"])
     if level < 3 and "enterprise" in task.label and "shippable" in task.label:
-        raise Exception(
+        print (
```
jonathanmendez, line 195: This seems wrong. No other verify functions just print instead of raising.

taskcluster/kinds/enterprise-release-push/kind.yml
```diff
@@ -0,0 +1,79 @@
...
+
+task-defaults:
+    description: "Stage a Firefox Enterprise OCI release in the registry (by digest)."
+    run-on-projects: [enterprise-firefox]
+    shipping-product: firefox-enterprise
+    shipping-phase: push
+    attributes:
+        shippable: true
+    from-deps:
+        # `fetch` is a kind-dependency only so the moa fetch artifact resolves;
```
jonathanmendez: Is "`fetch`" here referring to the `toolchain` entry in `kind-dependencies`? Stale comment?

taskcluster/kinds/enterprise-release-push/kind.yml
```diff
@@ -0,0 +1,79 @@
...
+            # transform also sets REGISTRY_SECRET, named after whichever
+            # repository the reference points at.
+            REGISTRY_REFERENCE: harbor.stage.fx-enterprise.nonprod.webservices.mozgcp.net/enterprise-shipit-dev/firefox
+    scopes:
+        # `{name}` is the repository of REGISTRY_REFERENCE with its slashes
+        # flattened; the transform renders it (see its SECRET_PATH).
+        - secrets:get:project/enterprise/level-{level}/oci-registry/{name}
+        # moa downloads the (private) enterprise repack artifacts referenced by
+        # the release definition.
+        # TODO: Re-enable - queue:get-artifact:project/enterprise/repacks/*
```
jonathanmendez: Is there a Bug associated with this TODO?

taskcluster/kinds/enterprise-release-ship/kind.yml
```diff
@@ -0,0 +1,66 @@
...
+
+task-defaults:
+    description: "Ship a Firefox Enterprise OCI release by tagging the staged digest."
+    run-on-projects: [enterprise-firefox]
+    shipping-product: firefox-enterprise
+    shipping-phase: ship
+    attributes:
+        shippable: true
+    from-deps:
+        # `fetch` is a kind-dependency only so the moa fetch artifact resolves;
```
jonathanmendez: Same comment about "`fetch`"

taskcluster/scripts/enterprise-oci-tag.sh
```diff
@@ -0,0 +1,48 @@
...
+set -o pipefail
+
+: "${REGISTRY_SECRET:?REGISTRY_SECRET must be set}"
+: "${MOZ_FETCHES_DIR:?MOZ_FETCHES_DIR must be set}"
+: "${TASKCLUSTER_PROXY_URL:?TASKCLUSTER_PROXY_URL must be set (enable the taskcluster proxy)}"
+
+moa="${MOZ_FETCHES_DIR}/moa"
+index_file="${MOZ_FETCHES_DIR}/release-index.json"
+
+chmod +x "${moa}"
```
jonathanmendez: Why is this needed (at all, but also why is it only in ship but not in push)?

taskcluster/scripts/enterprise-oci-upload.sh
```diff
@@ -0,0 +1,59 @@
...
+set -o pipefail
+
+: "${REGISTRY_REFERENCE:?REGISTRY_REFERENCE must be set}"
+: "${REGISTRY_SECRET:?REGISTRY_SECRET must be set}"
+: "${MOZ_FETCHES_DIR:?MOZ_FETCHES_DIR must be set}"
+: "${TASKCLUSTER_PROXY_URL:?TASKCLUSTER_PROXY_URL must be set (enable the taskcluster proxy)}"
+
+moa="${MOZ_FETCHES_DIR}/mozilla-oci-artifacts-cli/moa"
+definition="${MOZ_FETCHES_DIR}/release-definition.json"
+output="/builds/worker/artifacts/release-index.json"
```
jonathanmendez, line 38: `output` isn't used beyond making the directory. I'm guessing the last python invocation is meant to be captured into this path?

taskcluster/scripts/push-enterprise-release.sh
```diff
@@ -0,0 +1,59 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# Stage a Firefox Enterprise OCI release in the registry, addressed by digest.
+#
+# Consumes the `release-definition.json` produced by an upstream
+# `enterprise-release-definition` task and the prebuilt `moa` binary, both in
+# $MOZ_FETCHES_DIR, then runs `moa push release --no-tag`. This uploads all
+# blobs, per-variant manifests, the release index and referrers WITHOUT applying
+# the release-channel tag. The resulting OCI index digest is written to
+# `release-index.txt` and handed off to the ship task, which tags that digest.
```
jonathanmendez:
```suggestion
# `release-index.json` and handed off to the ship task, which tags that digest.
```

jonathanmendez (changes requested):
At the very least it looks like the push script is broken, but I have some general questions as it is difficult to follow everything.

taskcluster/gecko_taskgraph/util/verify.py
lissyx: That's just temporary to allow running things on try

taskcluster/kinds/enterprise-release-push/kind.yml
lissyx: No, this needs to happen before I merge, once I remove the `public/` prefix. Also something just to be able to work painless from try

taskcluster/scripts/enterprise-oci-tag.sh
lissyx: It was already there in @michaelvanstraten's first iteration. Not sure it's really required, I can remove this

taskcluster/scripts/enterprise-oci-upload.sh
lissyx: Right, I forgot it, and since `ship` is not really tested yet, I could not see things were broken

taskcluster/kinds/enterprise-release-push/kind.yml
lissyx: Stale comment from earlier versions

taskcluster/kinds/enterprise-release-ship/kind.yml
lissyx: Same, stale

taskcluster/gecko_taskgraph/actions/enterprise_release_promotion.py
lissyx: No, you're probably right. I thought it would be a good idea, but now that I get shipit to work locally ...

taskcluster/gecko_taskgraph/transforms/require_deps.py
lissyx: Leftover from early version from michael. I think we can remove this now.

taskcluster/gecko_taskgraph/transforms/enterprise_release_definition.py
lissyx: Which ones are complicated here?

michaelvanstraten: I think we can get rid of this post task generation collection entirely by creating the release definition task when we generate each repack task instead of trying to piece things back together here via the (partner, sub_config).

lissyx: This might be also an option indeed.

python/mozrelease/mozrelease/partner_repack.py
lissyx: Merged in Bug-2070996

taskcluster/gecko_taskgraph/transforms/enterprise_release_definition.py
lissyx: I have something pending at  https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=b5de55fc6519fd00bd1793c605aad01bfd6c28c7, but the PR is not yet updated. This allows us to simplify the task graph:
 - without, https://firefox-ci-tc.services.mozilla.com/tasks/groups/CBXVNUZZQXS4Jj6tCeJxww is 991 tasks
 - with, https://firefox-ci-tc.services.mozilla.com/tasks/groups/I0PGfpwrTR2JOkHrwQxAjQ is 803 tasks

taskcluster/docs/attributes.md
```diff
@@ -596,4 +596,17 @@ The label of the task this one was copied from by the `duplicate` transforms.
 A downstream kind that names this task's kind in `duplicate.chain-from` makes
 its own copies depend on this task rather than on the one it was copied from.
 
+## partner
+
+The partner an enterprise repack belongs to, as named in the enterprise
+manifest. Set on the release definition task, which is generated once per
+partner repack, and inherited by the tasks that push and ship that release.
+
+## sub_config
```
ahal: Instead of two separate attributes could we combine them into one? E.g:
```
"attributes": {
  "partner": {
    "name": <partner>,
    "config": <sub_config>
  }
}
```

At the very least this should probably be renamed to `partner_sub_config`

taskcluster/docs/kinds.md
```diff
@@ -68,6 +68,27 @@ This kind performs an artifact build: one based on precompiled binaries
...
+artifact and the prebuilt ``moa`` binary, runs ``moa push release --no-tag``,
+and emits the resulting index digest as ``public/build/release-index.json``.
+
+## enterprise-release-ship
+
+Ships a Firefox Enterprise OCI release by tagging the digest staged by the push
+task with the release channel (a manifest-only retag, no re-upload). One task
+per console consumes the ``enterprise-release-push`` ``release-index.json``
+handoff and runs ``moa tag``.
+
```
ahal, line 87: Having specific kinds / tasks that correspond to the various release phases is the wrong dynamic. I recommend you keep the tasks named after *what* they do rather than *when* they run.

So e.g, you have task labels like `enterprise-oci-tag` or `enterprise-oci-push` (or whatever). Then in each task you specify which shipping phase it runs in like:
https://searchfox.org/firefox-main/source/taskcluster/kinds/release-eme-free-repack/kind.yml#34

In general, a "kind" should group tasks that *do* similar things. With this layout, you'll end up throwing all kinds of unrelated stuff into this kind and the transforms will become a big mess down the road.

ahal (changes requested):

taskcluster/docs/attributes.md
lissyx: Likely with https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=b5de55fc6519fd00bd1793c605aad01bfd6c28c7 this should just go away

taskcluster/docs/kinds.md
lissyx: ok then `enterprise-oci-upload` and `enterprise-oci-tag` would make more sense

taskcluster/gecko_taskgraph/transforms/enterprise_release_definition.py
jonathanmendez: `releases` here is a dict keyed by a 2-tuple, whose values are a 2-tuple of a dict and a set, where the inner dict is itself keyed by a 2-tuple. That's a lot to try to hold in my head while navigating the relationship between the pieces of it while reading the code. FWIW there's a naming issue too with "labels" meaning "dependencies".

taskcluster/docs/kinds.md
lissyx: I'm renaming to those suggestion from above, but if you have other preferences @jonathanmendez and @michaelvanstraten, we can do what you feel bettter

taskcluster/gecko_taskgraph/transforms/enterprise_release_definition.py
lissyx: From https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=b5de55fc6519fd00bd1793c605aad01bfd6c28c7&searchStr=stageGCP it looks like it's working

jonathanmendez (approved):
Looks ok from what I can understand, but I defer to @ahal for a fuller review

taskcluster/scripts/enterprise-oci-upload.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Fetch the .env `moa` reads via dotenv. The secret stores it ready to use;
+# the only work left is unwrapping the {"secret": ..., "expires": ...} envelope
+# the secrets API responds with.
+curl -sSf "${TASKCLUSTER_PROXY_URL%/}/secrets/v1/secret/${REGISTRY_SECRET}" |
+    python3 -c 'import json,sys; sys.stdout.write(json.load(sys.stdin)["secret"]["content"])' \
+    > .env
+
+index=$("${moa}" push release \
+    --concurrency 1 \
```
lissyx: this is not relevant anymore with @michaelvanstraten's 0.2.0 release

taskcluster/gecko_taskgraph/util/verify.py
ahal: This check seems wrong overall though. Shippable Firefox tasks are allowed on try, what was the reason for adding this to enterprise?

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -2374,6 +2374,16 @@ def add_index_routes(config, tasks):
...
+        # Index namespaces embed the head ref, and on try that is a branch
+        # name of any length, URL-encoded into the route. With the long job
+        # names of the enterprise repacks it goes over the 249 character limit
+        # Taskcluster puts on routes. Nothing reads the try namespaces, so skip
+        # them rather than truncate.
+        if config.params["project"] in ENTERPRISE_TRY_PROJECTS:
+            del task["index"]
+            yield task
+            continue
+
```
ahal: It seems better for `index` to not get set in the first place in these cases.. but I guess this is fine for convenience.

Maybe worth filing a Taskcluster issue about this.

taskcluster/gecko_taskgraph/transforms/enterprise_release_definition.py
```diff
@@ -0,0 +1,322 @@
...
+
+def _build_policies(raw_base, revision, build_assets):
+    ftl_files = {
+        "en-US": f"{raw_base}/{revision}/{build_assets['policies-ftl-source']}",
+    }
+
+    changesets_path = os.path.join(GECKO, build_assets["enterprise-l10n"])
+    try:
+        with open(changesets_path) as fh:
+            changesets = json.load(fh)
```
ahal: Because this is ultimately called from within a task loop, we're loading this json file over and over. Would be better to put it in a helper with `@functools.cache`

ahal (approved):
Looks good, have a few comments / suggestions. And obviously please remove all the temp stuff for testing on try before landing.

taskcluster/gecko_taskgraph/util/verify.py
lissyx: @ahal I added this out of safety, because we were making **repacks** from the decison task and having them indexed, and as much as I remember we had a case of pushing a PR that had overwritten something.

Moving to release promotion action, maybe we can drop this enforcement?

What's your take @jonathanmendez ?

taskcluster/gecko_taskgraph/transforms/task.py
lissyx: Sure, I'll look if I can even make a patch maybe

lissyx: https://bugzilla.mozilla.org/show_bug.cgi?id=2074344

taskcluster/gecko_taskgraph/util/verify.py
lissyx: I've been thinking again and I think it's the right move: with moving enterprise repacks to be triggered from an action that will be gated properly, there's no need for this verify step

lissyx:
Recent pushes were good enough without much hacks that everything now works: https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=b4d4f61ec6a9edc49d77c68c7560676a41f6e42d&searchStr=release

This is still pending:
 - [ ] Asserting the state of things for thunderbird
 - [ ] `mozilla-oci-artifacts-cli` new release to avoid the `sed` hack
 - [ ] TaskCluster scopes that I currently bypass on my local instance of `ShitIt`
 - [ ] Handling nightlies with a cron job
 
 Please be aware https://github.com/mozilla/enterprise-console/actions/workflows/push-release.yaml will **not** be usable anymore once this lands

PR #1319, opened 2026-08-21
Bug 2065489 - Account for github repo as well

lissyx:
### Description 

Bugzilla: Bug-2065489

### DIFF
--- a/toolkit/components/telemetry/tests/unit/test_TelemetrySession.js
+++ b/toolkit/components/telemetry/tests/unit/test_TelemetrySession.js
@@ -214,7 +214,7 @@ function checkPayloadInfo(data, reason) {
   // Check for a valid revision.
   if (data.revision != "") {
     const revisionUrlRegEx =
-      /^http[s]?:\/\/hg.mozilla.org(\/[a-z\S]+)+(\/rev\/[0-9a-z]+)$/g;
+      /^http[s]?:\/\/(hg\.mozilla\.org(\/[\S]+)+(\/rev\/[0-9a-z]+)|github\.com\/mozilla\/enterprise-firefox(-try)?\/commit\/[0-9a-f]+)$/g;
     Assert.ok(revisionUrlRegEx.test(data.revision));
   }
 


toolkit/components/telemetry/tests/unit/test_TelemetrySession.js
```diff
@@ -214,7 +214,7 @@ function checkPayloadInfo(data, reason) {
   // Check for a valid revision.
   if (data.revision != "") {
     const revisionUrlRegEx =
-      /^http[s]?:\/\/hg.mozilla.org(\/[a-z\S]+)+(\/rev\/[0-9a-z]+)$/g;
+    /^http[s]?:\/\/(hg\.mozilla\.org(\/[a-z\S]+)+(\/rev\/[0-9a-z]+)|github\.com\/mozilla\/enterprise-firefox\/commit\/[0-9a-f]+)$/g;
```
jonathanmendez: Optionally, since you need to lint anyway and you fixed `.` to `\.`, you could consider removing the redundant `a-z` combined with `\S`
```suggestion
    /^http[s]?:\/\/(hg\.mozilla\.org(\/[\S]+)+(\/rev\/[0-9a-z]+)|github\.com\/mozilla\/enterprise-firefox\/commit\/[0-9a-f]+)$/g;
```

jonathanmendez: I don't know the full context of where the `data.revision` comes from, but will this fail on PR builds with a URL to the user fork rather than to `/mozilla/enterprise-firefox`?

jonathanmendez (approved):

toolkit/components/telemetry/tests/unit/test_TelemetrySession.js
lissyx: That's possible, I'm not sure how we can do better here. It should be green from try so that may be enough?
(and maybe I need to be less strict on the repo)

lissyx: Added try should be enough

PR #1321, opened 2026-08-21
Bug 2065612 - Set proper treeherder routes for enterprise-firefox-try…

ahal:
… actions


ahal:
Should fix busted actions on `enterprise-firefox-try`.. Though hard to test because this invalidates the action hooks.

lissyx (approved):

PR #1322, opened 2026-08-21
Bug 2059588 - Invoke crash reporter from Felt to add bearer authentication to crash reports

jonathanmendez:
### Description 

Bugzilla: Bug-2059588

TODO: This is WIP for an alternative to #1256

---

* Alternative to: #1256 

---

### Testing 

* [x] Added tests

### DIFF
--- a/toolkit/components/felt/FeltCrashReporter.sys.mjs
+++ b/toolkit/components/felt/FeltCrashReporter.sys.mjs
@@ -1,0 +1,231 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const lazy = {};
+
+ChromeUtils.defineESModuleGetters(lazy, {
+  ConsoleClient: "resource://gre/modules/enterprise/ConsoleClient.sys.mjs",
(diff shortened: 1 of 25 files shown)


toolkit/components/felt/content/FeltProcessParent.sys.mjs
afranchuk, line 704:
```suggestion
   * leaves the submission to us, so that the upload can be authenticated with a
```
s/client/submission. At the time of writing, GH isn't rendering this suggestion correctly, so hopefully this change is clear.

toolkit/components/felt/FeltCrashReporter.sys.mjs
afranchuk, line 136: If the crash reporter fails (non-zero exit code), there's no handling of this case. It should probably throw, since calling code expects an exception on failure.

afranchuk, line 135: I think `proc.wait()` is better than using `exitPromise` (though they have the same result as currently implemented).

toolkit/crashreporter/CrashAnnotations.yaml
afranchuk, line 511: This should be `scope: client` (the default if `scope` is omitted). It shouldn't actually be sent in reports.

afranchuk (changes requested):
This looks reasonable, but it is yet another different crash path to maintain.

jonathanmendez:
Closing this as we went with the original approach.

PR #1325, opened 2026-08-22
Bug 2065840 - Capture the active theme in browser_theme_undo.js

1rneh:
### Description 

Bugzilla: Bug-2065840

- Captures the active theme instead of testing with upstream's default theme in `test_undoTheme`.

---

### Testing 

* [X] Added tests (restored)


lissyx (approved):
Thanks, I'm sure I had the same patch somewhere ready and never pushed it :)

PR #1339, opened 2026-08-25
Bug 2068283 - Generic console support

fiji-flo:
### Description 

I'll edit this later, opening for sharing early.

Bugzilla: Bug-2068283

Every enterprise deployment currently requires a per-customer repack that bakes the console address into the AutoConfig file (`firefox.cfg`). This adds a *generic build*: one artifact that ships without a console address and gets configured at first startup, so the same build can serve any organization (downloads, evaluations, QA).

 **Implementation Details:**

- `--with-enterprise-console-url` still bakes a real address in for repacks. Without it the build is generic and `firefox.cfg` carries the placeholder `FIREFOX_ENTERPRISE_GENERIC`
- At startup, before profile selection, the placeholder is resolved in this order:
  1. the `MOZ_ENTERPRISE_CONSOLE_URL` environment variable (test harness override),
  2. the `consoleAddress` persisted in `felt.json` (UAppData) by the setup dialog,
  3. otherwise a modal pre-profile setup dialog (`chrome://felt/content/consoleSetup.xhtml`, FELT UI only; skipped under `MOZ_AUTOMATION` and in background-task mode) asks for the address, persists it to `felt.json`, and relaunches so the early consumers (crash reporter `ServerURL`, update URL, FELT connection) see the configured value from the start.
- `--reset-console-address` clears the stored address so the dialog runs again.
- The resolution logic lives once, in the new IO-free `enterprise-console` crate (`toolkit/components/enterprise/rust`), shared by all native consumers: browser startup through the `felt` crate's FFI (`XRE_ReadEnterpriseConsoleAddress` / `XRE_ParseEnterpriseServerURL` in `toolkit/xre/CreateAppData.cpp`), and the standalone crash reporter client. `ConsoleClient.sys.mjs` mirrors it in JS.
- mozrunner and Marionette set `MOZ_ENTERPRISE_CONSOLE_URL` to a closed local port by default so generic builds never block on the dialog under automation.

---

### Screenshots 

image

---

### Testing 

* [ ] Added tests

### DIFF
--- a/browser/moz.configure
+++ b/browser/moz.configure
@@ -76,7 +76,7 @@ set_define("MOZ_ENTERPRISE", enable_enterprise)
 imply_option("--with-crashreporter-url", "", when="--enable-enterprise")
 
 default_enterprise_console = dependable(
-    "https://stage.fx-enterprise.nonprod.webservices.mozgcp.net"
+    "FIREFOX_ENTERPRISE_GENERIC"
 )
 
 option(
@@ -94,9 +94,15 @@ option(
 def enterprise_console_url(value, default_url):
     # --disable-enterprise-console or an empty value falls back to the default.
     url = value[0] if value and value[0] else default_url
-    if not url.startswith("https://"):
+    if not any(
+        [
+            url.startswith("https://"),
+            url.startswith("http://"),
+            url == "FIREFOX_ENTERPRISE_GENERIC",
+        ]
+    ):
         die(
-            "--with-enterprise-console-url must be a https:// URL, got %s",
+            "--with-enterprise-console-url must be a http(s):// URL, got %s",
             url,
         )
     # The value is interpolated into a JS string literal in firefox.cfg.
--- a/testing/enterprise/base_test.py
+++ b/testing/enterprise/base_test.py
@@ -61,6 +61,7 @@ class Environment(Enum):
 class EnterpriseTestsBase(MarionetteTestCase):
     def setUp(self):
         os.environ.update({"MOZ_DISABLE_NONLOCAL_CONNECTIONS": "0"})
+        os.environ.update({"MOZ_ENTERPRISE_CONSOLE_URL": "http://127.0.0.1:1"})
 
         if getattr(self, "EXTRA_ENV", None):
             self._saved_env = deepcopy(os.environ)
@@ -128,6 +129,7 @@ class EnterpriseTestsBase(MarionetteTestCase):
             self.marionette.instance.prefs = None
 
         del os.environ["MOZ_DISABLE_NONLOCAL_CONNECTIONS"]
+        del os.environ["MOZ_ENTERPRISE_CONSOLE_URL"]
 
         self.marionette.quit(in_app=False, clean=True)
 
--- a/toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
+++ b/toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
@@ -35,6 +35,36 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
 export const CONSOLE_ADDRESS_PREF = "enterprise.console.address";
 
 /**
+ * Placeholder value the pref holds on generic (non-repacked) builds, where
+ * the AutoConfig file does not bake in a real console address. The actual
+ * address then comes from the MOZ_ENTERPRISE_CONSOLE_URL environment
+ * variable or from felt.json, filled in by the pre-profile console setup
+ * dialog. Keep in sync with CONSOLE_ADDRESS_PLACEHOLDER in the
+ * enterprise-console crate (toolkit/components/enterprise/rust), which holds
+ * this resolution logic for the native consumers; resolveConsoleAddress below
+ * mirrors it in JS.
+ */
+export const CONSOLE_ADDRESS_PLACEHOLDER = "FIREFOX_ENTERPRISE_GENERIC";
+
+async function resolveConsoleAddress(prefValue) {
+  if (prefValue !== CONSOLE_ADDRESS_PLACEHOLDER) {
+    return prefValue;
+  }
+  const envUrl = Services.env.get("MOZ_ENTERPRISE_CONSOLE_URL");
+  if (envUrl) {
+    return envUrl;
+  }
+  await lazy.FeltStorage.init();
+  const storedUrl = lazy.FeltStorage.getConsoleAddress();
+  if (!storedUrl) {
+    throw new Error(
+      "Console address is the generic placeholder and no stored address exists"
+    );
+  }
+  return storedUrl;
+}
+
+/**
  * Error logged when user needs to reauthenticate to obtain new token data
  */
 class ReauthRequiredError extends Error {
@@ -100,7 +130,7 @@ export const ConsoleClient = {
       this._consoleUriReadyPromise = new Promise((resolve, reject) => {
         try {
           const consoleURI = Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
-          resolve(consoleURI);
+          resolve(resolveConsoleAddress(consoleURI));
         } catch (e) {
           lazy.log.warn(`Missing console URI. Waiting on pref change.`);
           const consolePrefObserver = {
@@ -115,7 +145,7 @@ export const ConsoleClient = {
                   try {
                     const consoleURI =
                       Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
-                    resolve(consoleURI);
+                    resolve(resolveConsoleAddress(consoleURI));
                   } catch (ex) {
                     lazy.log.error(
                       `Critical misconfiguration: Missing console URI`
@@ -128,6 +158,13 @@ export const ConsoleClient = {
           };
           Services.prefs.addObserver(CONSOLE_ADDRESS_PREF, consolePrefObserver);
         }
+      });
+      // A failure (e.g. felt.json unreadable at that instant) must not be
+      // cached for the rest of the session: clear the slot so the next
+      // access retries the resolution. Callers still see the rejection.
+      this._consoleUriReadyPromise.catch(e => {
+        lazy.log.error("Failed to resolve the console address", e);
+        this._consoleUriReadyPromise = null;
       });
     }
     return this._consoleUriReadyPromise.then(url => new URL(url));
--- a/toolkit/components/enterprise/rust/enterprise-console/src/lib.rs
+++ b/toolkit/components/enterprise/rust/enterprise-console/src/lib.rs
@@ -1,0 +1,348 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Shared logic for obtaining the enterprise console address.
+//!
+//! A repack bakes the console address into the AutoConfig file
+//! (`firefox.cfg`); on generic builds the file holds
+//! [`CONSOLE_ADDRESS_PLACEHOLDER`] instead, and the address is resolved from
+//! the [`CONSOLE_ADDRESS_ENV`] environment variable or from the value the
+//! console setup dialog persisted in `felt.json`. This crate keeps that
+//! resolution identical for every native consumer: the browser (through the
+//! felt crate's FFI, used by CreateAppData.cpp) and the standalone crash
+//! reporter client. `resolveConsoleAddress` in ConsoleClient.sys.mjs mirrors
+//! it in JS.
+//!
+//! The crate is IO-free: callers hand in file contents and the environment
+//! value, so each consumer keeps its own file and environment access (the
+//! crash reporter mocks both in its tests).
+
+/// The pref holding the enterprise console address, set by the AutoConfig
+/// file.
+pub const CONSOLE_ADDRESS_PREF: &str = "enterprise.console.address";
+
+/// Placeholder address in the AutoConfig file of generic (non-repacked)
+/// builds. Keep in sync with CONSOLE_ADDRESS_PLACEHOLDER in
+/// ConsoleClient.sys.mjs and the placeholder in
+/// browser/branding/enterprise/byteshift.py.
+pub const CONSOLE_ADDRESS_PLACEHOLDER: &str = "FIREFOX_ENTERPRISE_GENERIC";
+
+/// Environment variable providing the console address on generic builds
+/// (used by test harnesses).
+pub const CONSOLE_ADDRESS_ENV: &str = "MOZ_ENTERPRISE_CONSOLE_URL";
+
+/// Storage file in the user application data directory where the console
+/// setup dialog persists the address on generic builds.
+pub const FELT_STORAGE_FILENAME: &str = "felt.json";
+
+/// Key holding the console address in the felt storage file. Keep in sync
+/// with FeltStorage.sys.mjs.
+pub const FELT_CONSOLE_ADDRESS_KEY: &str = "consoleAddress";
+
+/// Default byte shift applied to AutoConfig files
+/// (`general.config.obscure_value`). Keep in sync with OBSCURE_VALUE in
+/// browser/branding/enterprise/byteshift.py.
+pub const DEFAULT_OBSCURE_VALUE: u8 = 13;
+
+/// Pref-setting functions an AutoConfig file may use to set a string pref.
+const PREF_FUNCTIONS: &[&str] = &["lockPref", "defaultPref", "pref"];
+
+/// Extract the console address from the raw contents of an AutoConfig file
+/// without evaluating it.
+///
+/// AutoConfig files are byte shifted by `general.config.obscure_value` and
+/// have an intentionally unparseable first line; the shift is undone (trying
+/// the default value, then plaintext) and the file scanned for the pref call.
+/// On generic builds this yields [`CONSOLE_ADDRESS_PLACEHOLDER`]; pass the
+/// result through [`resolve_console_address`].
+pub fn console_address_from_autoconfig(contents: &[u8]) -> Option<String> {
+    for shift in [DEFAULT_OBSCURE_VALUE, 0] {
+        let decoded: Vec<u8> = contents.iter().map(|b| b.wrapping_sub(shift)).collect();
+        let Ok(text) = String::from_utf8(decoded) else {
+            continue;
+        };
+        for func in PREF_FUNCTIONS {
+            if let Some(value) = find_string_pref_call(&text, func, CONSOLE_ADDRESS_PREF) {
+                return Some(value.to_owned());
+            }
+        }
+    }
+    None
+}
+
+/// Why [`resolve_console_address`] or [`stored_console_address`] could not
+/// produce an address. Callers treat every variant as "console setup needed";
+/// the variant names the cause for logging.
+#[derive(Debug)]
+pub enum ResolveError {
+    /// felt.json could not be read. `std::io::ErrorKind::NotFound` is the
+    /// expected first-run state; other kinds are real problems.
+    FeltStorageRead(std::io::Error),
+    /// felt.json exists but is not parseable JSON.
+    FeltStorageCorrupt,
+    /// felt.json holds no (non-empty) console address.
+    NoStoredAddress,
+}
+
+impl std::fmt::Display for ResolveError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ResolveError::FeltStorageRead(e) => {
+                write!(f, "could not read the felt storage file: {e}")
+            }
+            ResolveError::FeltStorageCorrupt => {
+                write!(f, "the felt storage file is not parseable JSON")
+            }
+            ResolveError::NoStoredAddress => {
+                write!(f, "the felt storage file holds no console address")
+            }
+        }
+    }
+}
+
+impl std::error::Error for ResolveError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            ResolveError::FeltStorageRead(e) => Some(e),
+            _ => None,
+        }
+    }
+}
+
+impl ResolveError {
+    /// The expected setup-needed state: no felt.json yet, or one without an
+    /// address. Everything else points at a real problem worth surfacing.
+    pub fn is_expected_first_run(&self) -> bool {
+        match self {
+            ResolveError::FeltStorageRead(e) => e.kind() == std::io::ErrorKind::NotFound,
+            ResolveError::FeltStorageCorrupt => false,
+            ResolveError::NoStoredAddress => true,
+        }
+    }
+}
+
+/// Resolve a console address that may be [`CONSOLE_ADDRESS_PLACEHOLDER`].
+///
+/// A real address is returned unchanged. The placeholder resolves to
+/// `env_value` (the value of [`CONSOLE_ADDRESS_ENV`]) when non-empty, then to
+/// the address stored in felt.json, whose contents `read_felt_json` supplies
+/// (called only when needed). The error says why the placeholder could not be
+/// resolved; the browser then shows the console setup dialog.
+pub fn resolve_console_address(
+    address: &str,
+    env_value: Option<&str>,
+    read_felt_json: impl FnOnce() -> Result<Vec<u8>, std::io::Error>,
+) -> Result<String, ResolveError> {
+    if address != CONSOLE_ADDRESS_PLACEHOLDER {
+        return Ok(address.to_owned());
+    }
+    if let Some(url) = env_value {
+        if !url.is_empty() {
+            return Ok(url.to_owned());
+        }
+    }
+    stored_console_address(&read_felt_json().map_err(ResolveError::FeltStorageRead)?)
+}
+
+/// Read the console address out of felt.json contents.
+pub fn stored_console_address(felt_json: &[u8]) -> Result<String, ResolveError> {
+    let json: serde_json::Value =
+        serde_json::from_slice(felt_json).map_err(|_| ResolveError::FeltStorageCorrupt)?;
+    match json.get(FELT_CONSOLE_ADDRESS_KEY).and_then(|v| v.as_str()) {
+        Some(url) if !url.is_empty() => Ok(url.to_owned()),
+        _ => Err(ResolveError::NoStoredAddress),
+    }
+}
+
+/// Outcome of [`remove_stored_console_address`].
+pub enum RemoveStoredAddress {
+    /// No address was stored (including unparseable contents); nothing to
+    /// write back.
+    AlreadyAbsent,
+    /// The address was removed; the new felt.json contents to write back.
+    Removed(String),
+    /// The contents are not a JSON object; nothing can be removed.
+    Invalid,
+}
+
+/// Remove the console address from felt.json contents, keeping other keys.
+pub fn remove_stored_console_address(felt_json: &[u8]) -> RemoveStoredAddress {
+    let Ok(mut json) = serde_json::from_slice::<serde_json::Value>(felt_json) else {
+        return RemoveStoredAddress::AlreadyAbsent;
+    };
+    let Some(obj) = json.as_object_mut() else {
+        return RemoveStoredAddress::Invalid;
+    };
+    if obj.remove(FELT_CONSOLE_ADDRESS_KEY).is_none() {
+        return RemoveStoredAddress::AlreadyAbsent;
+    }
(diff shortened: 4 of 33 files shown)


toolkit/xre/CreateAppData.cpp
```diff
@@ -106,6 +140,26 @@ nsresult XRE_ParseEnterpriseServerURL(XREAppData& aAppData,
     NS_ENSURE_SUCCESS(rv, rv);
   }
 
+  if (serverUrl.EqualsLiteral(ENTERPRISE_CONSOLE_PLACEHOLDER)) {
```
lissyx: Out of safety I think this should be reversed: if the value is not the place holder, immediately return `NS_ERROR_NOT_AVAILABLE` (or maybe `NS_ERROR_INVALID_ARG` would be more meaningful

browser/branding/enterprise/distribution.ini
```diff
@@ -7,4 +7,7 @@ version=1.0
 about=Mozilla Firefox Enterprise Local Build
 
 [Preferences]
-enterprise.console.address=https://stage.fx-enterprise.nonprod.webservices.mozgcp.net
+# FIREFOX_ENTERPRISE_GENERIC is a placeholder: on first launch the console
+# setup dialog asks for the address and persists it in felt.json. Repacks
+# replace it with the actual console address.
+enterprise.console.address=FIREFOX_ENTERPRISE_GENERIC
```
lissyx: But then we break running local builds without requiring extra step from developpers, which adds friction

lissyx: I'd rather prefer that we do not change the default `distribution.ini` to such placeholder. We may at some point want to point it to a different instance than stage though ?

But I'd prefer if we had a dedicated repack that will set the placeholder value

This way, local build are still immediately actionable and the console setup step works like you want from the specific repack

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -105,7 +132,7 @@ export const ConsoleClient = {
       this._consoleUriReadyPromise = new Promise((resolve, reject) => {
         try {
           const consoleURI = Services.prefs.getStringPref(CONSOLE_ADDRESS_PREF);
-          resolve(consoleURI);
+          resolve(resolveConsoleAddress(consoleURI));
```
lissyx, line 133: I'd prefer we explicitely test before calling `resolveConsoleAddress` and remove the check for `CONSOLE_ADDRESS_PLACEHOLDER` from `resolveConsoleAddress`
```suggestion
          if (consoleURI === CONSOLE_ADDRESS_PLACEHOLDER) {
            resolve(resolveConsoleAddress(consoleURI));
            return;
          }
          
          resolve(consoleURI);
```

And the same in the later handling

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -39,6 +39,33 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+ * address then comes from the MOZ_ENTERPRISE_CONSOLE_ADDRESS environment
+ * variable or from felt.json, filled in by the pre-profile console setup
+ * dialog. Keep in sync with ENTERPRISE_CONSOLE_PLACEHOLDER in nsXULAppAPI.h.
+ */
+export const CONSOLE_ADDRESS_PLACEHOLDER = "FIREFOX_ENTERPRISE_GENERIC";
+
+async function resolveConsoleAddress(prefValue) {
+  if (prefValue !== CONSOLE_ADDRESS_PLACEHOLDER) {
+    return prefValue;
+  }
```
lissyx, line 52:
```suggestion
```

lissyx:
image
That looks nice!

browser/branding/enterprise/distribution.ini
lissyx: It can also make sure that I cannot inadvertenly have a wrong console in memory of my `felt.json` storage that gets picked up when running random builds, scoping the feature to those generic repacks

toolkit/components/felt/rust/src/lib.rs
lissyx, line 1: We need to be extra careful about reviewing further changes using those function, as we may end up in a case where `FeltStorage.sys.mjs` and those are manipulating the same files at the same time. I think for the time being it's OK, even on macOS where we may have running processes without a window.

But maybe a naïve locking mechanism may be good to protect us for the future?

toolkit/components/felt/FeltStorage.sys.mjs
```diff
@@ -23,9 +23,11 @@ export const FeltStorage = {
...
 
   async init() {
-    this._feltStorage = new lazy.JSONFile({
-      path: this.FELT_FILE_PATH,
-    });
+    if (!(this._feltStorage instanceof lazy.JSONFile)) {
+      this._feltStorage = new lazy.JSONFile({
+        path: this.FELT_FILE_PATH,
+      });
+    }
```
lissyx: Should we rather guard with a boolean, e.g. `this._initialized` ?

lissyx: Also why do we need this?

I assume it's because we hit now two call sites of `FeltStorage.init()` one in ConsoleClient and one in consoleSetup.js ?

toolkit/xre/CreateAppData.cpp
```diff
@@ -98,6 +101,37 @@ static nsresult ParseConsoleUrlFromDistribution(XREAppData& aAppData,
...
+  // into .../felt.json for every later consumer.
+  nsCOMPtr<nsIFile> file;
+  rv = dir->Clone(getter_AddRefs(file));
+  NS_ENSURE_SUCCESS(rv, rv);
+  rv = file->AppendNative("felt.json"_ns);
+  NS_ENSURE_SUCCESS(rv, rv);
+  nsAutoString path;
+  rv = file->GetPath(path);
+  NS_ENSURE_SUCCESS(rv, rv);
+  CopyUTF16toUTF8(path, aOutPath);
```
lissyx, line 107: `nsCString` is safe for windows paths ? What if the path contains UTF-8 or UTF-16 chars?

toolkit/xre/nsAppRunner.cpp
```diff
@@ -4889,9 +4959,20 @@ int XREMain::XRE_mainInit(bool* aExitFlag,
   }
 
 #if defined(MOZ_ENTERPRISE)
-  XRE_ParseEnterpriseServerURL(*mAppData);
-  // Ignoring nsresult. If console url is not found in a release build, the
-  // default server url is empty and crash reports will fail to submit.
+  if (CheckArg("reset-console-address") == ARG_FOUND) {
```
lissyx: So `firefox --reset-console-address` would allow to trigger again the dialog, but only if its distribution.ini contains the placeholder value.

At my first reading of this part of the code I was afraid, so it may be worth commenting here to make this context crystal clear

toolkit/xre/nsAppRunner.cpp
```diff
@@ -5991,6 +6072,30 @@ int XREMain::XRE_mainStartup(bool* aExitFlag,
 
   // We now know there is no existing instance using the selected profile.
 
+#if defined(MOZ_ENTERPRISE)
+  // Never block automation on the interactive setup dialog: test harnesses
+  // launch the browser before any console address exists (Marionette's first
+  // launch precedes the test setUp that provides one via distribution.ini or
+  // MOZ_ENTERPRISE_CONSOLE_ADDRESS) and would time out waiting for it.
```
lissyx: I think this becomes irrelevant if we move to making such a generic repack instead of changing the default `distribution.ini` value and the whole checks can be simplified

toolkit/xre/nsAppRunner.cpp
```diff
@@ -5991,6 +6072,30 @@ int XREMain::XRE_mainStartup(bool* aExitFlag,
...
+  ) {
+    rv = ShowEnterpriseConsoleSetup(mNativeApp);
+    if (rv == NS_ERROR_LAUNCHED_CHILD_PROCESS || rv == NS_ERROR_ABORT) {
+      *aExitFlag = true;
+      return 0;
+    }
+    if (NS_FAILED(rv)) {
+      return 1;
+    }
+  }
```
lissyx: Implementing via a generic repack instead of changing local builds values, we should be able to simplify to

```suggestion
  const bool enterpriseConsoleSetupAllowed = is_felt_ui()
#  ifdef MOZ_BACKGROUNDTASKS
      && !BackgroundTasks::IsBackgroundTaskMode()
#  endif
  ;
  
  if (gEnterpriseConsoleSetupNeeded && enterpriseConsoleSetupAllowed) {
    rv = ShowEnterpriseConsoleSetup(mNativeApp);
    if (rv == NS_ERROR_LAUNCHED_CHILD_PROCESS || rv == NS_ERROR_ABORT) {
      *aExitFlag = true;
      return 0;
    }
    if (NS_FAILED(rv)) {
      return 1;
    }
  }
```

lissyx:
@fiji-flo Thanks for taking care of this, I think it's overall right, I just really want that we do not change the default build output and rather make such a "generic" repack that will run on CI

toolkit/xre/CreateAppData.cpp
fiji-flo: Taken from https://searchfox.org/enterprise-main/source/toolkit/xre/nsUpdateDriver.cpp#100

toolkit/xre/CreateAppData.cpp
fiji-flo: That's not the flow here. If the value is not the place holder we need to continue. `NS_ERROR_NOT_AVAILABLE` is only returned if there's no URL stored / in evn.

toolkit/xre/nsAppRunner.cpp
fiji-flo: Can you suggest an improvement to my comments?

toolkit/components/felt/rust/src/lib.rs
fiji-flo: Yes, we might want to think about this outside of this PR. Multiple Enterprise instances would read/write from that file concurrently right now.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
fiji-flo: That's would be much more code be cause this right now bubbles the exceptions. I think about giving it a better name maybe.

toolkit/xre/nsAppRunner.cpp
fiji-flo: I'll look at this part tomorrow.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
lissyx: I'm not sure the naming is wrong, it's really just that we add / hide the "normal" flow that I feel weird, but if it's going to generate much more exception handling, we can keep it this way

toolkit/components/felt/rust/src/lib.rs
lissyx: Can we file a follow up for a trivial locking system (e.g. `.lock` file) at least? To protect us?

Other option, but it would start to get out of scope so a follow up would be nice, would be to convert the `FeltStorage` code to directly rely on rust so the locking could be done there?

toolkit/xre/nsAppRunner.cpp
lissyx: I'll try to come up with something yes

toolkit/xre/nsAppRunner.cpp
```diff
@@ -4888,6 +4958,33 @@ int XREMain::XRE_mainInit(bool* aExitFlag,
     }
   }
 
+#if defined(MOZ_ENTERPRISE)
+  if (CheckArg("reset-console-address") == ARG_FOUND) {
```
```suggestion
#if defined(MOZ_ENTERPRISE)
  // The flag allows to trigger again the dialog, if autoconfig file contains the placeholder value.
  // It should be safe enough on shippable builds where the autoconfig file is part of the distribution
  // and thus should be immune to any change.
  if (CheckArg("reset-console-address") == ARG_FOUND) {
```

testing/enterprise/test_felt_console_error.py
```diff
@@ -95,7 +95,7 @@ def test_felt_unreachable_ip_shows_connection_error(self):
             },
             default_branch=True,
         ):
-            self.login_location.value = "http://127.0.0.1:1"
+            self.login_location.value = "https://127.0.0.1:1"
```
lissyx: If we allow `http` at build time, do we need this change?

browser/moz.configure
```diff
@@ -94,9 +94,15 @@ option(
 def enterprise_console_url(value, default_url):
     # --disable-enterprise-console or an empty value falls back to the default.
     url = value[0] if value and value[0] else default_url
-    if not url.startswith("https://"):
+    if not any(
+        [
+            url.startswith("https://"),
+            url.startswith("http://"),
```
lissyx, line 100: I'm wondering if we (c|sh)ould make sure a build on CI does not allow `http`, out of safety

testing/mozbase/mozrunner/mozrunner/utils.py
```diff
@@ -101,6 +101,13 @@ def test_environment(
     # override the user's choice here.  See bug 1049688.
     env.setdefault("MOZ_DISABLE_NONLOCAL_CONNECTIONS", "1")
 
+    # Give enterprise builds a console address so generic builds never start
+    # unconfigured (or block on the console setup dialog) under a test
+    # harness. A closed local port, because non-local connections are
+    # disabled above. Set it to the empty string to exercise the
+    # unconfigured state.
+    env.setdefault("MOZ_ENTERPRISE_CONSOLE_URL", "https://127.0.0.1:1")
```
lissyx: I dont think it should be taken care of here. We have other places to setup our env that does not involves changing `mozrunner`.

cf https://searchfox.org/enterprise-main/rev/b43f65e4868d2595f2b8c11a8d1df99292742dc8/testing/enterprise/base_test.py#63

testing/marionette/client/marionette_driver/geckoinstance.py
```diff
@@ -441,6 +441,12 @@ def _get_runner_args(self):
         if env.get("MOZ_REMOTE_ALLOW_SYSTEM_ACCESS") is None:
             env.update({"MOZ_REMOTE_ALLOW_SYSTEM_ACCESS": "1"})
 
+        # Give enterprise builds a console address so generic builds never
+        # start unconfigured (or block on the console setup dialog) under
+        # Marionette. Set it to the empty string to exercise the unconfigured
+        # state.
+        env.setdefault("MOZ_ENTERPRISE_CONSOLE_URL", "https://127.0.0.1:1")
```
lissyx: Similar, I do not think it's OK we do this kind of changes

toolkit/components/felt/rust/src/lib.rs
lissyx: @fiji-flo I do not see any locking, not follow up filed for it, and I'm worried this may trigger weird bugs ahrd to pinpoint

toolkit/crashreporter/client/app/src/config.rs
lissyx, line 1: This should be reviewed by @jonathanmendez or @afranchuk

toolkit/crashreporter/client/app/src/glean.rs
lissyx, line 1: Similar, @jonathanmendez should have a look here

toolkit/components/enterprise/rust/enterprise-console/src/lib.rs
```diff
@@ -0,0 +1,348 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Shared logic for obtaining the enterprise console address.
+//!
+//! A repack bakes the console address into the AutoConfig file
+//! (`firefox.cfg`); on generic (non-repacked) builds the file holds
```
lissyx: Since we want to have generic repacks for l10n at least
```suggestion
//! (`firefox.cfg`); on generic builds the file holds
```

browser/moz.configure
```diff
@@ -76,7 +76,7 @@ set_define("MOZ_ENTERPRISE", enable_enterprise)
 imply_option("--with-crashreporter-url", "", when="--enable-enterprise")
 
 default_enterprise_console = dependable(
-    "https://stage.fx-enterprise.nonprod.webservices.mozgcp.net"
+    "FIREFOX_ENTERPRISE_GENERIC"
```
lissyx, line 79: This will make any build out of CI and out of developper that do not pass `--with-enterprise-console-url` value a generic build. I think it's fine, but probably we will want to cover this in any dev build docs we have to limit friction here and people know they need to either bake a value in their mozconfig, and make sure there's a clearly documented instance to use.

Would existing stage be good for this or would we want to have something dedicated ?

toolkit/components/enterprise/rust/enterprise-console/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+/// (used by test harnesses).
+pub const CONSOLE_ADDRESS_ENV: &str = "MOZ_ENTERPRISE_CONSOLE_URL";
+
+/// Storage file in the user application data directory where the console
+/// setup dialog persists the address on generic builds.
+pub const FELT_STORAGE_FILENAME: &str = "felt.json";
+
+/// Key holding the console address in the felt storage file. Keep in sync
+/// with FeltStorage.sys.mjs.
+pub const FELT_CONSOLE_ADDRESS_KEY: &str = "consoleAddress";
```
lissyx, line 41: Out of scope but worth a follow up: to avoid the risk of drifting, we should just move all of the `FeltStorage` actual logic into the crate, and make the JS module just an interface to it?

toolkit/components/enterprise/rust/enterprise-console/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+
+/// Default byte shift applied to AutoConfig files
+/// (`general.config.obscure_value`). Keep in sync with OBSCURE_VALUE in
+/// browser/branding/enterprise/byteshift.py.
+pub const DEFAULT_OBSCURE_VALUE: u8 = 13;
```
lissyx, line 46: Similar but I think @mkaply wants to rework all of that anyway to already avoid the duplication we introduced between Firefox and Thunderbird

toolkit/components/felt/FeltStorage.sys.mjs
```diff
@@ -12,8 +12,10 @@ ChromeUtils.defineESModuleGetters(lazy, {
  * Storage helper for reading and writing felt-related profile data to felt.json
  */
 export const FeltStorage = {
+  _initialized: false,
```
lissyx, line 15:
```suggestion
  _initialized: false,
  
  _feltStorage: null,
```

lissyx (changes requested):
We really should not touch things `mozrunner` and `geckoinstance`

browser/branding/enterprise/distribution.ini
gcp: We discussed this a lot - this only makes sense if you want to assume there is some "right" console for developers and that default specifically is our "stage", which is not even clear. If you want to bake a default console for your own builds, that belongs in the mozconfig, IHMO.

_We may at some point want to point it to a different instance than stage though ?_

and

_local build are still immediately actionable_

Both kind of illustrate this: local builds are only actionable if you **actually have an account on stage**.

lissyx: @gcp this comment is not relevant, now that we have build config it's not a problem anymore, the only problem we have is picking a console, and for the time being it's fine sticking docs to our stage instance, but maybe later we want to spin up a different instance where anyone at mozilla / mzla has access by default ?

browser/moz.configure
gcp: As I tried to point out here: https://github.com/mozilla/enterprise-firefox/pull/1339#discussion_r3933330254

Stage isn't particularly useful as a default and nothing really changes here. If you don't have an account on stage, you can't use it anyway.

lissyx: > As I tried to point out here: [#1339 (comment)](https://github.com/mozilla/enterprise-firefox/pull/1339#discussion_r3933330254)
> 
> Stage isn't particularly useful as a default and nothing really changes here. If you don't have an account on stage, you can't use it anyway.

yes, but my point was about the fact that it's something we now have the opportunity to make better especially as we may have to collaborate more on our code with people outside of our team.

gcp: I agree but fixing that is a separate thing from this PR, no? I would say that this improves things, since being given the "specify a console screen" is a bit more explicit hint that you need to do something than showing our Mozilla SSO and then giving an error, which is what happens now.

lissyx: > I agree but fixing that is a separate thing from this PR, no? I would say that this improves things, since being given the "specify a console screen" is a bit more explicit hint that you need to do something than showing our Mozilla SSO and then giving an error, which is what happens now.

yes, it's separate, I thought it was obvious from my comment. My goal here was just mentionning it and getting feedback on the future change we may want here

toolkit/components/enterprise/rust/enterprise-console/src/lib.rs
fiji-flo: Yes, I agree, but I felt to much for this PR.

lissyx: yep, can you just make sure to file follow up (for this and elsewhere mentionned) ?

testing/enterprise/base_test.py
```diff
@@ -61,6 +61,7 @@ class Environment(Enum):
 class EnterpriseTestsBase(MarionetteTestCase):
     def setUp(self):
         os.environ.update({"MOZ_DISABLE_NONLOCAL_CONNECTIONS": "0"})
+        os.environ.update({"MOZ_ENTERPRISE_CONSOLE_URL": "http://127.0.0.1:1"})
```
lissyx, line 64: nit:
```suggestion
        os.environ.update({
            "MOZ_DISABLE_NONLOCAL_CONNECTIONS": "0",
            "MOZ_ENTERPRISE_CONSOLE_URL": "http://127.0.0.1:1",
         })
```

toolkit/components/enterprise/rust/enterprise-console/src/lib.rs
lissyx: Just make sure to file the follow up, thanks :)

lissyx (approved):
r=me with green tests

toolkit/crashreporter/client/app/src/config.rs
```diff
@@ -269,33 +272,41 @@ impl Config {
...
+                Ok(dir) => self.app_data_dir = Some(dir),
+                Err(e) => log::warn!("could not determine the application data directory: {e:#}"),
+            }
+        }
+
         // Enterprise builds normally get the submission endpoint from the
         // ServerURL annotation. Before the browser derives it from the console
         // address (e.g. a startup crash) it is still the domainless placeholder
         // from application.ini (`/submit?...`), so fall back to reading the
-        // console address from AutoConfig when the annotation isn't usable.
```
afranchuk, line 276: Accidental comment removal?

toolkit/crashreporter/client/app/src/config.rs
```diff
@@ -504,24 +515,50 @@ impl Config {
...
+    /// The crash reporter data directory, `Crash Reports` inside the user
+    /// application data directory.
+    fn get_data_dir(&self, vendor: &str, product: &str) -> anyhow::Result<PathBuf> {
+        let app_data_dir = self.get_app_data_dir(vendor, product)?;
+        #[cfg(all(not(mock), target_os = "macos"))]
+        std::fs::create_dir_all(&app_data_dir).with_context(|| {
+            self.build_string("crashreporter-error-creating-dir")
+                .arg("path", app_data_dir.display().to_string())
+                .get()
+        })?;
```
afranchuk, line 528: I can't recall why this was necessary on macOS, but it feels like it's a good thing to do on all platforms.

toolkit/crashreporter/client/app/src/config.rs
```diff
@@ -269,33 +272,41 @@ impl Config {
...
+        // file, it'd simplify the data_dir logic and things like glean initialization (which
+        // relies on the data dir).
+        let vendor = extra[VENDOR_KEY].as_str().unwrap_or(DEFAULT_VENDOR);
+        let product = extra[PRODUCT_KEY].as_str().unwrap_or(DEFAULT_PRODUCT);
+        if self.data_dir.is_none() {
+            self.data_dir = Some(self.get_data_dir(vendor, product)?);
+            self.update_log_file();
+        }
+        // Computed rather than derived from `data_dir`, which may be an
+        // arbitrary directory passed in the environment.
```
afranchuk, line 286: FWIW [this code](https://searchfox.org/firefox-main/rev/3c6e9f0700b5a702c6d3fde4306554d364adb213/toolkit/crashreporter/nsExceptionHandler.cpp#2305-2308) always overwrites the data dir env, so it shouldn't be arbitrary (and all that calculation code is probably not used often). But this is still perfectly fine. I kinda hope we can get rid of all these env variables setting things up to avoid all this conditional behavior.

browser/moz.configure
lissyx, line 79: Might be too late now, so we can fix this later when we'll re-unite this code, but `FIREFOX_ENTERPRISE_GENERIC` looks weird to Thunderbird folks :), and they would prefer something not referring to Firefox.

toolkit/crashreporter/client/app/src/glean.rs
```diff
@@ -150,9 +160,11 @@ impl InitOptions {
...
         #[cfg(all(not(mock), feature = "enterprise"))]
         {
-            init_glean.configuration.server_endpoint = Some(
-                crate::enterprise_prefs::console_glean_url(server_url.as_deref())?,
-            );
+            init_glean.configuration.server_endpoint =
+                Some(crate::enterprise_prefs::console_glean_url(
+                    server_url.as_deref(),
+                    app_data_dir.as_deref(),
+                )?);
```
afranchuk: I don't think it makes sense to have this call in the `glean` module. A method like `InitOptions::set_server_url` would allow setting with the result of this method externally (and you'd probably be able to avoid all the `cfg` `not(mock)` additions as a beneficial side effect).

toolkit/crashreporter/client/app/src/send_ping.rs
```diff
@@ -7,6 +7,17 @@
...
+///
+/// `CrashManager` always passes `UAppData/Crash Reports`, so the parent of the
+/// given path is the directory holding `felt.json`.
+#[cfg(all(not(mock), feature = "enterprise"))]
+fn app_data_dir(data_path: &::std::ffi::OsStr) -> Option<::std::path::PathBuf> {
+    ::std::path::Path::new(data_path)
+        .parent()
+        .map(::std::path::Path::to_path_buf)
+}
+
```
afranchuk, line 20: Maybe it would be cleaner to pass `UAppData` as the argument to the program and add the `Crash Reports` path component for the glean data dir?

afranchuk (approved):
I reviewed the crashreporter changes.

I think they're functionally fine, though I had some ideas for improvement (and one comment line seems to have been erroneously deleted).

toolkit/crashreporter/client/app/src/send_ping.rs
fiji-flo: I agree but It would change more of the upstream code.

toolkit/crashreporter/client/app/src/config.rs
fiji-flo: That's just code moved, so I'd keep it like that?

afranchuk: Fair enough for now.

toolkit/crashreporter/client/app/src/enterprise_prefs.rs
```diff
@@ -10,52 +10,67 @@
...
+use enterprise_console::{
+    console_address_from_autoconfig, resolve_console_address, CONSOLE_ADDRESS_ENV,
+    CONSOLE_ADDRESS_PREF, FELT_STORAGE_FILENAME,
+};
 use url::Url;
 
 /// The enterprise AutoConfig file name. Enterprise builds point
 /// `general.config.filename` at this (set in `firefox-branding.js`); it is not
 /// a default value anywhere else.
 const ENTERPRISE_AUTOCONFIG_FILENAME: &str = "firefox.cfg";
```
lissyx: Looks like my comment got lost when reviewing, but I had asked for this to be changed at the risk of breaking thunderbird: it's in fact broken.

PR #1345, opened 2026-08-26
Bug 2066822 - Enables enterprise storage encryption by default

1rneh:
### Description 

Bugzilla: Bug-2066822

- Flips `security.storage.encryption.enabled` to true
- Deprecates the policy `EnterpriseStorageEncryption`, and since it's not consumed yet we can remove it entirely.

### Testing 

* [X] Added tests


modules/libpref/init/StaticPrefList.yaml
```diff
@@ -19246,11 +19246,13 @@
...
   mirror: always
 
+#if defined(MOZ_ENTERPRISE) && !defined(MOZ_THUNDERBIRD)
 # Enable storage encryption functionality.
 - name: security.storage.encryption.enabled
   type: bool
-  value: false
+  value: true
   mirror: always
+#endif
```
lissyx: My suggestion on matrix was to stick to what we do for the sqlite pref and the comment from copilot goes into the same direction


```suggestion
# Enable storage encryption functionality only on Firefox Enterprise
- name: security.storage.encryption.enabled
  type: bool
#if defined(MOZ_ENTERPRISE) && !defined(MOZ_THUNDERBIRD)
  value: true
#else
  value: false
#endif
  mirror: always
```

lissyx (approved):
r=me with the suggested changes

1rneh:
@lissyx You're just too fast. I had removed the requested review just a few seconds after putting it 😄

toolkit/components/enterprise/modules/EnterpriseStorageEncryption.sys.mjs
```diff
@@ -13,7 +11,6 @@ ChromeUtils.defineESModuleGetters(lazy, {
 export const EnterpriseStorageEncryption = {
   async init() {
     if (
-      AppConstants.MOZ_ENTERPRISE &&
       Services.prefs.getBoolPref(
         "security.storage.encryption.enabled",
         false
```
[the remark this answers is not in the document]
lissyx: This module is not supposed to be packaged outside of `MOZ_ENTERPRISE` builds. Checking the pref should be enough and reflect the expected state.

Flipping the pref on non enterprise build will have no impact because the module is not loaded.

toolkit/components/enterprise/tests/xpcshell/xpcshell.toml
```diff
@@ -0,0 +1,3 @@
+[DEFAULT]
+
+["test_storage_encryption.js"]
```
lissyx, line 16: This would run and fail I suspect on thunderbird right now since encryption is not enabled there

```suggestion
["test_storage_encryption.js"]
`run-if = ["buildapp == 'browser'"]`
```

lissyx (approved):

toolkit/components/passwordmgr/test/mochitest/mochitest.toml
```diff
@@ -429,6 +429,7 @@ run-if = [
 ]
 skip-if = [
   "xorigin", # Tests desktop prompts and bug 1333264
+  "enterprise", # On enterprise builds the primpary password is provisioned by the console and keeps the token logged in.
```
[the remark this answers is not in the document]
1rneh: I like "primpary" better.

security/manager/ssl/SecretDecoderRing.cpp
```diff
@@ -318,7 +321,8 @@ SecretDecoderRing::Login(const nsACString& password, bool* success) {
 // LoginHelper.sys.mjs, using the identical MOZ_ENTERPRISE + pref gate.
 static bool EnterpriseManagesPrimaryPassword() {
 #if defined(MOZ_ENTERPRISE)
-  return mozilla::StaticPrefs::security_storage_encryption_enabled();
+  return mozilla::StaticPrefs::security_storage_encryption_enabled() &&
+         is_felt_browser();
```
lissyx, line 325: I just verified this locally and as I was thinking, we are using encryption during FELT. @beurdouche what was the rationale for this ?

(verified by `sqlitebrowser` on `permissions.sqlite`: no table visible)

lissyx, line 325:
```suggestion
   (is_felt_browser() || is_felt_ui())
```

toolkit/components/passwordmgr/LoginHelper.sys.mjs
```diff
@@ -1623,7 +1623,11 @@ export const LoginHelper = {
   isEnterpriseManagedPrimaryPassword() {
     return (
       AppConstants.MOZ_ENTERPRISE &&
-      Services.prefs.getBoolPref("security.storage.encryption.enabled", false)
+      Services.prefs.getBoolPref(
+        "security.storage.encryption.enabled",
+        false
+      ) &&
+      Services.felt.isFeltBrowser()
```
lissyx, line 1642:
```suggestion
   (Services.felt.isFeltBrowser() || Services.felt.isFeltUI())
```

security/manager/ssl/SecretDecoderRing.cpp
beurdouche: No real reason not to do it since not doing it would make things more complex. The Kek in the felt case is a local KEK because you don’t know who the user is at the beginning so you can’t protect the profile with a remote KEK.

1rneh: > I just verified this locally and as I was thinking, we are using encryption during FELT. @beurdouche what was the rationale for this ?
>
> (verified by sqlitebrowser on permissions.sqlite: no table visible)


@lissyx The preferences `security.storage.encryption.enabled` and `security.storage.encryption.sqlite.enabled` do two different things.

- `security.storage.encryption.enabled` controls whether to fetch and use the console provisioned (remote) kek.
- `security.storage.encryption.sqlite.enabled` controls whether to encrypt sqlite databases

My patch **only** changes the default whether to fetch the remote kek. And it's restricted to only enable that feature in the launched browser. So there is no remote kek in felt but sqlite encryption remains enabled.

lissyx: Thanks for reminding me, we should have picked different, more explicit prefs names, it's always confusing to me when I come back later on it.

lissyx: So we do not need this then

toolkit/components/passwordmgr/LoginHelper.sys.mjs
lissyx: And we do not need this either

toolkit/components/passwordmgr/LoginHelper.sys.mjs
```diff
@@ -1635,7 +1635,11 @@ export const LoginHelper = {
   isEnterpriseManagedPrimaryPassword() {
     return (
       AppConstants.MOZ_ENTERPRISE &&
-      Services.prefs.getBoolPref("security.storage.encryption.enabled", false)
+      Services.prefs.getBoolPref(
+        "security.storage.encryption.enabled",
+        false
+      ) &&
```
lissyx, line 1641: Maybe wroth a comment to make sure next readers have correct context on remote vs local `kek`

security/manager/ssl/SecretDecoderRing.cpp
lissyx: And so maybe worth a comment as well here, so the confusion is defused for next readers

1rneh: I could rename the preference to `security.storage.encryption.remote_kek.enabled` to remove confusion.

lissyx: But likely in a follow up, and this may need to be dealt with at m-c level, some of this landed on m-c

1rneh: Most of `security.storage.encryption.sqlite.enabled` is in upstream, but `security.storage.encryption.enabled` is enterprise-only. But yes, this can be a follow-up.

toolkit/components/enterprise/modules/EnterpriseStorageEncryption.sys.mjs
```diff
@@ -13,12 +11,11 @@ ChromeUtils.defineESModuleGetters(lazy, {
...
       ) &&
-      !Services.felt?.isFeltUI()
+      Services.felt?.isFeltBrowser()
     ) {
       await this.load();
```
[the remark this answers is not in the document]
1rneh: Filed Bug-2073615.

PR #1352, opened 2026-08-27
Bug 2040418 - Address enterprise localization review comments

gcp:
Addresses the review feedback from https://github.com/mozilla-l10n/enterprise-firefox-l10n/pull/27:

- Uses Device ID instead of Machine ID for the about:support label.
- Clarifies that Endpoint Detection and Response is an industry term that must remain in English.

Validation: python3.9 ./mach lint toolkit/content/aboutSupport.xhtml toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl

### DIFF
--- a/toolkit/content/aboutSupport.xhtml
+++ b/toolkit/content/aboutSupport.xhtml
@@ -144,7 +144,7 @@
           </tr>
 
           <tr id="machine-id-row" class="no-copy">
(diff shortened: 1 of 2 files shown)


lissyx (approved):

flodolo:
Could you uplift this to beta, possibly release?

It would prevent this string for lingering around for 2 or 4 more weeks on our side.

PR #1355, opened 2026-08-27
Bug 2062380 - Add admin-provided rule messages to block and warn dialogs

jonathanmendez:
### Description 

Bugzilla: Bug-2062380

The DLP rules allow an admin to specify a message for when the rule is triggered. This PR adds the surfacing of such messages to the block or warn dialogs.

This PR also updated the in-tree test wasm to an implementation that behaves according to the input, so that tests can effectively mock behavior. The source for the new wasm is included in-tree.

---

### Screenshots 

With message in block dialog: 
image

With message in warn dialog:
image

---

### Testing 

* [x] Added tests
* [x] TODO Manual testing performed


toolkit/components/contentanalysis/nsIContentAnalysis.idl
```diff
@@ -86,6 +86,12 @@ interface nsIContentAnalysisResponse : nsIContentAnalysisResult
   // e.g. this is a synthetic response Firefox produced itself.
   readonly attribute AString ruleName;
 
+  // The admin-authored message from the rule that determined `action` (see
+  // nsIContentAnalysisRule.message). Empty if that rule supplied no message,
+  // or if the verdict came from a backend that has no rule configuration to
+  // draw one from (e.g. an external agent).
+  readonly attribute AString ruleMessage;
```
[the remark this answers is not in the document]
gregstoll: FWIW it's fine to do this but I don't think it's necessary; we don't really guarantee binary compatibility for anything. (this was a bigger deal back in the days where addons could directly call into XPCOM stuff)

browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
```diff
@@ -932,7 +943,8 @@ export const ContentAnalysis = {
     aUserActionId,
     aCAResult,
     aIsSyntheticResponse,
-    aRequestCancelError
+    aRequestCancelError,
+    aRuleMessage
```
gregstoll, line 947: This is fine for now but sheesh the signature of this function is getting hairy :-(

toolkit/components/contentanalysis/nsIContentAnalysis.idl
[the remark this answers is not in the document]
jonathanmendez: Yeah agents love to point this out and I just swat them away.

browser/components/contentanalysis/content/ContentAnalysis.sys.mjs
```diff
@@ -1184,4 +1199,25 @@ export const ContentAnalysis = {
...
+   * rule supplied no message. Only rules from the enterprise-only
+   * DataLossPrevention policy carry one, so the label lives in enterprise.ftl
+   * and is unavailable in other builds.
+   *
+   * @param {string} aBody
+   * @param {string} aRuleMessage
+   */
+  _appendRuleMessage(aBody, aRuleMessage) {
+    const ruleMessage = aRuleMessage?.trim();
+    if (!ruleMessage || !AppConstants.MOZ_ENTERPRISE) {
```
gregstoll, line 1215: (FYI I double-checked and "" is evaluated as false in Javascript, so hooray :-))

toolkit/components/contentanalysis/tests/browser/browser_content_analysis_block_dialog.js
gregstoll: Does this pass? `String.includes()` seems to be case sensitive but the message starts with "You're".

toolkit/components/contentanalysis/tests/wasm/src/lib.rs
gregstoll: Since this is only used for tests, I think it would be more appropriate to return an error (somehow?) if an entry is malformed.

toolkit/components/contentanalysis/tests/xpcshell/test_wasm_backend.js
gregstoll: This comment is confusing because this function only uploads the file once

toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl
gregstoll, line 111: hypernit: maybe add a ":" at the end of this string?

gregstoll (approved):
just a few small things, thanks!

toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl
jonathanmendez: I think I'll leave it off because a) it works _ok_ for now but b) the future styling from our UX designer uses other aspects to separate headers instead of punctuation as seen by 
image

toolkit/components/contentanalysis/tests/xpcshell/test_wasm_backend.js
jonathanmendez: Oof, missed that when originally introduced. I can just delete this.

toolkit/components/contentanalysis/tests/browser/browser_content_analysis_block_dialog.js
jonathanmendez: This is actually testing against the _body_ which has "Under your organization's data protection policies, **you're not permitted to paste this content**." (The _title_ has just the part starting with "You're not permitted..").

That's confusing, so I'll see if I can update the test to use ftl values instead of hardcoding a copy of the text.

PR #1375, opened 2026-09-01
Bug 2051904 - Report disk encryption in device posture and about:support

gcp:
### Description

Bugzilla: [Bug 2051904](https://bugzilla.mozilla.org/show_bug.cgi?id=2051904)

Add disk-encryption status to enterprise device posture:

```json
{
  "diskEncryption": {
    "status": "full",
    "method": "dm-crypt"
  }
}
```

The first patch moves `run_command_bounded` out of the EDR checker and into a shared process module. Disk-encryption detection reuses this machinery to run platform probes with bounded execution time and output.

Firefox checks the operating-system volume, other mounted fixed volumes, and active Linux swap. Linux mounts at `/boot`, `/boot/efi`, and `/efi` are excluded; data mounts beneath those paths still count. This is not a root-only or remotely attested signal: an encrypted `/` with an unencrypted, separately mounted `/home` is reported as `partial`.

The same result is shown in the Security Software section of `about:support` for troubleshooting.

#### Posture values

| Field | Value | Meaning |
|---|---|---|
| `status` | `full` | Every relevant volume was inspected and is encrypted |
| `status` | `enabled` | The operating-system volume is encrypted and no plaintext volume was found, but inspection of another relevant volume or encryption mapping was incomplete |
| `status` | `partial` | The operating-system volume is encrypted, but another fixed volume or active swap is confirmed unencrypted; this takes precedence if a different secondary volume is converting |
| `status` | `disabled` | The operating-system volume is not encrypted |
| `status` | `in-progress` | Encryption or decryption is in progress and no secondary volume is confirmed unencrypted |
| `status` | `unknown` | Encryption of the operating-system volume cannot be determined |
| `method` | `filevault` | FileVault on macOS |
| `method` | `bitlocker` | BitLocker on Windows |
| `method` | `dm-crypt` | dm-crypt on Linux |
| `method` | `zfs` | Native ZFS encryption |
| `method` | `null` | No method could be determined |

#### Platform coverage

| Platform | Detection |
|---|---|
| Windows | Checks BitLocker protection for the operating-system volume and other mounted fixed volumes |
| macOS | Checks FileVault for the operating-system volume and encryption of other mounted internal volumes |
| Linux | Walks device-mapper stacks to detect dm-crypt on the root filesystem, other mounted fixed filesystems, and active swap |
| Linux/Btrfs | Resolves a Btrfs filesystem to its member devices and checks the storage stack beneath every member |
| Linux/ZFS | Detects native ZFS encryption or checks every pool device for dm-crypt |

Btrfs is not reported as an encryption method because Btrfs does not provide native filesystem encryption. An encrypted Btrfs filesystem is reported as `dm-crypt` when its member devices are protected by dm-crypt.

Detection runs off the main thread. Concurrent callers share a scan, and results are cached to limit repeated platform probes.

---

### Testing

- [x] Added tests
- [ ] Manual testing performed

Tests cover status aggregation, dm-crypt device stacks, multi-device Btrfs, native and dm-crypt-backed ZFS, Linux swap, FileVault, BitLocker, the JavaScript API, device-posture reporting, and `about:support`.


toolkit/components/felt/rust/src/disk_encryption.rs
```diff
@@ -0,0 +1,423 @@
...
+                return Err(NS_ERROR_NOT_AVAILABLE);
+            }
+            running.waiting.push(callback);
+            return Ok(());
+        }
+        *sweep = Some(Sweep {
+            started: Instant::now(),
+            waiting: vec![callback],
+        });
+        drop(sweep);
```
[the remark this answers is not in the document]
gcp: Leaving `SWEEP` populated is deliberate. Clearing it would start a second sweep while the wedged worker is still alive; when that worker eventually returns, its `done()` takes whichever sweep is current and delivers a stale result to the new callers, or drains their waiters before the new task finishes. Failing fast to `unknown` avoids that race, and the comment above `WEDGED_SWEEP_TIMEOUT` documents the trade-off. Callbacks queued behind the wedged sweep are released when it completes.

browser/base/content/test/about/browser_aboutSupport.js
```diff
@@ -75,6 +75,117 @@ add_task(async function () {
   );
 });
 
+add_task(
+  { skip_if: () => !AppConstants.MOZ_ENTERPRISE },
+  async function test_disk_encryption_row() {
```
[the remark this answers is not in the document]
gcp: `AppConstants` is provided to every browser mochitest scope by the harness (`testing/mochitest/browser-test.js` imports it and the test scope closes over it; the harness docs use `skip_if: () => !AppConstants.DEBUG` as their example). Other tests in this directory, e.g. `browser_aboutCertError.js`, rely on it the same way, and this test passes locally.

gcp:
**Accepted limitations**
- The signal is self-reported. A user who controls the Firefox install or profile can fabricate any status. Only remote attestation would change this.
- `enabled` means "boot volume encrypted, inspection incomplete or unattested". It covers plain dm-crypt mappings Firefox cannot verify without root, sweeps that ran out of time, and volumes it could not classify. Consoles that need proof should treat only `full` as compliant and `enabled` as needing attention.
- `unknown` is cheap to induce and is the expected result for snap/flatpak installs, whose root is a loop device. Alerting on transitions into `unknown` and on fleet-anomalous `unknown` is more useful than treating it as a fixed verdict.
- Results are point-in-time and cached (10 min for `full`/`partial`/`disabled`, 60 s otherwise). A change right after a sweep is reported up to a TTL plus one poll later.
- Root or admin can hide a data volume: on Linux by mounting it at exactly `/boot` or `/efi`, on Windows by removing its mount point (unmounted fixed volumes are skipped because the EFI and recovery partitions are always unmounted and unencrypted), on macOS possibly by mounting under `/System/Volumes/` (unverified under the sealed system volume). Same trust class as the first item.
- On Windows the BitLocker state comes from Explorer's `System.Volume.BitLockerProtection` shell property, which is readable without elevation. Per-user COM class hijacking of the shell could spoof it, again equivalent to controlling the process.
- `plist` 1.7.0 to 1.10.0 deltas are audited `safe-to-run` only; a `safe-to-deploy` delta audit should be filed so the `[policy.felt]` exemption can be removed.

gcp:
Verified on macOS and Linux, need to test on Windows still.

browser/components/BrowserGlue.sys.mjs
```diff
@@ -816,14 +816,16 @@ BrowserGlue.prototype = {
...
         );
-        Troubleshoot.snapshot().then(snapshotData => {
-          // for privacy we remove crash IDs and all preferences (but bug 1091944
-          // exists to expose prefs once we are confident of privacy implications)
-          delete snapshotData.crashes;
-          delete snapshotData.modifiedPreferences;
-          delete snapshotData.printingPreferences;
-          channel.send(snapshotData, target);
-        });
+        Troubleshoot.snapshot({ includeEnterpriseSecurity: false }).then(
```
lissyx, line 754: Is this a change that needs to live on Thunderbird ?

toolkit/components/felt/rust/FeltDiskEncryptionWin.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "mozilla/mscom/Utils.h"
+#include "nsCOMPtr.h"
+
+// Explorer's BitLocker property. Older SDKs omit it from propkey.h.
+static const PROPERTYKEY kVolumeBitLockerProtection = {
+    {0x2d15a9a1,
+     0xa556,
+     0x4189,
+     {0x91, 0xad, 0x02, 0x74, 0x58, 0xf1, 0x1a, 0x07}},
+    1717};
```
lissyx: For further handling, it would be nice to know precisely where this is coming from

toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl
```diff
@@ -12,6 +12,54 @@ app-basics-device-id = Device ID
...
+# Variables:
+#   $method (String): "filevault", "bitlocker", "dm-crypt" or "zfs".
+security-software-disk-encryption-full =
+    { $method ->
+        [bitlocker] Enabled (BitLocker)
+        [dm-crypt] Enabled (dm-crypt)
+        [filevault] Enabled (FileVault)
+        [zfs] Enabled (ZFS)
+       *[other] Enabled
+    }
```
lissyx: Is this fine by l10n people? I'm never sure with that construct :(

lissyx: (I'm not sure also having a hardcoded list this way is a good thing future proof)

toolkit/modules/tests/xpcshell/test_DiskEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+const { AppConstants } = ChromeUtils.importESModule(
+  "resource://gre/modules/AppConstants.sys.mjs"
+);
+const { MockRegistrar } = ChromeUtils.importESModule(
+  "resource://testing-common/MockRegistrar.sys.mjs"
+);
+
+const enterpriseOnly = () => ({ skip_if: () => !AppConstants.MOZ_ENTERPRISE });
```
lissyx: Can we skip at the manifest level rather than within the test ? Logs will be easier to read

toolkit/modules/tests/xpcshell/test_DiskEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+  "full",
+  "enabled",
+  "partial",
+  "disabled",
+  "in-progress",
+  "unknown",
+];
+
+let DiskEncryption;
+if (AppConstants.MOZ_ENTERPRISE) {
```
lissyx: We can remove this if we `run-if = ["enterprise"]` instead

toolkit/modules/tests/xpcshell/xpcshell.toml
```diff
@@ -50,6 +50,8 @@ skip-if = [
 ["test_DeferredTask_window.js"]
 support-files = ["test_DeferredTask_window/*"]
 
+["test_DiskEncryption.js"]
```
lissyx, line 53:
```suggestion
["test_DiskEncryption.js"]
run-if = ["enterprise"]
```

toolkit/modules/moz.build
```diff
@@ -229,6 +229,7 @@ MOZ_SRC_FILES += [
 
 if CONFIG["MOZ_ENTERPRISE"]:
     EXTRA_JS_MODULES.enterprise += [
+        "DiskEncryption.sys.mjs",
         "EdrDetection.sys.mjs",
         "MachineId.sys.mjs",
```
lissyx, line 233: We should have placed those under `toolkit/components/enterprise/modules` but it's likely a bit too late. Worth a follow up to refactor this cleanly later ?

toolkit/modules/DiskEncryption.sys.mjs
```diff
@@ -0,0 +1,82 @@
...
+            onComplete(status, method) {
+              if (!VALID_STATUSES.has(status) || !VALID_METHODS.has(method)) {
+                finish(UNKNOWN);
+                return;
+              }
+              finish({ status, method });
+            },
+          });
+      } catch (e) {
+        finish(UNKNOWN);
```
lissyx, line 79: Worth logging the  error to `console` ?

toolkit/components/felt/rust/src/disk_encryption.rs
```diff
@@ -0,0 +1,435 @@
...
+    }
+}
+
+/// Aggregates the boot volume and other mounted fixed volumes. `None` means the
+/// backend could not enumerate all other volumes.
+pub(crate) fn aggregate(boot: VolumeState, others: Option<&[VolumeState]>) -> EncryptionStatus {
+    match boot {
+        VolumeState::Unknown => EncryptionStatus::Unknown,
+        VolumeState::Converting => EncryptionStatus::InProgress,
+        VolumeState::Unencrypted => EncryptionStatus::Disabled,
```
lissyx, line 78: Would it mean a `/boot` not encrypted on linux would make this report disabled even if `/` is encrypted ?

toolkit/components/felt/rust/src/disk_encryption_win.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+/// The mounted fixed volumes. `complete` is false when a volume could not be
+/// examined and may be missing from `list`.
+struct Volumes {
+    list: Vec<Volume>,
+    complete: bool,
+}
+
+/// Shell property reads are synchronous, so the deadline is checked between calls.
```
lissyx, line 43: Where are the shell calls, I do not see them ?

toolkit/components/felt/rust/src/disk_encryption.rs
```diff
@@ -0,0 +1,435 @@
...
+
+impl EncryptionStatus {
+    pub fn as_str(self) -> &'static str {
+        match self {
+            EncryptionStatus::Full => "full",
+            EncryptionStatus::Enabled => "enabled",
+            EncryptionStatus::Partial => "partial",
+            EncryptionStatus::Disabled => "disabled",
+            EncryptionStatus::InProgress => "in-progress",
+            EncryptionStatus::Unknown => "unknown",
```
lissyx, line 51: Could those strings may be defined in the idl and re-used here and everywhere? Or is that only doable with WebIDL ? I never know

toolkit/components/felt/rust/src/disk_encryption.rs
lissyx: I think the test cases later answer this.

lissyx (approved):
LGTM, with relevants comments addressed

toolkit/components/felt/rust/src/process.rs
```diff
@@ -0,0 +1,135 @@
...
+            Ok(None) => {
+                if start.elapsed() >= budget {
+                    let _ = child.kill();
+                    let _ = child.wait();
+                    return None;
+                }
+                thread::sleep(PROBE_POLL_INTERVAL);
+            }
+            Err(_) => {
+                // Dropping a Child neither kills nor reaps it.
```
jonathanmendez: nit: This comment feels out of place, especially since there's not a matching comment in the `Ok(None)` case. Reads as "this is why I added this code" rather than "this is an explanation for how the code works".

I think context is enough to understand that the intent is to kill/reap a child that didn't exit normally, so the comment could just be dropped entirely.

toolkit/components/felt/rust/src/disk_encryption.rs
```diff
@@ -0,0 +1,435 @@
...
+}
+
+static SWEEP: Mutex<Option<Sweep>> = Mutex::new(None);
+
+fn cache_is_usable(cache: &Option<(Instant, DiskEncryption)>, now: Instant) -> bool {
+    match cache {
+        Some((at, result)) => {
+            let ttl = if matches!(
+                result.status,
+                EncryptionStatus::Enabled
```
jonathanmendez, line 146: I'm guessing this is considered "unstable" for the case where one of the non-boot volumes is unknown, but the other scenarios that yield "Enabled" seem to all be "stable".
I'm assuming that since this is just a cache TTL you're leaning toward catching that one case, but maybe it's worth a small comment here since it looks odd that `Enabled` is yielding `UNKNOWN_CACHE_TTL`

toolkit/components/felt/rust/FeltDiskEncryptionWin.cpp
jonathanmendez: I wonder if this couldn't be done via `PSGetPropertyKeyFromName` with "System.Volume.BitLockerProtection" ([StackOverflow post](https://stackoverflow.com/a/47192128) with similar overall code). But if nothing else, yes some extra detailed clarification of this magic would be useful.

jonathanmendez (approved):
LGTM with comments addressed

toolkit/components/felt/rust/FeltDiskEncryptionWin.cpp
gcp: Implemented @jonathanmendez 's method.

toolkit/components/felt/rust/src/disk_encryption.rs
gcp: Explained this in the comment about what `UNKNOWN_CACHE_TTL` is for.

toolkit/components/felt/rust/src/process.rs
gcp: Claude just loves to add these silly comments about what things are not, missed this one :-/

toolkit/components/felt/rust/src/disk_encryption.rs
gcp: XPIDL only has integer constants. WebIDL can do this, but this is XPIDL, so I guess nope.

toolkit/components/felt/rust/src/disk_encryption_win.rs
gcp: `let boot = volume_state(&boot_volume.mount_path);`

ends up calling `felt_read_bitlocker_protection` which calls https://github.com/mozilla/enterprise-firefox/pull/1375/changes#diff-237c6feb2452fb9dc439bdabcfe07ed72da38c9fa734f46d6624c40db68e58d9R24

which are Windows Shell COM API functions.

toolkit/components/felt/rust/src/disk_encryption.rs
gcp: "boot" is `/` here, not `/boot, /boot/efi, and /efi` which are excluded because they can't be encrypted (AFAIK).

toolkit/modules/moz.build
gcp: Probably, simple enough.

toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl
gcp: It should work but I simplified it a ton.

toolkit/components/felt/rust/moz.build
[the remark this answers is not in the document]
gcp: BS

PR #1376, opened 2026-09-01
Bug 2068057 - Make console URL configurable at build time

lissyx:
### Description 

Bugzilla: Bug-2068057

### DIFF
--- a/browser/moz.configure
+++ b/browser/moz.configure
@@ -75,6 +75,42 @@ set_define("MOZ_ENTERPRISE", enable_enterprise)
 # Clear the default value to avoid mozilla.com url compiled into product
 imply_option("--with-crashreporter-url", "", when="--enable-enterprise")
 
+default_enterprise_console = dependable(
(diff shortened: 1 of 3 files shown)


browser/moz.configure
```diff
@@ -75,6 +75,42 @@ set_define("MOZ_ENTERPRISE", enable_enterprise)
 # Clear the default value to avoid mozilla.com url compiled into product
 imply_option("--with-crashreporter-url", "", when="--enable-enterprise")
 
+default_enterprise_console = dependable(
+    "https://stage.fx-enterprise.nonprod.webservices.mozgcp.net"
+)
+
+option(
+    "--enable-enterprise-console",
```
mkaply: I don't think the name makes sense. We're setting a value, not enabling the console.

Maybe something like:

--with-enterprise-console-url

Because the console is already enabled.

mkaply (approved):

PR #1383, opened 2026-09-01
Bug 2068176 - Add macOS 27 to the list of known macOS versions, to return the OS name in the posture

julienw:
### Description

Bugzilla: Bug-2068176

macOS 27 is newly released and is named Golden Gate. So that the code name is posted in the posture, this patch adds this version to the array matched when the posture is sent.

Not sure I need comprehensive tests for this, but you'll tell me :-)


lissyx (approved):

julienw:
Thanks for the review!

PR #1422, opened 2026-09-07
Bug 2069825 - Anchor confirmation hint to enterprise badge

1rneh:
### Description 

Bugzilla: Bug-2069825

Updates the confirmation hint after sending  a tab to be anchored to the enterprise toolbar button instead of the fxa toolbar button.


lissyx (approved):
LGTM

PR #1427, opened 2026-09-07
Bug 2069904 - Skip content sandbox test on windows for access to felt.json

1rneh:
### Description 

Bugzilla: Bug-2069904

Skips the content sandbox 

`browser_content_sandbox_fs_tests.js` asserts that a web content process cannot read the felt storage. felt.json lives in UAppData rather than the profile, and the Windows content sandbox hardens the profile directory but not UAppData. Hence felt.json is currently readable from content on Windows.

This wraps the felt.json read assertion in a non-Windows check. It also corrects the `FeltStorage` docs comment to state that felt.json is stored in UAppData.

Follow-up work done in Bug-2069930.

### DIFF
--- a/security/sandbox/test/browser_content_sandbox_fs_tests.js
+++ b/security/sandbox/test/browser_content_sandbox_fs_tests.js
@@ -117,18 +117,23 @@ async function testFileAccessAllPlatforms() {
   }
 
   if (AppConstants.MOZ_ENTERPRISE) {
-    const uAppDataDir = GetUserAppDataDir();
-    const feltJson = uAppDataDir.clone();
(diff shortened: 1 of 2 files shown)


lissyx (approved):

lissyx:
I was looking at this test earlier today and I was telling myself we should go ask again bob or david on that to fix it correctly :)

1rneh:
> I was looking at this test earlier today and I was telling myself we should go ask again bob or david on that to fix it correctly :)

I filed Bug-2069930 to investigate address the issue for felt.json. That can also entails pining them to investigate a generic fix.

PR #1430, opened 2026-09-07
Bug 2061664 - Clean up the About dialog on enterprise builds

jporter-dev:
### Description 

Bugzilla: Bug-2061664

Removes consumer-oriented content from the About dialog on MOZ_ENTERPRISE builds, adds an enterprise-specific notice, and updates the release notes URL to the path for enterprise release notes.

> Note: the release notes path is not yet live, but based on convo with @bsmth we can go ahead and use the URL from the JIRA link in the bug since it will be live soon.

---

### Screenshots 

image

_Before_

image

_After_

---

### Testing 

* [x] Added tests


browser/base/content/aboutDialog.js
bsmth, line 105: Should we get rid of the update channel sentence, also? i.e., what does `default update channel` mean?

bsmth:
The "is experimental and may be unstable" sentence is only visible in the Nightly build, right?

bsmth:
> The "is experimental and may be unstable" sentence is only visible in the Nightly build, right?

Note that the string in Nightly mentions the release channel, but ours is `brand name`:

image

browser/base/content/aboutDialog.js
jporter-dev: Thanks for pointing that out, Brian! Replying to both of your comments here:

> Should we get rid of the update channel sentence?

[This string](build: https://searchfox.org/enterprise-main/source/browser/locales/en-US/browser/aboutDialog.ftl#70-72) uses the `channel` from `app.update.channel` / `MOZ_UPDATE_CHANNEL`, which falls back to `default` when not set here: https://searchfox.org/enterprise-main/source/build/moz.configure/init.configure#1193-1197. This is also only shown for non-release and non-packaged app installs: https://searchfox.org/enterprise-main/source/browser/base/content/aboutDialog.js#112-117, shown below on my Windows release install:

image

This shows `nightly-enterprise` on my macOS nightly install: 

image

Since this does show a proper channel identifier in nightly, and is not shown in release, I think we should probably keep it since it's probably useful for QA.

---

> The "is experimental and may be unstable" sentence is only visible in the Nightly build, right?

Yes, this is driven by the version string (e.g. `155.0a1` will show it, `155.0` will not): https://searchfox.org/enterprise-main/source/browser/base/content/aboutDialog.js#48-79

bsmth: All LGTM but one comment that it would be good to figure out if it's easy to have:

```diff
+The Nightly release channel is experimental and may be unstable
-Firefox Enterprise is experimental and may be unstable
```

This is really more accurate as opposed to the product, i.e., we don't say "Firefox is experimental and may be unstable", we limit this to the release channel, and that's true for Nightly (and Enterprise Nightly).

jporter-dev: Yeah, I think the main problem here is that we use "Firefox Enterprise" as our `brand name` on our nightly builds instead of `Firefox Enterprise Nightly` or something along those lines, which would read a bit better. 

I'll look into fixing this up, thanks!

jporter-dev: Added this as an enterprise-specific string in place of the usual experimental warning

browser/branding/enterprise/pref/firefox-branding.js
```diff
@@ -20,17 +20,17 @@ pref("app.update.promptWaitTime", 691200);
...
-  pref("app.releaseNotesURL.aboutDialog", "https://www.firefox.com/%LOCALE%/firefox/%VERSION%/releasenotes/?utm_source=firefox-browser&utm_medium=firefox-desktop&utm_campaign=about-dialog");
+  pref("app.releaseNotesURL.aboutDialog", "https://www.firefox.com/%LOCALE%/firefox/enterprise/%VERSION%/releasenotes/?utm_source=firefox-browser&utm_medium=firefox-desktop&utm_campaign=about-dialog");
 #else
   pref("app.update.url.manual", "https://www.mozilla.org/%LOCALE%/firefox/new?reason=manual-update");
   pref("app.update.url.details", "https://www.mozilla.org/%LOCALE%/firefox/notes");
   pref("app.releaseNotesURL", "https://www.firefox.com/%LOCALE%/firefox/%VERSION%/releasenotes/?utm_source=firefox-browser&utm_medium=firefox-desktop&utm_campaign=whatsnew");
-  pref("app.releaseNotesURL.aboutDialog", "https://www.firefox.com/%LOCALE%/firefox/%VERSION%/releasenotes/?utm_source=firefox-browser&utm_medium=firefox-desktop&utm_campaign=about-dialog");
+  pref("app.releaseNotesURL.aboutDialog", "https://www.firefox.com/%LOCALE%/firefox/enterprise/%VERSION%/releasenotes/");
```
1rneh: FYI I'm working on removing the unused `MOZ_UPDATE_CHANNEL == beta`/ `MOZ_ESR` branches in `browser/branding/enterprise/firefox-branding.js` as part of Bug-2068879. And I was wondering if we already have more enterprise specific urls for the other ones?

toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl
```diff
@@ -5,6 +5,13 @@
 # Access Connector is an Enterprise feature name and must not be translated.
 -enterprise-feature-access-connector = Access Connector
 
+# Shown in the About dialog on enterprise builds.
+aboutdialog-managed-notice = Your browser is being managed by your organization.
+
+# Shown in the About dialog on enterprise nightly builds.
+# "Nightly" is a release channel name and must not be translated.
+aboutdialog-nightly-channel-warning = The Nightly release channel is experimental and may be unstable.
```
1rneh: Let's not hardcode a brand string here. How about we introduce `-brand-short-name-nightly-channel = Firefox Enterprise Nightly` in https://searchfox.org/enterprise-main/source/browser/branding/enterprise/locales/en-US/brand.ftl#20 ?

browser/base/content/aboutDialog.xhtml
```diff
@@ -116,9 +125,16 @@
                        data-l10n-attrs="{&quot;channel&quot;: &quot;&quot;}">
             <label id="currentChannel" data-l10n-name="current-channel"/>
           </description>
+#endif
+#ifdef MOZ_ENTERPRISE
+          <description class="text-blurb" id="managedDesc" data-l10n-id="aboutdialog-managed-notice"/>
 #endif
           <vbox id="experimental" hidden="true">
+#ifdef MOZ_ENTERPRISE
+            <description class="text-blurb" id="warningDesc" data-l10n-id="aboutdialog-nightly-channel-warning"></description>
```
1rneh: `nit`
```suggestion
            <description class="text-blurb" id="warningDesc" data-l10n-id="warningDesc-version-enterprise-nightly"></description>
```

1rneh (changes requested):

browser/base/content/aboutDialog.xhtml
jporter-dev: Would love to keep this name consistent, but unfortunately it fails fluent lint :( the existing `warningDesc-version` is exempted in https://searchfox.org/enterprise-main/source/tools/lint/fluent-lint/exclusions.yml#50

```
  12:1  error  Identifiers may only contain lowercase characters and - (ID: warningDesc-version-enterprise-nightly)  ID01 (fluent-lint)
```

I can update it to `warningdesc-version-enterprise-nightly` to get as close as possible, but prefixing with `aboutdialog` makes it clearer where it's used when looking at `enterprise.ftl`

toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl
jporter-dev: https://github.com/mozilla/enterprise-firefox/pull/1430#discussion_r3958809093

Brian suggested this version of the warning string, slotting in `Firefox Enterprise Nightly` into it sounds a little weird. Do you think we should include the brand name in there, or just keep it as `Nightly`? 

`The Firefox Enterprise Nightly release channel is experimental and may be unstable`
vs
`The Nightly release channel is experimental and may be unstable`

Alternatively, I just noticed the message that says `You are currently on the nightly update channel` - we could possibly pass that in and say:

`The nightly-enterprise release channel is expierimental and may be unstable`

jporter-dev: Or we go back to the original string and just fix the brand for it and say `Firefox Enterprise Nightly is experimental and may be unstable` 😅

browser/branding/enterprise/pref/firefox-branding.js
jporter-dev: Not that I'm aware of, the release notes change is tracked [here](https://mozilla-hub.atlassian.net/browse/WT-1590). It does look like the defined `app.update.url.details` redirects to the `app.releaseNotesURL`. `app.update.url.manual` just lands me at the Firefox homepage..

Since you're removing the beta/esr branches, I'll just leave those two branches alone and just change the paths below them.

toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl
1rneh: Thanks for laying out all the options :)

I don't really have a strong preference. Any case we go (the one you updated to is fine with me) we should add that brand string in `brand.ftl` for even if it's "Nightly".

browser/base/content/aboutDialog.xhtml
1rneh: > I can update it to warningdesc-version-enterprise-nightly to get as close as possible, but prefixing with aboutdialog makes it clearer where it's used when looking at enterprise.ftl

Right, but with the comment you added it's also really clear now.

1rneh (approved):
Thank you!

PR #1433, opened 2026-09-08
Bug 2071363 - Enterprise: end-to-end GitHub dispatch and monitoring

lissyx:
Bugzilla: Bug-2071363

Trigger Enterprise console's End-to-end tests from TaskCluster, report the status. Should mirror the logs as well.

### DIFF
--- a/taskcluster/gecko_taskgraph/transforms/task.py
+++ b/taskcluster/gecko_taskgraph/transforms/task.py
@@ -2067,6 +2067,26 @@ def setup_raptor(config, tasks):
 
 
 @transforms.add
+def setup_enterprise_end2end(config, tasks):
+    """Add options that are specific to enterprise_end2end jobs (identified by suite=enterprise_end2end).
+
+    This variant uses a separate set of transforms for manipulating the tests at the
+    task-level. Currently only used for setting the taskcluster proxy setting and
+    the scopes required for enterprise end2end github api secrets.
+    """
+    from gecko_taskgraph.transforms.test.enterprise import (
+        task_transforms as enterprise_end2end_transforms,
+    )
+
+    for task in tasks:
+        if task.get("extra", {}).get("suite", "") != "enterprise-end2end":
+            yield task
+            continue
+
+        yield from enterprise_end2end_transforms(config, [task])
+
+
+@transforms.add
 def task_name_from_label(config, tasks):
     for task in tasks:
         taskname = task.pop("name", None)
--- a/taskcluster/gecko_taskgraph/transforms/test/enterprise.py
+++ b/taskcluster/gecko_taskgraph/transforms/test/enterprise.py
@@ -1,0 +1,43 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from taskgraph.transforms.base import TransformSequence
+
+task_transforms = TransformSequence()
+
+
+@task_transforms.add
+def force_optimization(config, tasks):
+    for task in tasks:
+        task["optimization"] = {"never": None}
+        yield task
+
+
+@task_transforms.add
+def force_test_manifests(config, tasks):
+    for task in tasks:
+        task["attributes"]["test_manifests"] = ["empty.toml"]
+        yield task
+
+
+@task_transforms.add
+def add_scopes_and_proxy(config, tasks):
+    for task in tasks:
+        task.setdefault("worker", {})["taskcluster-proxy"] = True
+        task.setdefault("scopes", []).append(
+            "secrets:get:project/enterprise/level-{level}/enterprise-console-backend-apitoken"
+        )
+        yield task
+
+
+@task_transforms.add
+def add_upstream_tasks(config, tasks):
+    for task in tasks:
+        task["worker"].setdefault("env", {})["UPSTREAM_TASKIDS"] = {
+            # We only want signing related tasks here, not build (used by mac builds for signing artifact resolution)
+            "task-reference": " ".join([
+                f"<{dep}>" for dep in task["dependencies"] if ("build" in dep)
+            ])
+        }
+        yield task
--- a/taskcluster/gecko_taskgraph/transforms/test/other.py
+++ b/taskcluster/gecko_taskgraph/transforms/test/other.py
@@ -214,13 +214,14 @@ def set_download_symbols(config, tasks):
     on demand for everything else. ASAN builds shouldn't download
     symbols since they don't product symbol zips see bug 1283879"""
     for task in tasks:
-        if task["test-platform"].split("/")[-1] == "debug":
-            task["mozharness"]["download-symbols"] = True
-        elif "asan" in task["build-platform"] or "tsan" in task["build-platform"]:
-            if "download-symbols" in task["mozharness"]:
-                del task["mozharness"]["download-symbols"]
-        else:
-            task["mozharness"]["download-symbols"] = "ondemand"
+        if task["suite"] != "enterprise-end2end":
+            if task["test-platform"].split("/")[-1] == "debug":
+                task["mozharness"]["download-symbols"] = True
+            elif "asan" in task["build-platform"] or "tsan" in task["build-platform"]:
+                if "download-symbols" in task["mozharness"]:
+                    del task["mozharness"]["download-symbols"]
+            else:
+                task["mozharness"]["download-symbols"] = "ondemand"
         yield task
 
 
@@ -818,6 +819,7 @@ def ensure_spi_disabled_on_all_but_spi(config, tasks):
         has_no_setpref = (
             "gtest",
             "cppunit",
+            "enterprise-end2end",
             "jittest",
             "junit",
             "raptor",
@@ -1064,6 +1066,7 @@ def enable_webrender(config, tasks):
             "gtest",
             "jittest",
             "raptor",
+            "enterprise-end2end",
         ]:
             extra_options.append("--setpref=layers.d3d11.enable-blacklist=false")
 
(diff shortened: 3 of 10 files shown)


testing/mozharness/scripts/enterprise-end2end.py
```diff
@@ -0,0 +1,250 @@
...
+            "gecko_log": dirs["abs_blob_upload_dir"],
+            "this_chunk": self.config.get("this_chunk", 1),
+            "total_chunks": self.config.get("total_chunks", 1),
+            "repo": self.config.get("repo"),
+            "owner": self.config.get("owner"),
+            "workflow_id": "EMPTY",
+        }
+
+        # python = self.query_python_path("python")
+        # cmd = [python, "-u", os.path.join(dirs["abs_enterprise_end2end_dir"], "runtests.py")]
```
jonathanmendez: Leftover?
```suggestion
```

testing/mozharness/scripts/enterprise-end2end.py
```diff
@@ -0,0 +1,250 @@
...
+            # Make sure that the logging directory exists
+            self.fatal("Could not create blobber upload directory")
+
+        env = {}
+        env["MOZ_UPLOAD_DIR"] = self.query_abs_dirs()["abs_blob_upload_dir"]
+
+        if not os.path.isdir(env["MOZ_UPLOAD_DIR"]):
+            self.mkdir_p(env["MOZ_UPLOAD_DIR"])
+
+        env = self.query_env(partial_env=env)
```
jonathanmendez: It looks like env gets built but never used (except for `env["MOZ_UPLOAD_DIR"]` within these lines which could be turned into a local)? But I'm not sure if `self.query_env` has some side effect..

testing/mozharness/scripts/enterprise-end2end.py
```diff
@@ -0,0 +1,250 @@
...
+        error_summary_file = os.path.join(
+            dirs["abs_blob_upload_dir"], "enterprise_end2end_errorsummary.log"
+        )
+
+        config_fmt_args = {
+            "raw_log_file": raw_log_file,
+            "error_summary_file": error_summary_file,
+            "gecko_log": dirs["abs_blob_upload_dir"],
+            "this_chunk": self.config.get("this_chunk", 1),
+            "total_chunks": self.config.get("total_chunks", 1),
```
jonathanmendez: Looks like only `repo`, `owner`, and `workflow_id` are used?
```suggestion
```

testing/mozharness/scripts/enterprise-end2end.py
```diff
@@ -0,0 +1,250 @@
...
+        if status_code != 200:
+            self.log(f"GitHub Workflow list failed: {data.get('message')}")
+            raise ValueError(f"github workflow list failed: {status_code}")
+        self.log(f"GitHub Workflows URL: {workflows_url}: returned {status_code}")
+
+        config_fmt_args["workflow_id"] = next(
+            w
+            for w in data["workflows"]
+            if w["state"] == "active" and w["name"] == "Test"
+        )["id"]
```
jonathanmendez: Instead of querying workflows and filtering by name, the [single-workflow endpoint accepts the workflow file name as an option](https://docs.github.com/en/rest/actions/workflows?apiVersion=2026-03-10#get-a-workflow).
```suggestion
        workflow_url = f"{base_url}/workflows/test.yaml"
        self.log(f"GitHub Workflow URL: {workflow_url}")
        status_code, data = self.github_api(workflow_url, headers)
        if status_code != 200:
            self.log(f"GitHub Workflow get failed: {data.get('message')}")
            raise ValueError(f"github workflow get failed: {status_code}")
        self.log(f"GitHub Workflow URL: {workflow_url}: returned {status_code}")

        config_fmt_args["workflow_id"] = data["id"]
```

testing/mozharness/scripts/enterprise-end2end.py
```diff
@@ -0,0 +1,250 @@
...
+            "ref": "main",
+            "inputs": {
+                "fxe_task_id": f"{build_task}",
+                "affected_scopes": json.dumps('["e2e"]'),
+            },
+        }
+
+        self.log(
+            f"GitHub Workflow Dispatches: {dispatches_url}: returned {status_code}"
+        )
```
jonathanmendez: This looks leftover; it looks like the real version is below (appropriately after the API call for dispatches).
```suggestion
```

testing/mozharness/scripts/enterprise-end2end.py
```diff
@@ -0,0 +1,250 @@
...
+        )["id"]
+
+        dispatches_url = (
+            base_url + "/workflows/%(workflow_id)s/dispatches" % config_fmt_args
+        )
+        payload = {
+            "ref": "main",
+            "inputs": {
+                "fxe_task_id": f"{build_task}",
+                "affected_scopes": json.dumps('["e2e"]'),
```
jonathanmendez: The entire payload is going to get passed to `json.dumps`, is extra this one needed too?

taskcluster/gecko_taskgraph/transforms/test/enterprise.py
```diff
@@ -0,0 +1,43 @@
...
+            "secrets:get:project/enterprise/level-{level}/enterprise-console-backend-apitoken"
+        )
+        yield task
+
+
+@task_transforms.add
+def add_upstream_tasks(config, tasks):
+    for task in tasks:
+        task["worker"].setdefault("env", {})["UPSTREAM_TASKIDS"] = {
+            # We only want signing related tasks here, not build (used by mac builds for signing artifact resolution)
```
jonathanmendez: This comment doesn't seem to match the code.. is it just a stale/copied comment?

taskcluster/gecko_taskgraph/transforms/task.py
```diff
@@ -2066,6 +2066,26 @@ def setup_raptor(config, tasks):
         yield from raptor_transforms(config, [task])
 
 
+@transforms.add
+def setup_enterprise_end2end(config, tasks):
+    """Add options that are specific to enterprise_end2end jobs (identified by suite=enterprise_end2end).
```
jonathanmendez: nit:
```suggestion
    """Add options that are specific to enterprise_end2end jobs (identified by suite=enterprise-end2end).
```

taskcluster/gecko_taskgraph/transforms/test/other.py
```diff
@@ -214,13 +214,14 @@ def set_download_symbols(config, tasks):
...
     symbols since they don't product symbol zips see bug 1283879"""
     for task in tasks:
-        if task["test-platform"].split("/")[-1] == "debug":
-            task["mozharness"]["download-symbols"] = True
-        elif "asan" in task["build-platform"] or "tsan" in task["build-platform"]:
-            if "download-symbols" in task["mozharness"]:
-                del task["mozharness"]["download-symbols"]
-        else:
-            task["mozharness"]["download-symbols"] = "ondemand"
+        if task["suite"] != "enterprise-end2end":
```
jonathanmendez: It feels a bit odd to have to check specifically for this suite here. If there's not some other way to avoid this, could you add to the comment above at least?

Separate question that would potentially make this unnecessary: do we want/need both opt and debug runs of the end2end tests?

testing/mozbase/moztest/moztest/resolve.py
```diff
@@ -562,6 +568,7 @@ def WebglSuite(name):
     ("browser-chrome", "screenshots"): "mochitest-browser-screenshots",
     ("browser-chrome", "translations"): "mochitest-browser-translations",
     ("chrome", "gpu"): "mochitest-chrome-gpu",
+    ("enterprise-end2end", "integration"): "enterprise-end2end",
```
jonathanmendez: Is this needed/used? I don't know where a subsuite would come into play here.

testing/mozharness/scripts/enterprise-end2end.py
```diff
@@ -0,0 +1,250 @@
...
+            },
+        }
+
+        self.log(
+            f"GitHub Workflow Dispatches: {dispatches_url}: returned {status_code}"
+        )
+        status_code, data = self.github_api(dispatches_url, headers, payload=payload)
```
jonathanmendez: It appears that the workflow itself is always trying to download `target.tar.xz` which only exists for Linux. I'm not sure if the platform should become a payload argument or if the workflow should be figuring it out from the task id/task.

taskcluster/kinds/test/enterprise-end2end.yml
```diff
@@ -0,0 +1,21 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+enterprise-end2end:
+    run-on-projects: built-projects
+    tier: default
+    attributes:
+        code-review: true
+    worker-type: default
```
jonathanmendez: Since this just invokes the github action and polls, maybe it should specify a specific/cheap worker type.

jonathanmendez (changes requested):

testing/mozharness/scripts/enterprise-end2end.py
lissyx: Yes, that's how the workflow is designed upstream.

taskcluster/gecko_taskgraph/transforms/test/other.py
lissyx: Likely we want, because of race that can happen on opt vs debug, and also debug has more assertions that can help find issues

testing/mozbase/moztest/moztest/resolve.py
lissyx: it's used when computing things, I need to remember precisely where

testing/mozharness/scripts/enterprise-end2end.py
lissyx: This is a limitation of the current workflow: its only supporting linux right now :(

lissyx: Console CI lacks macOS and Windows runners

taskcluster/kinds/test/enterprise-end2end.yml
lissyx: I think it's already defaulting to `enterprise-t` default which should be rather cheap. I have pending https://github.com/mozilla/enterprise-firefox/pull/1348 for using cheaper ones especially for `try-status` mirroring, but it requires deployment on fxci-config side so I cannot yet make use of it

PR #1434, opened 2026-09-08
Bug 2040659 - Name the browser run that reports a posture

gcp:
Device posture now includes a random `clientSessionId` that stays constant for one managed browser run and changes on relaunch. This lets the console detect restarts even when no other posture values change.

The ID is kept only in `DevicePosture` module state, so it works with telemetry disabled and is never persisted. Felt waits for outstanding refreshes before replacing it and clearing the monitor's baseline, preventing stale posture from being reused.

Tested locally: `test_felt_device_posture_session_id.py` and the existing posture and restart tests pass.

### DIFF
--- a/testing/enterprise/manifest-generic.toml
+++ b/testing/enterprise/manifest-generic.toml
@@ -18,6 +18,8 @@ subsuite = "enterprise"
 
 ["test_felt_device_posture_refresh_rejected.py"]
 
+["test_felt_device_posture_session_id.py"]
+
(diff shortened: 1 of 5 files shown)


gcp:
Try (marionette-enterprise): https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=cfc0e4f71fc794dd86a49f060497db5dcbc9b60b

lissyx (approved):

PR #1442, opened 2026-09-08
Bug 2064561 - Check policies required at startup from felt and pass to browser via env vars

jonathanmendez:
### Description 

Bugzilla: Bug-2064561

Use felt to check for specific policies (from console) that normally get checked prior to policies setup (via checking the Windows registry for policies). When the policies are set, use environment variables to pass to the browser, which is updated to check for the env var in addition to checking the windows registry.

---

* Depends on: #1463 

---

### Testing 

* [x] Added tests
  * Required for third-party module blocking, where setting up an automated test would be unreasonable (if even possible). ~~However, initial manual testing shows that third-party module blocking isn't currently possible in Fx Enterprise anyway, which I'm starting investigation on.~~ *Fixed in Bug-2070625*


testing/enterprise/test_felt_browser_startup_policies.py
lissyx: I'm not sure this is going to behave well: some of the marionette harness we have here really assumes we only have one test class per file. you probably want to move this to a second file, or at least triple check it does what you want

testing/enterprise/test_felt_startup_policies.py
lissyx, line 18: Same question about the two classes

toolkit/components/felt/content/FeltProcessParent.sys.mjs
lissyx: This is going to fetch all policies and add them all into `this._startupPolicies`. It sounds overkill here. We probably do not have a choice but to keep the `getRemotePolicies()` call but we should explicitely retain only the policies we want to handle at startup. 

And likely this should be isolated somewhere outside of `FeltProcessParent` that should just be a consumer

lissyx (changes requested):

testing/enterprise/test_felt_browser_startup_policies.py
jonathanmendez: Thanks. I did see some discrepancies in the log that were probably ok but seems safest to separate them, which I've done.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jonathanmendez: I've moved this to a StartupPolicies class that limits what's stored to just the particular startup policies.

lissyx (approved):

PR #1443, opened 2026-09-08
Bug 2066101 - Refactor enterprise partner configs to limit GitHub requests

lissyx:
### Description 

Bugzilla: Bug-2066101

### DIFF
--- a/taskcluster/gecko_taskgraph/util/partners.py
+++ b/taskcluster/gecko_taskgraph/util/partners.py
@@ -643,17 +643,44 @@ def _pad_macos_attribution_code(attribution_string):
     return attribution_string
 
 
+# Cache key for the one enterprise manifest crawl every enterprise kind shares.
+# Deliberately not a kind name, so it can't be picked up by
+# get_partner_config_by_kind(). There is already caching handled by
+# get_partner_config_by_url() but it will still hit the network once for each
+# "kind" required to be populated, which can be reduced to only once for all
+# the kinds.
+ENTERPRISE_MANIFEST_FAKE_KIND = "enterprise-manifest"
+
+# The repack.cfg platform keys each enterprise kind turns into build platforms.
+ENTERPRISE_PLATFORM_MAPPINGS = {
+    "repackage-deb": {
+        "linux-x86_64": "linux64-enterprise-shippable",
+        "linux-aarch64": "linux64-aarch64-enterprise-shippable",
+    },
+    "repackage-msix": {"win64-aarch64": "win64-aarch64-enterprise-shippable"},
+    "repackage-msi": {"win64": "win64-enterprise-shippable"},
+    "enterprise-repack-repackage": {
+        "linux-x86_64": "linux64-enterprise-shippable",
+        "linux-aarch64": "linux64-aarch64-enterprise-shippable",
+        "mac": "macosx64-enterprise-shippable",
+        "win64": "win64-enterprise-shippable",
+        "win64-aarch64": "win64-aarch64-enterprise-shippable",
+    },
+    "enterprise-repack-mac-signing": {"mac": "macosx64-enterprise-shippable"},
+    "enterprise-repack-mac-notarization": {"mac": "macosx64-enterprise-shippable"},
+}
+
+
 def make_enterprise_repack_from_url(parameters, graph_config, kind, platform_mapping):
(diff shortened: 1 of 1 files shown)


taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -643,17 +644,43 @@ def _pad_macos_attribution_code(attribution_string):
...
 
 
+# Cache key for the one enterprise manifest crawl every enterprise kind shares.
+# Deliberately not a kind name, so it can't be picked up by
+# get_partner_config_by_kind().
+ENTERPRISE_MANIFEST_KIND = "enterprise-manifest"
+
+# The repack.cfg platform keys each enterprise kind turns into build platforms.
+# Declaration order matters: get_partner_config_by_kind() resolves a kind to the
+# first entry that is a prefix of it.
```
jonathanmendez: If I'm understanding things correctly, I think this comment is now incorrect, since `get_partner_config_by_kind()` resolves to the longest entry rather than the first?

taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -399,12 +399,13 @@ def get_partner_config_by_kind(config, kind):
...
 
     # TODO eme-free should be a partner; we shouldn't care about per-kind
-    for k in partner_configs:
-        if kind.startswith(k):
-            kind_config = partner_configs[k]
-            break
-    else:
+    # The longest matching key wins, because one key can prefix another:
+    # "repackage-msi" also prefixes the "repackage-msix" kind, and only the
+    # longer key holds the msix config.
```
jonathanmendez: I'm confused by this comment, because I can't derive intent from it. It describes that the code will take the longest matching key, and gives the msi/msix example; but I don't know why we want msix or more generally why we want the longest matching key?

taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -643,17 +644,43 @@ def _pad_macos_attribution_code(attribution_string):
     return attribution_string
 
 
+# Cache key for the one enterprise manifest crawl every enterprise kind shares.
+# Deliberately not a kind name, so it can't be picked up by
+# get_partner_config_by_kind().
+ENTERPRISE_MANIFEST_KIND = "enterprise-manifest"
```
jonathanmendez: It's weird to name this a `_KIND` when it's explicitly not a kind name. Maybe `_FAKE_KIND` or just `_CACHE_KEY`?

taskcluster/gecko_taskgraph/util/partners.py
```diff
@@ -643,17 +644,43 @@ def _pad_macos_attribution_code(attribution_string):
...
+}
+
+
 def make_enterprise_repack_from_url(parameters, graph_config, kind, platform_mapping):
     token = get_token({"level": parameters["level"], "source": "enterprise"})
     partner_url_config = get_partner_url_config(parameters, graph_config)
-    get_partner_config_by_url(
-        manifest_url=partner_url_config["enterprise-repack"],
+    return get_partner_config_by_url(
+        manifest_url=partner_url_config.get("enterprise-repack"),
```
jonathanmendez: nit: Switching to `.get()` lets `get_partner_config_by_url` report an error if the url is missing, but the error message will say the url for "enterprise-manifest" is not defined. This could be misleading since it's not an actual kind, whereas leaving it with `["enterprise-repack"]` would at least surface "enterprise-repack" in the KeyError.

jonathanmendez (approved):
As best I can understand, this looks functionally correct, but I've requested some comment/naming updates for clarity.

taskcluster/gecko_taskgraph/util/partners.py
lissyx: Because of the ordering of keys: previously we would match with `startswith()`, so `repackage-msi` would always match before `repackage-msix`

taskcluster/gecko_taskgraph/util/partners.py
lissyx: Yes removing

taskcluster/gecko_taskgraph/util/partners.py
lissyx: I think i'll remove this part

PR #1435, opened 2026-09-08
Bug 2070179 - Restrict MOZ_BYPASS_FELT, and remove MOZ_AUTOMATION checks

lissyx:
### Description 

Bugzilla: Bug-2070179

MOZ_BYPASS_FELT is still required to perform some automation execution, but previously it was gated against MOZ_AUTOMATION. Remove this and limit allowance to MOZ_UPDATE_CHANNEL being non existent or default value: this will exclude MOZ_BYPASS_FELT from working on shippable builds.

Also update checks that were done against MOZ_AUTOMATION to allow some features and gate then the sameway: marionette, remote debugging and devtools.

PR #1444, opened 2026-09-08
Bug 2068746 - Refuse remote agent system access on shipping enterprise builds

farmisen:
Fix: https://bugzilla.mozilla.org/show_bug.cgi?id=2068746

## Description

Gates RemoteAgent.allowSystemAccess on enterprise builds carrying a non-default update channel.

Verified across all three request routes (env Marionette, flag Marionette, flag BiDi): each grants privileged scope without the fix and is refused with it.

### DIFF
--- a/remote/components/RemoteAgent.sys.mjs
+++ b/remote/components/RemoteAgent.sys.mjs
@@ -20,7 +20,8 @@ ChromeUtils.defineESModuleGetters(lazy, {
 ChromeUtils.defineLazyGetter(lazy, "logger", () => lazy.Log.get());
 
 // A remote client with system access can run privileged JavaScript in the parent
-// process and undo enterprise policy. Deny it on non-default enterprise builds.
+// process and any other privileged context, such as WebExtension processes, and
+// undo enterprise policy. Deny it on non-default enterprise builds.
 const PREVENT_SYSTEM_ACCESS =
   AppConstants.MOZ_ENTERPRISE && AppConstants.MOZ_UPDATE_CHANNEL !== "default";
 
--- a/remote/components/RemoteAgent.sys.mjs
+++ b/remote/components/RemoteAgent.sys.mjs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { AppConstants } from "resource://gre/modules/AppConstants.sys.mjs";
+
 const lazy = {};
 
 ChromeUtils.defineESModuleGetters(lazy, {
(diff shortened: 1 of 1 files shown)


remote/components/RemoteAgent.sys.mjs
```diff
@@ -17,6 +19,11 @@ ChromeUtils.defineESModuleGetters(lazy, {
 
 ChromeUtils.defineLazyGetter(lazy, "logger", () => lazy.Log.get());
 
+// A remote client with system access can run privileged JavaScript in the parent
+// process and undo enterprise policy. Deny it on non-default enterprise builds.
+const PREVENT_SYSTEM_ACCESS =
+  AppConstants.MOZ_ENTERPRISE && AppConstants.MOZ_UPDATE_CHANNEL !== "default";
```
lissyx, line 26: In #1435 I've limited remote CLI args, isn't it enough already?

farmisen: I think they are patching 2 different env values/--args combinations that both leads to a similar problem - I have a script I used to repro/verify [that] 2068746 [was fixed] and I am going to run it on your branch.

farmisen: confirmed I was still able to reproduce the bug on enterprise-main post #1435 merge

remote/components/RemoteAgent.sys.mjs
```diff
@@ -51,7 +58,9 @@ class RemoteAgentParentProcess {
   constructor() {
     this.#allowHosts = null;
     this.#allowOrigins = null;
-    this.#allowSystemAccess = Services.env.get(ENV_ALLOW_SYSTEM_ACCESS) == "1";
+    this.#allowSystemAccess =
+      !PREVENT_SYSTEM_ACCESS &&
+      Services.env.get(ENV_ALLOW_SYSTEM_ACCESS) == "1";
```
lissyx, line 64: This is the only part I'm worried about because I do not get the risks here, so my review will depend on feedback from webdriver people on that point

lissyx (approved):
r=me with if webdriver people thinks it's OK

remote/components/RemoteAgent.sys.mjs
```diff
@@ -17,6 +19,11 @@ ChromeUtils.defineESModuleGetters(lazy, {
 
 ChromeUtils.defineLazyGetter(lazy, "logger", () => lazy.Log.get());
 
+// A remote client with system access can run privileged JavaScript in the parent
+// process and undo enterprise policy. Deny it on non-default enterprise builds.
```
whimboo: Note that this is not only for the parent process but any other privileged process as well like webextensions etc.

remote/components/RemoteAgent.sys.mjs
whimboo: Not allowing system access will basically prevent any command in both WebDriver BiDi and Marionette (with todays Firefox Nightly) to interact with a privileged page that has more than the default principle. If you do not want to test Firefox UI or require any call to a Gecko API then you can disable it for those builds where you think it's not appropriate.

remote/components/RemoteAgent.sys.mjs
farmisen: clarified - thanks!

remote/components/RemoteAgent.sys.mjs
farmisen: that's the intent of this PR - disabling it for the shipping enterprise builds and keeping it open for the default channel so local and CI test builds are unaffected.

lissyx: Thanks @whimboo I just wanted to be cautious here, it's indeed what we want to do.

PR #1414, opened 2026-09-04
Bug-2059720, Bug-2059723, Bug-2059726 - Clean up Relay, breach alerts, and data collection sections in about:preferences

jporter-dev:
### Description 

Bugzilla: Bug-2059720, Bug-2059723, Bug-2059726

This PR spans a few different bugs related to `about:preferences` and either removes or updates different sections for Enterprise builds.

- Bug-2059720: Removed with `signon.firefoxRelay.feature="unavailable"`
- Bug-2059723: Removed with `browser.urlbar.trustPanel.breachAlerts=false` and `browser.urlbar.trustPanel.breachAlerts.featureGate=false`
- Bug-2059726:
  - Only remaining item is the Crash Report checkbox. This is togglable by default, and is locked + checked/unchecked by the `CrashReportsSubmit` policy.
  - Hid the health report, usage ping, add-on recommendations, studies, and Nimbus rollouts checkboxes with MOZ_ENTERPRISE guards in `permissions-data.mjs`
  - Flipped pref defaults for associated checkboxes
    - `datareporting.usage.uploadEnabled=false` - Disable daily usage ping
    - `app.shield.optoutstudies.enabled=false` - Disable feature studies
    - `nimbus.rollouts.enabled=false` - Disable remote rollouts / Nimbus
    - `browser.discovery.enabled=false` - Disable add-on recommendations
    - `datareporting.healthreport.uploadEnabled=true` - left this one alone as it disables all Glean telemetry, which the current Enterprise metrics ride on. This should probably be locked, but deferring that to bug-2046816.
    - Added Enterprise variants of the section heading (no consumer-focused description) and the crash reports string (organization-focused copy, no Learn More link).

---

### Dependencies / Related Issues 

- `datareporting.healthreport.uploadEnabled` deferred to Bug-2046816

---

### Screenshots 

|||
|-|-|
|image<br>_Before: Data Collection_|image<br>_After: Data Collection_|
|image<br>_Before: Breach Alerts_|image<br>_After: Breach Alerts_|
|image<br>_Before: Firefox Relay_|image<br>_After: Firefox Relay_|

PR #1449, opened 2026-09-09
Bug 2070454 - Restore Nimbus opt-out pref defaults in enterprise tests

jporter-dev:
### Description 

Bugzilla: Bug-2070454

Bug-2059726 (#1414) set `nimbus.rollouts.enabled=false` and `app.shield.optoutstudies.enabled=false` which turns off `ExperimentAPI.enabled`, and stops the enrollment paths the Nimbus tests depend on.

- Restored both defaults in `NimbusTestUtils` on enterprise. Set on the default branch since several tests clear the user branch during cleanup. 
- Skipped `test_backgroundtask_experiments.js`. `BackgroundTask_message.sys.mjs` clears `app.shield.optoutstudies.enabled` at startup in its own process, so nothing the parent sets survives.

---

* Regressed by: #1414 

---

### Testing 

* [ ] Added tests


toolkit/components/backgroundtasks/tests/xpcshell/xpcshell.toml
```diff
@@ -27,6 +27,7 @@ run-if = [
   "buildapp == 'browser'",
 ]
 reason = "ASRouter is Firefox-only."
+skip-if = ["enterprise"] # The task process resets app.shield.optoutstudies.enabled to the enterprise default (false), so no enrollment happens and the test can't pass.
```
lissyx: i'd prefer we fold this into the existing `run-if`

toolkit/components/nimbus/test/NimbusTestUtils.sys.mjs
```diff
@@ -34,6 +35,18 @@ const lazy = XPCOMUtils.declareLazy({
...
+// enrollment paths Nimbus tests exercise. Restored on the default branch because
+// tests clear the user branch during cleanup.
+if (AppConstants.MOZ_ENTERPRISE) {
+  Services.prefs
+    .getDefaultBranch("")
+    .setBoolPref("nimbus.rollouts.enabled", true);
+  Services.prefs
+    .getDefaultBranch("")
+    .setBoolPref("app.shield.optoutstudies.enabled", true);
+}
```
lissyx, line 48: if nimbus is disabled we should just skip those tests? Potentially add an enterprise specific test that verifies it's disabled ?

lissyx (approved):
r=me with comments addressed

toolkit/components/nimbus/test/NimbusTestUtils.sys.mjs
jporter-dev: Nimbus isn't truly disabled, just the default for `app.shield.optoutstudies.enabled` is now false. If we should truly disable Nimbus I can file a followup bug?

lissyx: Please yes

jporter-dev: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=2073149 for followup

PR #1454, opened 2026-09-09
Bug 2068752 - Uninstall add-ons listed in Extensions.Uninstall at every startup

johnreitano:
### Description

Bugzilla: Bug-2068752

`Extensions.Uninstall` is now reconciled against the installed add-ons on every startup instead of once per recorded list value. The `runOncePerModification` record for the list only detects an edited list. An add-on that a policy installed (persisted install source `enterprise-policy`) is left alone while the list is unchanged and the policy also has an Install list, which keeps the documented way of updating an add-on by listing it in both `Uninstall` and `Install` (Bug 1495749) and survives the add-on updating itself. An ID that `ExtensionSettings` force- or normal-installs is skipped and reported as a conflict against `Extensions`. The Install step now follows the uninstall step explicitly and re-runs only when the Uninstall list changed.

The `runOncePerModification` documentation states that its record is user-editable profile state and must only gate settings the user may change afterwards, never enforcement. The other call sites are either initial-state settings by design or tracked separately.

---

* Related: Bug 2068736 (meta)

---

### Screenshots

Not applicable, no UI changes.

---

### Testing

* [x] Added tests

New `test_extensions_uninstall.js` (xpcshell) covers: a pre-existing record, re-enforcement after a reinstall, no churn for a policy-installed add-on before and after a self-update, the both-lists update flow, a changed Install URL with a pre-existing record, and the `ExtensionSettings` conflict. `browser_policy_extensions.js` now asserts the record already names the add-on before the uninstall task. `test_runOnce_helper.js` covers `runOncePerModification` and `clearRunOnceModification`.

Local results on an enterprise build: xpcshell 6 files passed, 0 failed; browser-chrome 82 passed, 0 failed, 3 todo. Browser-chrome on the enterprise build needs `./mach mochitest --headless --setenv MOZ_AUTOMATION=1 --setenv MOZ_BYPASS_FELT=1`.

**Steps to verify changes:**
1. Install an add-on into a profile, then deploy a `policies.json` with `Extensions.Uninstall` listing its ID and start the browser: the add-on is removed.
2. Reinstall the add-on and restart with the same policy: it is removed again.
3. List the same ID in both `Uninstall` and `Install` and restart twice: the policy-installed add-on stays and is not re-downloaded on the second start.

**Expected result:** the Uninstall list is enforced at every startup regardless of the recorded list value, while add-ons the policy itself installs are not churned.

https://claude.ai/code/session_01PsrknLtSrjBpXejVumvsE4

### DIFF
--- a/browser/components/enterprisepolicies/Policies.sys.mjs
+++ b/browser/components/enterprisepolicies/Policies.sys.mjs
@@ -53,6 +53,8 @@ ChromeUtils.defineESModuleGetters(lazy, {
   pemToBase64: "resource://gre/modules/PoliciesHelpers.sys.mjs",
   processMIMEInfo: "resource://gre/modules/PoliciesHelpers.sys.mjs",
   replacePathVariables: "resource://gre/modules/PoliciesHelpers.sys.mjs",
+  isRunOnceModificationApplied:
+    "resource://gre/modules/PoliciesHelpers.sys.mjs",
   reportFailure: "resource://gre/modules/PoliciesHelpers.sys.mjs",
   setDefaultPermission: "resource://gre/modules/PoliciesHelpers.sys.mjs",
   runOncePerModification: "resource://gre/modules/PoliciesHelpers.sys.mjs",
@@ -78,18 +80,24 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
 /**
  * Uninstalls the add-ons an Extensions.Uninstall list names whenever they are
  * present, so a run-once marker pre-seeded in the profile cannot keep one
- * installed. While the list is unchanged and the policy also installs add-ons,
- * an add-on that a policy installed is left alone: that is what lets an
- * administrator update an add-on by listing it in both Uninstall and Install.
+ * installed. An ID that ExtensionSettings installs is reported as a conflict
+ * and skipped, whether or not it is installed yet.
+ *
+ * While neither the Uninstall list nor the Install list has changed since it
+ * was last applied, an add-on that a policy installed is left alone. That is
+ * what lets an administrator update an add-on by listing it in both Uninstall
+ * and Install without it being downloaded again at every startup. As soon as
+ * either list changes, everything listed is uninstalled, so an add-on the
+ * Install list no longer covers does not linger.
  *
  * @param {string[]} ids
  *        The IDs of the add-ons to uninstall.
- * @param {boolean} hasInstallList
- *        Whether the policy also has a non-empty Install list.
+ * @param {string[]} [installList]
+ *        The policy's Install list, if it has one.
  * @returns {Promise<boolean>}
  *        Whether the Uninstall list changed since it was last applied.
  */
-async function uninstallListedAddons(ids, hasInstallList) {
+async function uninstallListedAddons(ids, installList = []) {
   let listChanged = false;
   lazy.runOncePerModification(
     "extensionsUninstall",
@@ -98,24 +106,31 @@ async function uninstallListedAddons(ids, hasInstallList) {
       listChanged = true;
     }
   );
-  const addons = await lazy.AddonManager.getAddonsByIDs([...new Set(ids)]);
+  const keepPolicyInstalls =
+    !listChanged &&
+    !!installList.length &&
+    lazy.isRunOnceModificationApplied(
+      "extensionsInstall",
+      JSON.stringify(installList)
+    );
+  const candidates = [...new Set(ids)].filter(id => {
+    const mode = Services.policies.getExtensionSettings(id)?.installation_mode;
+    if (mode == "force_installed" || mode == "normal_installed") {
+      lazy.reportFailure(
+        "Extensions",
+        `Not uninstalling ${id} because ExtensionSettings installs it`
+      );
+      return false;
+    }
+    return true;
+  });
+  const addons = await lazy.AddonManager.getAddonsByIDs(candidates);
   for (const addon of addons) {
     if (!addon) {
       continue;
     }
-    const mode = Services.policies.getExtensionSettings(
-      addon.id
-    )?.installation_mode;
-    if (mode == "force_installed" || mode == "normal_installed") {
-      lazy.reportFailure(
-        "Extensions",
-        `Not uninstalling ${addon.id} because ExtensionSettings installs it`
-      );
-      continue;
-    }
     if (
-      !listChanged &&
-      hasInstallList &&
+      keepPolicyInstalls &&
       addon.installTelemetryInfo?.source == "enterprise-policy"
     ) {
       continue;
@@ -1914,7 +1929,7 @@ export var Policies = {
       if ("Uninstall" in param) {
         uninstallingPromise = uninstallListedAddons(
           param.Uninstall,
-          !!param.Install?.length
+          param.Install
         );
       }
       if ("Install" in param) {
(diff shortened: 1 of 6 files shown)


johnreitano:
@mozilla/enterprise-firefox-engineering this is ready for review. I don't have permission to add a reviewer on this repo, so a formal review request from someone on the team would be appreciated.

browser/components/enterprisepolicies/Policies.sys.mjs
lissyx, line 1: This will require a porting bug to Thunderbird, please file follow up blocking `thunderbird-client-poc` meta bug and assign to me (or if you are willing to do the port yourself, you need to build and run `comm/mail/components/enterprisepolicies/tests/browser/browser_policies_differences.js` and make sure to adapt `Policies.sys.mjs` accordingly on Thunderbird side)

toolkit/components/enterprisepolicies/PoliciesHelpers.sys.mjs
```diff
@@ -603,6 +603,12 @@ export function runOnce(actionName, callback) {
  * If the callback that was passed is an async function, you can await on this
  * function to await for the callback.
  *
+ * The record of the last applied value is an ordinary user preference in the
```
gcp: This sounds like a caveat on, I'm guessing here `actionName`, but I'm not sure. Wouldn't it make sense to leave a small note on the parameter itself?

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -75,6 +75,61 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+    }
+    try {
+      await addon.uninstall();
+    } catch (e) {
+      // This can fail for add-ons that can't be uninstalled.
+      lazy.log.debug(`Add-on ID (${addon.id}) couldn't be uninstalled.`);
+    }
+  }
+  return listChanged;
+}
```
lissyx: All this code should live in `PoliciesHelper` I think

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -75,6 +75,61 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+      lazy.reportFailure(
+        "Extensions",
+        `Not uninstalling ${addon.id} because ExtensionSettings installs it`
+      );
+      continue;
+    }
+    if (
+      !listChanged &&
+      hasInstallList &&
+      addon.installTelemetryInfo?.source == "enterprise-policy"
```
gcp: This doesn't check whether the add-on in install and uninstall match, so you can have a pending admin-forced update:

```
  {
    "Uninstall": ["addon-a@example.com"],
    "Install": ["https://example.com/addon-a.xpi"]
  }
```

replace install

```
  {
    "Uninstall": ["addon-a@example.com"],
    "Install": ["https://example.com/addon-b.xpi"]
  }
```

and now `addon-a` stays installed and not updated.

browser/components/enterprisepolicies/Policies.sys.mjs
```diff
@@ -75,6 +75,61 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
...
+    () => {
+      listChanged = true;
+    }
+  );
+  const addons = await lazy.AddonManager.getAddonsByIDs([...new Set(ids)]);
+  for (const addon of addons) {
+    if (!addon) {
+      continue;
+    }
+    const mode = Services.policies.getExtensionSettings(
```
gcp: This should go before the `AddonManager` call in order for the code below to work correctly on a fresh profile. It needs to resolve the requests pending, which is not necessarily already reflected in `AddonManger`.

browser/components/enterprisepolicies/tests/xpcshell/test_extensions_uninstall.js
```diff
@@ -0,0 +1,336 @@
...
+    await addon.uninstall();
+    await uninstalled;
+  }
+}
+
+// The Uninstall and Install steps of the Extensions policy keep running after
+// the engine reports the policies applied; give them time to finish.
+async function settle() {
+  await AddonManager.getAddonsByIDs([ADDON_ID]);
+  // eslint-disable-next-line mozilla/no-arbitrary-setTimeout
```
gcp: _Frowning intensifies._

gcp (changes requested):

browser/components/enterprisepolicies/tests/xpcshell/test_extensions_uninstall.js
johnreitano: removed

toolkit/components/enterprisepolicies/PoliciesHelpers.sys.mjs
johnreitano: I clarified this comment a bit to help a developer decide whether to use this function.
I also added to comment on `actionName`. Let me know what you think.

toolkit/components/enterprisepolicies/PoliciesHelpers.sys.mjs
```diff
@@ -603,6 +603,12 @@ export function runOnce(actionName, callback) {
...
+ * profile, which the user can edit before startup to make the callback look
+ * like it already ran. Only use this for a setting the user is allowed to
+ * change afterwards. A policy that enforces something must instead check the
+ * state it controls at every startup.
+ *
  * @param {string} actionName
  *        A given name which will be used to track if this callback has run.
  *        This string will be part of a pref name.
```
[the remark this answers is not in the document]
johnreitano: fixed

browser/components/enterprisepolicies/Policies.sys.mjs
johnreitano: Good catch. The exemption now also requires the Install list to be unchanged since it was last applied, so replacing addon-a.xpi with addon-b.xpi uninstalls addon-a and installs addon-b (covered by test_changed_install_list_removes_policy_installed_addon). We can't match an Uninstall ID against an Install URL without downloading it, so the rule is "leave policy installs alone only when neither list changed".

browser/components/enterprisepolicies/Policies.sys.mjs
johnreitano: In the pending-install case the lookup returns null and the loop skips the ID, so the add-on was never at risk of being uninstalled, but the conflict went unreported until the next startup. With my latest commit, the check now runs on the configured IDs before getAddonsByIDs, so it reports at every startup, fresh profile included (see test_extensionsettings_conflict_reported_before_install).

browser/components/enterprisepolicies/Policies.sys.mjs
johnreitano: Moved to PoliciesHelpers as uninstallListedAddons

browser/components/enterprisepolicies/Policies.sys.mjs
johnreitano: I filed the corresponding bug for Thunderbird at https://bugzilla.mozilla.org/show_bug.cgi?id=2071109 and assigned to you.

johnreitano: Note uninstallListedAddons and isRunOnceModificationApplied exist only in this fork's PoliciesHelpers, so the new code needs a MOZ_ENTERPRISE guard until it lands upstream.

1rneh (commented):
Bug-2070529 just landed which changed the policy engine to reject deprecated, superseded or incompatible policies. The policy `Extensions` is one of them (see [thread](https://mozilla.slack.com/archives/C0976G3HW9L/p1789117540315649)).

johnreitano:
Closing this: With Bug 2070529 / PR #1460, Extensions is now rejected on Firefox Enterprise builds, so the handler this PR modifies is unreachable here.

Administrators wanting an add-on removed should use ExtensionSettings, which stays supported and uninstalls on every run with no run-once marker in the path.

Extensions is still supported upstream, so the every-startup reconciliation is worth having in mozilla-central — filing separately. The runOncePerModification JSDoc from this PR will be carried into a follow-up rather than dropped.

PR #1460, opened 2026-09-10
Bug 2070529 - Reject incompatible policies

1rneh:
### Description 

Bugzilla: Bug-2070529

The policy engine validates the policy's compatibility and rejects a policy whose current build variant is incompatible (version_added === false).

Also lets `modifySchemaForTests` backfill the `x-compatibility` and `x-restart-required` defaults the meta-schema mandates, so custom test schemas need not declare them and skips tests that rely on enterprise incompatible policies

---

### Testing 

* [X] Added tests

### DIFF
--- a/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
+++ b/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
@@ -622,6 +622,14 @@ EnterprisePoliciesManager.prototype = {
       return { isValid: false, parsedParams: null };
     }
 
+    if (!this._isPolicyCompatible(policyName)) {
+      this._reportPolicyError(
+        policyName,
+        `Policy ${policyName} is not supported.`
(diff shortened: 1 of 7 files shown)


lissyx (approved):

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
lissyx, line 697: I hoped we could avoid but turns out we need to detect Thunderbird there as well

1rneh: Right, we need to add `x-compatibility` data for thunderbird in `policies-schema.json`.

lissyx: I have a local patch for it and will take care of this in Bug-2071226

PR #1462, opened 2026-09-10
Bug 2068739 - Do not auto-load profile PKCS#11 modules on enterprise builds

cmkoehler:
Bug 2068739.

On enterprise builds, initialize NSS without the profile's PKCS#11 module DB,
so modules named in the profile are not auto-loaded at startup. Administrator-
configured modules still load via the SecurityDevices policy (load-without-
persist), and safe mode still refuses module loads.

Adds a regression test, and updates `test_broken_fips` for enterprise builds
(where the profile module DB — and thus the FIPS-mode recovery it exercises —
is never loaded).

Try: `user/chkoehler/bug-2068739-enterprise-pkcs11` — xpcshell across
Linux/macOS/Windows. All enterprise chunks pass. The failing non-enterprise
policy/permission tests (`test_permissions`, `test_policy_descriptions`,
`test_popups_cookies_addons`) are pre-existing on `enterprise-main` and
unrelated to this change (its code path is not compiled on non-enterprise
builds).


lissyx:
I lack proper understanding here of the extents of just refusing to load those modules: I know they are used for very legitimate purpose in the wild, and I can see usecase where this may break end-users of our existing pilot users. I am not sure if those use-cases can be covered by `SecurityDevices` policy for example.

gcp:
The key here would be the location where they are loaded from: it should not be a user-controllable location, because that likely indeed would mean arbitrary DLL injection into our process.

Is there currently a place where they are supposed to be placed that is admin-writable only? The https://firefox-admin-docs.mozilla.org/reference/policies/securitydevices/ policy has the admin indicate some directory, so that indeed seems to be the intended use case.

Regardless of the above, this is a policy that may well break current deployments, and I'm wary that there may even be third party card readers etc whose installers expect to be able to drop those files in the user profile. For those cases, some kind of hash/signature based allowlist may be required.

gcp:
@hafta I know you spent a lot of time looking at PCKS#11 usage, do you know if we have any data on where those modules are typically installed?

cmkoehler:
Thanks both — the location/allowlist framing is right, and I think it reframes this PR rather than opposing it.

**Deny vs. allowlist isn't either/or.** The profile `pkcs11.txt` is read inside `NSS_Initialize`, ~80 ms before the policy engine exists — so no per-module check ("deny all" or "allow if X") can happen in NSS at load time; it has to be in PSM once policies load. That's what this PR already does: it stops the pre-policy auto-load (`DoNotLoadModules` on enterprise) and routes loading through a post-init `SECMOD_LoadUserModule` path. An allowlist just changes that rule from "load none" to "load the ones the admin blessed" — so this PR is the mechanism an allowlist needs, and I'd fold it in on top.

**Proposed surface** — extend `SecurityDevices`:
```json
"SecurityDevices": {
  "Add": { ... },
  "Delete": [ ... ],
  "AllowProfileModules": [
    { "Name": "OpenSC",     "SHA256": "9f86d0…" },
    { "Name": "VendorCard", "SignedBy": "Developer ID Application: Vendor (TEAMID)" }
  ]
}
```
Matches on `SHA256` or `SignedBy`. It's admin-sourced (`policies.json`/GPO/plist), so the profile stays untrusted — @lissyx, this keeps legit deployments working: an installer can still drop a module in the profile, and it loads iff it's allowlisted.

**Enforcement:** `SecurityDevices.onProfileAfterChange` → a new `nsIPKCS11ModuleDB` method → PSM checks each profile `pkcs11.txt` entry against the allowlist and loads matches via the `SECMOD_LoadUserModule` path already here. Reuses `nsICryptoHash` (SHA-256), `Authenticode.cpp` (Windows signer), and `nsMacUtilsImpl::GetSignatureType` (macOS).

Both `SHA256` and `SignedBy` land in this PR (`SHA256` covers exact bytes but breaks on vendor updates; `SignedBy` is the update-stable blessing). Platform caveats: Linux has no code signing, so it's SHA256-only there; and our macOS helper reports signature *type*, not signer identity, so full `SignedBy` matching is Windows-first with macOS best-effort.

**Question:** extend `SecurityDevices` with `AllowProfileModules` as above, or a separate policy?

lissyx:
I get the feeling this requires more investigation on the definition of the policy itself

gcp:
https://bugzilla.mozilla.org/show_bug.cgi?id=2007195

Remote PKCS#11 _may_ (emphasis is important here) land in Firefox 158, which changes the attack surface significantly. A rogue module would then sit in a process that can be sandboxed (but seeing how strict that sandbox can be is something we haven't found out yet) and can't directly change the main process.

The timing is annoying here, we almost certainly can't block on that landing or being sufficiently sandboxed, but it *may* mean we could be OK with breaking a few modules or deployments...

security/manager/ssl/PKCS11ModuleDB.cpp
```diff
@@ -345,6 +379,16 @@ PKCS11ModuleDB::AddModule(const nsAString& aModuleName,
   auto promiseHolder =
       MakeRefPtr<nsMainThreadPtrHolder<Promise>>("AddModule promise", promise);
 
+#ifdef MOZ_ENTERPRISE
+  // Refuse in safe mode, where no PKCS#11 modules load. Checked here because
```
gcp, line 383: Does this imply upstream Firefox loads PKCS#11 modules in Safe mode? That sounds like a separate upstream bug.

cmkoehler: from what i can tell, no. upstream doesnt auto load profile modules in safe mode. the guard exists only because the enterprise add-path uses SECMOD_LoadUserModule, which loads regardless of being in safe mode. upstream's add-path (SECMOD_AddNewModule) instead fails in safe mode because it can't persist — which is why test_pkcs11_safe_mode.js sees a throw. so the guard just reproduces upstream's "addModule fails in safe mode."

cmkoehler:
following up. did we want to land this pr or reevaluate the approach?

hafta:
> @hafta I know you spent a lot of time looking at PCKS#11 usage, do you know if we have any data on where those modules are typically installed?

I chatted with @gcp about some possibilities and it looks like we've got a plan now.

> Both `SHA256` and `SignedBy` land in this PR (`SHA256` covers exact bytes but breaks on vendor updates; `SignedBy` is the update-stable blessing). Platform caveats: Linux has no code signing, so it's SHA256-only there; and our macOS helper reports signature _type_, not signer identity, so full `SignedBy` matching is Windows-first with macOS best-effort.

For macOS, is the helper being referred to here the nsMacUtilsImpl::GetSignatureType()? That was just intending to get the signature type on macOS to understand if what percentage of modules used are signed. With a long term goal of only loading signed modules. So it didn't consider signing identity. For this you would want to use something else such as verifying the team ID of the binary or doing the string check on the leaf certificate and verifying the chain. It doesn't have to be a best-effort on macOS.

gcp:
I think the conclusion from the discussion was a) for security reasons, we're OK with breaking some backwards compat now in FxE. If it causes issues for a customer, we'll know exactly which modules we have to unbreak b) longer term, the module loading is moving to a separate process, at which point the module being loaded into that process will (if we can sufficiently sandbox it) not have any security implications for the other processes.

cmkoehler:
Thanks @hafta @gcp — plan understood: land the deny on Firefox Enterprise now, accept some breakage (unbreak specific modules if a customer hits one), with the move to a separate, sandboxable module process (bug 2007195) as the long-term fix. I'll drop the allowlist from this PR.

Two follow-ups:

1. **macOS signer matching** — thanks for the correction. Noted that `nsMacUtilsImpl::GetSignatureType()` was only for measuring the signed-module percentage, and that proper identity matching on macOS (Team ID, or leaf-cert string check + chain verification) is doable rather than best-effort. Not needed here, but good to have straight for the 2007195 work.

2. **"We'll know which module to unbreak" needs a small addition.** The existing `pkcs11.third_party_modules_loaded` / `third_party_module_profile_entries` telemetry enumerates the *loaded* module list (`SECMOD_GetDefaultModuleList()`) at init. With this PR the profile's `pkcs11.txt` is never read, so declined modules aren't loaded and won't appear there — we'd lose visibility into exactly the modules we're now blocking. So I'll add a startup log of the third-party `pkcs11.txt` entries we decline to load, so a blocked module is identifiable from logs/support. Surfacing it as telemetry (fleet-wide visibility) would need a separate data review, so I'd do that as a follow-up rather than here.

cmkoehler:
are we good to approve and merge?

cmkoehler:
Correction to my earlier comment about diagnosability — that framing was wrong.

There's no need for a bespoke startup log or a separate data review. The existing `pkcs11.third_party_module_profile_entries` metric already reports "the PKCS#11 module filenames in the profile at launch" and is already data-reviewed. The only problem is that it's populated by enumerating *loaded* modules, so with this PR (profile modules no longer auto-loaded) it would stop capturing them.

So instead of a log, I've sourced that metric from the profile's `pkcs11.txt` directly (macOS, matching its existing collection), so a blocked module stays visible with no new data collection — and added a test asserting it. Extending that visibility to Windows/Linux would be new collection there, so that's out of scope for this PR.

security/manager/ssl/PKCS11ModuleDB.cpp
mkaply: When SECMOD_FindModule returns null this falls through to NS_OK, so deleteModule("does-not-exist") succeeds here but throws on non-enterprise, where SECMOD_DeleteModule returns SECFailure. No current caller depends on that -- device_manager.js deletes a selected module, the tests delete what
they added, and SecurityDevices.Delete swallows errors deliberately -- but it is an undocumented divergence in an XPCOM method's contract between build configs. Either return NS_ERROR_FAILURE when the module isn't found, or note in the comment that a missing module is treated as already-deleted.

(For anyone reviewing the refcounting: this is correct, not a double-free. SECMOD_UnloadUserModule calls SECMOD_DeleteModuleEx(NULL, mod, &atype, PR_FALSE), which releases the list element's reference, not the caller's from SECMOD_FindModule -- same pattern as the osclientcerts unload in nsNSSComponent.cpp. permdb=PR_FALSE is also the right choice, since there's no module DB to write to.)

mkaply: Worth a comment that this escaping is required rather than stylistic, because it looks droppable. NormalizeModuleNameIn doesn't sanitize quotes, so the module name reaches DoAddModule verbatim; upstream has no injection surface because SECMOD_AddNewModule takes name and path as separate arguments, but building a spec string creates one. A name like

    x" library="/tmp/evil.so

would otherwise redirect the load. Escaping both values is correct.

This is also why reusing LoadUserModuleAt in NSSCertDBTrustDomain.cpp would be wrong: it escapes only the library path, since its name argument is a compile-time constant.

Nit: return nsCString rather than nsAutoCString -- the inline buffer does nothing in a return value.

mkaply: Accurate today -- every in-tree caller passes 0/0 (head_psm.js, test_pkcs11_moduleDB.js, test_pkcs11_remote_process.js, load_device.js). But silently discarding a non-zero value is a trap for whoever passes one next, and the IDL still advertises the arguments. Consider returning NS_ERROR_INVALID_ARG when either is non-zero, so the limitation fails loudly instead of being documented in a comment.

security/manager/ssl/PKCS11ModuleDB.cpp
mkaply: Confirming the answer, with the specifics: the guard is load-bearing. test_pkcs11_safe_mode.js asserts addModule throws in safe mode. Upstream gets that for free because SECMOD_AddNewModule can't persist, whereas SECMOD_LoadUserModule would succeed -- so without this the existing test would fail on enterprise. Not an upstream bug: upstream doesn't auto-load profile modules in safe mode either (InitializeNSSWithFallbacks already passes DoNotLoadModules).

The main-thread justification for checking here rather than in DoAddModule also holds -- GetInSafeMode() does a do_GetService on @mozilla.org/xre/runtime;1.

mkaply:
The enterprise DoAddModule path also runs in the PKCS#11 utility process via RecvAddModule, so with security.utility_pkcs11_module_process.enabled set you get a mixed state: admin modules load non-persistently there, while that process still auto-loads the profile's pkcs11.txt, because ConfigureNSSInPKCS11UtilityProcess passes PKCS11DBConfig::LoadModules unconditionally. That's the follow-up bug, but the PR description should say this changes utility-process behavior too.

On upstreamability: this puts #ifdef MOZ_ENTERPRISE inside security/manager/ssl. If the intent is ever to upstream rather than carry a permanent delta, a generic build option is the shape upstream is likely to accept -- folding it into MOZ_NO_SMART_CARDS, or adding something like --disable-profile-pkcs11-modules -- with enterprise simply being the
configuration that enables it. Same behavior, no product-specific conditional in shared security code, and SECMOD_LoadUserModule becomes a capability upstream can reason about independently.

mkaply (changes requested):

browser/moz.configure
```diff
@@ -72,10 +72,37 @@ def enable_enterprise(value):
...
+
+@depends("--enable-profile-pkcs11-modules")
+def disable_profile_pkcs11_modules(value):
+    if not value:
+        return True
+
+
+set_config("MOZ_DISABLE_PROFILE_PKCS11_MODULES", disable_profile_pkcs11_modules)
+set_define("MOZ_DISABLE_PROFILE_PKCS11_MODULES", disable_profile_pkcs11_modules)
+
```
lissyx: Please move as much as possible of this to `toolkit/moz.configure` and make sure to file a follow up bug similar to https://bugzilla.mozilla.org/show_bug.cgi?id=2062894 to make sure that we do not forget when I upstream `--enable-enterprise`.

PKCS#11 is also used on Thunderbird and it may even be more relevant there, so I suspect we can to handle it as well there.

security/manager/ssl/tests/unit/test_broken_fips.js
lissyx: This comment is referring to Enterprise builds but now the check is against disable profile pkcs11

security/manager/ssl/tests/unit/test_pkcs11_profile_module.js
lissyx: Similar, update the comment to reflect the new build flag

security/manager/ssl/tests/unit/xpcshell-smartcards.toml
lissyx: Is this pref required? If so, please make sure to leave a comment justifying

security/manager/ssl/nsNSSComponent.cpp
lissyx:
```suggestion
  PKCS11DBConfig profileModuleDBConfig =
  #if defined(MOZ_DISABLE_PROFILE_PKCS11_MODULES)
                                             false
  #else
                                             !safeMode
  #endif
                                             ? PKCS11DBConfig::LoadModules
                                             : PKCS11DBConfig::DoNotLoadModules;
```

lissyx: I feel it's easier to read this way

security/manager/ssl/PKCS11ModuleDB.cpp
lissyx: What is the "matching way" here?

mkaply:
Nice update!

Claude doesn't think the #ifdef will work though

AppConstants.MOZ_DISABLE_PROFILE_PKCS11_MODULES is always going to be false here.
The defines that AppConstants.sys.mjs can see come from those explicit lists in
toolkit/modules/moz.build, and the new one isn't in any of them, so the #ifdef
never fires. It worked before because MOZ_ENTERPRISE is already in that list.
Which means both test_pkcs11_profile_module.js and test_broken_fips.js will take
the non-enterprise branch and fail on enterprise -- so we'll want a new try run
too, the one in the description is from before the switch. Also, now that it's a
generic option rather than enterprise-only, should it move to
toolkit/moz.configure next to MOZ_NO_SMART_CARDS? Everything using it is
toolkit-level.

lissyx:
Thanks @cmkoehler, this is looking better, but I'd like a bit of changes. Should some people from security also have a look at the changes?

mkaply:
I'll ping dkeeler

cmkoehler:
Verified this actually fires. The built `AppConstants.sys.mjs` in the objdir resolves `MOZ_DISABLE_PROFILE_PKCS11_MODULES: true` on an enterprise build, and both test assertions take the disabled branch and pass. `AppConstants.sys.mjs` is an `EXTRA_PP_JS_MODULES`, whose preprocessor receives all `set_define`s (ACDEFINES) — the same path `#ifdef MOZ_ENTERPRISE` uses — so it doesn't need to be in the `_BOOL` lists in `toolkit/modules/moz.build` (those are only for the `@X_BOOL@` substitution form, which this doesn't use).

Moved the option to `toolkit/moz.configure` per the suggestion (browser implies it off for enterprise). A fresh try run is coming — the one in the description predates this.

cmkoehler:
i think we should be all good outside the suggested security review that was mentioned.

security/manager/ssl/tests/unit/test_broken_fips.js
mozkeeler: It might make more sense to just skip this test in this build configuration, if that's possible.

security/manager/ssl/tests/unit/test_pkcs11_moduleDB.js
mozkeeler: I'm not sure this test makes sense - we essentially trust the code that's calling `addModule`, right? And moreover, if someone wanted to load `evilModule.dll` rather than `goodModule.dll`, couldn't they just replace the file `goodModule.dll` with `evilModule.dll`?

security/manager/ssl/tests/unit/test_broken_fips.js
cmkoehler: done. it now early retunrs with MOZ_DISABLE_PROFILE_PKCS11_MODULES is set. i used a runtime AppConstants skip rather than a manifest skip-if, to avoid adding the flag to mozinfo just for this.

security/manager/ssl/tests/unit/test_pkcs11_moduleDB.js
cmkoehler: youre right. i reframed it as a correctness/regression guard for the spec string quoting (@mkaply had asked for escaping). happy to remove it entirely if youd rather that.

lissyx (approved):
LGTM, assuming security review is positive.

cmkoehler:
@mozkeeler are we good to merge?

mkaply:
I think if you addressed their comments, we're good.

cmkoehler:
> I think if you addressed their comments, we're good.

im blocked from merging because you are still flagged as requesting changes. im not sure if im allowed to merge via command line

mkaply (approved):

mkaply:
> > I think if you addressed their comments, we're good.
> 
> im blocked from merging because you are still flagged as requesting changes. im not sure if im allowed to merge via command line

I think I fixed it.

PR #1463, opened 2026-09-10
Bug 2070625 - Use launcher process when spawning browser from felt

jonathanmendez:
### Description 

Bugzilla: Bug-2070625

On Windows, Firefox's initial process is a Launcher process that launches the browser process in a suspended state so that it can patch the DLL loading mechanism to block modules and to set up pre-launch CIG guards, etc. Normally this happens because the initial process detects that its parent process is not Firefox, so it becomes a Launcher. However, since we launch the browser from Felt, the auto-detection does not work. Similarly by launching using `Subprocess`, we don't go through the typical setup with which Firefox propagates the protections to subprocesses.

Instead we can pass the "--launcher" flag to specify a launcher process and get the protections.

Note: Normally, the initial launcher process exits leaving the spawned browser running. However, with how Felt spawns the process there are problems with this approach:
1. By default, Subprocess will kill its process tree when its process exits
2. Felt depends on the exit code from the browser process, which is not captured if the launcher process exits

The extra launcher process has minimal overhead (~8MB from a debug build on my system) so we will just keep it alive.

---

* Blocks: #1442 

---

### Testing 

* [x] Added tests
  * I've done testing of the specific third-party module blocking and some restart scenarios, but we'll need QA to be on the alert for any issues that could show up.

### DIFF
--- a/testing/enterprise/manifest-browser.toml
+++ b/testing/enterprise/manifest-browser.toml
@@ -31,6 +31,15 @@ disabled = "Bug 1996558: scaffolding for the browsing-child encryption-mismatch 
 
 ["test_felt_browser_fxa.py"]
 
+["test_felt_browser_launcher_process.py"]
+run-if = [
+  "os == 'win'",
+]
+skip-if = [
+  "asan",
+  "os == 'win' && arch == 'aarch64'",
+]
+
 ["test_felt_browser_new_window_from_cli.py"]
 
 ["test_felt_browser_rapid_new_tabs_from_cli.py"]
--- a/toolkit/components/felt/content/FeltProcessParent.sys.mjs
+++ b/toolkit/components/felt/content/FeltProcessParent.sys.mjs
@@ -874,8 +874,20 @@ export class FeltProcessParent extends JSProcessActorParent {
       profileArgs = ["-P", foundProfile.name];
     }
 
+    // Use the launcher process so the browser gets the DLL blocklist and other
+    // security enhancements patched in while suspended.
+    // --wait-for-browser keeps the launcher alive to relay the browser's exit code
+    // --no-deelevate keeps the browser at felt's integrity level so that the ipc
+    // socket is accessible to both the browser and felt. (Felt running elevated
+    // is probably bad, but unlikely due to felt's own launcher process)
+    let launcherArgs = [];
+    if (Services.appinfo.OS == "WINNT" && lazy.isBuildAppBrowser()) {
+      launcherArgs = ["--launcher", "--wait-for-browser", "--no-deelevate"];
+    }
+
     lazy.log.debug(`Using profileArgs: ${profileArgs}`);
     const firefoxRunArgs = [
+      ...launcherArgs,
       "--foreground",
       ...profileArgs,
       "-felt",
--- a/testing/enterprise/test_felt_browser_launcher_process.py
+++ b/testing/enterprise/test_felt_browser_launcher_process.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python3
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_tests import FeltTests
+
+
+class BrowserLauncherProcess(FeltTests):
+    def test_browser_started_via_launcher(self):
+        """Felt must start the browser through the launcher process, which
+        patches security enhancements into the browser process. The dynamic
+        blocklist is only available if the browser was started through the
+        launcher process, so we can check for that to verify it."""
+        self.run_felt_base()
+        self.connect_child_browser()
+
+        with self._child_driver.using_context(self._child_driver.CONTEXT_CHROME):
+            available = self._child_driver.execute_script(
+                """
+                return Cc["@mozilla.org/about-thirdparty;1"].getService(
+                  Ci.nsIAboutThirdParty
+                ).isDynamicBlocklistAvailable;
+                """
+            )
+
+        self._logger.info(f"isDynamicBlocklistAvailable: {available}")
+
+        assert available, (
+            "Expected the browser to be started by a launcher process, but it "
+            "has no shared section (isDynamicBlocklistAvailable is false)"
+        )
--- a/testing/enterprise/test_felt_restart_works.py
+++ b/testing/enterprise/test_felt_restart_works.py
@@ -65,12 +65,20 @@ class AppRestartWorks(FeltTests):
         felt_ui_pid = self._driver.session_capabilities["moz:processID"]
         target_exe = psutil.Process(felt_ui_pid).exe()
         felt_browser_pids = set(self._get_felt_browser_pids(target_exe))
-        unexpected = felt_browser_pids - {new_browser_pid}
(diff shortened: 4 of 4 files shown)


lissyx:
Happily surprised it is this simple

lissyx:
This looks to be regressing: https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=40dcc726d8fae95b9e8f1bfe3f39715264081339&selectedTaskRun=WoeHwMtCQye-pQph5Hxakg.0
`TEST-UNEXPECTED-FAIL | testing/enterprise/test_felt_restart_works.py AppRestartWorks.test_app_restart_works | AssertionError: Extra FELT browser process(es) after restart (double relaunch?): {7316} `

I assume we need to update this to take into account the launcher process we let alive? This is in fact a good opportunity to make sure the launcher process stays alive and reports browser exit code correctly over multiple restarts for example

lissyx (changes requested):
LGTM but you need to fix the regressed tests. I'd love we take the opportunity to triple check that with the launcher process we do not loose any exit code or anything in various cases. It should already be covered in existing tests, right?

testing/enterprise/manifest-browser.toml
```diff
@@ -72,6 +72,11 @@ run-if = [
 
 ["test_felt_browser_starts.py"]
 
+["test_felt_browser_third_party_blocklist.py"]
+run-if = [
+  "os == 'win'",
+]
```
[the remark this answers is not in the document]
gregstoll: yeah, probably a good idea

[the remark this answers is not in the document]
lissyx: (we have no asan for now)

toolkit/components/felt/content/FeltProcessParent.sys.mjs
gregstoll: This comment could use a bit of cleaning up (I'd prefer just explaining what the flags do instead of "Neither flag is removable" :-) )

In particular, is there a reason we need `--no-deelevate`?

testing/enterprise/test_felt_browser_third_party_blocklist.py
```diff
@@ -0,0 +1,35 @@
...
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_tests import FeltTests
+
+
+class BrowserThirdPartyBlocklist(FeltTests):
+    def test_browser_dynamic_blocklist_available(self):
+        """The user-level blocklist only reaches the browser through the shared
```
gregstoll: nit: I think it might be worth mentioning that we also want the launcher process for other security enhancements, and this is just a convenient way of testing that.

(pretty neat that we can test this!)

gregstoll (approved):

jonathanmendez:
> LGTM but you need to fix the regressed tests. I'd love we take the opportunity to triple check that with the launcher process we do not loose any exit code or anything in various cases. It should already be covered in existing tests, right?

I updated the regressed restart test to actively check for the launcher process when applicable.

The other scenarios I can think of are covered by existing tests. Passing back an exit code is specifically exercised by the crash tests. We have tests exercising various quit scenarios, etc. or passing along --safe-mode for that direction. Let me know if you think of a specific gap to add.

testing/enterprise/test_felt_browser_third_party_blocklist.py
jonathanmendez: Actually renamed the test/class to better reflect the more general purpose.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -874,8 +874,20 @@ export class FeltProcessParent extends JSProcessActorParent {
       profileArgs = ["-P", foundProfile.name];
     }
 
+    // Use the launcher process so the browser gets the DLL blocklist and other
+    // security enhancements patched in while suspended.
+    // --wait-for-browser keeps the launcher alive to relay the browser's exit code
+    // --no-deelevate keeps the browser at felt's integrity level so that the ipc
+    // socket is accessible to both the browser and felt. (Felt running elevated
+    // is probably bad, but unlikely due to felt's own launcher process)
```
lissyx, line 882: Does it means we should file a follow up / improve some sandboxing somewhere to avoid `--no-deelevate` ? If so, we should do it.

lissyx (approved):
r=me with comments addressed

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jonathanmendez: Hopefully cleaned up the comment well enough.

> In particular, is there a reason we need `--no-deelevate`?

Discussed this. The `--no-deelevate` is likely to not be important (but harmless) in expected scenarios, because the top-level launcher process that launches felt will de-elevate felt itself.
However, in the rare situation that someone does decide to force launching firefox elevated, a) getting to the browser elevated seems to match intent, and b) de-elevating would currently yield an unhelpful error dialog about authentication:
image

We could potentially revisit actively rejecting an elevated browser at the felt level (with a better dialog and error handling), but I'm not sure it feels like felt's job.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
jonathanmendez: See some extra context in [my reply here](https://github.com/mozilla/enterprise-firefox/pull/1463#discussion_r3993382236). I can file a follow-up but my point stands that it doesn't feel like felt's job to me.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
lissyx: @jonathanmendez this error is the socket issue mentionned in the comment? you replied to greg's one, but I would personnally feel safer if we had no such arg passed. Allowing the IPC socket is likely a solvable problem in a follow up, and once the socket is authorized, we can remove `--no-deelevate`

jonathanmendez: Yes the image is what happened when I removed the `--no-deelevate` from `FeltProcessParent`, then invoked firefox.exe from an elevated terminal session passing `--no-deelevate` (so that felt itself would stay elevated). I didn't study the logs too hard but I thought I saw something about the ipc connection failing.

I searched for `--no-delevate` and found various support cases where people seemed to need to be able to run Firefox this way (mostly related to drag-and-drop). We recommend it as a (non-preferred) option at https://support.mozilla.org/en-US/kb/windows-administrator-launcher-process-error-fix

So I'm not sure what the end goal should be. If we leave `--no-delevate` off in `FeltParentProcess` then we remove the escape hatch for users to run the browser elevated, i.e. that becomes unsupported in Firefox Enterprise. Is that the intent you're aiming for?

lissyx: I think running with the least privileges is always better, this is what `--no-deelevate` allows right?

jonathanmendez: We discussed this directly. The main concern was whether the launcher might *elevate* the spawned process. After double-checking the code I am confident it does not (and doing so is contrary to its purpose).

So for now, it is ok to leave `--no-deelevate` to respect if a user is trying to intentionally run elevated and bypass the normal de-elevation that happens in *Felt's* launcher. I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=2072012 to follow-up on the IPC socket connection if we want to pursue that sometime.

PR #1465, opened 2026-09-11
Bug 2071207 - Prepare updates when the console schedules a browser restart

gcp:
Fixes [Bug 2071207](https://bugzilla.mozilla.org/show_bug.cgi?id=2071207): a console-requested restart could relaunch Firefox before FELT had prepared its update. Firefox now requests preparation when the console supplies a restart deadline, while the user continues browsing.

### Behavior

- **Restart:** Any browser restart applies a ready update by restarting FELT. Unfinished downloads never delay the restart, so an immediate deadline can still precede update readiness.
- **Retry:** A console deadline requests an update check immediately, so a restart scheduled to apply a freshly published update finds it. While the deadline stands, Firefox re-requests a check after five minutes, then doubles the spacing up to `app.update.interval` (six hours by default). Once an update is downloading or staged, a re-check is local and does not contact the update server. A withdrawn and re-issued deadline starts over with an immediate check. Startup and captive-portal recovery checks are unaffected.
- **Notification:** Staging an update while browsing triggers the existing “Restart to update” badge on staging-enabled builds.
- **Lifecycle:** Update work is serialized; reopening FELT refreshes its UI immediately. Captive portals suspend preparation. Elevation-bookkeeping errors cannot prevent an update restart.

### Validation

- Native/resource builds and lint pass.
- Four enterprise xpcshell files and the browser enforcer test pass (30 browser assertions).
- Integration coverage passes for update-check IPC, update restarts with elevation failure, and captive portals.
- [CI for the current revision](https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-pr&revision=f7959efbb480) is running.

### DIFF
--- a/toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
+++ b/toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
@@ -18,6 +18,7 @@ ChromeUtils.defineLazyGetter(lazy, "log", () => {
 });
 
 const MS_PER_MINUTE = 60 * 1000;
+const UPDATE_CHECK_INTERVAL_MS = 5 * MS_PER_MINUTE;
 
 // Grace granted to a freshly launched session when the console names none.
 const DEFAULT_GRACE_PERIOD_MINUTES = 10;
@@ -55,7 +56,7 @@ const REPLACEABLE_IDS = [
  */
 export const RelaunchEnforcer = {
   _schedule: null,
-  _updateCheckRequested: false,
+  _lastUpdateCheck: null,
   _restartTask: null,
   _escalationTask: null,
   _countdownTask: null,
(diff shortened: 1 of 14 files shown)


toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -173,6 +175,23 @@ export const RelaunchEnforcer = {
...
+    if (
+      this._lastUpdateCheck === null ||
+      now - this._lastUpdateCheck >= UPDATE_CHECK_INTERVAL_MS ||
+      now < this._lastUpdateCheck
+    ) {
+      try {
+        this._requestUpdateCheck();
+        this._lastUpdateCheck = now;
+      } catch (e) {
+        if (e.result === Cr.NS_ERROR_NOT_CONNECTED) {
```
lissyx, line 198: If browser is not connected to felt, all bets are off :(

lissyx (approved):
r=me, it's a bit of a changes, but I do not think it's that risky

lissyx:
Should this be uplifted to beta?

gcp:
>I do not think it's that risky

I'm a bit nervous after one of the bots complained this interferes with captive portal handling tbh.

toolkit/components/felt/content/FeltProcessParent.sys.mjs
```diff
@@ -191,101 +200,93 @@ export class FeltProcessParent extends JSProcessActorParent {
...
+                  lazy.log.error("ParentProcess: getReadyUpdate failed", err);
+                  return null;
+                });
+                const pendingUpdate = [
+                  "pending",
+                  "pending-service",
+                  "pending-elevate",
+                  "applied",
+                  "applied-service",
+                ].includes(readyUpdate?.state);
```
lissyx, line 248: On the bug you said:
> I think I'll make Felt *always* check for an update before restarting Firefox. This avoids "I just relaunched and now I need to again" which you can currently get and is super annoying.

But this code does not perform any force check here

gcp: We discussed this; it only checks the moment it gets the notification from the console that a restart is requested. We don't check every restart (yet), that would need some discussion wrt pacing/backoff.

lissyx (approved):

PR #1466, opened 2026-09-11
Bug 2071248 - Add Thunderbird handling for build variant

lissyx:
### Description 

Bugzilla: Bug-2071248

### DIFF
--- a/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
+++ b/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
@@ -689,16 +689,21 @@ EnterprisePoliciesManager.prototype = {
   /**
    * The build variant key used by "x-compatibility" in policies-schema.json.
    *
-   * @returns {"firefox_enterprise"|"firefox_esr"|"firefox"} variant key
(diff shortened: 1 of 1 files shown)


1rneh (approved):
When the `x-compatibility` states `version_added !== false` then the policy is categorized as supported which comes in handy for thunderbird right now, as no policies are rejected hence no behavior changes. 

But we need to add `thunderbird` and `thunderbird_esr` variants for each policy compatibility.

lissyx:
> When the `x-compatibility` states `version_added !== false` then the policy is categorized as supported which comes in handy for thunderbird right now, as no policies are rejected hence no behavior changes.
> 
> But we need to add `thunderbird` and `thunderbird_esr` variants for each policy compatibility.

Yes, it should be in https://phabricator.services.mozilla.com/D325360

lissyx:
> > When the `x-compatibility` states `version_added !== false` then the policy is categorized as supported which comes in handy for thunderbird right now, as no policies are rejected hence no behavior changes.
> > But we need to add `thunderbird` and `thunderbird_esr` variants for each policy compatibility.
> 
> Yes, it should be in https://phabricator.services.mozilla.com/D325360

it's ready for review, and since the current PR is merged, there is nothing blocking it anymore :)

PR #1470, opened 2026-09-11
Enterprise main merge 2026091100

jonathanmendez:
### Description

Bugzilla: NO BUG  

Resolved conflicts:

*   .cargo/config.toml.in
*   supply-chain/config.toml
*   taskcluster/kinds/generate-profile/kind.yml
*   taskcluster/kinds/instrumented-build/kind.yml


jporter-dev (approved):

jonathanmendez:
I amended the final commit to trigger a new PR since toolchains failed too many retried in the original. The original try is at https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=94e48dbab901cf33faa433f35842237d7c290e06

jonathanmendez:
There was an almost consistent new test failure in `test_felt_device_posture_refresh_rejected.py` (just one of the various platforms' runs passed, the rest failed). Used claude to diagnose the failure as an existing race condition.

### Previously:
The test launched the browser, then edited the profiles `extensions.json` to fake that a new extension had been installed to trigger a change in posture. However, that file is overwritten by its owner during startup, which blaps the faked extension in a race condition (it looks like this merge changed the timing to make this fairly likely).

I used test instrumentation to prove that the faked extension was being blapped, although I did not directly verify what was doing the blapping. The late-added items were extra themes, which don't trigger a posture change.

### Now:
Instead of faking an extension, use the existing felt test helpers to force-install an extension by policy to trigger the change in posture. Since it's a true extension installation it goes through the normal ownership of the file instead of an outside write that could be blapped.

PR #1471, opened 2026-09-11
Bug 2068749 - Read AutoConfig inputs from the default branch on enterprise builds

farmisen:
Fix: https://bugzilla.mozilla.org/show_bug.cgi?id=2068749
Should also close: https://bugzilla.mozilla.org/show_bug.cgi?id=1632281
## Description

Reads autoadmin.global_config_url from the default branch so a user profile 
can no longer redirect AutoConfig and bypass the managed policies.
A user value, from the profile or from pref() in the .cfg, is reported to the
console and dropped. `general.config.vendor` and `general.config.filename` keep
working with pref() in the .cfg: their profile-set user values are dropped
before the .cfg is evaluated, so user.js cannot fail the vendor check.

Three xpcshell tests cover the user.js URL, the user.js vendor and filename,
and pref()/defaultPref() from the .cfg.

Supersedes #1452.

### DIFF
--- a/extensions/pref/autoconfig/src/nsReadConfig.cpp
+++ b/extensions/pref/autoconfig/src/nsReadConfig.cpp
@@ -220,9 +220,18 @@ nsresult nsReadConfig::readConfigFile() {
   // file we allow for the preference to be set (and locked) by the creator
   // of the cfg file meaning the file can not be renamed (successfully).
 
+  // On enterprise builds the AutoConfig inputs below are read from the default
+  // branch so a profile's user.js can neither redirect AutoConfig to another
+  // file, make the vendor check fail, nor point autoadmin.global_config_url at
+  // an arbitrary script. An administrator configures these from
+  // the shipped firefox.cfg, which sets them on the default branch.
   nsCOMPtr<nsIPrefBranch> prefBranch;
+#if defined(MOZ_ENTERPRISE)
+  prefBranch = defaultPrefBranch;
+#else
   rv = prefService->GetBranch(nullptr, getter_AddRefs(prefBranch));
   NS_ENSURE_SUCCESS(rv, rv);
+#endif
 
   int32_t obscureValue = 0;
   (void)defaultPrefBranch->GetIntPref("general.config.obscure_value",
--- a/extensions/pref/autoconfig/test/unit/autoconfig-admin.cfg
+++ b/extensions/pref/autoconfig/test/unit/autoconfig-admin.cfg
@@ -1,0 +1,2 @@
+// The first line is skipped by the AutoConfig parser.
+lockPref("_autoconfig_.test.admin", "cfg");
--- a/extensions/pref/autoconfig/test/unit/autoconfig-remote.jsc
+++ b/extensions/pref/autoconfig/test/unit/autoconfig-remote.jsc
@@ -1,0 +1,2 @@
+// Stands in for a remote AutoConfig file a profile's user.js points at.
+pref("_autoconfig_.test.remote", "remote");
--- a/extensions/pref/autoconfig/test/unit/test_autoconfig_user_url.js
+++ b/extensions/pref/autoconfig/test/unit/test_autoconfig_user_url.js
@@ -1,0 +1,79 @@
+/* Any copyright is dedicated to the Public Domain.
+   http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// On an enterprise build a profile's user.js cannot point AutoConfig at
+// another file through autoadmin.global_config_url. Once the shipped
+// firefox.cfg has been evaluated the remaining AutoConfig inputs are never
+// consulted, so the user-set URL is never fetched.
+
+const { AppConstants } = ChromeUtils.importESModule(
+  "resource://gre/modules/AppConstants.sys.mjs"
+);
+
+function run_test() {
+  if (!AppConstants.MOZ_ENTERPRISE) {
+    info("Skipping on non-enterprise build (MOZ_ENTERPRISE is not set)");
+    ok(true, "skipped: not an enterprise build");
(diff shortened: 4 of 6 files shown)


mkaply:
This shouldn't be enterprise specific. Setting this as a default pref is a bug we had a long time ago and never fixed:

https://bugzilla.mozilla.org/show_bug.cgi?id=1632281

farmisen:
> This shouldn't be enterprise specific. Setting this as a default pref is a bug we had a long time ago and never fixed:
> 
> https://bugzilla.mozilla.org/show_bug.cgi?id=1632281

thanks for the pointer to Bug 1632281. Reworked in the second commit:
- `autoadmin.global_config_url` is now read from the default branch on all builds. A user value, whether from user.js or from a pref() call in the .cfg, is reported to the console and cleared, so a  misconfigured pref() no longer sits in the profile looking like it works.
- `general.config.vendor` and `general.config.filename` have their profile-set user values are dropped in readConfigFile before the .cfg  is evaluated. That keeps pref() working for them from a .cfg while user.js  can no longer fail the vendor check.

mkaply:
Thanks!

I think Claude was probably a little overzealous in the commenting here :). Might be worth asking it to be a little less verbose. And also verify that all those tests are necessary.

Have you done a change to core Firefox yet? If not, I can work with you to do the corresponding Firefox change.

mkaply (changes requested):
I'm sorry, I've completely changed my tune after looking at this more.

I think we should just clear all the values and not switch the read to default. I am worried that this patch breaks existing deployments and there is no way to know.

So we should just clear everything before the cfg is parsed:

    Preferences::ClearUser("general.config.filename");
    Preferences::ClearUser("general.config.vendor");
    Preferences::ClearUser("autoadmin.global_config_url");
    Preferences::ClearUser("autoadmin.offline_failover");
    Preferences::ClearUser("autoadmin.append_emailaddr");
    Preferences::ClearUser("autoadmin.refresh_interval");
    Preferences::ClearUser("autoadmin.failover_to_cached");

The four extra `autoadmin.` prefs are the rest of the hole — `nsAutoConfig`
reads all of them off the root branch, so a profile can still defang AutoConfig
even once the URL is locked down. Put the block after the
`general.config.filename` default-branch check so non-AutoConfig builds don't
wipe profile values on the way to the early return. Note `append_emailaddr`'s
suffix comes from `mail.*` prefs that are legitimately profile data, so this
controls whether a suffix is appended, not what's in it.

Then drop the `HasUserValue` guard — `Preferences::ClearUser` already does that
lookup and returns `NS_OK` when there's no user value (`Preferences.cpp:6490`).

**Tests are heavier than the change.** There are four contracts: profile values
can't fail the vendor check, profile values don't set the URL, `.cfg` `pref()`
still sets vendor, and `pref()` vs `defaultPref()` for the URL.
`test_autoconfig_cfg_pref.js` phase 2 already covers the first three. One file
with four sections and one shared setup would cover the same ground, and since
all three are `run-sequentially` (good catch on the `GreD/autoconfig.cfg`
collision) it turns three serialized xpcshell startups into one.

Also please cite `nsAppRunner.cpp:6183-6199` in the comment instead of asserting
the ordering bare - the `prefservice:before-read-userprefs` topic name says the
opposite and sent me down the wrong path for a while. I confirmed the real order
by having a `.cfg` `getPref()` a value only user.js set; the `.cfg` saw it.

Also, can you file a followup in regular Firefox to just remove general.config.vendor completely? It's not really needed.

farmisen:
> I'm sorry, I've completely changed my tune after looking at this more.
> 
> I think we should just clear all the values and not switch the read to default. I am worried that this patch breaks existing deployments and there is no way to know.
> 
> So we should just clear everything before the cfg is parsed:
> 
> ```
> Preferences::ClearUser("general.config.filename");
> Preferences::ClearUser("general.config.vendor");
> Preferences::ClearUser("autoadmin.global_config_url");
> Preferences::ClearUser("autoadmin.offline_failover");
> Preferences::ClearUser("autoadmin.append_emailaddr");
> Preferences::ClearUser("autoadmin.refresh_interval");
> Preferences::ClearUser("autoadmin.failover_to_cached");
> ```
> 
> The four extra `autoadmin.` prefs are the rest of the hole — `nsAutoConfig` reads all of them off the root branch, so a profile can still defang AutoConfig even once the URL is locked down. Put the block after the `general.config.filename` default-branch check so non-AutoConfig builds don't wipe profile values on the way to the early return. Note `append_emailaddr`'s suffix comes from `mail.*` prefs that are legitimately profile data, so this controls whether a suffix is appended, not what's in it.
> 
> Then drop the `HasUserValue` guard — `Preferences::ClearUser` already does that lookup and returns `NS_OK` when there's no user value (`Preferences.cpp:6490`).
> 
> **Tests are heavier than the change.** There are four contracts: profile values can't fail the vendor check, profile values don't set the URL, `.cfg` `pref()` still sets vendor, and `pref()` vs `defaultPref()` for the URL. `test_autoconfig_cfg_pref.js` phase 2 already covers the first three. One file with four sections and one shared setup would cover the same ground, and since all three are `run-sequentially` (good catch on the `GreD/autoconfig.cfg` collision) it turns three serialized xpcshell startups into one.
> 
> Also please cite `nsAppRunner.cpp:6183-6199` in the comment instead of asserting the ordering bare - the `prefservice:before-read-userprefs` topic name says the opposite and sent me down the wrong path for a while. I confirmed the real order by having a `.cfg` `getPref()` a value only user.js set; the `.cfg` saw it.
> 
> Also, can you file a followup in regular Firefox to just remove general.config.vendor completely? It's not really needed.

Addressed everything but the followup - on it this morning.

mkaply (approved):

PR #1472, opened 2026-09-11
Bug 2071459 - Add forced signout and relaunch delegation for Thunderbird

jporter-dev:
### Description 

Bugzilla: Bug-2071459

Add application-specific delegation points needed by Thunderbird for enterprise forced sign-out and relaunch enforcement. Thunderbird requires handling forced signouts differently since it is missing some browser-specific pieces. Additionally, Thunderbird does not support several components of `RelaunchEnforcer.sys.mjs` such as `InfoBar.sys.mjs`, `navigator:browser`, etc. 

To support Thunderbird while keeping Firefox behavior unchanged, this PR adds two delegation points:

- Pre-shutdown hook registration in `ConsoleClient.sys.mjs` to allow application-specific pre-shutdown logic. `quitIgnoringCanClose()` is now async and awaits the hook (quitting anyway on failure) before requesting the quit.
- Warning UI delegate registration in `RelaunchEnforcer.sys.mjs` to allow an application to register its own warning UI. The delegate must implement `showOrUpdate()`, `hide()`, and `isVisible()`, and is kept up to date throughout the relaunch warning lifecycle.

The existing InfoBar warning UI moves out of toolkit into the new `EnterpriseForcedQuit.sys.mjs`, which registers both the hook and the delegate for Firefox at `app-startup` via `components.conf`


lissyx:
> Additionally, Thunderbird does not support several components of `RelaunchEnforcer.sys.mjs` such as `InfoBar.sys.mjs`, `navigator:browser`, etc.

I thought there was an alternative InfoBar?

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -691,21 +696,48 @@ export const ConsoleClient = {
...
+    if (typeof aHook !== "function") {
+      throw new TypeError("The before-forced-quit hook must be a function.");
+    }
+    this._beforeForcedQuitHook = aHook;
+    return () => {
+      if (this._beforeForcedQuitHook === aHook) {
+        this._beforeForcedQuitHook = null;
+      }
+    };
+  },
```
lissyx: LGTM but worth @1rneh having a look as well ?

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -691,21 +696,48 @@ export const ConsoleClient = {
...
+    if (this._beforeForcedQuitHook) {
+      try {
+        await this._beforeForcedQuitHook(aFlags);
+      } catch (error) {
+        lazy.log.error("Pre-forced-quit hook failed; quitting anyway.", error);
+      }
+    } else {
+      for (const win of Services.wm.getEnumerator("navigator:browser")) {
+        win.skipNextCanClose = true;
+      }
```
lissyx: If we introduce application specific hook, we should not `if / else` here, and rather make use of it in Firefox as well

lissyx: i.e.,
```
      for (const win of Services.wm.getEnumerator("navigator:browser")) {
        win.skipNextCanClose = true;
      }
```

Should be part of the `aHook` on Firefox

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -66,6 +66,48 @@ export const RelaunchEnforcer = {
...
+   * a callback for the warning's restart action. It returns whether the warning
+   * is visible.
+   *
+   * @param {object} aDelegate - The warning UI delegate.
+   * @param {function(object): (boolean|Promise<boolean>)} aDelegate.showOrUpdate
+   * @param {function(): void} aDelegate.hide
+   * @param {function(): boolean} aDelegate.isVisible
+   * @returns {void}
+   */
+  registerWarningUIDelegate(aDelegate) {
```
lissyx, line 79: Why not making use of it for Firefox as well?

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -478,29 +590,39 @@ export const RelaunchEnforcer = {
...
+      // hand. Skipping the animation removes the element, and runs InfoBar's
+      // own teardown, before returning.
+      let removed = false;
+      for (const win of Services.wm.getEnumerator("navigator:browser")) {
+        for (const id of [WARNING_ID, IMMINENT_ID]) {
+          const bar = win.gNotificationBox?.getNotificationWithValue(id);
+          if (bar) {
+            win.gNotificationBox.removeNotification(bar, true);
+            removed = true;
+          }
```
lissyx: I do not think it's wise to special case like this, and we should rather use `warningUIDelegate` on Firefox as well

lissyx (changes requested):
I think the idea holds, assuming the ConsoleClient changes are also OK by @1rneh, but I do not think we we should special case between Firefox and Thunderbird here, rather make use of the new system for Firefox as well.

jporter-dev:
> I think the idea holds, assuming the ConsoleClient changes are also OK by @1rneh, but I do not think we we should special case between Firefox and Thunderbird here, rather make use of the new system for Firefox as well.

Good call, I had planned to do that as part of my refinement of this, but just wanted to make sure it was the right approach before moving on. I added `browser/components/enterprise/EnterpriseForcedQuit.sys.mjs` to handle this for Firefox, and Thunderbird should be able to do the same.

browser/components/enterprise/tests/browser/browser.toml
```diff
@@ -0,0 +1,8 @@
+[DEFAULT]
+environment = ["MOZ_BYPASS_FELT=1"]
```
lissyx: This is already set by harness, please do not.

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
```diff
@@ -0,0 +1,267 @@
...
+      content: {
+        priority: isImminent
+          ? win.gNotificationBox.PRIORITY_CRITICAL_HIGH
+          : win.gNotificationBox.PRIORITY_INFO_HIGH,
+        type: "universal",
+        dismissable: false,
+        text: {
+          string_id: isImminent
+            ? "enterprise-relaunch-imminent-message"
+            : "enterprise-relaunch-warning-message",
```
lissyx, line 98: Those strings lives currently in `toolkit` and they refer to "tabs": https://searchfox.org/enterprise-main/rev/c745341954c2ebdbad1cd8b9c27f927e0e26ca51/toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl#137-147.

We probably want to make sure either they are in `browser` (moving files, we need to check with l10n how it works with IDs) or reword to not mention "tabs"

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -63,14 +50,56 @@ export const RelaunchEnforcer = {
...
   _shownMinutes: null,
   _shownDeadlineMinute: null,
   _restarting: false,
   _awaitingSessionRestore: false,
   // Serializes bar updates against the shown-state above.
   _refreshChain: Promise.resolve(),
+  // Delegate the warning UI to the application, if it registered one.
+  _warningUIDelegate: null,
+  // Invalidates delegated updates and restart actions when the warning hides.
+  _warningUIGeneration: 0,
```
lissyx: I'm not sure I grasp the scope of this

lissyx: It looks like a bad workaround for a case of improperly balanced show/hide calls ?

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -63,14 +50,56 @@ export const RelaunchEnforcer = {
...
+      );
+    }
+    if (this._warningUIDelegate) {
+      throw new Error("A warning UI delegate is already registered.");
+    }
+    if (this._hasProcessedConsolePoll) {
+      throw new Error(
+        "The warning UI delegate must be registered before console polling starts."
+      );
+    }
```
lissyx: Why? Looks like a potential race here, if there's such a dependency, the ordering should be fixed directly?

lissyx: Checking how `onConsolePoll()` is called, I'm not sure this is worth throwing ...
https://searchfox.org/enterprise-main/rev/c745341954c2ebdbad1cd8b9c27f927e0e26ca51/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs#1611-1612,1639

Especially we call `registerWarningUIDelegate()` from `EnterpriseForcedQuit`'s
```
  init() {
    lazy.RelaunchEnforcer.registerWarningUIDelegate(this.warningUI);
```

So there's already a comment:
```
/**
 * The app-startup entry point (see components.conf): both registrations must
 * land before "policies-startup" delivers the first console poll, which runs
 * right after the app-startup category.
 */
```

This needs to be tripled checked. Also given `onConsolePoll` will depend on the network round-trip, how can we be certain that this dependency is respected? Sounds like a recipe for race condition. Maybe it should use promise or/and properly wait for events?

If we throw because of a race condition, we may:
 - not have a way to be notified (logs, error message, crash etc.), so the exception is thrown for nothing
 - it silently breaks relaunch enforcement notification

And why does the **registration** should depend on the policies processing? It's just registering handlers to show things later.

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -337,209 +367,94 @@ export const RelaunchEnforcer = {
...
+  async _updateDelegatedWarning() {
     if (!this._schedule || this._restarting) {
       return;
     }
-
-    const win = this._barWindow();
-    if (!win) {
-      // The first poll precedes session restore; the next poll retries.
+    const delegate = this._warningUIDelegate;
+    if (!delegate) {
```
lissyx, line 387: Having no delegate should instead be considered a critical condition, because we will fail to notify the user, which is the whole point

lissyx: Are there legitimate cases where having no delegate is an acceptable state?

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
jporter-dev: Removed the ordering dependency. It instead now updates on delegate registration instead, and a relaunch deadline with no delegate registered raises an error.

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
jporter-dev: This was leftover from before I moved the Firefox implementation to leverage the delegation, in that scenario this made sense but it no longer applies. 

Updated this to throw if no delegate is registered since we cannot notify a user without one now.

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
jporter-dev: Good call out, how did you handle this for the browser -> toolkit FELT move? Did you create a migration? https://firefox-source-docs.mozilla.org/l10n/migrations/fluent.html#moving-messages-to-a-different-file

I think it may make sense to keep these application-specific, at least in some sense, since Thunderbird may want to add their own qualifying text. Maybe keep the main portion strings in `toolkit` and move the suggestion "Tabs will reopen." at the end to `browser` and `mail`? 

e.g. https://searchfox.org/enterprise-main/source/toolkit/locales/en-US/toolkit/enterprise/enterprise.ftl#137-149

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
```diff
@@ -0,0 +1,262 @@
...
+  IMMINENT_ID,
+  "COMPULSORY_RESTART_SCHEDULED",
+  "INFOBAR_ACTION_86",
+  "INFOBAR_DEFAULT_AND_PIN_87",
+  "INFOBAR_LAUNCH_ON_LOGIN",
+  "INFOBAR_LAUNCH_ON_LOGIN_FINAL",
+  "MULTIPROFILE_DATA_COLLECTION_CHANGED_INFOBAR",
+  "PREF_OBSERVER_MESSAGE_94",
+  "updated-privacy-notice-notification-infobar",
+];
```
lissyx: I'm not confident with this hardcoded list :(

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
```diff
@@ -0,0 +1,262 @@
...
+            : "enterprise-relaunch-warning-message",
+        },
+        buttons: [
+          {
+            label: { string_id: "enterprise-relaunch-restart-now" },
+            action: { type: "RESTART_APP", dismiss: false },
+          },
+        ],
+        attributes: isImminent ? { minutes } : { datetime: restartAt },
+        canReplace: REPLACEABLE_IDS,
```
lissyx: Given the message we are delivering, I think we should be replacing any previous one?

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
lissyx: I dont think we had a problem back then, but I'd feel better with l10n approval here

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
lissyx: At least in the "imminent" case?

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -337,209 +361,89 @@ export const RelaunchEnforcer = {
...
+      throw new Error(
+        "No relaunch warning UI delegate is registered; the user cannot be notified."
+      );
     }
-
     const { restartAt } = this._schedule;
     const remaining = restartAt - Date.now();
     const isImminent = remaining <= IMMINENT_THRESHOLD_MS;
```
lissyx: If we cannot always replace / show our message, we should change the `imminent` threshold, 5 minutes is likely not enough to give people enough time

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
jporter-dev: It's just ported over from `RelaunchEnforcer` https://searchfox.org/enterprise-main/source/toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs#42-53

lissyx: OK, at the very lease we should have a follow up to clean this as soon as we can. Or at least a way to make sure we can keep those in sync, but given it's sourced from random various places ...

testing/enterprise/test_felt_browser_forced_signout.py
```diff
@@ -0,0 +1,73 @@
...
+
+sys.path.append(os.path.dirname(__file__))
+
+from felt_browser_starts import FeltStartsBrowser
+
+
+class BrowserForcedSignout(FeltStartsBrowser):
+    def test_forced_signout_waits_for_hook(self):
+        self.run_felt_base()
+        self.run_felt_browser_started()
```
lissyx: If we change this to just connecting, we can make the whole test browser-agnostic and running on thunderbird (dont forget to move to `manifest-generic.toml`)
```suggestion
        self.connect_child_browser()
```

testing/enterprise/test_felt_browser_forced_signout.py
```diff
@@ -0,0 +1,73 @@
...
+
+        marker_path = os.path.join(self._child_profile_path, "forced-signout-hook")
```
lissyx: Please find something more decriptive that `marker`

browser/components/enterprise/tests/browser/browser.toml
lissyx: This should have a `[DEFAULT]` section ?

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -691,21 +696,47 @@ export const ConsoleClient = {
   },
 
   /**
-   * Quit Firefox, ignoring any callbacks installed by the page
-   * preventing the tab/window from closing.
+   * Registers application-specific pre-shutdown logic to run before a forced
+   * quit, such as flushing state or suppressing close vetoes. Registration is
+   * last-write-wins: a later hook replaces the current one.
```
lissyx: I'm wondering if it's a correct design: maybe we should in fact not allow any overwrite?
And provide a `unregisterBeforeForcedQuitHook` as well?
And assert/throw to ensure balanced calls?

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
jporter-dev: I agree, hardcoding this isn't great, but not seeing an immediate way to make this easily dynamic. I can file a bug for followup to try to improve this bit.

testing/enterprise/test_felt_browser_forced_signout.py
jporter-dev: Done, thanks for noting this, will have to keep in mind for future shared tests. Renamed this to remove `browser` from the name too.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
jporter-dev: Updated this to match the approach I used in `RelaunchEnforcer` which does exactly what you described

browser/components/enterprise/components.conf
```diff
@@ -0,0 +1,16 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+Classes = [
+    {
+        'cid': '{32448648-9fb5-4e17-8b71-31ce4183a18c}',
+        'contract_ids': ['@mozilla.org/browser/enterprise-forced-quit;1'],
+        'esModule': 'resource:///modules/enterprise/EnterpriseForcedQuit.sys.mjs',
+        'constructor': 'EnterpriseForcedQuitStartup',
+        'categories': {'app-startup': 'EnterpriseForcedQuit'},
```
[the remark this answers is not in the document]
jporter-dev: Moved it from `app-startup` to `profile-after-change`

browser/components/enterprise/tests/browser/browser.toml
```diff
@@ -0,0 +1,3 @@
+[DEFAULT]
+
+["browser_relaunch_ui.js"]
```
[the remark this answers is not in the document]
jporter-dev: https://github.com/mozilla/enterprise-firefox/pull/1472#discussion_r4044109668

It was noted in a previous comment here, that this is not needed?

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
```diff
@@ -0,0 +1,262 @@
...
+      )
+    ) {
+      return true;
+    }
+
+    const win = this._barWindow();
+    if (!win) {
+      // The first poll precedes session restore; the next poll retries.
+      return false;
+    }
```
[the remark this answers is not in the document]
jporter-dev: `isVisible()` short-circuits on `!!this._notification` before reaching `lazy.InfoBar`, so it is non-null only once `InfoBar` has already loaded. On the first poll with no window it is null, the lazy getter is never dereferenced, and no module loads. `_barWindow()` is guarded the same way, the `lazy.InfoBar` reference is the third operand of an && inside the loop over navigator:browser, so it is unreachable with zero windows. The lazy-loading guarantee documented in `_barWindow()` holds.

toolkit/components/enterprise/modules/RelaunchEnforcer.sys.mjs
```diff
@@ -337,209 +361,89 @@ export const RelaunchEnforcer = {
...
-
-    const win = this._barWindow();
-    if (!win) {
-      // The first poll precedes session restore; the next poll retries.
-      return;
+    const delegate = this._warningUIDelegate;
+    if (!delegate) {
+      throw new Error(
+        "No relaunch warning UI delegate is registered; the user cannot be notified."
+      );
```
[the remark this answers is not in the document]
jporter-dev: I moved both delegates to lookup-only categories that are referenced at our delegation points.

browser/components/enterprise/tests/browser/browser.toml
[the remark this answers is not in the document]
lissyx: Yes, only the env var is not needed :)

lissyx (approved):

lissyx (changes requested):
"enterprise-forced-quit-hook"

lissyx:
> "enterprise-forced-quit-hook"

GitHub's UI was failing to find it

lissyx (approved):

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
jporter-dev: Resolving this since we determined this can stay as it is with mentioning "tabs" since Thunderbird has tabs as well. 

I did reach out to l10n for clarification on how moving this would be handled in our repo, and they confirmed the migration scripts will work, but there is no formal process or automation. Had I moved this and created a migration, it would have been run manually. If we'll have more string movement in the future, they'd want to set up a process for this, and we would need to choose an enterprise-specific location for our migration scripts (theirs are [here](https://searchfox.org/firefox-main/source/python/l10n/fluent_migrations)).

browser/components/enterprise/EnterpriseForcedQuit.sys.mjs
jporter-dev: Created bug-2074600 for followup

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -691,21 +707,113 @@ export const ConsoleClient = {
...
+            }, timeoutMs);
+          }),
+        ]);
+      } catch (error) {
+        lazy.log.error("Pre-forced-quit hook failed; quitting anyway.", error);
+      } finally {
+        lazy.clearTimeout(timeoutId);
+      }
     }
     Services.startup.quit(aFlags);
```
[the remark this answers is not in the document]
jporter-dev: Correct on the mechanism, but suppressing close vetoes is inherently application-specific.. skipNextCanClose is a `navigator:browser` window property, and toolkit has no equivalent for an application whose windows it knows nothing about. Moving it behind the hook is the point of the patch; a toolkit-side fallback would put browser logic right back here.

I've updated the JSDoc to explain this to any callers.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -691,23 +711,130 @@ export const ConsoleClient = {
...
+        ChromeUtils.importESModule(url).registerForcedQuitHook();
+      } catch (e) {
+        lazy.log.error(
+          `Failed to register the forced-quit hook from ${url}:`,
+          e
+        );
+      }
+    }
+    return this._beforeForcedQuitHook;
+  },
```
lissyx: How much does this live on `ConsoleClient` now? Maybe move it to its own module, reduces the changes to ConsoleClient

lissyx (approved):

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
```diff
@@ -691,23 +709,85 @@ export const ConsoleClient = {
...
+  _appForcedQuitHook() {
+    if (this._forcedQuitHook !== undefined) {
+      return this._forcedQuitHook;
+    }
+    this._forcedQuitHook = null;
+    for (const { value: url } of Services.catMan.enumerateCategory(
+      FORCED_QUIT_HOOK_CATEGORY
+    )) {
+      try {
+        this._forcedQuitHook = ChromeUtils.importESModule(url).beforeForcedQuit;
```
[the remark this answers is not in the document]
jporter-dev: This is fine, that prototype is just for reference.

toolkit/components/enterprise/modules/ConsoleClient.sys.mjs
jporter-dev: Thanks, that's a good call since this has grown a bit. I did remove the "registration" bits from here and instead leverage the added category instead. But still moved it out to `ForcedQuitHandler` to clean this up.

lissyx (approved):
With the recent changes, we are not touching anymore ConsoleClient, so I think we are fine there

PR #1473, opened 2026-09-11
Bug 2068747 - Only honor toolkit.policies.perUserDir only in automation

farmisen:
Fix: https://bugzilla.mozilla.org/show_bug.cgi?id=2068747

## Description
Honor toolkit.policies.perUserDir only under Cu.isInAutomation.

Supersedes https://github.com/mozilla/enterprise-firefox/pull/1452.

### DIFF
--- a/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
+++ b/toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
@@ -1288,14 +1288,14 @@ class JSONPoliciesProvider extends PoliciesProvider {
 
     try {
       let configFile;
-      // A deployment choice (Bug 1583466), so only the default branch may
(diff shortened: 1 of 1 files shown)


mkaply:
This was only added for testing, so it shouldn't be enterprise specific. We should only allow it when in automation.

farmisen:
> This was only added for testing, so it shouldn't be enterprise specific. We should only allow it when in automation.

done

mkaply:
Nit:

+// For testing, load browser policy per-user from
+// /run/user/$UID/appname so this is only honored in automation.

This doesn't read correctly.

Should probably be:

Load browser policy per-user from /run/user/$UID/appname for testing only.

Or something like that.

mkaply (approved):
Approved with fix to comment.

And let's get a Firefox bug opened up for this,.

PR #1479, opened 2026-09-14
Bug 2071381 - Ignore null-valued local policies and fail closed on managed init failure

johnreitano:
Bugzilla: Bug-2071381

A null-valued policy in a local `policies.json` aborted enterprise policy initialization. `isEmptyObject()` evaluated `Object.keys(null)` (because `typeof null == "object"`) and threw; since `_updateStatus()` ran outside `_initialize()`'s `try/catch`, the exception escaped **after** a successful console-policy fetch, silently dropping the fetched policies and leaving a managed (felt) browser running unmanaged — a fail-open bypass of enterprise policy enforcement. A local, unprivileged user could trigger it by planting a null-valued local policy.

## Fix

1. **`isEmptyObject()` null-guard** — reject `null` before `Object.keys()`, so a malformed local policy value is ignored rather than fatal.
2. **Fail closed on managed init failure** — move status evaluation and startup activation inside `_initialize()`'s `try`; on any caught failure discard the partial state (status `FAILED`, no partially scheduled policy set) so the engine never advertises `ACTIVE` or applies a partial set, and, for a felt/managed browser, `initiateShutdown()` instead of continuing with the fetched policies unapplied. Consumer builds log and continue. The null-guard comes first so the malformed value is handled harmlessly rather than forcing a shutdown loop.

## Testing

New xpcshell regression tests (all pass locally):
- `test_null_local_policy_ignored.js` — a null-valued local policy is ignored while a valid sibling policy applies (console-free / upstream-shared case).
- `live/test_null_local_policy_preserves_console.js` — a successfully fetched console policy is still applied alongside a null local policy, including a null nested inside a known local policy.
- `live/test_managed_init_fails_closed.js` — a managed (felt) init failure fails closed via `initiateShutdown()`, and a consumer (non-felt) init failure does not shut down; both discard the partial state.

Also verified end-to-end against a live console in a managed FELT browser: with the fix, a planted null-valued local policy is ignored and the console `WebsiteFilter` stays enforced (before the fix it was silently dropped, engine left `UNINITIALIZED`).

https://claude.ai/code/session_01VgtSoD72MLY6iQDj18LvZd


johnreitano:
Superseded by #1480 — identical changes, renamed branch `bug-2071381-crash-on-null-local-policy`. Closing this one.

PR #1480, opened 2026-09-14
Bug 2071381 - Ignore null-valued local policies and fail closed on managed init failure

johnreitano:
Bugzilla: Bug-2071381

A null-valued policy in a local `policies.json` aborted enterprise policy initialization. `isEmptyObject()` evaluated `Object.keys(null)` (because `typeof null == "object"`) and threw; since `_updateStatus()` ran outside `_initialize()`'s `try/catch`, the exception escaped **after** a successful console-policy fetch, silently dropping the fetched policies and leaving a managed (felt) browser running unmanaged — a fail-open bypass of enterprise policy enforcement. A local, unprivileged user could trigger it by planting a null-valued local policy.

## Fix

1. **`isEmptyObject()` null-guard** — reject `null` before `Object.keys()`, so a malformed local policy value is ignored rather than fatal.
2. **Fail closed on managed init failure** — move status evaluation and startup activation inside `_initialize()`'s `try`. On any caught failure discard the partial state through a `_discardPolicies()` helper shared with `_resetEngine()`: status `FAILED`, no parsed policies, no provider (so the remote polling that starts on `EnterprisePolicies:Initialized` cannot re-apply the fetched set and flip the engine back to `ACTIVE`), and only the callbacks scheduled by the failed attempt dropped, so the cleanup scheduled for the previous session's policies still runs. A felt/managed browser then calls `initiateShutdown()` instead of continuing with the fetched policies unapplied; a browser that is not felt managed logs and continues. The null-guard comes first so the malformed value is handled harmlessly rather than forcing a shutdown loop.

The error text stays interpolated into the logged message: about:policies lists only the first argument of each console event, so passing the error solely as a second argument would drop it from the Errors tab.

## Testing

New xpcshell regression tests (all pass locally):
- `test_null_local_policy_ignored.js` — a null-valued local policy is ignored while a valid sibling policy applies (console-free / upstream-shared case).
- `live/test_null_local_policy_preserves_console.js` — a successfully fetched console policy is still applied alongside a null local policy (the scenario from the bug).
- `live/test_managed_init_fails_closed.js` — a managed (felt) init failure fails closed via `initiateShutdown()`, discards the partial state, and a later policy update does not re-activate it; an init failure in a browser that is not felt managed does not shut down, and the previous session's cleanup still runs.

Also verified end-to-end against a live console in a managed FELT browser: with the fix, a planted null-valued local policy is ignored and the console `WebsiteFilter` stays enforced (before the fix it was silently dropped, engine left `UNINITIALIZED`).

https://claude.ai/code/session_01VgtSoD72MLY6iQDj18LvZd


toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
bsmth, line 248: Can you check this part - why do we change the interpolation? We should make sure we preserve errors on the `about:policies` page.

mkaply:
The isEmptyObject null fix looks right, and moving _updateStatus and _activateStartupPolicies inside the try is the right call.

One thing though: the catch doesn't null out this._provider. We still notify EnterprisePolicies:Initialized after this, which starts remote polling, and _updatePolicies only bails when _provider is null. So the first poll that sees changed policies re-applies everything and _updateStatus puts us back to ACTIVE. In the felt case that window is real since performSignout is async and FELT quits us later. Adding this._provider = null like _resetEngine does should cover it.

Also the loop that clears _callbacks wipes the _cleanup callbacks that _cleanupPolicies just scheduled, and that already cleared browser.policies.applied, so the cleanup never runs again. Minor today since nothing in production listens for EnterprisePolicies:Cleanup, but it's a silent change.

The catch is basically a subset of _resetEngine now. Might be worth pulling out a shared helper so they don't drift.

Small thing: test_consumer_init_failure_does_not_shut_down runs with enterprise.policies.live.enabled set from live/head.js, so it's not really a consumer build. The assertion is fine, just the name.

toolkit/components/enterprisepolicies/EnterprisePoliciesParent.sys.mjs
johnreitano: Good catch: my change dropped the error text.  Restored the ${e} interpolation.

johnreitano:
Thanks @mkaply, all four addressed in the updated commit.

> the catch doesn't null out this._provider

Nulled now, along with `RemotePoliciesProvider.dropInstance()`. Dropping the remote instance also removes its `EnterprisePolicies:Initialized` observer, so a FAILED engine does not start polling at all, and `_updatePolicies()` bails on the null provider if an update fires anyway. The test now fires `EnterprisePolicies:Update` after each failed init and asserts the engine stays FAILED with no active policies.

> the loop that clears _callbacks wipes the _cleanup callbacks

The callback lists are snapshotted before the provider is built and restored on failure, so the `_cleanup` callbacks scheduled by `_cleanupPolicies()` survive and only what the failed attempt scheduled is dropped. A new test task checks the previous session's cleanup still runs (`EnterprisePolicies:Cleanup` for `onBeforeAddons`) after a failed init.

> Might be worth pulling out a shared helper

Pulled into `_discardPolicies()` (parsed policies, both hash maps, remote instance, provider), used by both the catch and `_resetEngine()`.

> it's not really a consumer build

Renamed to `test_non_felt_init_failure_does_not_shut_down` and reworded the comment in the catch, since a non-felt enterprise build takes that branch too.

mkaply (approved):
This looks good but "feels" overtested. Can you see if any of these are redundant?

johnreitano:
> This looks good but "feels" overtested. Can you see if any of these are redundant?

Yeah, there was a bit of redundancy there, which I fixed in my latest commit.

PR #1481, opened 2026-09-15
Bug 2072087 - Sign repackaged update MAR file

lissyx:
### Description 

Bugzilla: Bug-2072087

Produce signed MAR update our of repackages, so that the AutoConfig file is properly populated


lissyx:
This should land only once nightly is 159

jonathanmendez (approved):
It generally looks good/correct to me. I don't know if there are nuanced details of the names/symbols that someone like ahal might be better to review.

PR #1497, opened 2026-09-17
Bug 2072727 - Disable FOG init on shutdown during marionette tests

lissyx:
### Description 

Bugzilla: Bug-2072727

FOG init on shutdown adds some shutdown delay that breaks some of our testing where we wait not long enough for the FELT window to re-appear


1rneh (approved):
> FOG init on shutdown adds some shutdown delay that breaks some of our testing where we wait not long enough for the FELT window to re-appear

Have you looked into the option of getting rid of testing with an arbitrary timeout? This fix of disabling FOG seems very prone to regress.

lissyx:
> > FOG init on shutdown adds some shutdown delay that breaks some of our testing where we wait not long enough for the FELT window to re-appear
> 
> Have you looked into the option of getting rid of testing with an arbitrary timeout? This fix of disabling FOG seems very prone to regress.

That's one of the options I was exploring locally, yes, but I am also waiting on feedback from glean team in case this is not hiding another more troublesome issue, to be certain there's no underlying problem hidden here

PR #1503, opened 2026-09-17
Bug 2073027 - Pass auth token over a pipe instead of environment

gregawoods:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=2073027 (previously 2072386)

## Description

Auth tokens were being sent to the crash reporter child process via the environment, which was potentially readable by any same-user process. Moves the auth token sending to an inherited pipe to minimize risk. 

This change only applies to linux/macos, the windows path remains the same.

Tested the fix on a native arm64 Ubuntu 24.04 VM with an enterprise build.


lissyx:
I'm marking @afranchuk and/or @jonathanmendez as reviewers since they know this code quite better than I do

lissyx:
@gregawoods Can you issue a `mach try fuzzy` ? I guess you want to select crashreporter things, with `'enterprise`. Then force push your PR to make sure try status can be reflected on the PR

gregawoods:
@lissyx Thanks! Have a try kicked off [here](https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=d59c17cb7deed209070b0f376587841368acf423), will keep an eye on it. I guess my previous one didn't link up because the commit hash changed.

gregawoods:
Looks like a [fresh rust advisory](https://rustsec.org/advisories/RUSTSEC-2026-0293.html) is causing that cargo audit failure.

lissyx:
> Looks like a [fresh rust advisory](https://rustsec.org/advisories/RUSTSEC-2026-0293.html) is causing that cargo audit failure.

yes unfortunately

lissyx:
> Looks like a [fresh rust advisory](https://rustsec.org/advisories/RUSTSEC-2026-0293.html) is causing that cargo audit failure.

Outside of this known upstream issue, it looks good :) Thanks!

toolkit/crashreporter/client/app/src/net/auth.rs
afranchuk: I might have just made `ACCESS_TOKEN` a mocked value (even just by calling `mock::hook` in `access_token()`), but it's not a big deal. Plus the way it's done now tests the IO/FD code better.

toolkit/crashreporter/nsExceptionHandler.cpp
afranchuk: AFAIK `snprintf` won't allocate except for some floating point conversions, but I suppose we can't guarantee that. Thanks C/C++. `itoa` is another one, but is non-standardized.

afranchuk (approved):

jonathanmendez (approved):
This looks ok to me as an improvement from current state, but I wonder if we need to investigate something more general/structural?
- Is there any path in mind for covering Windows?
- I recently discovered that the workflow for a content crash ping is not currently being authenticated, and since it's spawned in js via Subprocess.call, it needs another mechanism.
- Maybe all paths (including Windows) could use stdin? My claude is suggesting it's possible but with some caveats I haven't fully explored yet.

If you want to land this and explore a more general solution elsewhere, then we can move this discussion to the Bug.

gregawoods:
>     * Is there any path in mind for covering Windows?

My understanding is that the problem this addresses (environment being readable via `/proc/<pid>/environ`) isn't a concern on windows, hence why it was left alone. I could see some value in all platforms taking a consistent path though.

Another potential angle is generating a limited access token that gets passed to the crash reporter and only has access to the API it needs and nothing else. I haven't checked on the api side yet but I imagine there'd be additional plumbing required for that.

I feel ok merging this in and having more discussion on what to do next.

PR #1506, opened 2026-09-17
Bug 2060944 - Select the proper browser window when opening external links on Windows

farmisen:
Fix: https://bugzilla.mozilla.org/show_bug.cgi?id=2060944

## Description

Opening an external URL on Windows doesn't bring the browser window to the front. When the remoting layer hands off foreground rights, it gives them to the Felt UI process rather than to the browsing child that actually owns the window, so a `window.focus()` call from the child has no effect.

This change threads the foreground permission through to the right process:

- The browsing child now includes its pid in the `FeltReady` message.
- Before forwarding an external URL, the UI calls `AllowSetForegroundWindow` with that pid, granting the child a one-time right to take the foreground.
- The child then uses that right by calling `SetForegroundWindow` on its most recently active browser window.

Clicking an external link while the browser is running now raises the browser window, matching the behavior on other platforms.

## Notes

  - Verified manually on Win 11 by opening a link via `-osint -url` against a running instance and
  comparing `GetForegroundWindow()` to the browser's HWND: PASS on the patched build, deterministic FAIL
  (foreground on the UI's hidden remoting window) without it.


gcp:
I was surprised this didn't work because I wrote the original code and tested this. It seems the critical part is _"SetForegroundWindow rules grant the right to a process that "was started by the foreground process". That condition is **evaluated at process creation**, so the browsing child only inherits it **if the UI process held the foreground at the moment it spawned the child**."_

The explicit `AllowSetForegroundWindow` call fixes this.

gcp:
Note this will interact with https://bugzilla.mozilla.org/show_bug.cgi?id=2064561 which adds back the Launcher Process (but because you capture the pid, it might work?)

farmisen:
> Note this will interact with https://bugzilla.mozilla.org/show_bug.cgi?id=2064561 which adds back the Launcher Process (but because you capture the pid, it might work?)

Verified that I cut my branch and done my testing after https://github.com/mozilla/enterprise-firefox/pull/1463 got merged in- and yes since we capture the proper fx window pid it will _(should)_ work.

toolkit/components/felt/rust/FeltWidgetUtilsWin.cpp
```diff
@@ -0,0 +1,39 @@
...
+  nsCOMPtr<mozIDOMWindowProxy> navWin;
+  winMediator->GetMostRecentBrowserWindow(getter_AddRefs(navWin));
+  if (!navWin) {
+    return;
+  }
+  nsCOMPtr<nsIWidget> widget = mozilla::widget::WidgetUtils::DOMWindowToWidget(
+      nsPIDOMWindowOuter::From(navWin));
+  if (!widget) {
+    return;
+  }
```
lissyx: nit: I think some `NS_WARNING()` could be useful here to debug in case things go wild

toolkit/components/felt/rust/src/utils.rs
```diff
@@ -84,6 +86,24 @@ pub static TOKENS: LazyLock<Arc<RwLock<Tokens>>> =
...
+#[cfg(target_os = "windows")]
+pub static BROWSER_PID: AtomicU32 = AtomicU32::new(0);
+
+#[cfg(target_os = "windows")]
+pub fn allow_browser_foreground() {
+    let pid = BROWSER_PID.load(Ordering::Relaxed);
+    if pid == 0 {
+        trace!("allow_browser_foreground(): no browser pid yet, not granting");
+        return;
+    }
```
lissyx: I think we should error here, there's no good reason we would be sending `OpenURL` without a running browser, so the PID should be there.

toolkit/components/felt/rust/src/utils.rs
```diff
@@ -84,6 +86,24 @@ pub static TOKENS: LazyLock<Arc<RwLock<Tokens>>> =
     LazyLock::new(|| Arc::new(RwLock::new(Default::default())));
 pub static CONSOLE_URL: OnceLock<Arc<String>> = OnceLock::new();
 
+#[cfg(target_os = "windows")]
+pub static BROWSER_PID: AtomicU32 = AtomicU32::new(0);
+
+#[cfg(target_os = "windows")]
+pub fn allow_browser_foreground() {
```
lissyx: make this return a `Result`

toolkit/components/felt/rust/src/components.rs
```diff
@@ -333,6 +333,8 @@ impl FeltXPCOM {
             disposition,
             focus_hint
         );
+        #[cfg(target_os = "windows")]
+        utils::allow_browser_foreground();
```
lissyx: We should get return value here and return a failure to the caller of `OpenURL`

toolkit/components/felt/rust/src/components.rs
```diff
@@ -443,7 +445,19 @@ impl FeltXPCOM {
...
+                        if matches!(
+                            msg,
+                            Ok(FeltMessage::Restarting
+                                | FeltMessage::Exiting
+                                | FeltMessage::LogoutShutdown)
+                                | Err(ipc_channel::IpcError::Disconnected)
+                        ) {
+                            trace!("FeltServerThread::felt_server::ipc_loop(): browser gone, clearing pid");
+                            crate::utils::BROWSER_PID.store(0, Ordering::Relaxed);
+                        }
```
lissyx: Would likely be easier to read with a `match` here as well

lissyx (changes requested):
LGTM, but the fact this flew under the radar justifies by itself that we absolutely need testing coverage here for this

toolkit/components/felt/rust/FeltWidgetUtilsWin.cpp
farmisen: done

toolkit/components/felt/rust/src/utils.rs
farmisen: done

toolkit/components/felt/rust/src/components.rs
farmisen: addressed

toolkit/components/felt/rust/src/components.rs
farmisen: moved pid reset calls inside the match bellow

farmisen:
> LGTM, but the fact this flew under the radar justifies by itself that we absolutely need testing coverage here for this

added a marionette test / verified it fails on enterprise-main and pass on this branch

testing/enterprise/test_felt_browser_external_link.py
```diff
@@ -14,13 +14,7 @@
...
 
 
-class FeltStartsBrowserExternalLink(FeltTests):
-    def test_browser_external_link(self):
-        self.run_felt_base()
-        self._external_link = f"http://localhost:{self.console_port}/ping"
-        self.run_felt_browser_started()
-        self.run_felt_open_external_link()
-
+class BaseBrowserExternalLink(FeltTests):
```
lissyx: Two classes in the same file will make marionette do very weird things, I'm even surprised it  works. You need to refactor so the base class is in a file named not `test_` and not having any `test_*` methods

toolkit/components/felt/rust/src/components.rs
```diff
@@ -452,21 +457,29 @@ impl FeltXPCOM {
                         match rx.recv() {
                             Ok(FeltMessage::Restarting) => {
                                 trace!("FeltServerThread::felt_server::ipc_loop(): Restarting");
+                                #[cfg(target_os = "windows")]
```
lissyx: we do not need this elsewhere than windows, but the amount of `#[cfg()]` it adds make me wondering if it would not be easier to just do it on any platform?

toolkit/components/felt/rust/src/message.rs
```diff
@@ -64,7 +64,7 @@ pub enum FeltMessage {
     /// first IPC handshake, ahead of `AccessToken` / `Ready`.
     PrimarySecret(String),
     RefreshTokens,
-    FeltReady,
+    FeltReady(u32),
```
lissyx: Especially because we do not specialize for Windows here, so maybe just report the browser PID everywhere

toolkit/components/felt/rust/src/utils.rs
```diff
@@ -84,6 +90,34 @@ pub static TOKENS: LazyLock<Arc<RwLock<Tokens>>> =
...
 pub static CONSOLE_URL: OnceLock<Arc<String>> = OnceLock::new();
 
+#[cfg(target_os = "windows")]
+pub static BROWSER_PID: AtomicU32 = AtomicU32::new(0);
+
+/// Lets the browsing child bring itself to the front once. Returns whether
+/// Windows accepted the grant, which it refuses when we don't hold the
+/// foreground right ourselves. Errors if the browser never sent its pid.
+#[cfg(target_os = "windows")]
+pub fn allow_browser_foreground() -> Result<bool, nserror::nsresult> {
```
lissyx: and make this work built on all platform but only doing real stuff on windows

lissyx (changes requested):
I think if we simplify all the `#[cfg()]` that we can, this would not be a big deal on other platforms, it could even be useful for other uses, and it would make following the code easier, the only things that needs to be windows specific here is really the activation step / allow_browser_foreground step

testing/enterprise/test_felt_browser_external_link.py
farmisen: done

toolkit/components/felt/rust/src/components.rs
farmisen: done

toolkit/components/felt/rust/src/utils.rs
farmisen: done

lissyx (approved):
r=me with the latest comment addressed: if it's too much work for this PR, maybe a follow up blocking `firefox-client-mvp` ? And we can uplift if we decide it's important enough

toolkit/components/felt/rust/src/utils.rs
farmisen: addressed l94-97

PR #1507, opened 2026-09-17
Bug 2073249 - Fill in the posture build, os and security fields that TelemetryEnvironment no longer provides

gcp:
Fixes [bug 2073249](https://bugzilla.mozilla.org/show_bug.cgi?id=2073249). Upstream [bug 2068624](https://bugzilla.mozilla.org/show_bug.cgi?id=2068624) removes `environment.system.sec`, replaces most of `environment.build` with schema placeholders (`version` becomes `"00."`) and drops `locale`, `windowsUBR`, `installYear`, `hasPrefetch` and `hasSuperfetch` from `environment.system.os`. `DevicePosture.collect` read all three sections from `TelemetryEnvironment.currentEnvironment`, so the console receives no security section and a bogus browser version.

**Commit 1** keeps reading what the environment still provides and layers the missing fields on top from `Services.appinfo`, `Services.sysinfo`, `AppConstants` and the Windows registry, so the payload is unchanged. The `DevicePosture` typedef is updated to describe the sections by content. The bug suggests moving off `TelemetryEnvironment` entirely; that would mean duplicating the environment's remaining `os` and `build` collection here, so this keeps the shared part shared.

**Commit 2** adds a check to `test_felt_device_posture_elements.py` that the `build`, `os` and `security` sections are populated and that `build.version` matches `Services.appinfo.version`, in both Felt (before sign-in, when it collects posture and still has a window to script) and the browser.

Verified locally on macOS: the Marionette posture suite passes 7/7. The Windows-only fields are untested here.


toolkit/components/enterprise/modules/DevicePosture.sys.mjs
```diff
@@ -345,8 +436,11 @@ export const DevicePosture = {
 
     const devicePosturePayload = {
       os,
-      security: lazy.TelemetryEnvironment.currentEnvironment.system.sec,
-      build: lazy.TelemetryEnvironment.currentEnvironment.build,
+      security: getSecurityInfo(),
```
jporter-dev, line 439: `data.sec = this._getSecurityAppData();` in ` toolkit/components/telemetry/app/TelemetryEnvironment.sys.mjs` was within a `if (AppConstants.platform === "win")` branch, so I _think_ this shape changes slightly from `security: null` to `security: {antivirus: null, antispyware: null, firewall: null}` on non-Windows. https://searchfox.org/enterprise-main/source/toolkit/components/telemetry/app/TelemetryEnvironment.sys.mjs#1870-1884

I haven't looked at how this is used in the Console, but if nothing relies on `security` yet, this seems harmless, and it's already documented at `getSecurityInfo` as expected on anything not Windows.

jporter-dev (approved):
The changes look sound to me! Left one note about the change in how the Windows security payload is returned for non-Windows.

toolkit/components/enterprise/modules/DevicePosture.sys.mjs
gcp: @fiji-flo I assume making this consistent between OSes is an improvement from your side, but just highlighting this.

PR #1508, opened 2026-09-17
Bug 2073025 - Validate the FELT UI scratch profile directory before use

vstepmoz:
The FELT UI launcher reuses a scratch profile at a fixed path in the OS temporary directory (felt-<MOZ_UPDATE_CHANNEL>). Require that directory to be a private per-user directory before adopting or wiping it.

Add ValidateFeltScratchDir(), which on Unix requires the path to be a directory (not a symlink) owned by the current user with mode 0700, and call it from both SelectProfile (adoption) and ResetFeltUIScratchProfile (wipe). It is a no-op where the OS temporary directory is already per-user (macOS, Windows).

Startup fails closed when the check does not pass; moving the scratch profile to a per-user location (e.g. XDG_RUNTIME_DIR) is left as follow-up.

Add testing/enterprise/test_felt_ui_scratch_profile_refuses.py, which launches the binary without a forced profile and asserts the refusal for the symlink and non-0700 cases.

### Description 

Bugzilla: Bug-2073025

The FELT UI launcher reuses a scratch profile at a fixed path in the OS temporary directory (`felt-<MOZ_UPDATE_CHANNEL>`). Require that directory to be a private per-user directory before adopting or wiping it.

  Add `ValidateFeltScratchDir()`, which on Unix requires the path to be a directory (not a symlink) owned by the current user with mode 0700, and call it from both `SelectProfile` (adoption) and `ResetFeltUIScratchProfile` (wipe). It is a no-op where the OS temporary directory is already per-user (macOS, Windows).

  Startup fails closed when the check does not pass; moving the scratch profile to a per-user location (e.g. `XDG_RUNTIME_DIR`) is left as follow-up.

---

* Depends on: none

---

### Screenshots 

Not applicable — no UI changes.

---

### Testing 

* [x] Added tests


toolkit/xre/nsAppRunner.cpp
```diff
@@ -2952,6 +2952,25 @@ static nsresult ProfileEncryptionMismatchDialog(const char* aMsgKey,
...
+// Returns NS_OK only if |aDir| is a private per-user directory: a directory
+// (not a symlink) owned by the current user with mode 0700. Unix-only; a no-op
+// elsewhere, where the OS temporary directory is already per-user.
+static nsresult ValidateFeltScratchDir(nsIFile* aDir) {
+#  if defined(XP_UNIX)
+  nsAutoCString path;
+  MOZ_TRY(aDir->GetNativePath(path));
+
+  // lstat rather than stat so a symlink is reported as a symlink instead of
+  // being followed to its target.
```
lissyx, line 2964: dont we want to check both in this case, out of safety?

toolkit/xre/nsAppRunner.cpp
```diff
@@ -3645,6 +3666,13 @@ static nsresult SelectProfile(nsToolkitProfileService* aProfileSvc,
         // many (thousands) of existing directories, which is unlikely to
         // happen.
         MOZ_TRY(file->CreateUnique(nsIFile::DIRECTORY_TYPE, 0700));
+      } else if (NS_FAILED(ValidateFeltScratchDir(file))) {
+        // Only reuse an existing scratch directory if it is a private per-user
+        // directory.
+        Output(true,
+               "Error: refusing to use the Felt UI scratch profile: it is not "
+               "a private directory owned by the current user.\n");
+        return NS_ERROR_ABORT;
```
lissyx: nit: maybe we could verify that always, not just when we have existing dir, just to be sure `CreateUnique` from above did what it should have done

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
```diff
@@ -0,0 +1,125 @@
...
+# requires a display -- provided by the marionette-enterprise CI environment.
+FELT_SCRATCH_DIR_NAME = "felt-default"
+REFUSAL_MESSAGE = "refusing to use the Felt UI scratch profile"
+
+
+@unittest.skipIf(
+    sys.platform == "win32",
+    "The scratch-profile validation is XP_UNIX-only; Windows uses a per-user "
+    "temporary directory and ValidateFeltScratchDir is a no-op there.",
+)
```
lissyx: Please use `run-if` or `skip-if` in the manifest rather

lissyx: (its already the case, so just remove this)

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
```diff
@@ -0,0 +1,125 @@
...
+            env=self._child_env(),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            text=True,
+            timeout=90,
+            check=False,
+        )
+        self._logger.info(f"Bare launch exited: {proc.returncode}")
+        return proc.stdout
+
```
lissyx, line 80: We should not just check `stdout` but exit code here. Also verify nothing has been used in the broken profile (`mtime` checks ?)

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
```diff
@@ -0,0 +1,125 @@
...
+    def test_refuses_directory_symlink(self):
+        # A symlink at the scratch path must be rejected (lstat/!S_ISDIR) rather
+        # than followed, so neither adoption nor the wipe traverses it.
+        self._decoy_path = tempfile.mkdtemp(prefix="felt-refuse-decoy")
+        os.symlink(self._decoy_path, self._scratch_path)
+
+        output = self._launch_bare_and_capture()
+        assert REFUSAL_MESSAGE in output, (
+            f"Expected refusal for a symlinked scratch profile; output was:\n{output}"
+        )
```
lissyx, line 138: Also assert exit code

lissyx (approved):
r=me with comments addressed

lissyx (changes requested):
Looks like there are tests regressions: https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=8d01a11be8bbe7ac23b322fa4918eb94ae57fb1e

The test is failing on macOS: https://treeherder.mozilla.org/logviewer?job_id=593743900&repo=enterprise-firefox-try&task=Zg9mGm7DT22wc8uVYWqb8A.0&lineNumber=10825

There may be new failures on windows:
https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=8d01a11be8bbe7ac23b322fa4918eb94ae57fb1e&selectedTaskRun=IPMQCjUQQ9iATuokNukPAQ.0
https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=8d01a11be8bbe7ac23b322fa4918eb94ae57fb1e&selectedTaskRun=JABB3yOvTHCN91zHuZ-1sA.0

toolkit/xre/nsAppRunner.cpp
```diff
@@ -2952,6 +2952,25 @@ static nsresult ProfileEncryptionMismatchDialog(const char* aMsgKey,
 }
 
 #if defined(MOZ_ENTERPRISE)
+// Returns NS_OK only if |aDir| is a private per-user directory: a directory
+// (not a symlink) owned by the current user with mode 0700. Unix-only; a no-op
+// elsewhere, where the OS temporary directory is already per-user.
+static nsresult ValidateFeltScratchDir(nsIFile* aDir) {
+#  if defined(XP_UNIX)
```
lissyx: If macOS is fine then we should use `XP_LINUX` and make sure we do not care for Android here?

```suggestion
#  if defined(XP_LINUX) && !defined(ANDROID)
```

testing/enterprise/manifest-generic.toml
```diff
@@ -38,6 +38,9 @@ subsuite = "enterprise"
 
 ["test_felt_token_refresh_5xx.py"]
 
+["test_felt_ui_scratch_profile_refuses.py"]
+skip-if = ["os == 'win'"]
```
lissyx:
```suggestion
run-if = ["os == 'linux'"]
```

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
vstepmoz: Added exit code check. Also added mtime check for the whole scratch directory tree.

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
vstepmoz: Added

toolkit/xre/nsAppRunner.cpp
vstepmoz: I like your idea. Changed to check that always.

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
vstepmoz: Updated.

toolkit/xre/nsAppRunner.cpp
vstepmoz: Replaced with XP_LINUX and ANDROID check

vstepmoz: Added that. Thank you.

testing/enterprise/manifest-generic.toml
vstepmoz: Updated.

toolkit/xre/nsAppRunner.cpp
vstepmoz: ValidateFeltScratchDir accepts the scratch directory only if all of:
- it's a directory, not a symlink (lstat + !S_ISDIR),
- it's owned by the current user (st_uid == geteuid()),
- its mode is exactly 0700.
On lstat vs stat: that's deliberate. lstat stats the path itself instead of following it, so a symlink fails the S_ISDIR check and is rejected — we stop there rather than follow the link.

I might actually be getting this wrong — could you please clarify what "both" refers to here? If you meant something else, I'm happy to add it.

vstepmoz:
@lissyx thanks for the review! I've addressed all the comments (replied inline on each) in dedca599. The two remaining red jobs look unrelated to this change — macOS test_felt_restart_is_quit and Windows test_felt_browser_updates_forward, both process-teardown assertions in other tests; our test is run-if linux (skipped on both) and the C++ check is a no-op off Linux. Mind taking another look?

lissyx:
> @lissyx thanks for the review! I've addressed all the comments (replied inline on each) in [dedca59](https://github.com/mozilla/enterprise-firefox/commit/dedca599e1507f5c57d8f18be229a2c89d0a9866). The two remaining red jobs look unrelated to this change — macOS test_felt_restart_is_quit and Windows test_felt_browser_updates_forward, both process-teardown assertions in other tests; our test is run-if linux (skipped on both) and the C++ check is a no-op off Linux. Mind taking another look?

I think you're right but I'm a bit curious especially of the one about `Sqlite.sys.mjs`. I've retriggered them ...

lissyx:
@vstepmoz Still pending execution of https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=32e7f38034f55e1b53d3cc2d1b52122c0cf0ccbb&selectedTaskRun=Savr6RipSvqBE5Aal-ofKA.0 but the windows failure is intermittent, so I think you're good. Do you mind keeping an eye on this macOS test failure retrigger and if it's confirmed to be an intermittent, I think you can land this? Hopefully it should be no more than a few hours ...

vstepmoz:
Will do! I'll update here once it finishes.

vstepmoz:
macOS failed again with the same error. I'll try to figure out if it's somehow related to my changes.

vstepmoz:
@lissyx  It looks like it's an intermittent failure too, just success rate is lower. Here are few jobs with the same error: [job_1](https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=9907a8eed9eebe475a4785afa946cf5800ea77a3&selectedTaskRun=ed2390yjTfqD6mVK89ehSg.0), [job_2](https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=6384eeb19246918f1fa3ca6a886e8f29fcd598ea&selectedTaskRun=TD4wBQWTT3i-9NNFDLg3xQ.0l), [job_3](https://treeherder.mozilla.org/jobs?repo=enterprise-firefox-try&revision=9a22005d0588152c29fd9d8db091e14e77c96f08&selectedTaskRun=BlafsUZcSJiobsk-NH1pew.0). Probably worth filing a separate bug to address that. Do you think the current patch is good and we could land that? Please let me know if there's anything else needed here.

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
```diff
@@ -0,0 +1,163 @@
...
+        proc = subprocess.Popen(
+            [binary],
+            env=self._child_env(),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            text=True,
+            start_new_session=True,
+        )
+        try:
+            output, _ = proc.communicate(timeout=90)
```
lissyx: 90s of timeout is likely way too much, I doubt more than 10s is required

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
```diff
@@ -0,0 +1,163 @@
...
+        except subprocess.TimeoutExpired:
+            os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
+            proc.communicate()
+            raise
```
lissyx: Please raise a clear exception with a clear error message, so that debugging on CI is eased

toolkit/xre/nsAppRunner.cpp
lissyx: Checking both the symlink and the target ?

Also the comment is not clear "so a symlink is reported as a symlink instead of being followed to its target"

lissyx (approved):
Thanks, this looks good

r=me with the remainder of the comments addressed

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
vstepmoz: Dropped to 10s

testing/enterprise/test_felt_ui_scratch_profile_refuses.py
vstepmoz: Updated. Raising exceptions with clear message now.

toolkit/xre/nsAppRunner.cpp
vstepmoz: My comment was misleading, I've dropped it. The function header already states the rule: "Returns NS_OK only if |aDir| is a private per-user directory: a directory (not a symlink) owned by the current user with mode 0700."
That scratch directory (/tmp/felt-<channel>) is only ever created by our code, as a real directory with mode 0700 — we never create a symlink there. So finding a symlink at that path is already proof that something other than us put it there. That is enough to refuse.
 Checking where it points wouldn't add anything: whoever planted the link still controls it and can repoint it at any time.

vstepmoz:
@lissyx thanks again! Addressed all three in c638653 (replied inline on each). Jobs are done: 54 pass, 3 fail. The two marionette reds are the same intermittents as before. The third isn't: source-test-mozlint-cargo-audit is RUSTSEC-2026-0293 against ringbuf. It's a Rust lint check, so it won't clear on a retrigger. But it isn't related to this PR either — it's tree-wide.

lissyx:
> @lissyx thanks again! Addressed all three in [c638653](https://github.com/mozilla/enterprise-firefox/commit/c6386532062041ee87e22d92121d0ce520357e84) (replied inline on each). Jobs are done: 54 pass, 3 fail. The two marionette reds are the same intermittents as before. The third isn't: source-test-mozlint-cargo-audit is RUSTSEC-2026-0293 against ringbuf. It's a Rust lint check, so it won't clear on a retrigger. But it isn't related to this PR either — it's tree-wide.

Thanks!

PR #1525, opened 2026-09-18
Bug 2073585 - Only run stub installer test when stub installer is built

1rneh:
### Description 

Bugzilla: Bug-2073585

Regressed by Bug-2068879 which skipped building the test stub installer since the unused stub installer assets were removed from the branding folder.


lissyx (approved):
r=me with comments addressed

browser/installer/windows/nsis/moz.build
```diff
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-XPCSHELL_TESTS_MANIFESTS += [
-    "test/xpcshell/xpcshell.toml",
-]
+if CONFIG["MOZ_STUB_INSTALLER"]:
```
lissyx, line 5: Just make it clear we can because there's a single test that is only for stub installer and we have no mozinfo for the stub installer here.

maybe safer alternative would be `skip-if` on the stub installer test itself in the manifest

lissyx: https://searchfox.org/enterprise-main/rev/42d649c34b67e9b2819dd87af8d0c2612fda8d48/browser/installer/windows/nsis/test/xpcshell/xpcshell.toml#1-9 skipping here on `enterprise`

